PR #1, opened 2022-08-19
Add quickstart to Readme

Lagoja:
Signed-off-by: John Lago <750845+Lagoja@users.noreply.github.com>


loreto:
You have to do the change in axiom (opensource/devbox) and it'll get published here. 

If you do the PR in the devbox repo it'll get clobbered when we republish from axiom.

PR #37, opened 2022-08-29
devbox: fix deduping packages and remove lo

gcurtis:
## Summary

Fix Devbox.Add to deduplicate packages instead of removing the ones with duplicates. Extract the two functions that we were using in lo to remove it as a dependency.

## How was it tested?

`go test`


loreto (approved):
LGTM

Although, I'd prefer if the implementations of `unique` and `exclude` are in a different file (or package). In general I'm trying to make sure `devbox.go` only (or primarily) contains the public interface to the library.

gcurtis:
Sounds good. I'll move them somewhere else before merging.

PR #40, opened 2022-08-29
[devbox] Add anonymized telemetry

loreto:
## Summary
Add anonymized telemetry to understand usage of devbox and further improve it. We allow users to opt-out via an environment variable.

To add telemetry I implemented a small "midcobra" framework, that make it possible to add "middlware" to cobra CLIs. The plan is to re-use the midcobra functionality in other binaries like `envsec`.

## How was it tested?
Built locally, ran, and checked telemetry logs.


.goreleaser.yaml
```diff
@@ -10,6 +10,7 @@ builds:
       - -s -w -X go.jetpack.io/devbox/build.Version={{.Version}}
       - -s -w -X go.jetpack.io/devbox/build.Commit={{.Commit}}
       - -s -w -X go.jetpack.io/devbox/build.CommitDate={{.CommitDate}}
+      - -s -w -X go.jetpack.io/devbox/build.TelemetryKey={{ .Env.TELEMETRY_KEY }}
```
gcurtis, line 13: Does this actually need to be kept secret? It'll still be visible in the binary. Can we just put it in a const instead of a linker flag?

gcurtis (approved):
I just want to make sure we're not prematurely architecting things with the middleware. It's a little tricky to follow as opposed to something like:

```go
t := tracker.Track(...)
defer t.done()

// run the cobra stuff
```

.goreleaser.yaml
loreto: I'm gonna leave as is for this PR since it's already working, but we can hard-code it in a follow up clean up.

I'm open either way, last time I asked on Slack, folks leaned towards hiding this, so that's why I implemented this way.

Issue #8, opened 2022-08-26
Feature Request: support building containers for nodejs apps


crisdosyago:
Just a question. Understand if you're not planning this. 

Cool looking project!

robrich:
You can add [nodejs](https://search.nixos.org/packages?query=nodejs) package to your Devbox shell, and then dotnet run inside Devbox. We do need to build the Node.js detection when you generate an image though. Thanks for the feature request.

loreto:
As @robrich mentioned:
+ For the shell functionality, we already support nodejs – and you can install [any of the versions listed here](https://github.com/jetpack-io/devbox/issues/8).
+ For the container support, we are planning to add nodejs pretty soon. I'll repurpose this ticket to track the container support and we'll let you know as soon as we add it.

MasterOdin:
@loreto It looks like the link you put just links back to this ticket?

loreto:
Sorry!

Here's the right link: https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=nodejs

Issue #2, opened 2022-08-24
Allow overwrites on install, build and start command


LucilleH:
Right now `devbox plan` does automatic language detection and sets the install, build and start commands. But it only supports Golang for now.

I'd like to try it on a python project and build a container with it. Would love to see those three fields being exposed in `devbox.json` so someone can go ahead and modify them. Thoughts?

loreto:
Closing, will be released tomorrow.

PR #41, opened 2022-08-30
[planners] First pass on node js planner

LucilleH:
## Summary

Node JS planner initial pass.

TODOs:
- [x] Dockerfile optimization (NodeJS projects absolutely need this, as build time can be exceedingly long without it)
- [x] NodeJS examples
- [x] Node version detector
- [x] Base images for start stage (pending https://github.com/jetpack-io/devbox/pull/52)

See https://github.com/jetpack-io/devbox/issues/8

Example devbox.json
```
{
  "install_stage": {
    "command": "yarn install"
  },
  "build_stage": {
    "command": "yarn build"
  },
  "start_stage": {
    "command": "yarn start"
  },
  "packages": []
}
```

Output from devbox plan
```
{
  "install_stage": {
    "command": "yarn install",
    "input_files": [
      "package.json",
      "yarn.lock"
    ]
  },
  "build_stage": {
    "command": "yarn build",
    "input_files": [
      "."
    ]
  },
  "start_stage": {
    "command": "yarn start",
    "input_files": [
      "."
    ]
  },
  "dev_packages": [
    "nodejs",
    "yarn"
  ],
  "runtime_packages": [
    "nodejs",
    "yarn"
  ]
}
```

## How was it tested?
`devbox shell`
`devbox plan`
`devbox build` on `examples/python-poetry`
`devbox build` on `examples/nodejs`


hgezim:
Thanks for this @LucilleH! What else needs to be done. I need this :D

LucilleH:
@hgezim I'm waiting on https://github.com/jetpack-io/devbox/issues/2 since I don't feel very comfortable making assumptions about the build and start command. I'm hoping @loreto can get that in soon so I can build on top of it!

generate.go
```diff
@@ -20,7 +20,7 @@ var tmplFS embed.FS
 
 func generate(rootPath string, plan *planner.BuildPlan) error {
 	// TODO: we should also generate a .dockerignore file
-	files := []string{".gitignore", "Dockerfile", "shell.nix", "default.nix"}
+	files := []string{".gitignore", "Dockerfile", "shell.nix", "default.nix", "Dockerfile.dockerignore"}
```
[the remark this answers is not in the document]
hgezim: Does `DOCKER_BUILDKIT=1` needs to be set (author sets it here: https://github.com/moby/moby/issues/12886#issuecomment-480575928 ).

hgezim:
I think assumptions for build and start commands are completely fine for first pass. Furthermore, I think we should set a default for these even if we allow the user to change them via config. @LucilleH thoughts?

planner/nodejs_planner.go
```diff
@@ -0,0 +1,54 @@
...
+	return &BuildPlan{
+		Packages:       []string{},   // This should be what users set in nix packages. Leaving it empty for now.
+		InstallFiles:   installFiles, // TODO: add this to Dockerfile template.
+		InstallCommand: fmt.Sprintf("%s install", pkgManager),
+		// This does not need to be npm build or yarn build, since build script is set by users.
+		// For now we assume users will set the build script as "build", but it could be really anything.
+		// In fact, some apps don't even require a build step.
+
+		// Lucille's controversial opinion: For NodeJS projects, we should require BuildCommand and StartCommand be set by users.
+		BuildCommand: fmt.Sprintf("%s build", pkgManager),
```
hgezim, line 49: `npm build` is not valid (at least for npm@8).

generate.go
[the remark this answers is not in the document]
LucilleH: We set it here https://github.com/jetpack-io/devbox/blob/main/docker/docker.go#L67. It does work with `BUILDKIT=1` I'm assuming `BUILDKIT` and `DOCKER_BUILDKIT` can be used interchangeably? @loreto to confirm.

LucilleH:
> Another thing, don't we need to install a `nodejs` package like `go_planner` does here? https://github.com/jetpack-io/devbox/blob/main/planner/go_planner.go#L28

Planning to change how we do `go_planner`. We should use the `packages` field defined in `devbox.json` instead, so that the shell environment and the container is as closely mapped as possible.

planner/nodejs_planner.go
LucilleH: Yeah so for `build` to be valid, it has to be a command defined in your `package.json` - Hence why I don't feel comfortable making any assumptions 😄

hgezim:
> We should use the `packages` field defined in `devbox.json` instead, so that the shell environment and the container is as closely mapped as possible.

Heck yea! Makes so much more sense.

planner/nodejs_planner.go
```diff
@@ -0,0 +1,121 @@
...
+package planner
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"go.jetpack.io/devbox/cuecfg"
+)
+
+type NodeJsPlanner struct{}
```
gcurtis, line 13:
```suggestion
type NodeJSPlanner struct{}
```

planner/plan.go
```diff
@@ -42,8 +42,11 @@ type SharedPlan struct {
 }
 
 type Stage struct {
-	Command string `cue:"string" json:"command"`
-	Image   string `json:"-"`
+	Command    string   `cue:"string" json:"command"`
+	InputFiles []string `cue:"[...string]" json:"input_files,omitempty"`
+	Scripts    []string `cue:"[...string]" json:"scripts,omitempty"`
```
loreto, line 47: Two thoughts:
1. Could we just make these part of command for now to avoid introducing a new field?
i.e. your command can be:
```
script1 && script2 && do-something
```
We could even allow command to be an in-line script (see https://docs.docker.com/engine/reference/builder/#example-creating-inline-files)

Which would let you write a script without all the `&&`:
```
StartCommand: heredoc.Doc(`
    step 1
    step 2
    step 3
    command
`)
```

2. If we really do need a field like this, I'd prefer to call it something like `preCommandHooks`

generate.go
```diff
@@ -21,9 +21,7 @@ import (
 var tmplFS embed.FS
 
 var shellFiles = []string{".gitignore", "shell.nix"}
-
-// TODO: we should also generate a .dockerignore file
-var buildFiles = []string{".gitignore", "development.nix", "runtime.nix", "Dockerfile"}
+var buildFiles = []string{".gitignore", "development.nix", "runtime.nix", "Dockerfile", "Dockerfile.dockerignore"}
```
gcurtis, line 24: Does the name `Dockerfile.dockerignore` work? I thought it had to be named `.dockerignore`.

planner/nodejs_planner.go
```diff
@@ -0,0 +1,121 @@
...
+
+		SharedPlan: SharedPlan{
+			InstallStage: &Stage{
+				InputFiles: inputFiles,
+				Command:    fmt.Sprintf("%s install", pkgManager),
+			},
+
+			BuildStage: &Stage{
+				// Copy the rest of the directory over, since at install stage we only copied package.json and its lock file.
+				InputFiles: []string{"."},
```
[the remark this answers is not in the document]
LucilleH: If we do that, go and python planner will copy the whole directory twice. Is that what we want?

generate.go
LucilleH: See https://github.com/jetpack-io/devbox/pull/41#discussion_r958745018

tmpl/Dockerfile.tmpl
```diff
@@ -30,56 +30,70 @@ RUN --mount=type=cache,target=/nix/store,from=base,source=/nix/store \
...
 # ------------------------------------------------------
-
+{{ if .StartStage }}
+  {{ if .StartStage.InputFiles }}{{ range $i, $files := .StartStage.InputFiles }}
+COPY --link --from=builder /scratch/{{ $files }} ./
+  {{- end }}{{- end }}
+  {{ if not .StartStage.InputFiles }}
+COPY --link --from=builder /scratch/. ./
+  {{- end }}
+  {{ if .StartStage.Scripts }}{{ range $i, $script := .StartStage.Scripts }}
```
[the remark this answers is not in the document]
loreto: I dunno, I was assuming that would be a "no-op", but I don't know if that's the case.

planner/plan.go
LucilleH: This is mostly for the `Entrypoint` field in Dockerfile - do we want the `Entrypoint` to be `script1 && script2 && startCommand`? Or should it be `Run script1 && script2` and `Entrypoint startCommand`?

tmpl/Dockerfile.tmpl
```diff
@@ -30,56 +30,55 @@ RUN --mount=type=cache,target=/nix/store,from=base,source=/nix/store \
...
-WORKDIR /app
 
+WORKDIR /app
 # 4. START STAGE
 #    Execute the application inside the final container.
 # ------------------------------------------------------
-
+{{ if .StartStage }}
+  {{ if .StartStage.InputFiles }}{{ range $i, $files := .StartStage.InputFiles }}
+COPY --link --from=builder /scratch/{{ $files }} ./
```
loreto, line 79: Because of: https://github.com/moby/moby/issues/37965

You might want to add a dummy `RUN true` after each `COPY`

loreto (approved):
LGTM

loreto:
Note that one of the tests is failing with a segfault

hgezim:
Any ideas on when this will make it out to a release?

loreto:
@hgezim We're planning to do a release tomorrow. 🤞

Issue #33, opened 2022-08-29
Devbox build unable to handle different versions of Go


RickyGrassmuck:
### Issue
When adding go_1_19 to my projects Devbox configuration, the `devbox build` command fails with the error

```
#0 38.04 error: collision between `/nix/store/i0rdvlkvc5sz0dr87s379j0i4xmnwrp5-go-1.18.4/bin/go' and `/nix/store/w205lbrnh8cm9680pqzw2pc07rhi43an-go-1.19/bin/go'
#0 38.15 error: builder for '/nix/store/9vrr8h6l7n1gjrl8ybb24qrvpbirvd33-devbox.drv' failed with exit code 25
------
error: failed to solve: executor failed running [/bin/sh -eux -o pipefail -c nix-env -if ./.devbox/gen/default.nix]: exit code: 100
```

### Expected Result
A devbox container image that includes Go version 1.19

### Additional Information
I dug into the code a bit and the issue is being caused by the language detection feature for Go. The Go planner builder does not contain any logic for detecting that a golang nix package has already been requested which results in both the `go_1_19` and `go` Nix packages being added to the list of packages used for populating the `default.nix` and `shell.nix` template files.

Contents of devbox.json 
```
{
  "packages": [
    "go_1_19"
  ]
}
```

Generated default.nix
```
let
  pkgs = import (fetchTarball {
    # Commit hash as of 2022-08-16
    # `git ls-remote https://github.com/nixos/nixpkgs nixos-unstable`
    url = "https://github.com/nixos/nixpkgs/archive/af9e00071d0971eb292fd5abef334e66eda3cb69.tar.gz";
    sha256 = "1mdwy0419m5i9ss6s5frbhgzgyccbwycxm5nal40c8486bai0hwy";
  }) {};
in with pkgs;
buildEnv {
  name = "devbox";
  paths = [
    go_1_19
  
    go
  ];
  pathsToLink = [ "/bin" ];
}
```
Generated shell.nix
```
let
  pkgs = import (fetchTarball {
    # Commit hash as of 2022-08-16
    # `git ls-remote https://github.com/nixos/nixpkgs nixos-unstable`
    url = "https://github.com/nixos/nixpkgs/archive/af9e00071d0971eb292fd5abef334e66eda3cb69.tar.gz";
    sha256 = "1mdwy0419m5i9ss6s5frbhgzgyccbwycxm5nal40c8486bai0hwy";
  }) {};
in with pkgs;
mkShell {
[...]

PR #43, opened 2022-08-30
[devbox] Allow for custom install/build/start commands

loreto:
## Summary
Allow for custom install/build/start commands by customizing them in `devbox.json`. This makes it possible for users to build docker images in the cases when they need to do some customization, or when we have not yet written a language planner for that language.

## How was it tested?
Built, and ran `devbox plan` and `devbox generate` against a few examples.

### DIFF
--- a/planner/go_planner.go
+++ b/planner/go_planner.go
@@ -22,14 +22,20 @@ func (g *GoPlanner) IsRelevant(srcDir string) bool {
 	return fileExists(goModPath)
 }
 
-func (g *GoPlanner) Plan(srcDir string) *BuildPlan {
-	return &BuildPlan{
+func (g *GoPlanner) GetPlan(srcDir string) *Plan {
+	return &Plan{
 		Packages: []string{
 			"go",
 		},
-		InstallCommand: "go get",
-		BuildCommand:   "CGO_ENABLED=0 go build -o out",
-		StartCommand:   "./out", // TODO: Move gin specific stuff elsewhere.
+		InstallStep: &Step{
+			Command: "go get",
+		},
+		BuildStep: &Step{
+			Command: "CGO_ENABLED=0 go build -o app",
+		},
+		StartStep: &Step{
+			Command: "./app",
+		},
 	}
 }
 
--- a/planner/go_planner.go
+++ b/planner/go_planner.go
@@ -27,13 +27,13 @@ func (g *GoPlanner) GetPlan(srcDir string) *Plan {
 		Packages: []string{
 			"go",
 		},
-		InstallStep: &Step{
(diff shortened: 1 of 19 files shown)


loreto:
This addresses https://github.com/jetpack-io/devbox/issues/2

gcurtis (approved):
LGTM

planner/planner.go
```diff
@@ -6,21 +6,21 @@ package planner
 type Planner interface {
 	Name() string
 	IsRelevant(srcDir string) bool
-	Plan(srcDir string) *BuildPlan
+	GetPlan(srcDir string) *Plan
```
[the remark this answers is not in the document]
loreto: I wanted to use `Plan` for the struct (since `BuildPlan` feels like a misnomer when we also use it to start the shell, and then `go` complained that the struct couldn't be called `Plan` at the same time as the method)

Thoughts?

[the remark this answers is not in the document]
gcurtis: Ah, I see. I'd just go ahead and merge then. We can think about the names later.

planner/go_planner.go
```diff
@@ -22,14 +22,20 @@ func (g *GoPlanner) IsRelevant(srcDir string) bool {
...
-	return &BuildPlan{
+func (g *GoPlanner) GetPlan(srcDir string) *Plan {
+	return &Plan{
 		Packages: []string{
 			"go",
 		},
-		InstallCommand: "go get",
-		BuildCommand:   "CGO_ENABLED=0 go build -o out",
-		StartCommand:   "./out", // TODO: Move gin specific stuff elsewhere.
+		InstallStep: &Step{
```
LucilleH, line 30: For packages, will mergo append `go` on top of the `packages` definition in `devbox.json` config?

LucilleH (approved):

planner/go_planner.go
LucilleH: If so, should we delete the `go` package in `go_planner`?

loreto: Not sure I followed everything.

+ Packages will get appended and duplicates removed (see planner/plan_test.go)
+ We need `go` in order to call `go get` even in the case the user has not requested that package.

Is that what you were asking?

loreto: That would cause an error (see https://github.com/jetpack-io/devbox/issues/33), but I believe the fix is to improve the planner code to detect the version of go (which is stated in the `go.mod`), and add the right package

LucilleH: That would not work for NodeJS. I'm guessing for Node I can just use the user specified one.

Issue #42, opened 2022-08-30
Allow setting ENV and other custom scripts on shell start


LucilleH:
On shell start, I want to set some ENV variables and configure some other things. For example, I want to set some personal aws credentials, or kube configs.

Would be great if I can set a hook to run a script of my choice on `devbox shell`.

Issue #15, opened 2022-08-26
Q: Support for multiple devbox environments?


icosahedron:
Is it possible to have different environments with different shells? I move to a different directory and create a new `devbox.json` but running `devbox shell` still contains the installed packages from the original `devbox.json`.

Perhaps this is a question pertinent only to Rust? I installed Clojure via one of the shells and it didn't show in the other shell.

Edit: Okay, I think I have it sorted, but perhaps some clarification on how it works?

icosahedron:
Rust doesn't seem to work on macOS in a devbox shell.  A simple default app created with `cargo` doesn't link properly.  Not sure what the issue is. Probably on the Rust side?

loreto:
@icosahedron We do support different shells with different environments. I suspect you ran into an issue by accidentally calling `devbox shell` inside `devbox shell`. It's happened to us a few times internally.

We just released a new version of `devbox` that prevents you from making that mistake (shell-inception is what we call it)

Re: rust compilation, I can't tell if that's due to the nested shells  ... or something else. Would you mind making sure you're not in a nested shell, and then trying to compile again? If it doesn't work:
1.  Would you be able to send me the output of `env`? (do scrub any sensitive env variables first) Mainly looking to see if there's something else interfering.
2. Is there a sample rust project you could point me that is representative of what you're trying to do? That way I could try building it myself to see if I can get it to work.

icosahedron:
I too have been accidentally a few shells deep ¯\_(ツ)_/¯, but the rust issue isn't related to that. I've pushed a repo with the little project that I'm working on.

https://codeberg.org/icosahedron/rust-roguelike/src/branch/devbox

It fails linking with the errors
[cargo-build-error.txt](https://github.com/jetpack-io/devbox/files/9436785/cargo-build-error.txt)

And here are the requested environment variables.
[devbox-env.txt](https://github.com/jetpack-io/devbox/files/9436799/devbox-env.txt)

icosahedron:
I should state that I put `rustup` and then used that to install the Rust toolchain from inside `nix shell`. It seems to install correctly, in that `cargo` and `rustc` are in nix directories?

Lagoja:
Hey @icosahedron, I did some digging and testing on my end, and I think the error you posted above is because AppKit framework is not available in your devbox shell. I was able to add it to my devbox by running: 

`devbox add darwin.apple_sdk.frameworks.AppKit`

Can you try that in your project and see if that helps?

[...]

PR #45, opened 2022-08-31
Simple change to support shell hooks

venkytv:
## Summary

A simple change to support custom shell hooks. This is something I often need and currently use [direnv](https://github.com/direnv/direnv) for, but I find it convenient to have the configuration within `devbox.json`.

Fully understand if you decide not to pull this in. :)  This is just a very quick hack, and in general, JSON is not that great a fit for storing multi-line strings. But I find devbox very useful, and this feature makes it a bit more useful for me.

## How was it tested?

go build
go test
devbox init
devbox shell

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -91,7 +91,8 @@ func (d *Devbox) Build(opts ...docker.BuildOptions) error {
 // environment.
 func (d *Devbox) Plan() *planner.BuildPlan {
 	basePlan := &planner.BuildPlan{
-		Packages: d.cfg.Packages,
+		Packages:  d.cfg.Packages,
+		ShellHook: d.cfg.ShellHook,
 	}
 	return planner.MergePlans(basePlan, planner.Plan(d.srcDir))
 }
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,8 +12,6 @@ on:
         default: false
         type: boolean
   # Nightly releases
-  schedule:
-    - cron: "45 8 * * 1-5"
   push:
     tags:
       - "*" # Tags that trigger a new release version
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,7 @@ on:
 
 permissions:
   contents: read
-  pull-requests: read
+  pull-requests: none
 
 jobs:
   golangci-lint:
--- a/config.go
+++ b/config.go
@@ -12,7 +12,8 @@ import (
 type Config struct {
 	// Packages is the slice of Nix packages that devbox makes available in
 	// its environment.
-	Packages []string `cue:"[...string]" json:"packages,omitempty"`
+	Packages  []string `cue:"[...string]" json:"packages,omitempty"`
+	ShellHook string   `cue:"string" json:"shell-hook"`
 }
 
 // ReadConfig reads a devbox config file.
--- a/planner/plan.go
+++ b/planner/plan.go
@@ -13,6 +13,7 @@ import (
 // or whether it should be the same structure as devbox.Config.
 type BuildPlan struct {
 	Packages       []string `cue:"[...string]" json:"packages"`
+	ShellHook      string   `cue:"string" json:"shell-hook,omitempty"`
 	InstallCommand string   `cue:"string" json:"install_command,omitempty"`
 	BuildCommand   string   `cue:"string" json:"build_command,omitempty"`
 	StartCommand   string   `cue:"string" json:"start_command,omitempty"`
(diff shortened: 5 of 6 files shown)


loreto:
@venkytv Thanks for sending.

Could you help me understand the use-case a tiny bit more? An example `devbox.json` using the shell hook would be great.

On the multi-line strings: I've considered letting people pick the format of the config file (i.e. if instead of having a `devbox.json` you had a `devbox.yaml` or a `devbox.toml`, we would detect that and use that file instead). What are your thoughts on that idea?

venkytv:
@loreto Sure. Essentially, a dev environment for us also includes setting a few environment-specific parameters, typically environment variables.  A sample `devbox.json` could look like this:
```
{
  "packages": [
    "python3"
  ],
  "shell-hook": "export MYPARAM1=something\nexport MYPARAM2=42"
}
```

So, a `devbox shell` would launch with these parameters already set for the environment.
```bash
$ devbox shell
$ echo $MYPARAM
something
```

(Just noticed https://github.com/jetpack-io/devbox/issues/42 which is requesting something similar to this.)

I could use direnv for this of course, but having it in `devbox.json` makes it easier to distribute a single file for use across the team.

And yes, YAML especially would be perfect for this as it has multiple ways to represent multiline strings!

By the way, I noticed that I have corrupted the PR with some additional commits I pushed into my fork after I submitted this. Do let me know if you want to include this, and I will clean up the commits and re-submit it.

Thanks again for this very useful tool!

gcurtis:
Thanks @venkytv!

One thing we should be aware of (and somewhat related to #15) is that devbox currently runs two shells:

1. nix-shell always launches bash, which is where the hooks run.
2. From that shell, we launch whatever is in `SHELL` to try and preserve the user's preferred shell. This means that any aliases and unexported variables or functions set in the hooks won't persist into the new shell.

If we implement something like this, we'd probably want to run the hooks inside the 2nd shell (which should be possible after I finish the work in #15) or limit it to exported env vars to avoid confusion.

loreto:
@venkytv One more question for you. I open to the hook idea – I think it can be useful for some advanced use cases.

That said, for setting env variables, I was also thinking that, if there's a `.env` file present in your directory, `devbox` could automatically use it when starting the shell.  Thoughts on that approach vs the shell hook approach for the env variables use-case?

devbox.go
```diff
@@ -91,7 +91,8 @@ func (d *Devbox) Build(opts ...docker.BuildOptions) error {
 // environment.
 func (d *Devbox) Plan() *planner.BuildPlan {
 	basePlan := &planner.BuildPlan{
-		Packages: d.cfg.Packages,
+		Packages:  d.cfg.Packages,
+		ShellHook: d.cfg.ShellHook,
```
LucilleH, line 95: I'm assuming `ShellHook` is only applicable on `devbox shell` command, and not `devbox build`? If so, we don't need to add it to the planner

LucilleH (commented):
I personally really like this idea since I need it too myself, and this is a great way to get some quick ENV and iterm display config setup. Bravo!

I wonder if for multi-lines shell hook, we recommend developers to put them in a `script.sh` and reference the script directly?

devbox.go
loreto: @LucilleH `Plan` is currently called by both `devbox shell` and `devbox build`

LucilleH: Oh so in this case the shell hook will be run in both scenario. Ok that's a bit different from https://github.com/jetpack-io/devbox/issues/42 since that ask is for shell only

LucilleH: @loreto sorry to clarify, the shell hook in this case won’t actually be called in the Dockerfile?

loreto: It won't run because there's nothing that executes it. This behaves as you want. I'm just saying the same struct that includes the hook is passed to both shell and build.

venkytv:
> @venkytv One more question for you. I open to the hook idea – I think it can be useful for some advanced use cases.
> 
> That said, for setting env variables, I was also thinking that, if there's a `.env` file present in your directory, `devbox` could automatically use it when starting the shell. Thoughts on that approach vs the shell hook approach for the env variables use-case?

@loreto That would work too, but I prefer the idea of having one file that contains everything the environment needs. The instructions to use it would be as simple as "download this devbox.yaml file and run `devbox shell`". If the config refers to external files, it becomes a bit more complicated. "Download devbox.json and the right .env file for it, make sure they are in the same directory, and then run `devbox shell`". Just makes things a bit more fragile, IMO.

The same goes for including a separate file and calling that from the `shell-hook`... one more thing which could potentially go out of sync. I'd rather use a YAML format for the config instead and stick the entire shell-hook script inline inside it.

loreto:
@venkytv We're gonna move forward with a feature for shell hooks as you requested. That said, I'm waiting for @gcurtis PR to land, as it might affect how we go about implementing that.

venkytv:
Awesome! Thank you.

loreto:
@venkytv We have a new implementation of this functionality in PR https://github.com/jetpack-io/devbox/pull/93

I'm gonna close this PR in favor of that one – that said, hook functionality coming soon!

PR #48, opened 2022-08-31
[planners] Add Python Poetry detection

mikeland86:
## Summary

* Implements Python detection for Poetry projects https://github.com/jetpack-io/devbox/issues/35
* Add new --debug flag
* Ensure we add stack to docker CMD errors
* Fix path being relative (previously `devbox build some/path` did not work)
* Adds example poetry project
* Use pex to create python executable.
* Specify new prod image that has python.

RFC:
* Add image to state. Currently only implemented for StartStage. Not exported to devbox.json

## How was it tested?

`devbox init examples/python-poetry`

`devbox shell examples/python-poetry`

image

`devbox build examples/python-poetry`

`docker run -p 8080:8080 devbox`

image


loreto:
Thinking about this a bit more: any reason why we can't execute the wheel directly? And that way we get rid of the Prepare Command?

i.e. the build command creates the wheel, the start command executes it.

mikeland86:
@loreto in order to execute a wheel you need a __main__.py file.

loreto:
Oh, I see: it used to be that you could execute `eggs` directly, and I thought `wheels` would have the same functionality ... but I guess not.

It seems like the right way to do this these days is `pex`: https://steve.dignam.xyz/2019/11/17/pex-and-poetry/ (I've run apps built by `pants` and it works. That would let you get rid of the extra entrypoint.

If it's easy enough, I would do that ... if that seems like too much of a change since you already got everything working, then let's just break apart the internal struct use by the planners, vs the struct that's serialized to the `devbox.json` (you can have a shared struct for the fields that are common to both to avoid having to do a lot of copying field by field)

If we don't change the schema `devbox.json` I'm less concerned on what the planners are passing around and we can evolve that without affecting `devbox.json`

Happy to chat tomorrow too.

mikeland86:
Agree something like pex might be ideal. I got it working on mac shell, but on nix I'm getting a permission error even when using a `PEX_ROOT` that nix has access to. It seems like the nix python can't access some directories that pex is creating. Need to look into that.

planner/python_poetry_planner.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+		InstallStage: &Stage{
+			Command: "poetry install --no-dev --no-interaction --no-ansi",
+		},
+		BuildStage: &Stage{
+			Command: "poetry build",
+		},
+		// TODO parse pyproject.toml to get the start command?
+		StartStage: &EntrypointStage{
+			Entrypoint:     "python main.py",
```
LucilleH, line 44: I'm not sure if we should call it `Entrypoint` because often times I actually want the entrypoint to be `/bin/sh` and the start command to be `yarn start`, so that I can sh into the running container easily and test things out. 

I think separating the concept here would help with that. I'd vote for naming this `Command` and use it as an `Entrypoint` for now. But I see that assumption changing later down the line and using the `Command` as `CMD`. Thoughts?

planner/python_poetry_planner.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (g *PythonPoetryPlanner) GetPlan(srcDir string) *Plan {
+	version := g.PythonVersion(srcDir)
+	return &Plan{
+		Packages: []string{
+			fmt.Sprintf("python%s", version.majorMinorConcatenated()),
+			"poetry",
+		},
+		InstallStage: &Stage{
+			Command: "poetry install --no-dev --no-interaction --no-ansi",
```
LucilleH, line 37: Just wondering: does poetry build properly in all scenarios without dev dependencies?

planner/python_poetry_planner.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+		BuildStage: &Stage{
+			Command: "poetry build",
+		},
+		// TODO parse pyproject.toml to get the start command?
+		StartStage: &EntrypointStage{
+			Entrypoint:     "python main.py",
+			Image:          fmt.Sprintf("python:%s-alpine", version.exact()),
+			PrepareCommand: "pip install dist/*.whl",
```
LucilleH, line 46: Sad that this needs to use `pip` but install uses `poetry`. I wish we can have nix runtime as base image so there will be no discrepancy.

planner/plan.go
```diff
@@ -27,11 +27,18 @@ type Plan struct {
...
-	StartStage *Stage `json:"start_stage,omitempty"`
+	StartStage *EntrypointStage `json:"start_stage,omitempty"`
 }
 
 type Stage struct {
 	Command string `cue:"string" json:"command"`
+	Image   string `cue:"string" json:"image"`
+}
+
+type EntrypointStage struct {
```
LucilleH, line 38: `StartStage`? Entrypoint may not be immediately intuitive imo.

Also, can this inherit the `Stage` struct?

planner/python_poetry_planner.go
mikeland86: If we keep calling it command I think we should use `CMD` instead. I think calling it `Command` and using it for `ENTRYPOINT` is confusing.

planner/python_poetry_planner.go
mikeland86: `build` int he poetry context is just creating a wheel so it would. But I switched to using pex so no longer building with poetry.

loreto:
Let me know when ready for review

mikeland86:
@loreto @LucilleH PTAL 

(approach changed a bunch, so you may need to review from scratch)

boxcli/midcobra/debug.go
```diff
@@ -0,0 +1,19 @@
+package midcobra
+
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+type Debug bool
```
loreto, line 9: I think I'd prefer a struct with a bool field.

boxcli/root.go
```diff
@@ -41,6 +43,14 @@ func RootCmd() *cobra.Command {
 	command.AddCommand(RemoveCmd())
 	command.AddCommand(ShellCmd())
 	command.AddCommand(VersionCmd())
+
+	command.PersistentFlags().BoolVar(
```
loreto, line 47: Should we mark this flag as hidden? It tends to pollute help messages because as a persistent flag it'll show up at the end of every command, and most users won't know what to do with the stack trace, unless we're asking for it (at which point we can tell them to re-run with `--debug`)

planner/version.go
```diff
@@ -0,0 +1,55 @@
+package planner
```
loreto, line 1: Could you add
```
// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
// Use of this source code is governed by the license in the LICENSE file.

```

To any new file? (if there's a way we could enforce that via linting, that'd be awesome)

loreto (approved):
LGTM

boxcli/args.go
```diff
@@ -3,13 +3,19 @@
...
 // Functions that help parse arguments
 
 // If args empty, defaults to the current directory
 // Otherwise grabs the path from the first argument
 func pathArg(args []string) string {
 	if len(args) > 0 {
-		return args[0]
+		p, err := filepath.Abs(args[0])
+		if err != nil {
+			panic(err) // What even triggers this?
```
gcurtis, line 16: It can fail if the user doesn't have permissions to read part of the path needed to make it absolute.

boxcli/root.go
gcurtis: Can we actually make this an environment variable? Something like `DEVBOX_DEBUG`? That's what Terraform does, and it's been really convenient since you don't need to remember to pass `--debug` all the time if you're a developer. You just set it once in your shell profile and forget about it.

gcurtis (approved):

planner/plan.go
```diff
@@ -32,6 +32,7 @@ type Plan struct {
 
 type Stage struct {
 	Command string `cue:"string" json:"command"`
+	Image   string `json:"-"`
```
LucilleH, line 35: I prefer removing this in favor of https://github.com/jetpack-io/devbox/pull/52

LucilleH (approved):

gcurtis:
Oh, one more thing. Since we have some temporary panics in there, can we recover somewhere toward the CLI root so we don't dump a stack trace to the user (unless `--debug` is set)?

planner/plan.go
mikeland86: I'll change it once #52 is merged (might be follow up)

PR #49, opened 2022-08-31
[devbox] Add version detection to the go planner

loreto:
## Summary
Add version detection to the go planner based on the contents of `go.mod`. When we don't have a corresponding package in `nixpkgs` we default to `1.19`

## How was it tested?
Compiled and ran against local project.

### DIFF
--- a/planner/go_planner.go
+++ b/planner/go_planner.go
@@ -19,7 +19,7 @@ var versionMap = map[string]string{
 	"1.17": "go_1_17",
 }
 
-const defaultPkg = "go_1_19" // Default for cases where we can't determine a version.
+const defaultPkg = "go_1_19" // Default to "latest" for cases where we can't determine a version.
 
 // GoPlanner implements interface Planner (compile-time check)
 var _ Planner = (*GoPlanner)(nil)
@@ -53,7 +53,7 @@ func (g *GoPlanner) GetPlan(srcDir string) *Plan {
 
 func getGoPackage(srcDir string) string {
 	goModPath := filepath.Join(srcDir, "go.mod")
-	goVersion := getVersion(goModPath)
+	goVersion := parseGoVersion(goModPath)
 	v, ok := versionMap[goVersion]
 	if ok {
 		return v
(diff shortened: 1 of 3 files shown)


RickyGrassmuck:
Awesome!

It's worth noting that if a user adds an additional go package(with a version that's different from the one in the go.mod file) to their devbox config, they will still run into a build error complaining about the conflict. IIRC, Nix has a way to set which package has priority so you can install multiple versions of a package so it may be worth adding a step to the build to handle that.

Not really sure how often someone will want multiple versions of a language in a devbox container but since it's a valid option it's probably worth having something in place to either handle it or warn the user when the scenario is detected.

planner/go_planner.go
```diff
@@ -39,6 +51,31 @@ func (g *GoPlanner) GetPlan(srcDir string) *Plan {
...
 
+func getGoPackage(srcDir string) string {
+	goModPath := filepath.Join(srcDir, "go.mod")
+	goVersion := getVersion(goModPath)
+	v, ok := versionMap[goVersion]
+	if ok {
+		return v
+	} else {
+		// Should we be throwing an error instead, if we don't have a nix package
+		// for the specified version of go?
```
gcurtis, line 62: Might be a good idea if we don't have a newer enough version.

planner/go_planner.go
```diff
@@ -39,6 +51,31 @@ func (g *GoPlanner) GetPlan(srcDir string) *Plan {
...
+		return defaultPkg
+	}
+}
+
+func getVersion(gomodPath string) string {
```
gcurtis, line 67: I think `getVersion` is kind of vague because this lives in the same package as other planners. Maybe `parseGoVersion`?

gcurtis (approved):

planner/go_planner.go
mikeland86: Can also be a method on `GoPlanner` 

python will have similar

loreto: Changing to `parseGoVersion` as requested.

planner/go_planner.go
loreto: Decided to keep: as long as we're updating this to always have the latest version of go, it should work ok. Later we can parse the versions and compare, so we only give the error if the version is greater.

Issue #46, opened 2022-08-31
Shell is not overridden on MacOS


TheGlenn88:
Hi,

I've install Nix, followed by Devbox, upon trying to do a `devbox add nodejs-18_x` followed by a `devbox shell` - I get dropped into a new zsh shell but has no access to node 18, it has access to node 16 from my main env.

Have I missed something or have I stumbled upon a bug? is there anything I can do to debug?

disclaimer: I had previously installed nix about a year back and didn't do too much with it.

Thanks

Lagoja:
Hey @TheGlenn88, we've noticed that in some cases, the Devbox path is not taking precedence over the user's path. We're working to fix this in an upcoming release

loreto:
@gcurtis merged a fixed. It's going out on the next release. Closing.

gcurtis:
@TheGlenn88 this fix is now available in [the most recent 0.0.4 release][1]. If you installed devbox with the install script, then it should auto-update to the new version the next time that you run it.

Thanks for taking the time to report this bug, and feel free to reopen this issue if you're still seeing the wrong version of Node after updating.

[1]: https://github.com/jetpack-io/devbox/releases/tag/0.0.4

TheGlenn88:
Thanks @gcurtis

I’ll test again tomorrow and report my findings. :)

TheGlenn88:
I’ve just managed to test actually and it worked perfectly.

Hats off to the smooth experience as well by the way, no messing around. Good job.

Issue #44, opened 2022-08-30
Feature request: Am I in a shell? a nested shell?


robrich:
Add a discover-inception command.  e.g. `devbox which` or `devbox whereami` outputs if I'm in a shell, and if that's in a shell and if that's in a shell recursively.  Perhaps add a `-v` that outputs the packages / env vars defined in that shell at each level.  If you're not in a shell it notes that too.

loreto:
We prohibited "shell-inception" in the `0.0.3` release.
We're planning to implement https://github.com/jetpack-io/devbox/issues/25

Do you still feel we should have a `which` or `whereami` command once we've done those two things?

Issue #25, opened 2022-08-27
devbox shell gives no indication


bogosj:
When I start a `devbox shell` I see no visual indication that I'm in the shell, but it definitely is working

![image](https://user-images.githubusercontent.com/10186620/187048347-c08c0e47-bda9-46c9-9b3a-d8ca0fdd370a.png)

```
# type lazygit
lazygit is hashed (/nix/store/alnmpxx1njfnv671sgnqxrws1bbari6r-lazygit-0.35/bin/lazygit)

# echo $PS1
\n\d \t\n\e[37;44m[\u@\h:\w]\[\e[00m\] \[\e[91m\]$(parse_git_branch)\[\e[00m\]\n#

# type parse_git_branch
parse_git_branch is a function
parse_git_branch ()
{
    git branch 2> /dev/null | sed -e '/^[^*]/d' -e 's/* \(.*\)/(\1)/'
}
```

loreto:
@bogosj , may I ask what shell you use? What do you see if you run `echo $SHELL`? Is it `bash`?

bogosj:
Yes, bash.

bogosj:
I tried zsh, same situation:

```
export SHELL=/usr/bin/zsh
zsh
bogosj@debian ~ % devbox shell
Installing nix packages. This may take a while...
Starting a devbox shell...
bogosj@debian ~ % type lazygit
lazygit is /nix/store/alnmpxx1njfnv671sgnqxrws1bbari6r-lazygit-0.35/bin/lazygit
bogosj@debian ~ % exit
bogosj@debian ~ % 
```

loreto:
Yeah, we need to fix on our side. I just wanted to know which shell so I can make sure it works for it. I'll take a look this coming week.

acrefoot:
Same issue with the PROMPT not being overridden on Mac :)
```
acrefoot@MacBook ~ % echo $SHELL
/bin/zsh
acrefoot@MacBook ~ % devbox shell
Installing nix packages. This may take a while...
Starting a devbox shell...
Restored session: Mon Aug 29 14:45:47 EDT 2022
acrefoot@MacBook ~ % echo $shell
/nix/store/7k3nqj8bpmpsk5vf8933hvs12q9307vc-bash-5.1-p16/bin/bash
```

loreto:
@gcurtis merged a fix. It's going out on the next release. Closing.

PR #51, opened 2022-09-01
shell,nix: fix shell PATH to prefer nix packages

### LINKED ISSUE #17: Devbox shell PATH should take precedence over existing binaries
Hey!

Just been testing devbox and the experience so far is really smooth. I stumbled upon 3 things that confused me:
1. `devbox add package` failed with `Package not found` until I opened a new terminal. It's kind of documented, but maybe there is a way for a better error? Anyway, that's a nit.
2. `devbox shell` first run can take quite some time and there is no progress bar, but we already have https://github.com/jetpack-io/devbox/issues/7
3. Which is this issue really about.

It seems the initial run of `devbox shell` does not spawn the shell, instead it exists once it completes package initialization. Here's the terminal log that shows the issue:

```sh
lucjansuski@MacBook-Pro-3 ~ % cd /tmp
lucjansuski@MacBook-Pro-3 /tmp % mkdir test
lucjansuski@MacBook-Pro-3 /tmp % cd test
lucjansuski@MacBook-Pro-3 test % devbox init
lucjansuski@MacBook-Pro-3 test % devbox add python
lucjansuski@MacBook-Pro-3 test % devbox shell
this path will be fetched (7.93 MiB download, 49.47 MiB unpacked):
  /nix/store/3j1bsklak6p752nhys18n0p9i3grq1bz-python-2.7.18
copying path '/nix/store/3j1bsklak6p752nhys18n0p9i3grq1bz-python-2.7.18' from 'https://cache.nixos.org'...
Starting a devbox shell...
lucjansuski@MacBook-Pro-3 test % which python
/Users/lucjansuski/.asdf/shims/python
lucjansuski@MacBook-Pro-3 test % devbox shell
Starting a devbox shell...
MacBook-Pro-3:test lucjansuski$ which python
/nix/store/3j1bsklak6p752nhys18n0p9i3grq1bz-python-2.7.18/bin/python
MacBook-Pro-3:test lucjansuski$
```

I'm on `macOS Monterey 12.5`.

gcurtis:
This fixes an issue where the user's rc (~/.bashrc, ~/.zshrc, etc.) files were prepending to the devbox shell PATH. This can cause the shell to unexpectedly run the non-devbox version of programs. It's especially noticeable with some version managers that point the PATH at their shims so they can control which version of a program is run.

The basic steps that result in this issue are:

1. Devbox launches Nix, which sets the PATH to point to the Nix store.
2. From within Nix, we launch the user's default shell.
3. The shell runs the user's init files and stomps on the Nix PATH.

Fixing this is a little tricky because there's no straightforward way to run additional commands _after_ the shell runs the init files. Instead, we need to copy the user's init files and append to them. Doing this requires knowing how each shell handles initialization, so we need to special-case the more common shells by performing the following steps:

1. Attempt to detect the user's default shell via the SHELL env var (we may add more sophisticated detection later).
2. If the shell is recognized as bash, zsh, ksh, dash, ash or sh, copy over the user's corresponding init files to a temp directory and append our own commands to it.
3. Build an `exec` shell command that invokes the new shell and tells it to run our temp init file instead of the user's.

If we can't detect the shell, we fall back to launching a vanilla Nix shell. If we don't recognize the detected shell, we fall back to launching it without overriding the init files.

Finally, since we now have a way of running our own commands in the user's shell, we can change the PS1 prompt to show the user that they're in devbox.

Fixes #17, #25, #44, #46.

### DIFF
--- a/nix/nix.go
+++ b/nix/nix.go
@@ -10,15 +10,59 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+
+	"go.jetpack.io/devbox/shell"
 )
 
 func Shell(path string) error {
-	cmd := exec.Command("nix-shell", path)
-	// Default to the shell already being used.
-	shell := os.Getenv("SHELL")
-	if shell != "" {
-		cmd.Args = append(cmd.Args, "--command", shell)
+	// nix-shell only runs bash, which isn't great if the user has a
+	// different default shell. Here we try to detect what their current
+	// shell is, and then `exec` it to replace the bash process inside
+	// nix-shell.
+	sh, err := shell.Detect()
+	if err != nil {
+		// Fall back to running the vanilla Nix bash shell.
+		return runFallbackShell(path)
 	}
+
+	// Naively running the user's shell has two problems:
+	//
+	// 1. The shell will source the user's rc file and potentially reorder
+	// the PATH. This is especially a problem with some shims that prepend
+	// their own directories to the front of the PATH, replacing the
+	// Nix-installed packages.
+	// 2. If their shell is bash, we end up double-sourcing their ~/.bashrc.
+	// Once when nix-shell launches bash, and again when we exec it.
+	//
+	// To workaround this, first we store the current (outside of devbox)
+	// PATH in ORIGINAL_PATH. Then we run a "pure" nix-shell to prevent it
+	// from sourcing their ~/.bashrc. From inside the nix-shell (but before
+	// launching the user's preferred shell) we store the PATH again in
+	// PURE_NIX_PATH. When we're finally in the user's preferred shell, we
+	// can use these env vars to set the PATH so that Nix packages are up
+	// front, and all of the other programs come after.
+	//
+	// ORIGINAL_PATH is set by sh.StartCommand.
+	// PURE_NIX_PATH is set by the shell hook in shell.nix.tmpl.
+	sh.SetInit(`
+# Update the $PATH so the user can keep using programs that live outside of Nix,
+# but prefer anything installed by Nix.
+export PATH="$PURE_NIX_PATH:$ORIGINAL_PATH"
+
+# Prepend to the prompt to make it clear we're in a devbox shell.
+export PS1="(devbox) $PS1"
+`)
+
+	cmd := exec.Command("nix-shell", path)
+	cmd.Args = append(cmd.Args, "--pure", "--command", sh.ExecCommand())
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func runFallbackShell(path string) error {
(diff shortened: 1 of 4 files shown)


tmpl/shell.nix.tmpl
```diff
@@ -13,10 +13,18 @@ mkShell {
...
       export IN_NIX_SHELL=0
       export DEVBOX_SHELL_ENABLED=1
+
+      # We set PURE_NIX_PATH in case the user's init files in the devbox shell
+      # end up modifying PATH.
+      export PURE_NIX_PATH="$PATH"
+
+      # Make sure we include a basic path for the devbox shell. Otherwise it may
+      # fail to start.
+      export PATH=$PATH:/bin:/usr/bin
```
loreto, line 23: Could we do the MacOS default:

```suggestion
      export PATH=$PATH:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
```

shell/shell.go
```diff
@@ -0,0 +1,144 @@
+// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+// Package shell detects the user's default shell and configures it to run in
+// Devbox.
+package shell
```
loreto, line 6: Totally cool with this package, but just to mention it, another option could be to place `shell.go` inside `nix`. It could make the nix package self-contained and potentially re-usable by others.

You can leave as is too if you prefer.

nix/nix.go
```diff
@@ -10,15 +10,59 @@ import (
...
+	//
+	// ORIGINAL_PATH is set by sh.StartCommand.
+	// PURE_NIX_PATH is set by the shell hook in shell.nix.tmpl.
+	_ = sh.SetInit(`
+# Update the $PATH so the user can keep using programs that live outside of Nix,
+# but prefer anything installed by Nix.
+export PATH="$PURE_NIX_PATH:$ORIGINAL_PATH"
+
+# Prepend to the prompt to make it clear we're in a devbox shell.
+export PS1="(devbox) $PS1"
```
loreto, line 53: Could you add the box emoji too?
```suggestion
export PS1="(📦 devbox) $PS1"
```

Question: Does PS1 work with all the shells?

loreto (approved):
LGTM, see comments below. Make sure to test it well.
On that note: I know you're working on this next, but we should figure how we'll test some of this stuff (and other devbox) functionality automatically. I'm worried we're reaching the point where it's complex enough that it's hard to test all the variations manually.

shell/shell.go
gcurtis: I think I like that idea. I'll try combining the two and see how it works out.

gcurtis: I'll merge this first so we have something working and then see about merging the packages in a followup.

gcurtis: @loreto I merged them in #93 and I think it works really well.

PR #53, opened 2022-09-01
tmpl: fix nil pointer error in Dockerfile.tmpl

gcurtis:
## Summary

Check that each stage is set in the plan before attempting to use its command. This fixes a nil pointer error that occurs when using an empty devbox.json in an unsupported language.

## How was it tested?

Ran `devbox plan` in a Go project with an empty `devbox.json`.

### DIFF
--- a/tmpl/Dockerfile.tmpl
+++ b/tmpl/Dockerfile.tmpl
@@ -26,12 +26,16 @@ RUN --mount=type=cache,target=/nix/store,from=base,source=/nix/store \
 # -----------------------------------------------
 
 COPY --link . ./
+{{ if .InstallStage }}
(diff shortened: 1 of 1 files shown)


loreto (approved):
LGTM

ebauger:
I got this problem on the 0.0.4

```
$ devbox build
Error: template: Dockerfile.tmpl:44:26: executing "Dockerfile.tmpl" at <.StartStage.Image>: nil pointer evaluating *planner.Stage.Image
Usage:
  devbox build [<dir>] [flags]

Flags:
      --engine string   Engine used to build the container: 'docker', 'podman' (default "docker")
  -h, --help            help for build
      --no-cache        Do not use a cache
```

loreto:
@ebauger We just released `0.0.5` and it includes this PR. Hopefully that fixes it!

PR #54, opened 2022-09-02
Do not generate dockerfile on shell command

LucilleH:
## Summary
Do not generate dockerfile on shell command, as someone who is using the shell command may not need to use language detector, or need to fill in `install`, `build` and `start` command if the language planner is absent.

## How was it tested?
devbox shell
devbox generate
devbox build

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -18,13 +18,12 @@ import (
 //go:embed tmpl/* tmpl/.*
 var tmplFS embed.FS
 
-func generate(rootPath string, plan *planner.Plan, includeBuildFiles bool) error {
-	// TODO: we should also generate a .dockerignore file
-	files := []string{".gitignore", "shell.nix", "default.nix"}
-	if includeBuildFiles {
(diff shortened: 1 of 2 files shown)


generate.go
```diff
@@ -18,9 +18,12 @@ import (
 //go:embed tmpl/* tmpl/.*
 var tmplFS embed.FS
 
-func generate(rootPath string, plan *planner.Plan) error {
+func generate(rootPath string, plan *planner.Plan, includeBuildFiles bool) error {
```
loreto, line 21: I don't love the `bool`.

How about this:
you change `generate` to take a string array.

you add `generateNix` (which calls `generate` with `.gitignore`, `shell.nix`, `default.nix`)

you add 'generateDockerfile` (which calls `generate` with `.gitignore`, `Dockerfile`)

loreto (approved):
LGTM

gcurtis (approved):

Issue #18, opened 2022-08-26
Feature Request: Allow alternative OCI image creation


matthewcmead:
It would be great to have pluggable container construction abilities so those on macOS using nerdctl/lima or buildah/krunvm could construct images transparently.

Thanks for a nice tool!

loreto:
Will look into this, other folks requested `Podman` as well.

loreto:
Reposting the podman request from #14:

Podman can be used as a (rootless and rootfull) drop-in replacement for Docker. Some users would like to take advantage of this, mainly on systems where the root filesystem is immutable and/or it is not beneficial to have both Docker and Podman installed.

All that would be required to support Podman afaik would be checking for a podman command and then using that if available (and possibly configurable in a configuration file and/or the CLI.)

Users can additionally work around this limitation of devbox by making either 1. an alias for Docker to pass to Podman, or 2. make a file in /usr/bin/ named docker that redirects to Podman. However, this isn't ideal, and it would most likely be trivial to implement support for Podman in the program itself.

I would love to help to implement this myself, however, I am inexperienced in Go, so I don't want to go tinkering where I shouldn't.

penyuan:
I'm not familiar with nerdctl/lima or buildah/krunvm, but Podman can be run rootless and is known to have some security advantages compared to Docker, so that's another good reason to support it! Can't wait for this.

orowith2os:
Podman support should be fairly trivial, just make it check for that and use it if it's the only method available.

As for how devbox should handle multiple container managers, it might be best to have it set as a CLI flag and/or inside of a configuration file (`$XDG_CONFIG_DIR/devbox/config.yml` or something along those lines), like I suggested in #14.

I'm not knowledgeable on how other container managers work, but having it use a default if available and be configurable might be best.

loreto:
Adding a note that another "engine" we might want to support is `nix`'s own `dockerTools.buildLayeredImage`.

As an update for everyone on this issue, I'm thinking this is implemented via a flag called `--engine`:
- `devbox build --engine=podman` builds using podman
- `devbox build --engine=docker` builds using docker
- `devbox build --engine=nix` builds using nix
- `devbox build` builds using the default. We could make it smart so if you only have one of the tools installed it defaults to the one you have installed.

I'm also thinking this feature would be marked as `experimental` for a few versions of devbox, to gain confidence that it's working well before marking it as a fully supported.

ParetoOptimalDev:
[...]

PR #55, opened 2022-09-02
Add support for podman

loreto:
## Summary
As requested in https://github.com/jetpack-io/devbox/issues/18 this PR adds support for podman via a new `--engine` flag.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -4,11 +4,13 @@
 package docker
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 
 	"github.com/imdario/mergo"
+	"golang.org/x/exp/slices"
 )
 
 // This package provides an API to build images using docker.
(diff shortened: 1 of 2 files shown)


docker/docker.go
```diff
@@ -88,3 +100,11 @@ func isFile(path string) bool {
 
 	return fileInfo.Mode().IsRegular()
 }
+
+func validateFlags(flags *BuildFlags) error {
+	engines := []string{"", "docker", "podman"}
```
gcurtis, line 105: Would we want to let the user specify whatever binary they want?

- Pros:
	- We could handle scenarios where the docker or podman binary is named something else (does this happen with multi-version installs?)
	- It doesn't need to be in the user's PATH.
- Cons:
	- It's easier to make the flag more restrictive at first and loosen it later if we need to.
	- We don't have control over what the build arguments should be, so all the engines would need to support the same exact flags.
	- Easier for the user to make typo's/mistakes that might lead to cryptic errors.

gcurtis (approved):

docker/docker.go
loreto: Yeah, my thought was to start restricted, and if it works well and people need custom binaries, we let it be arbitrary.

PR #57, opened 2022-09-02
shell,nix: set original PATH at start of init file

gcurtis:
## Summary

This fixes a bug that causes the devbox shell's init to fail when the user's rcfile tries to call a command outside of devbox. For example, the following line in a `~/.bashrc` will trigger an error if the devbox doesn't have Go installed:

	export PATH="$(go env GOPATH):$PATH"

This is because we're adding the ORIGINAL_PATH at the end of their shell init script, which means that anything before that only see Nix packages.

Fix this by adding a pre-init hook that lets us restore the PATH before the user's rcfile, while still forcing Nix packages to come first in the post-init hook.

## How was it tested?

Added a command that's only available outside of devbox to my ~/.bashrc and made sure it was called successfully.


gcurtis:
@LucilleH do you mind running this to verify that `devbox shell` is fixed for you?

nix/nix.go
```diff
@@ -44,14 +44,19 @@ func Shell(path string) error {
 	//
 	// ORIGINAL_PATH is set by sh.StartCommand.
 	// PURE_NIX_PATH is set by the shell hook in shell.nix.tmpl.
-	_ = sh.SetInit(`
-# Update the $PATH so the user can keep using programs that live outside of Nix,
-# but prefer anything installed by Nix.
-export PATH="$PURE_NIX_PATH:$ORIGINAL_PATH"
+	sh.PreInitHook = `
+# Update the $PATH so that the user's init script has access to all of their
```
loreto, line 48: Should both the PreInit and PostInit hooks set the path to: `$PURE_NIX_PATH:$ORIGINAL_PATH`?

gcurtis: I just switched both hooks to use the Nix packages.

I wasn't sure what the correct behavior should be:

- Prefer the non-devbox PATH to guarantee compatibility with the user's rcfile.
- Prefer the devbox PATH so that hooks use the Nix packages.

loreto:
@LucilleH are you able to test again?

Lagoja (approved):
Worked when I ran it!

loreto (approved):
LGTM

boxcli/midcobra/debug.go
```diff
@@ -27,17 +28,25 @@ func (d *DebugMiddleware) AttachToFlag(flags *pflag.FlagSet, flagName string) {
...
+func (d *DebugMiddleware) preRun(cmd *cobra.Command, args []string) {
+	if d == nil {
+		return
+	}
 
-func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr error) {
-	if runErr != nil && d.Debug() {
-		log.Printf("Error: %+v\n", runErr)
+	strVal := ""
+	if d.flag.Changed {
```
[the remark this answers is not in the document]
gcurtis: Well that explains it. I'll put up another PR to clean up the debug stuff and make the flag work.

PR #59, opened 2022-09-03
Add discord env variables to release job

loreto:
## Summary
We updated the release process to automatically announce on Discord, but we forgot to pass those secrets as environment variables. As a result, the release process fails and I'm unable to release version `0.0.4`. This PR fixes that.

## How was it tested?
I'll test by trying to release again.


loreto:
Note: since this is blocking the release I'm going to merge with a review, but feel free to review after the fact

PR #64, opened 2022-09-04
[CLI] Add IsBuildable to allow projects to be shell only

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/62

Add new `planner.IsBuildable` which returns false if project is not buildable. It assumes project is relevant.

## How was it tested?

* Created new relevant python project that is not buildable.
image


planner/planner.go
```diff
@@ -3,8 +3,13 @@
 
 package planner
 
+import "go.jetpack.io/devbox/boxcli/usererr"
+
 type Planner interface {
 	Name() string
+	// IsBuildable returns true if the planner can build the project.
+	// It assumes that IsRelevant() has already returned true.
+	IsBuildable(srcDir string) (bool, error)
```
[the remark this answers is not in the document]
mikeland86: Good idea, I can do this. I do think we need to check all stages are missing since some languages wont have a formal build stage (e.g. php does't have build maybe?)

planner/python_poetry_planner.go
```diff
@@ -136,3 +124,17 @@ func getPythonImage(version *version) string {
 	}
 	return fmt.Sprintf("python:%s-slim", version.exact())
 }
+
+func (g *PythonPoetryPlanner) isBuildable(srcDir string) (bool, string) {
+	project := g.PyProject(srcDir)
+	if project == nil {
+		return false, "Project is not buildable: pyproject.toml not found"
```
loreto, line 131: "Project is not buildable" feels to me like the interface of the planner is leaking into the messaging we give the user.

What about:
```
Could not build container for python application.
  pyproject.toml is missing and needed to install python dependencies.
```

planner/planner.go
```diff
@@ -18,23 +20,45 @@ func GetPlan(srcDir string) (*Plan, error) {
...
+		} else if !plan.Buildable() {
+			if plan.buildHint != "" {
+				return false, usererr.New(plan.buildHint)
+			}
+			return false, usererr.New("Unable to build project")
+		}
+		buildables = append(buildables, planner)
+	}
+	if len(buildables) > 1 {
+		return false, usererr.New("Multiple buildable plans found: %v", buildables)
```
loreto, line 51: "Multiple buildables" -> I worry the user won't understand what that is.

I don't love that the user has no way of recovering here (through no fault of their own). Have you seen this happen in practice?

PR #65, opened 2022-09-04
[PHP] Add PHP shell support

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/64

* Adds php planner (shell only)
* Adds php example service

## How was it tested?

image


gcurtis (approved):

README.md
```diff
@@ -142,6 +142,10 @@ Devbox currently detects the following languages:
 - Go
 - Python (Poetry)
 
+Shell only support:
```
loreto, line 145: Technically we let you create a shell for any language, you just need to declare the packages you need in `devbox.json`. So I'm slightly worried that by phrasing these as "shell-only" people will interpret it as, languages *not* in this list are not supported at all.

Thoughts?

PR #69, opened 2022-09-06
Fix README status badge

gcurtis:
## Summary

The tests workflow only runs on PRs, so we were showing the status of whichever PR was built last. Use the release workflow instead, which only runs on main and includes the test workflow.

## How was it tested?

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ### Instant, easy, and predictable shells and containers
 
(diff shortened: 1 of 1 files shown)


loreto:
I'm tempted to just remove it: I've noticed that it renders the badge as 'red' if a PR on a branch fails, even if that branch is *not* merged with main, and main is still "green". Thoughts?

savil:
> I'm tempted to just remove it: I've noticed that it renders the badge as 'red' if a PR on a branch fails, even if that branch is _not_ merged with main, and main is still "green". Thoughts?

@loreto isn't the intent of the PR to make the badge be "green" when main is working well, and avoid "red" status from PRs?

savil (approved):
approving based on my understanding that this PR actually addresses @loreto's concern

gcurtis:
@loreto yeah, that's what this is attempting to fix. It'll show the latest release build status instead. You can also include `?branch=main` if we ever run the release workflow on other branches.

loreto:
Let's include `?branch=main` just to be safe?

PR #71, opened 2022-09-06
Add links to devbox site & docs in readme

robrich:


savil (approved):

Lagoja:
I marked a few places where we need to add trailing slashes, otherwise the URLs won't work. Please add the slashes and then test the URLs to confirm the work, and then G2G!

PR #77, opened 2022-09-07
Add testdata/

loreto:
## Summary
Adds a testdata/ directory intended to store lots of example projects (and edge cases) we should regularly test against. I've gone ahead and added to cases where I know we currently fail but shouldn't.

I debated whether this should be just part of `examples/` but thought a separate directory makes sense since for testing we'll have lots of different variations to test various edge cases. Let me know too whether this will work well with the testing work you have coming up.

## How was it tested?
For now I manually ran devbox against the testdata projects (but it's failing).


gcurtis (approved):
Sounds good to me!

PR #78, opened 2022-09-07
Telemetry should send the failure message

loreto:
## Summary
Telemetry should send the failure message whenever there's an error. That will allow us to understand what the errors are and how to fix them.

## How was it tested?
Development version of segment.

### DIFF
--- a/boxcli/midcobra/telemetry.go
+++ b/boxcli/midcobra/telemetry.go
@@ -59,7 +59,11 @@ func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr 
 		return
 	}
 
-	segmentClient, _ := segment.NewWithConfig(m.opts.TelemetryKey, segment.Config{Verbose: false})
+	segmentClient, _ := segment.NewWithConfig(m.opts.TelemetryKey,
+		segment.Config{
+			Verbose:   false,
+			BatchSize: 1, // We're in a short lived program, so batching doesn't make sense
+		})
 
 	defer func() {
 		_ = segmentClient.Close()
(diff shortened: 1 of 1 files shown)


boxcli/midcobra/telemetry.go
```diff
@@ -70,15 +74,19 @@ func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr
...
 		Command:     subcmd.CommandPath(),
 		CommandArgs: subargs,
 		DeviceID:    deviceID(),
 		Duration:    time.Since(m.startTime),
-		Failed:      runErr != nil,
-	})
+	}
+	if runErr != nil {
+		evt.Failed = true
+		evt.FailedMsg = runErr.Error()
```
gcurtis, line 87: It might be a good idea to only track a distinct error type so that we can redact errors for analytics. Otherwise I worry that we might return an error with sensitive info somewhere else in the code and accidentally log it.

Maybe something like this?

```go
type RedactedError interface {
	Redact() string
}

func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr error) {
	// ...

	if runErr != nil {
		evt.Failed = true
		if redactErr, ok := runErr.(RedactedError); ok {
			evt.FailedMsg = redactErr.Redact()
		}
	}
}
```

loreto: I like the idea, but I'm also worried this means we will *not* capture a bunch of errors as a result. I'm trying to think if there's an in-between.

gcurtis: Another option is to do redacted format strings.

```go
// Not sure on the naming or if it should go in a new package.
type SafeError struct {
	format string
	args   []any
}

func (e *SafeError) Redacted() string {
	// This is ugly. You could probably do something fancier by substituting
	// the args with something else of the same type. Or use a template string
	// instead of fmt.
	return format
}

func (e *SafeError) Error() string {
	return fmt.Sprintf(e.format, e.args...)
}
```

loreto:
Abandoning in favor of Lucille's Sentry implementation.

PR #79, opened 2022-09-07
Packages should be unique in plan

LucilleH:
## Summary
Right now, two of the same packages can appear in the list because we do not dedup.

Before:
Screen Shot 2022-09-07 at 9 54 02 AM

After:
Screen Shot 2022-09-07 at 10 21 44 AM

## How was it tested?
devbox plan


loreto (approved):
LGTM

utils/slice.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+// Package devbox creates isolated development environments.
+package utils
```
gcurtis, line 5: Could we name this something like `pkgslice`?

```suggestion
// Package pkgslice manipulates slices of devbox package names.
package pkgslice
```

utils/slice_test.go
```diff
@@ -1,4 +1,8 @@
-package devbox
+// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+// Package devbox creates isolated development environments.
```
gcurtis, line 4: We only need to put the package doc comment in one file.

```suggestion
```

gcurtis (approved):
What do you think of doing the deduplication whenever we (un)marshal the config? It might help ensure we don't forget to do it somewhere.

LucilleH:
@gcurtis this dedup happens on merging the plans, not on unmarshal configs. The config could only have one `nodejs` package, but the plan may merge another `nodejs` in there. Thoughts?

gcurtis:
Ah, I see. I thought we were checking if a package already existed in the slice when merging, but that only happens in `Devbox.Add`. Never mind then!

gcurtis (approved):

PR #83, opened 2022-09-07
boxcli: add basic tests for devbox shell

gcurtis:
## Summary

Add tests that check that `devbox shell` launches successfully on the platforms (Linux and macOS) and shells (bash, dash, and zsh) that we support. Specifically, each test runs the equivalent of typing the following into your terminal:

	cd `mktemp -d`
	devbox init
	devbox add hello
	devbox shell
	hello
	exit
	exit

The bulk of this change is the addition of some test helpers that handle launching shells and interacting with them. For now they live in the same file as the `boxcli` shell tests, but we can move them into their own `shelltest` package later on if necessary. With these test helpers in place, it should be a lot easier to add tests for more complicated scenarios.

In order to run these tests in CI, the GitHub test jobs now install Nix along with some additional shells. They also run on both ubuntu-latest and macos-12 to make sure things work on both systems. The lint job also runs on Ubuntu and macOS to make sure that we don't inadvertently add platform-specific code that won't compile.

## How was it tested?

`go test`

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,11 +16,13 @@ permissions:
 
 jobs:
   golangci-lint:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-12]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
-      - name: Set up go
-        uses: actions/setup-go@v3
+      - uses: actions/setup-go@v3
         with:
           go-version-file: ./go.mod
           cache: false # use golangci cache instead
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yaml
```diff
@@ -41,3 +42,33 @@ jobs:
...
+        env:
+          HOMEBREW_NO_INSTALL_CLEANUP: 1
+          HOMEBREW_NO_ENV_HINTS: 1
+          HOMEBREW_NO_EMOJI: 1
+          HOMEBREW_NO_ANALYTICS: 1
+        run: |
+          brew update
+          brew install dash zsh
+      - name: Install Nix
+        run: sh <(curl -L https://nixos.org/nix/install) --daemon
```
LucilleH, line 65: Does nix get installed every single time the test runs? Can we cache this somehow

LucilleH: I ask because if we install nix every single time, devbox shell would be quite slow given we have to install the nix store?

loreto: Instead of installing nix, could we use a base image that has nix pre-installed? (for example, the same one we use in the dockerfile)

loreto: Oh, nvm, I see you're purposefully trying to test macos.

gcurtis: Installing Nix does slow things down a lot. With these tests the total CI time is about 5 minutes.

I want to look into caching the Nix and Homebrew packages with the cache action, but as a first step I wanted to get the tests working.

boxcli/shell_test.go
```diff
@@ -0,0 +1,286 @@
...
+	out = sh.devboxIO.read(t)
+	want := "Hello, world!"
+	if out != "Hello, world!" {
+		t.Errorf("Got hello command output %q, want %q.", out, want)
+	}
+}
+
+func TestShellHelloBash(t *testing.T) { testShellHello(t, "bash") }
+func TestShellHelloDash(t *testing.T) { testShellHello(t, "dash") }
+func TestShellHelloZsh(t *testing.T)  { testShellHello(t, "zsh") }
```
loreto, line 47: If we wanted to be really exhaustive, this could be the list:
- Bash
- Cmd
- Dash
- Elvish
- Fish
- Ion
- Nushell
- PowerShell
- Tcsh
- Xonsh
- Zsh

This is based on the list that `starship` works with: https://starship.rs/guide/#%F0%9F%9A%80-installation

loreto (approved):
LGTM

boxcli/shell_test.go
gcurtis: Yeah, there are a bunch more we could support. It's a bit of work because we have to implement the detection and using a custom shellrc. I should probably add some documentation on which shells we officially test and support.

PR #84, opened 2022-09-08
[python] Improve python poetry planner to auto-detect main module

mikeland86:
## Summary

This PR allows the poetry planner to detect and use `__main__` module. It does this by

1) Using the pypoetry.toml and auto-detect directories to try and find the `__main__.py`
2) It uses an inline python script in the build phase to determine the correct flag to pass to pex. (This part is a bit hacky).

I'm not 100% sure this will work in all cases, but it works in more cases than the previous implementation. 

Unrelated fix: fixed how we combine errors when plan fails.

cc: @Lagoja 

## How was it tested?

* `devbox build` with console script only
* `devbox build` with `__main__` module only
* `devbox build` without scripts or `__main__` module (failed, showed correct error).


loreto:
Could you add an appropriate test case in `testdata/python`?

mikeland86:
@loreto added a bunch

loreto (approved):
LGTM

PR #87, opened 2022-09-08
Add Node planner test and additional checks for planner fields

LucilleH:
## Summary
Add Node planner test and additional checks for planner fields

## How was it tested?
go test ./...


testdata/php/php8.1/plan.json
```diff
@@ -1,17 +1,25 @@
...
-  ],
-  "install_stage": {
-    "command": "composer install --no-dev --no-ansi"
-  },
-  "build_stage": {},
-  "start_stage": {
-    "command": "php -S 0.0.0.0:8080 -t public"
-  }
-}
+    "php81Packages.composer"
```
LucilleH, line 23: @mikeland86 I noticed this has changed. Is that expected?

mikeland86: yeah, we now support php building.

mikeland86:
ugh, I think we both did the same work?

LucilleH:
> ugh, I think we both did the same work?

That's ok 😄

PR #89, opened 2022-09-08
[planner] Optimize nodejs final image size

LucilleH:
## Summary
Optimize nodejs final image size by getting rid of dev packages in the final image.
I'm open to a better schema if postCommandHook is not the best way to go about it.

Tried it on a fairly complex NodeJS web app:
- size before: 837 MB
- size after (with yarn): 519 MB
- size after (with npm): 515 MB

Sample planner output with yarn:
```
{
  "install_stage": {
    "command": "yarn install",
    "input_files": [
      "package.json",
      "yarn.lock"
    ]
  },
  "build_stage": {
    "command": "yarn build && yarn install --production --ignore-scripts --prefer-offline",
    "input_files": [
      "."
    ],
  },
  "start_stage": {
    "command": "yarn start",
    "input_files": [
      "."
    ]
  },
  "dev_packages": [
    "nodejs",
    "yarn"
  ],
  "runtime_packages": [
    "nodejs",
    "yarn"
  ]
}
```

Sample planner output with npm:
```
"install_stage": {
    "command": "npm install",
    "input_files": [
      "package.json"
    ]
  },
  "build_stage": {
    "command": "npm prune --production",
    "input_files": [
      "."
    ]
  },
  "start_stage": {
    "command": "node index.js",
    "input_files": [
      "."
    ]
  },
  "dev_packages": [
    "nodejs"
  ],
  "runtime_packages": [
    "nodejs"
  ]
```

## How was it tested?
`devbox plan`
`devbox build`


mikeland86 (approved):
This is great!

Should we think about how to turn all of these extra docker steps into a framework? Something that can be exposed to users so they can do it themselves?

loreto:
Open to the current implementation.

Although remind me why we couldn't just do:
```
  "build_stage": {
    "command": "yarn build && yarn install --production --ignore-scripts --prefer-offline",
    "input_files": [
      "."
    ],
  },
 ```
 
 And:
 ```
   "build_stage": {
    "command": "npm prune --production",
    "input_files": [
      "."
    ],
  },
 ```

LucilleH:
@loreto because build command can be set and overwritten by the user (by default it is empty, but some apps require it). I feel like modifying what a user has set in `devbox.json` by appending `npm prune --production` is not great. Thoughts?

loreto:
Got it. So, I think we need to first sort out how we imagine devbox.json and planners interacting. I'm very on the fence about how they should interact with each other, and worry about edge cases / unexpected behavior.

Scenario 1:
If the user should be able to modify the `command`, and the planner should *still* work and minimize the image regardless of a custom command, I see the value of this change.

Scenario 2:
If when the user modifies the `command` it becomes _their_ responsibility to clean up. Then it's their choice (or not) to run a command that cleans up. I do think most users will get the cleanup wrong ... but if most users use the planner, and customization is only for advanced use cases, that might be ok.

Rather than reviewing this PR, should we figure out the expected interaction between planners and `devbox.json`? Once we've defined that, reviewing this PR should be easy.

LucilleH:
@loreto we can hold off until we figure out what the expected behavior should be between devbox config and planners 👍 

My initial thoughts are that for Node specifically,
1. Customization on the build command is not just for advanced use cases, as the number of apps that requires that step can be quite significant, but the exact command can also be anything. 
2. Most users will get the cleanup wrong. No doubt about that.

mikeland86:
@LucilleH @loreto middle ground:

command should be a slice (Commands). 

one thing that bugs me from devbox Dockerfiles is that because we try to condense all the logic into a single command the dockerfiles become unreadable. If instead each stage has 1 or more commands then we can implement this in a nicer way without dirtying the interface. 

(If we want to be clever, the json value of commands could be a string or an array) 

if the user replaces the commands, then that's up to them. But this could also make it easier for people to adapt generated plans.

and by the way my 2 cents on modifying the start command: Should we be asking the user to modify their package.json instead?  This is what we do in python and PHP. If they don't follow best practices, we simply tell them what to do. If they have some crazy constraint that prevents them from doing what they need to do, they can enter their plans manually, but that should not be the main way to do it.

mikeland86:
@LucilleH to unblock this (since it's a good improvement) I think you should do @loreto's suggestion and inline it. (This would be similar to the approach taken in python).

After/if we decide to make planner/json changes we can revisit?

loreto:
> @LucilleH to unblock this (since it's a good improvement) I think you should do @loreto's suggestion and inline it. (This would be similar to the approach taken in python).
> 
> After/if we decide to make planner/json changes we can revisit?

I like this. Makes the image smaller now, then we sort out an improved interface in a second pass.

LucilleH:
@mikeland86 @loreto I will go ahead and inline it. Please chime in if you think it should be otherwise!

loreto (approved):
LGTM

PR #93, opened 2022-09-08
nix: allow users to set shell startup hooks

gcurtis:
## Summary

Add a `shell.init_hook` field to devbox.json which can contain arbitrary commands to run at shell startup.

	{
	  "packages": ["go"],
	  "shell": {
	    "init_hook": "echo hello from a hook!"
	  }
	}

This change also merges the `shell` package into `nix`, making it clearer that it handles Nix shells specifically. The `nix.Shell` function is replaced by the `nix.Shell` struct, which does everything that the `shell.Shell` struct did.

The new API looks something like:

	sh, err := nix.DetectShell()
	sh.UserInitHook = cfg.Shell.InitHook
	sh.Run(nixPath)

Finally, we now build the devbox shellrc with a template since it was getting a bit complex.

## How was it tested?

Adding a hook and running a shell.


gcurtis:
Obviously also open to suggestions on how the `devbox.json` should look.

config.go
```diff
@@ -16,6 +16,12 @@ type Config struct {
 	// Packages is the slice of Nix packages that devbox makes available in
 	// its environment.
 	Packages []string `cue:"[...string]" json:"packages"`
+
+	// Shell configures the devbox shell environment.
+	Shell struct {
+		// InitHook contains commands that will run at shell startup.
+		InitHook string `json:"init_hook,omitempty"`
```
Lagoja, line 23: A few quick questions about this format: 
* Are there more non-hook fields that we foresee adding to the `shell` struct? 
* Should `stages` be in a `Docker` struct or something else, so that it's clearer they only affect Docker and hooks only affect shells?

LucilleH: Not a fan of the word `Docker` or `Container` in general. Feels like we are exposing implementation details. Some fields are for `shell`, others for `build`. At least that's how I've been thinking about them.

config.go
```diff
@@ -16,6 +16,12 @@ type Config struct {
 	// Packages is the slice of Nix packages that devbox makes available in
 	// its environment.
 	Packages []string `cue:"[...string]" json:"packages"`
+
+	// Shell configures the devbox shell environment.
+	Shell struct {
```
loreto, line 21: In terms of schema, I'm tempted to keep it flat for the time being and just make it `shell_hook` at the top-level (I might regret that later if we need more things inside shell, hehe)

loreto (approved):
LGTM, let's just agree on the right schema for `devbox.json`

nix/shell.go
```diff
@@ -0,0 +1,181 @@
...
+	case "bash":
+		sh.name = shBash
+		sh.userShellrcPath = rcfilePath(".bashrc")
+	case "zsh":
+		sh.name = shZsh
+		sh.userShellrcPath = rcfilePath(".zshrc")
+	case "ksh":
+		sh.name = shKsh
+		sh.userShellrcPath = rcfilePath(".kshrc")
+	case "dash", "ash", "sh":
```
savil, line 68: ah, `dash` is here!

config.go
gcurtis: For shell, we discussed possibly letting users pick which shell runs in devbox. I was thinking it might become something like:

```json
{
  "packages": ["go"],
  "shell": {
    "type": "zsh",
    "init_hook": "echo hello from a hook!"
  }
}
```

I'm not sure what we'd group the other fields under. They seem general enough that I think they're fine where they are. No strong opinion on that though.

nix/shell.go
```diff
@@ -0,0 +1,181 @@
...
+func (s *Shell) Run(nixPath string) error {
+	// Launch a fallback shell if we couldn't find the path to the user's
+	// default shell.
+	if s.binPath == "" {
+		cmd := exec.Command("nix-shell", nixPath)
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		debug.Log("Unrecognized user shell, falling back to: %v", cmd.Args)
```
savil, line 105: not sure: 
should the argument be `cmd.Args` - or `nix-shell nixPath` ?

nix/shell.go
```diff
@@ -0,0 +1,181 @@
...
+	// We need a temp dir (as opposed to a temp file) because zsh uses
+	// ZDOTDIR to point to a new directory containing the .zshrc.
+	tmp, err := os.MkdirTemp("", "devbox")
+	if err != nil {
+		return "", fmt.Errorf("create temp dir for shell init file: %v", err)
+	}
+
+	// This is a best-effort to include the user's existing shellrc. If we
+	// can't read it, then just omit it from the devbox shellrc.
+	userShellrc, _ := os.ReadFile(userShellrcPath)
```
savil, line 152: maybe only ignore error for `os.IsNotExist(err)`?
and set userShellrc to be empty-string in that case?

savil (approved):
nice! 

I think I could use this in the rust planner in #97: I need to call `rustup default <version>` after the shell starts....

devbox.go
```diff
@@ -130,7 +130,13 @@ func (d *Devbox) Shell() error {
 		return errors.WithStack(err)
 	}
 	nixDir := filepath.Join(d.srcDir, ".devbox/gen/shell.nix")
-	return nix.Shell(nixDir)
+	sh, err := nix.DetectShell()
+	if err != nil {
+		// Fall back to using a plain Nix shell.
+		sh = &nix.Shell{}
+	}
+	sh.UserInitHook = d.cfg.Shell.InitHook
```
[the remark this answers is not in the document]
gcurtis: If we eventually do, I'd want to flip it and put the config -> shell conversion in the calling package. I like Daniel's idea of keeping the `nix` package fairly isolated so it's usable by others.

nix/shell.go
gcurtis: Yep! dash and a bunch of shells fall under the "posix" camp where they only support initialization by putting a path to a file in the `ENV` environment variable. That's why there isn't an explicit dash enum.

nix/shell.go
gcurtis: They should be roughly the same. This will print something like `["nix-shell" "/blah/.devbox/gen"]`. I wanted to use `cmd.Args` to guarantee we're printing exactly what's being launched.

nix/shell.go
gcurtis: I think you're right here. It might be possible to return an error and a partially read file. I'll update it to explicitly set `userShellrc` to empty if there's an error so we don't risk running part of the user's shellrc.

venkytv:
Awesome! Thank you!

PR #92, opened 2022-09-08
[cuecfg] add toml support

savil:
## Summary

We need this to parse Cargo.toml files.

I chose this  library: [github.com/pelletier/go-toml](https://github.com/pelletier/go-toml). 
 - The alternative is https://github.com/BurntSushi/toml. This one has more github stars than `go-toml`.
 - `go-toml`'s interface is more similar to `encoding/json`, and is also used by `viper`.

Also, changed `v` to `value` in `cuecfg.go` to address linter complaint.

## How was it tested?

`go test ./cuecfg/...`

PR #97, opened 2022-09-09
[Rust Planner] part 1: devbox shell

savil:
## Summary

This PR adds basic support for Rust, focussing on `devbox shell`. The next PR will support `devbox build`.

**Rust nix package**
This PR uses the [Oxalica/rust overlay](https://github.com/oxalica/rust-overlay). This has prebuilt binaries which can be installed for the rust toolchain.  This is an improvement over mozilla's rust-overlay, but that involves compiling the sources upon installing the nix packages, which is not optimal.

The [nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#rust) suggests using a community overlay when one needs to pick different rust versions (as we do). There's also [enthusiasm](https://gutier.io/post/development-using-rust-with-nix/) for this approach in the community.

An alternate approach is to use the standard rust toolchain approach. This involves installing `rustup` and doing `rustup default <version>` to change versions. The downside of this is that we lose the benefits of the nix-store: The specific rust version's `rustc` and `cargo` binaries are dynamically downloaded. 

**Implementation**
As part of this, I needed to install support for overlays in the generated `shell.nix`

## How was it tested?

adding cases to `testdata/` for rust `stable` (currently, version `1.63.0`) and version `1.62.0`

**devbox shell**

```
> go install ./cmd/install

> devbox shell
> rustc --version
rustc 1.63.0 (4b91a6ea7 2022-08-08)



> cd testdata/rust/rust-1.62.0
> devbox shell
Starting a devbox shell...
info: using existing install for '1.62.0-x86_64-apple-darwin'
info: default toolchain set to '1.62.0-x86_64-apple-darwin'

  1.62.0-x86_64-apple-darwin unchanged - rustc 1.62.0 (a8314ef7d 2022-06-27)

(devbox)
> rustc --version
rustc 1.62.0 (a8314ef7d 2022-06-27)
(devbox)
```


savil:
* main
  * **PR #92** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/92" target="_blank">Graphite</a> 
    * **PR #97** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/97" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/97?utm_source=stack-comment).

planner/languages/rust/rust_planner.go
```diff
@@ -4,9 +4,16 @@
...
 import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/cuecfg"
 	"go.jetpack.io/devbox/planner/plansdk"
 )
 
+const cargoToml = "Cargo.toml"
```
loreto, line 15: Is it always capitalized?

planner/languages/rust/rust_planner.go
```diff
@@ -17,9 +24,82 @@ func (p *Planner) Name() string {
...
-	return false
+	cargoTomlPath := filepath.Join(srcDir, cargoToml)
+	return plansdk.FileExists(cargoTomlPath)
 }
 
 func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
-	return &plansdk.Plan{}
+	plan, err := p.getPlan(srcDir)
+	if err != nil {
+		fmt.Printf("[WARNING] error in RustPlanner.GetPlan: %s", err)
```
loreto, line 34: Don't print things inside a planner. If you want, you can pass a plan error

planner/languages/rust/rust_planner.go
```diff
@@ -17,9 +24,82 @@ func (p *Planner) Name() string {
...
+	// including them by default, but this could be likely optimized once we understand better.
+	//
+	// libiconv due to error:
+	//     ld: library not found for -liconv. clang-11: error: linker command failed with exit code 1
+	// https://discourse.nixos.org/t/nix-shell-rust-hello-world-ld-linkage-issue/17381/2
+	//
+	// gcc due to error:
+	//     linker `cc` not found
+	// https://github.com/NixOS/nixpkgs/issues/103642
+	packages := []string{"rustup", "libiconv", "gcc"}
```
loreto, line 52: Does rust by itself not pull the dependencies it needs?

planner/languages/rust/rust_planner.go
```diff
@@ -17,9 +24,82 @@ func (p *Planner) Name() string {
...
+		rustupVersion = version.Exact()
+	}
+
+	rustupDefaultCmd := fmt.Sprintf("rustup default %s", rustupVersion)
+
+	return &plansdk.Plan{
+		DevPackages:     packages,
+		RuntimePackages: packages,
+		Shell: plansdk.PlanShell{
+			InitHook: rustupDefaultCmd,
```
loreto, line 69: I don't love that the planner needs to use the shell hook – it feels like we're having to go outside the nix infrastructure to set the right version. Is there really no way to do it by specifying the right package instead of calling `rustup`?

planner/languages/rust/rust_planner.go
```diff
@@ -17,9 +24,82 @@ func (p *Planner) Name() string {
...
+	}, nil
+}
+
+func (p *Planner) rustVersion(srcDir string) (*plansdk.Version, error) {
+	cfg, err := p.cargoManifest(srcDir)
+	if err != nil {
+		return nil, err
+	}
+	if cfg.PackageField.RustVersion == "" {
+		return nil, nil
```
loreto, line 80: Should you not default to the latest version, or something like that when you can't figure it out?

planner/languages/rust/rust_planner.go
savil: Ah, I hadn't seen PlanErrors. Will use that.

planner/languages/rust/rust_planner.go
savil: 🤷🏾‍♂️ it doesn't seem to? This is one reason why I added a section to the PR summary about possibly needing more sophisticated approach to the nix-package.

planner/languages/rust/rust_planner.go
savil: I default it to `stable` in the caller, but I can't do that here since `plansdk.Version` needs a semver version.

savil: Alternatively, I could push that logic inside here, and make this return a `string` value

planner/languages/rust/rust_planner.go
savil: I thought it was https://github.com/rust-lang/cargo/issues/45
but when I tried `cargo build` on a `cargo.toml` project, it did work, so updated code to consider both.

planner/languages/rust/rust_planner.go
savil: I couldn't find a `rustup_<version>` package. I also feel this is odd, but not quite sure the best alternative fix for now...

savil:
Spoke to @loreto. May explore using the Oxalica overlay instead.

savil:
Moving back to draft until ready for re-review.

planner/languages/rust/rust_planner.go
```diff
@@ -17,9 +29,87 @@ func (p *Planner) Name() string {
...
+		return nil, errors.WithStack(err)
+	}
+
+	rustPkgDev := fmt.Sprintf("rust-bin.stable.%s.default", rustVersion)
+	rustPkgRuntime := fmt.Sprintf("rust-bin.stable.%s.minimal", rustVersion)
+
+	return &plansdk.Plan{
+		Overlays:        []string{RustOxalicaOverlay},
+		DevPackages:     []string{rustPkgDev},
+		RuntimePackages: []string{rustPkgRuntime},
```
savil, line 59: This RuntimePackages should be part of the next PR. I can drop it on my next update.

 I'll also look into whether we can avoid it entirely, similar to the go planner not needing any runtime packages.

planner/languages/rust/rust_planner.go
savil: Actually, `cargo` does complain about `cargo.toml`. If I do:
```
 > [builder 6/6] RUN --mount=type=cache,target=/nix/store,from=base,source=/nix/store cargo build --release:
#28 0.323 + cargo build --release
#28 0.343 error: could not find `Cargo.toml` in `/scratch` or any parent directory, but found cargo.toml please try to rename it to Cargo.toml
```

I think our code can support it since `cargo build` works, but an FYI

testdata/rust/rust-1.62.0/plan.json
```diff
@@ -0,0 +1,24 @@
...
+    "command": ""
+  },
+  "start_stage": {
+    "command": "",
+    "input_files": [
+      "."
+    ]
+  },
+  "nix_overlays": [
+    "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"
```
loreto, line 18: Would it be better to pin to a specific version of the overlay, instead of doing master? Or because the overlay provides such granular control on which version of rust to install, it's actually good to just point at master (so we get any latest version of rust automatically)

testdata/rust/rust-stable/src/main.rs
```diff
@@ -0,0 +1,14 @@
+use ferris_says::say;
+
+use std::io::{stdout, BufWriter};
+
+fn main() {
```
loreto, line 5: Would it be easy to change these to be closer to the go examples (check version via API, error out if the version is not the one the example expects)

testdata/rust/rust-1.62.0/plan.json
savil: This is a good question. I spent some time reading the docs. Here's my mental model:

- The `oxalica/rust-overlay` updates daily the Rust-toolchain binaries it references..
- The specific code of the `oxalica/rust-overlay` evolves with each commit, so we should _not_ use `master` but instead pull from the `stable` tag's commit. This would be a soft-pin.
- I don't think we want to do a hard-pin to a specific commit, because projects which don't specify a Rust version would expect to get the newest stable Rust version. By pulling from `oxalica/rust-overlay:stable` tag, that code will get updated over time to point to newer versions of Rust.

In summary, I'll update this to use `https://github.com/oxalica/rust-overlay/archive/stable.tar.gz`

testdata/rust/rust-stable/src/main.rs
savil: sure, I can look into doing that.

testdata/rust/rust-1.62.0/plan.json
savil: One more bullet point:
- for projects that want a specific version of Rust, they include `rust-version` in their `Cargo.toml`. And we continue to respect that in the form of the actual nix-package that is pulled in.

loreto (approved):
LGTM

But fix failing test.

PR #98, opened 2022-09-09
Add dummy planners for a bunch of languages

loreto:
## Summary
Add dummy planners for a bunch of languages; for now they are placeholders that don't do anything, but it sets up the whole structure so we can easily start editing them (or the community can contribute by finding the right file).

In the process refactored the common types and utility functions used by all planners into a `plansdk`.

Would appreciate a quick review since I'm fearful of constantly running into merge conflicts otherwise (I'm moving the location of the planners everyone is working on)

## How was it tested?
`go test -v ./...`


planner/empty/empty_planner.go
```diff
@@ -0,0 +1,26 @@
...
+package empty
+
+import "go.jetpack.io/devbox/planner/plansdk"
+
+type Planner struct{}
+
+// Implements interface Planner (compile-time check)
+var _ plansdk.Planner = (*Planner)(nil)
+
+func (g *Planner) Name() string {
```
LucilleH, line 16: Can it be `(p *Planner)`?

LucilleH (approved):

PR #99, opened 2022-09-09
Add Analyzer framework

loreto:
## Summary
Adds an `Analyzer` to the `plansdk`. Analyzers can be initialized on a root directory, and provide lots of utility functions to then understand (or analyze) the source code present in that root directory. Analyzer implements several functions that are really handy when implementing all sorts of planners.

This PR just introduces the library; future PRs would start using the library in the implementation of planners.

## How was it tested?
`go test -v ./...`


planner/plansdk/analyzer.go
```diff
@@ -0,0 +1,73 @@
...
+// Handy when implementing new Planners that need to analyze files in order
+// to determine what to do.
+type Analyzer struct {
+	rootDir string
+}
+
+func NewAnalyzer(rootDir string) *Analyzer {
+	abs, err := filepath.Abs(rootDir)
+	if err != nil {
+		abs = rootDir
```
savil, line 24: 🤔 why not error in this case? if a rootDir cannot be converted to an absolute-dir here, then we may run into issues later on.

planner/plansdk/analyzer.go
```diff
@@ -0,0 +1,73 @@
...
+		return path
+	}
+
+	return filepath.Join(a.rootDir, path)
+}
+
+// GlobFiles returns all the files matching the given glob patterns.
+// Patterns can be relative to the analyzer's root directory. Glob patterns
+// support "double star" matches.
+func (a *Analyzer) GlobFiles(patterns ...string) []string {
```
savil, line 46: private function?

savil (approved):

planner/plansdk/analyzer.go
loreto: I think this one might be used publicly

planner/plansdk/analyzer.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return &Analyzer{
+		rootDir: abs,
+	}
+}
+
+// AbsPath resolves the given path and turns it into an absolute path relative
+// to the root directory of the analyzer. If the given path is already absolute
+// it leaves it as is.
+func (a *Analyzer) AbsPath(path string) string {
```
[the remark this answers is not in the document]
loreto: I'll make it private. I thought someone might want to use it externally, but I can start private and if someone wants they can switch to public.

PR #103, opened 2022-09-09
Change the merging logic of devbox.json and planners

LucilleH:
## Summary
We have decided on the following merging logic between `devbox.json` and `planners`:

- devbox.json `packages` field will merge with planners `devPackages` and `runtimePackages`. Merge will be `appendWithSlice`.
- devbox.json `start_stage`, `build_stage` and `install_stage` subfields (shared plan):
  - if empty, will inherit the corresponding planners fields
  - if set, will override corresponding planners fields

TODO (Will follow up with separate PRs)
- devbox.json schema would be a separate one from the planner

## How was it tested?
go build -o devbox cmd/devbox/main.go
devbox plan

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -106,7 +106,12 @@ func (d *Devbox) Plan() *plansdk.Plan {
 		RuntimePackages: d.cfg.Packages,
 		SharedPlan:      d.cfg.SharedPlan,
 	}
-	return plansdk.MergePlans(basePlan, planner.GetPlan(d.srcDir))
+	plannerPlan := planner.GetPlan(d.srcDir)
+
+	plan := plansdk.MergePlans(basePlan, plannerPlan)
+	plan.SharedPlan = plansdk.OverrideSharedPlan(basePlan, plannerPlan)
+
+	return plan
 }
 
 // Generate creates the directory of Nix files and the Dockerfile that define
--- a/devbox.go
+++ b/devbox.go
@@ -101,17 +101,13 @@ func (d *Devbox) Build(opts ...docker.BuildOptions) error {
 // Plan creates a plan of the actions that devbox will take to generate its
 // environment.
 func (d *Devbox) Plan() *plansdk.Plan {
-	basePlan := &plansdk.Plan{
+	userPlan := &plansdk.Plan{
 		DevPackages:     d.cfg.Packages,
 		RuntimePackages: d.cfg.Packages,
 		SharedPlan:      d.cfg.SharedPlan,
 	}
-	plannerPlan := planner.GetPlan(d.srcDir)
 
-	plan := plansdk.MergePlans(basePlan, plannerPlan)
-	plan.SharedPlan = plansdk.OverrideSharedPlan(basePlan, plannerPlan)
-
-	return plan
+	return plansdk.MergeUserPlan(userPlan, planner.GetPlan(d.srcDir))
 }
 
 // Generate creates the directory of Nix files and the Dockerfile that define
--- a/devbox.go
+++ b/devbox.go
@@ -100,7 +100,7 @@ func (d *Devbox) Build(opts ...docker.BuildOptions) error {
 
 // Plan creates a plan of the actions that devbox will take to generate its
 // environment.
-func (d *Devbox) Plan() *plansdk.Plan {
+func (d *Devbox) Plan() (*plansdk.Plan, error) {
 	userPlan := &plansdk.Plan{
 		DevPackages:     d.cfg.Packages,
 		RuntimePackages: d.cfg.Packages,
@@ -113,7 +113,10 @@ func (d *Devbox) Plan() *plansdk.Plan {
 // Generate creates the directory of Nix files and the Dockerfile that define
 // the devbox environment.
 func (d *Devbox) Generate() error {
-	plan := d.Plan()
+	plan, err := d.Plan()
+	if err != nil {
+		return errors.WithStack(err)
+	}
 	if plan.Invalid() {
 		return plan.Error()
 	}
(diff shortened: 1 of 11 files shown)


devbox.go
```diff
@@ -104,9 +105,20 @@ func (d *Devbox) Plan() *plansdk.Plan {
...
+
+	// fields in the basePlan's SharedPlan:
+	//   if empty, will inherit the corresponding planners fields
+	//   if set, will override corresponding planners fields
+	plan.SharedPlan = d.cfg.SharedPlan
+	if err := mergo.Merge(&plan.SharedPlan, plannerPlans.SharedPlan); err != nil {
+		panic(err) // TODO: propagate error.
+	}
+
+	return plan
```
mikeland86, line 121: Should this not happen in `plansdk` instead?

as is right now, `plan` and `build` will use different implementations. (In fact, build is not merging base plan at all)

LucilleH: `build()` calls `generate()`, which calls `plan()`. I think all of them calls `plan()`, even `shell()` calls `plan()` too!

devbox.go
```diff
@@ -104,9 +105,20 @@ func (d *Devbox) Plan() *plansdk.Plan {
 	basePlan := &plansdk.Plan{
 		DevPackages:     d.cfg.Packages,
 		RuntimePackages: d.cfg.Packages,
-		SharedPlan:      d.cfg.SharedPlan,
 	}
-	return plansdk.MergePlans(basePlan, planner.GetPlan(d.srcDir))
+	plannerPlans := planner.GetPlan(d.srcDir)
```
loreto, line 109: I would say:
1. Change `planner.GetPlan` to return the plan from only a single planner (for now the first one that matches) instead of a merged plan.
2. Call `MergePlans` so that the fields from devbox.json override, and slices append. If the function doesn't behave like we want to, you can add a sister function `OverridePlan` or something like that. No need to call `mergo` in here.
3. If you could add a test for this logic, I'd really appreciate it!

LucilleH: @loreto I'm not sure about 1, but I will do 2 and 3 in this PR!

Changing 1 will affect the `Shell` command as well, which I don't think we want to only return the packages from one valid plan. I think we might have to think a bit more on that.

loreto: Aren't we trying to guarantee the shell behaves the same way as build and anything else, though?

LucilleH: Are we allowing a planner that only supports shell for the time being and not build?

devbox.go
mikeland86: my bad! I saw `d.Plan()` and miss-understood it to be `planner.GetPlan`

devbox.go
LucilleH: @loreto I'd be happy to do 1 as a follow-up. Want to think about it more though. Thoughts?

planner/languages/golang/golang_planner.go
```diff
@@ -42,13 +42,15 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 		},
 		SharedPlan: plansdk.SharedPlan{
 			InstallStage: &plansdk.Stage{
-				Command: "go get",
+				InputFiles: []string{"."},
```
mikeland86, line 45: This feels like it could be error prone. Thoughts on:

`plansdk.NewStageWithAllFiles`

or something like that? 

We could also do `plansdk.NewStage` copy all files by default. And we could do:

`plansdk.NewStage("cmd", plansdk.WithInputFiles(...))` to customize.

LucilleH: I'm still not sure how to do it gracefully. We need the default of `[]string{"."}` for install and start stage, but not build stage. (Copying everything in build stage results in unexpected behavior. Probably not desired)

planner/plansdk/plansdk.go
```diff
@@ -122,32 +122,51 @@ func MergePlans(plans ...*Plan) *Plan {
...
+
+func OverrideSharedPlan(dst *Plan, src *Plan) SharedPlan {
+	sharedPlan := &Plan{
+		SharedPlan: dst.SharedPlan,
+	}
+	// fields in the dst plan:
+	//   if empty, will inherit the corresponding src fields
+	//   if set, will override corresponding src fields
+	if err := mergo.Merge(sharedPlan, src); err != nil {
+		panic(err) // TODO: propagate error.
```
mikeland86, line 166: nit, let's avoid panics and instead structure the code to return errors.

planner/plansdk/plansdk.go
```diff
@@ -122,32 +122,51 @@ func MergePlans(plans ...*Plan) *Plan {
...
+			return p.SharedPlan
+		}
 	}
-	// Set default files for install stage to copy over from build step.
-	if plan.SharedPlan.StartStage.InputFiles == nil {
-		plan.SharedPlan.StartStage.InputFiles = []string{"."}
+	return SharedPlan{}
+}
+
+func OverrideSharedPlan(dst *Plan, src *Plan) SharedPlan {
```
mikeland86, line 158: Having a separate function we need to call in addition to MergePlans can lead to mistakes later on (if we forget to call it).

Thoughts on changing interface to:

`MergeAutomatedPlans` (or some similar name, designed to merge plans that come from planners.

`MergeBasePlan(automatedPlan, basePlan)`  (designed to merge json plan into automated one)


Sorry for all naming nits, I'm just getting confused with the planner plan naming and shared plan naming.

planner/languages/golang/golang_planner.go
LucilleH: @mikeland86 we use buildStage.inputFiles for `node_planner`

planner/plansdk/plansdk.go
LucilleH: The `panic` is everywhere at this point. Maybe we need a cleanup round to consolidate them all.

planner/languages/golang/golang_planner.go
mikeland86: got it! makes sense.

maybe we can add some syntactic sugar later.

planner/plansdk/plansdk.go
mikeland86: I think we should avoid adding new ones. (each new one makes the task of getting rid of them harder).

This function is only called by `devbox.Plan` which itself is only called by 3 functions (all of which return error). so it feels like a pretty minor change.

mikeland86 (approved):

PR #104, opened 2022-09-09
[Rust Planner] part 2: devbox build

savil:
## Summary

This PR adds support for `devbox build` for Rust projects.

1. Needed to add the `gcc` nix-package because Rust needs it for `libc` calls.
2. We directly execute the binary produced from `cargo build --release`, instead of `cargo run` to avoid needing to add the `cargo` nix package in the final runtime image.

## How was it tested?

1. added `plan.json` files to `testdata`.
2. In each `testdata/rust` directory, did:
   - `devbox build`
   - `docker run devbox` -> inspected output to ensure correct rust version was executed with.


savil:
* main
  * **PR #92** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/92" target="_blank">Graphite</a> 
    * **PR #97** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/97" target="_blank">Graphite</a> 
      * **PR #104** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/104" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/104?utm_source=stack-comment).

testdata/rust/rust-stable/plan.json
```diff
@@ -1,15 +1,18 @@
...
   "build_stage": {
-    "command": ""
+    "command": "cargo build",
+    "input_files": [
+      "."
+    ]
   },
   "start_stage": {
-    "command": "",
+    "command": "rustup default stable \u0026\u0026 cargo run \u0026\u0026 rustc --version",
```
loreto, line 15: Why do we need `\u0026`. Is that what the go json marshaller does? Can we make it output `&` instead?

savil: its from json marshaller. I'll see if there's a way to avoid it...

savil:
@loreto I added the most minimal `Overlays: []string` field in the `Plan` struct in previous PR (#97). 

As I look into `devbox build` implementation for Rust, I think we can get away without having any runtime-overlays, by copying the built binary (similar to golang_planner) from BuildStage to StartStage.


That said, I can see other planners wanting a runtime-overlay. Do we want to distinguish dev-overlays and runtime-ovarlays in the code now? 

If so, there are a couple of options on implementing it:
- `DevOverlays: []string`, `RuntimeOverlays: []string`  in addition to the existing `DevPackages` and `RuntimePackages`
-  Alternatively (and my preference), may be nicer to wrap this into a common struct so we have: `Dev: PackageInfo, Runtime: PackageInfo` where `PackageInfo` is a struct of `{ Packages: []string, Overlays: []string}`

WDYT? For now, I'm going to punt on this question (of dev versus runtime overlays) and consider Overlays as just DevOverlays, but wanted to raise it in case folks feel we should add it up-front.

loreto (approved):
LGTM

I would rename `Overlays` to `NixOverlays`. I don't think we need to distinguish between `Dev` and `Runtime` overlays for the time being.

planner/languages/rust/rust_planner.go
```diff
@@ -54,7 +58,20 @@ func (p *Planner) getPlan(srcDir string) (*plansdk.Plan, error) {
...
-		DevPackages: []string{rustPkgDev},
+		// 'gcc' added as a linker for libc (C toolchain)
+		// 1. https://softwareengineering.stackexchange.com/a/332254
+		// 2. https://stackoverflow.com/a/56166959
+		DevPackages: []string{rustPkgDev, "gcc"},
+		// 'gcc' needs to be present to execute the binary
+		RuntimePackages: []string{"gcc"},
+		SharedPlan: plansdk.SharedPlan{
+			BuildStage: &plansdk.Stage{
+				Command: "cargo build --release",
```
LucilleH, line 69: @savil when you rebase, I removed the default install stage `inputFiles` field, so you'd have to add
```
InstallStage: {
    InputFiles: []string{"."}
}
```

planner/languages/rust/rust_planner.go
```diff
@@ -54,7 +58,20 @@ func (p *Planner) getPlan(srcDir string) (*plansdk.Plan, error) {
...
+			},
+			StartStage: &plansdk.Stage{
```
LucilleH, line 71: Similarly, add `InputFiles` for `StartStage`

LucilleH (commented):
Try rebasing and run `go test ./...` again

savil:
image size:
1. with runtime packages = {`rust-overlay.minimal`, `gcc`} = 800MB
2. with runtime pkgs = {`gcc`} = 300MB
3. with no runtime pkgs = 30MB

The current implementation is (1). I want to explore alternate implementations.

loreto:
With that big of a difference, I would say forget the suggestion of checking the version within the testdata example, and let's just generate the small images.

savil:
Updated to just have `gcc` as dependency. This reduces the image size but at 300MB it is still fairly heavy. Will followup with improvements.

PR #105, opened 2022-09-10
[nginx]Add nginx static file planner DEV-1147

mikeland86:
## Summary

Implements static nginx planner for both shell and build commands. The strategy is a bit different for each:

* For both shell and build nginx is installed as a package. (Maybe we should consider using [services](https://nixos.wiki/wiki/Nginx) for build instead?)
* For shell, we compile with a few flags to ensure nginx can run with limited commands. Some inspiration borrowed from [here](https://docs.replit.com/tutorials/build-with-nix). It creates a script wrapper to ensure flags are correct.
* For build, it's a bit more straightforward (see readme for more details). We ensure we create the nginx user following inspiration from [nginx docs](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-docker/) (we may want to follow more of that doc in the future)
* Marking this as experimental, since it will probably change. 

## How was it tested?

* unit tests
* devbox shell followed by calling `nginx-shell`
* devbox build, followed by `docker run -p 80:80`


LucilleH:
I'm not sure if we need to support nginx in a shell. I do also have a working version with just some customization on `devbox.json`

mikeland86:
@LucilleH this PR is build only. But I think I can make it work in shell

mikeland86:
@LucilleH @loreto now ready for review and supports build and shell.

planner/languages/nginx/README.md
```diff
@@ -0,0 +1,28 @@
...
+directory. It defaults to `nginx.conf` for building and `shell-nginx.conf` for 
+shell. If it can't find the correct config it uses the other one.
+
+## How nginx works in shell
+
+To run shell nginx you can use the `shell-nginx` wrapper. This wrapper calls nginx 
+with a few options. If you want to see what this wrapper does, use `cat $(which shell-nginx)`
+
+In shell everything is local so you should avoid pointing to assets or files outside 
+the directory because the nix shell might not have access. For example your root
```
gcurtis, line 17: Which directory?

planner/languages/nginx/nginx_planner.go
```diff
@@ -0,0 +1,96 @@
...
+	return "nginx.Planner"
+}
+
+func (p *Planner) IsRelevant(srcDir string) bool {
+	return plansdk.FileExists(filepath.Join(srcDir, "nginx.conf")) ||
+		plansdk.FileExists(filepath.Join(srcDir, "shell-nginx.conf"))
+}
+
+func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
+	fmt.Println(srcDir)
```
gcurtis, line 29: Intentional?

planner/languages/nginx/nginx_planner.go
```diff
@@ -0,0 +1,96 @@
...
+var startCommand = strings.TrimSpace(`
+	addgroup --system --gid 101 nginx && \
+	adduser --system --ingroup nginx --no-create-home --home /nonexistent --gecos "nginx user" --shell /bin/false --uid 101 nginx && \
+	mkdir -p /var/cache/nginx/client_body && \
+	mkdir -p /var/log/nginx/ && \
+	echo Starting nginx with command \"nginx -c /app/%[1]s -g 'daemon off;'\" && \
+	nginx -c /app/%[1]s -g 'daemon off;'
+`)
+
+const customNginxDefintion = `
```
gcurtis, line 79:
```suggestion
const customNginxDefinition = `
```

planner/languages/nginx/nginx_planner.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+		StartStage: &plansdk.Stage{
+			// Create user/group and directories
+			Command:    fmt.Sprintf(startCommand, p.buildConfig(srcDir)),
+			InputFiles: plansdk.AllFiles(),
+		},
+		// These definitions are only used in shell. Since nix is lazy, it won't
+		// actually build them at all if they are not used.
+		Definitions: []string{
+			customNginxDefintion,
```
gcurtis, line 50:
```suggestion
			customNginxDefinition,
```

planner/plansdk/plansdk.go
```diff
@@ -16,12 +16,24 @@ type PlanError struct {
...
 }
 
+// TODO: Plan currently has a bunch of fields that it should not export.
+// Two reasons why we need this right now:
+// 1/ So that individual planners can use the fields
+// 2/ So that we print them out correctly in `devbox plan`
+//
+// (1) can be solved by using a WithOption pattern, (e.g. NewPlan(..., WithWelcomeMessage(...)))
+// (2) can be solved by using a custom JSON marshaler.
 type Plan struct {
```
gcurtis, line 26: Let's put an empty line between this comment and Plan so it doesn't show up in godocs.

planner/plansdk/plansdk.go
```diff
@@ -116,6 +108,10 @@ func (p *Plan) WithError(err error) *Plan {
 	return p
 }
 
+func (p *Plan) ShellAndDevPackages() []string {
+	return pkgslice.Unique(append(p.DevPackages, p.ShellPackages...))
```
gcurtis, line 112: You probably want to copy to a new slice here, otherwise the returned slice and DevPackages will share the same backing array.

LucilleH:
I don't know if this is the right way to solve it. I pulled the branch and tested with Docusaurus and it doesn't serve as expected (broken pages w no proper css files, etc). I might have misconfigured it perhaps.

But more importantly, we may want to use Nix services instead? Then we can generalize it to support services in general, nginx, Apache, LAMP stack, etc. See https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/nginx.nix

Thoughts?

planner/plansdk/plansdk.go
```diff
@@ -16,12 +16,24 @@ type PlanError struct {
...
+
 	NixOverlays []string `cur:"[...string]" json:"nix_overlays,omitempty"`
 
 	// DevPackages is the slice of Nix packages that devbox makes available in
-	// its development environment.
+	// its development environment. They are also available in shell.
 	DevPackages []string `cue:"[...string]" json:"dev_packages"`
+
+	// ShellPackages are only available in the shell environment.
+	ShellPackages []string `cue:"[...string]" json:"shell_packages"`
```
gcurtis, line 36: Does including shell packages in the dev environment add a lot of extra stuff?

I think trying to keep a single set of packages would be more reliable. If the shell has extra packages, then it means your dev environment no longer matches what gets built.

mikeland86:
@LucilleH @gcurtis reimplemented using same nginx for shell and build.

The user needs to add some config options when running in shell. For simplicity I generate the file and tell them to include it in the welcome message. It's not a fantastic experience, but it prevents us from having two different nginx builds.

We could try to automate the nginx.conf construction, but I feel like that could be really error prone.  

Ideally we can add the options in the command line call but I'm not sure that's possible.

LucilleH:
Tested locally, when running shell-nginx I got the following error:
```
Starting nginx with command:
nginx -p . -c nginx.conf -e /tmp/error.log -g "pid /tmp/mynginx.pid;daemon off;"
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in ./nginx.conf:1
nginx: [emerg] open() "./mime.types" failed (2: No such file or directory) in ./nginx.conf:4
```

planner/plansdk/stage.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (s *Stage) GetInputFiles() []string {
+	if s == nil {
+		return []string{}
+	}
+	return s.InputFiles
+}
+
+func AllFiles() []string {
+	return []string{"."}
```
LucilleH, line 24: oh this is nice. Should've done this earlier

LucilleH:
Doing devbox build and docker run, I got the following error:
```
docker run -p 80:80 --expose 80 -ti devbox
Starting nginx with command "nginx -c /app/nginx.conf -g daemon off;"
nginx: [emerg] open() "/app/mime.types" failed (2: No such file or directory) in /app/nginx.conf:4
```
It was working in the previous version 🤔

planner/languages/nginx/nginx_planner.go
```diff
@@ -0,0 +1,102 @@
...
+Use \"shell-nginx\" to start the server
+`
+
+var startCommand = strings.TrimSpace(`
+	addgroup --system --gid 101 nginx && \
+	adduser --system --ingroup nginx --no-create-home --home /nonexistent --gecos "nginx user" --shell /bin/false --uid 101 nginx && \
+	mkdir -p /var/cache/nginx/client_body && \
+	mkdir -p /var/log/nginx/ && \
+	PKG_PATH=$(readlink -f $(which nginx) | sed -r "s/\/bin\/nginx//g") && \
+	ln -s /app/%[1]s $PKG_PATH/conf/devbox-%[1]s && \
```
LucilleH, line 91: wow ok

gcurtis:
@mikeland86 I resolved the conflicts so I could test things out. Is it okay if I push them here?

mikeland86:
@gcurtis go for it

gcurtis (approved):
👍 I was able to run both a shell and a built image. I think this is a good first iteration.

PR #106, opened 2022-09-10
nix: allow user's shellrc to update PATH; add tests for shellrc template

gcurtis:
## Summary

At the end of the devbox shellrc file, we're setting:

	export PATH="$PURE_NIX_PATH:$ORIGINAL_PATH"

This will overwrite anything that the user's shellrc might've added to PATH. Change ORIGINAL_PATH to PATH to preserve those changes. There's also no need for ORIGINAL_PATH to be an environment variable, so set it directly in the template instead.

To prevent regressions, some tests that generate a devbox shellrc file and compare it to an expected golden file. This ensures we don't unknowingly change the way the shellrc is generated.

If we do intentionally change the devbox shellrc, the tests can automatically update the golden files with -update:

	go test ./nix -update

It's up to the author of the change to verify that any new shellrc files still work in a real shell.

## How was it tested?

`go test` and `devbox shell`

### DIFF
--- a/nix/shell.go
+++ b/nix/shell.go
@@ -118,7 +118,7 @@ func (s *Shell) Run(nixPath string) error {
 
 // execCommand is a command that replaces the current shell with s.
 func (s *Shell) execCommand() string {
-	shellrc, err := writeDevboxShellrc(s.userShellrcPath, s.UserInitHook)
+	shellrc, err := writeDevboxShellrc(s.userShellrcPath, s.UserInitHook, os.Environ())
 	if err != nil {
 		debug.Log("Failed to write devbox shellrc: %v", err)
 		return "exec " + s.binPath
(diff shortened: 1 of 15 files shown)


nix/shell.go
```diff
@@ -154,7 +159,22 @@ func writeDevboxShellrc(userShellrcPath string, userHook string) (path string, e
...
 
-	path = filepath.Join(tmp, filepath.Base(userShellrcPath))
+	var envPath []string
+	for _, kv := range env {
+		key, val, _ := strings.Cut(kv, "=")
+		if key == "PATH" {
+			envPath = filepath.SplitList(val)
+			break
+		}
+	}
```
gcurtis, line 169: This part actually goes away in #124, which is based off of this PR. It was just a bit cumbersome to backport it here.

PR #107, opened 2022-09-10
[json] Make input files unexported

mikeland86:
## Summary

This makes input files not show up in json. One side effect is that they don't show up in `devbox plan` either.

Another minor side effect is that input files can never be empty, but we could fix that if needed. 

@LucilleH @loreto thoughts?

## How was it tested?
* Unit tests
* Tested go and node examples

### DIFF
--- a/devbox_test.go
+++ b/devbox_test.go
@@ -67,9 +67,9 @@ func assertPlansMatch(t *testing.T, expected *plansdk.Plan, actual *plansdk.Plan
 	// Check that input files are the same for all stages.
 	// Depending on where the test command is invoked, the input file paths can be different.
 	// We will compare the file name only.
-	assert.ElementsMatch(expected.InstallStage.GetInputFiles(), getFileNames(actual.InstallStage.GetInputFiles()), "InstallStage.InputFiles should match")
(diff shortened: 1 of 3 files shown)


mikeland86:
Looks like `mergo` does't work on unexported fields.  So if we want to continue using mergo we would need it to be exported sadly. 

We may want to make merging if input fields custom anyway since we would want to dedup etc. Also if any planner explicitly says `.` then we may want to drop all other inputs (since they would be duplicates)

LucilleH:
I think the logic can be better and I can take a stab at refactoring it after doing some more work on https://github.com/jetpack-io/devbox/pull/103/files. I do think tho, that input files field should not be hidden from `devbox plan`, because it is kinda nice to just run devbox plan and know exactly what the behavior would be

loreto:
My take is that we should just have two different structs: one for `devbox.json` and the `Plan` one for the planners. `devbox plan` should show `Plan`. When creating a plan, we'll copy over the appropriate fields from `devbox.json` to a base plan.

(does that make sense?)

mikeland86:
> My take is that we should just have two different structs

@loreto 100% . Embedding structs is really confusing when marshaling/unmarshaling is involved (in fact, I find it always confusing and a bit hard to work with).

mikeland86:
I'm going to close this, will revisit once https://github.com/jetpack-io/devbox/pull/103 lands

PR #112, opened 2022-09-12
[Bug] Fix nodejs npm build step

LucilleH:
## Summary
Quick fix nodejs npm build step, as the command is different between npm and yarn. This needs to go in before the next release.

## How was it tested?
go test ./...


loreto (commented):
By chance is there an example in `testdata/` you could add that demonstrates the bug?

loreto:
Oh, is it `nodejs-npm/plan.json` that you're already updating?

LucilleH:
> Oh, is it `nodejs-npm/plan.json` that you're already updating?

Yes

loreto (approved):
LGTM

PR #113, opened 2022-09-12
[PHP] Add extension support and better version detection

mikeland86:
## Summary

Use composer.json to figure out if any extensions should be installed. Also, improved version detection (using `require` instead of `config`)

@loreto should we be using `systemPackages` instead of packages for languages?

## How was it tested?

```
devbox shell
php -m | grep mbstring 
```

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -107,7 +107,11 @@ func (d *Devbox) Plan() (*plansdk.Plan, error) {
 		SharedPlan:      d.cfg.SharedPlan,
 	}
 
-	return plansdk.MergeUserPlan(userPlan, planner.GetPlan(d.srcDir))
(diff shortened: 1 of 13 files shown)


loreto:
Could you add a `testdata/` example that uses as many extensions as possible? 

Does `development.nix.tmpl` not need to change too?

Could you elaborate on:
> should we be using systemPackages instead of packages for languages?

mikeland86:
> should we be using systemPackages instead of packages for languages?

in nix files "packages" are user packages. "systemPackages" belong to system. I guess since we are mostly doing single user setups it doesn't really matter

mikeland86:
@loreto added tests, updated development.nix, and removed a few panics. PTAL

loreto (approved):
LGTM

PR #120, opened 2022-09-13
[CSharp Planner] part 1: devbox shell

savil:
## Summary

This PR adds a C# Planner.

Do note that the nix package we use is that of the Dot Net Framework. And a 
particular Dot Net Framework version (say, 6) can support its latest C# version
and older versions. For example, dotnet-6 supports C# 10 and C# 9, and possibly older C# as well.

So, this Planner is in essence a DotNetPlanner, more than a CSharpPlanner. When
we look to add support for other DotNet languages like FSharp, we can either 
reuse this planner by renaming to DotNetPlanner, or have some mixin abstraction
for CSharpPlanner and FSharpPlanner to share logic.

## How was it tested?

added to `testdata/csharp`.

In each folder, I did:
```
> devbox shell

# verify version is correct
> dotnet --version

# expect Hello World
> dotnet run
```


savil:
* main
  * **PR #120** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/120" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/120?utm_source=stack-comment).

planner/languages/csharp/csharp_planner.go
```diff
@@ -17,9 +29,88 @@ func (p *Planner) Name() string {
...
+	return a.HasAnyFile("*.csproj")
 }
 
 func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
-	return &plansdk.Plan{}
+	plan, err := p.getPlan(srcDir)
+	if err != nil {
+		// Added this Printf because `devbox shell` was silently swallowing this error.
+		// TODO savil. Have `devbox shell` error out or print it instead.
+		fmt.Printf("error in getPlan: %s\n", err)
```
loreto, line 45: The issue is this: imagine there's 2 planners, one of them errors, one of them succeeds – `devbox shell` succeeds because it's able to use the successful plan – but this would still print out an error. 

I'm wondering if instead we should log an error after we've run all the planners and are trying to pick what plan to use.

Question: in practice, what type of error where you seeing here?

planner/languages/csharp/csharp_planner.go
```diff
@@ -17,9 +29,88 @@ func (p *Planner) Name() string {
...
+		// We should log that an error has occurred.
+		return nil, err
+	}
+	paths := a.GlobFiles("*.csproj")
+	if len(paths) < 1 {
+		return nil, errors.Errorf("expected to find a .csproj file in directory %s", srcDir)
+	}
+	projectFilePath := paths[0]
+
+	content, err := os.ReadFile(projectFilePath)
```
loreto, line 80: Instead of all the logic below this line, could you do `a.ParseFile`?

planner/languages/csharp/csharp_planner.go
```diff
@@ -17,9 +29,88 @@ func (p *Planner) Name() string {
...
+		return nil, err
+	}
+
+	return proj, nil
+}
+
+// The TargetFramework is more complicated than below, but I'm picking out what
+// seem to be the common uses.
+// https://docs.microsoft.com/en-us/dotnet/standard/frameworks
+func dotNetNixPackage(proj *Project) (string, error) {
```
loreto, line 97: Wondering if you can add `testdata/` for all the cases here.

I think you're missing:
- A project with a missing target framework
- donet 7
- dotnet 3
- Unknown/unrecognized version

testdata/csharp/csharp_10-dotnet_6/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
```
loreto, line 2: Let's go with a meaningful message. At minimum print the name of the example (i.e. "C# 10 using .Net 6"). Ideally you can do something closer to the `go` examples where you check the version that is running using some library and error if it's an unexpected version.

testdata/csharp/csharp_10-dotnet_6/devbox.json
```diff
@@ -0,0 +1,4 @@
+{
+  "packages": [],
+  "shell": {}
```
loreto, line 3: Is `devbox init` adding this by default? (if yes, could you add an `omitempty` to it really quickly?

planner/languages/csharp/csharp_planner.go
savil: Added these cases except dotnet-7

I couldn't make a project for dotnet-7, because its nixpkg has been [published on 2022-08-22](https://hydra.nixos.org/job/nixos/trunk-combined/nixpkgs.dotnet-sdk_7.x86_64-linux/all).

But we've pinned our nixpkgs to that of 2022-08-16
ref. from `shell.nix.tmpl` file:
```
  pkgs = import (fetchTarball {
    # Commit hash as of 2022-08-16
    # `git ls-remote https://github.com/nixos/nixpkgs nixos-unstable`
    url = "https://github.com/nixos/nixpkgs/archive/af9e00071d0971eb292fd5abef334e66eda3cb69.tar.gz";
    sha256 = "1mdwy0419m5i9ss6s5frbhgzgyccbwycxm5nal40c8486bai0hwy";
  }) {};
```

testdata/csharp/csharp_10-dotnet_6/Program.cs
savil: Added version checks, which error when the version is unexpected.

testdata/csharp/csharp_10-dotnet_6/devbox.json
savil: 🤔 looks like `omitempty` is already specified in `devbox.Config` for Shell.

I think we may have to make a struct-pointer for that:
https://stackoverflow.com/questions/18088294/how-to-not-marshal-an-empty-struct-into-json-with-go

planner/languages/csharp/csharp_planner.go
savil: This happened when I was trying out dotnet-6 but hadn't added support for it in the code below. 

I expected `devbox shell` to error with "unsupported version", but it just continued on -- which I suppose is the designed for behavior. But maybe we need to revisit that in some form: once a Planner has been deemed relevant, then it should not error.

PR #122, opened 2022-09-13
Separate devbox config struct from planner struct

LucilleH:
## Summary
This is a continuation of https://github.com/jetpack-io/devbox/pull/103. It hides the `inputFiles` fields in each stage and only expose the `command` field.

This PR 
- Get rid of `SharedPlan`
- Separate stage struct for devbox.json config
- Only expose `command` field in each stage

It does not include:
- Getters and setters
- immutability

## How was it tested?
go test ./...

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -5,7 +5,6 @@
 package devbox
 
 import (
-	"encoding/json"
 	"path/filepath"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


devbox.go
```diff
@@ -156,3 +156,27 @@ func (d *Devbox) saveCfg() error {
 	cfgPath := filepath.Join(d.srcDir, configFilename)
 	return cuecfg.WriteFile(cfgPath, d.cfg)
 }
+
+func (d *Devbox) convertToPlan() (*plansdk.Plan, error) {
+	configStages := []*Stage{d.cfg.InstallStage, d.cfg.BuildStage, d.cfg.StartStage}
+	planStages := []*plansdk.Stage{{}, {}, {}}
+
+	for i, stage := range configStages {
```
loreto, line 164: Why do you marshal and unmarshal from JSON? Is it to "copy" over the fields? I think you should just copy the right fields one by one, but if you're trying to make it more generic, use `mergo` instead?

loreto (approved):
LGTM

PR #124, opened 2022-09-13
nix: load env before Nix shellHook; clean up PATH

gcurtis:
## Summary

Depends on #106.

When launching a devbox shell, we need to configure the environment so that things don't break in weird ways. We need to copy some variables directly (like LANG/LC_* locale variables) and others we need to modify (such as PATH). Our current method of setting these variables has a few problems:

1. We don't set things early enough. By the time we set variables in the shellrc, a bunch of commands have already run (and possibly broken).
2. Setting variables with /usr/bin/env runs into command length limits, especially with the PATH that nix-shell sets.
3. Non-interactive shells don't run the shellrc at startup, which will be problematic for implementing `devbox run` or piping scripts to stdin.

To fix this, instead set environment variables when launching the `nix-shell` process and use the `--keep` flag to tell Nix to propagate them into the new shell. This ensures that the Nix shell, the Nix shell hooks, and the devbox shell all get the correct environment. See `envToKeep` for the full list of variables we copy over.

Setting the correct PATH is the most complicated part of this change, but to summarize:

1. When the user runs `devbox shell`, we take a snapshot of PATH and NIX_PROFILES.
2. We split and clean (per filepath.Clean) each path element in these variables.
3. We remove anything from PATH that is empty, relative, or a subdirectory of anything in NIX_PROFILES.
4. We stick this new PATH in a variable named PARENT_PATH. In the nix-shell hook, we set `PATH="$PATH:$PARENT_PATH`.

This ensures that the PATH is correct by the time the devbox shell runs. However, there's still the possibility that the user's shellrc will change the PATH again. This is common with version managers like asdf. To make sure Nix packages always have priority before giving the user a prompt, we run some commands in the devbox shellrc that find all PATH entries that are in the Nix store and put them in front of everything else.

## How was it tested?

`go test ./...` and running a `devbox shell`.


boxcli/shell.go
```diff
@@ -47,3 +48,11 @@ func runShellCmd(cmd *cobra.Command, args []string) error {
 	}
 	return err
 }
+
+func nixShellPersistentPreRunE(cmd *cobra.Command, args []string) error {
```
loreto, line 52: If you want to simplify the name, something like `shellPreRun` would be fine too.

gcurtis: This was actually unrelated to this PR. I had to merge again to fix the diff.

PR #125, opened 2022-09-13
[java planner] part1: shell packages for OpenJDK + Maven

mohsenari:
## Summary
Added shell planner for Java. 
Includes java versions that are used and available in nix packages. (open to expand the list)
Includes adding maven package by default. This can change later depending on which file exists in the project (e.g., `build.gradle`)

## How was it tested?
1. build and compile devbox
```shell
  go build -o ./devbox cmd/devbox/main.go
  ```
2. create an empty directory and put the sample `pom.xml` below inside it. (pom files are usually long.)
3. create an empty devbox.json and check the output of `./devbox plan`
4. change the version number in `<maven.compiler.source>` to 11 or 8 and verify the output of `./devbox plan` again.

pom file:
```xml
<?xml version="1.0" encoding="UTF-8"?>

<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>com.devbox.mavenapp</groupId>
  <artifactId>devbox-maven-app</artifactId>
  <version>1.0-SNAPSHOT</version>

  <name>devbox-maven-app</name>
  <!-- FIXME change it to the project's website -->
  <url>http://www.example.com</url>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <maven.compiler.source>17</maven.compiler.source>
    <maven.compiler.target>17</maven.compiler.target>
  </properties>

  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.11</version>
      <scope>test</scope>
    </dependency>
  </dependencies>

  <build>
    <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
      <plugins>
        <!-- clean lifecycle, see https://maven.apache.org/ref/current/maven-core/lifecycles.html#clean_Lifecycle -->
        <plugin>
          <artifactId>maven-clean-plugin</artifactId>
          <version>3.1.0</version>
        </plugin>
        <!-- default lifecycle, jar packaging: see https://maven.apache.org/ref/current/maven-core/default-bindings.html#Plugin_bindings_for_jar_packaging -->
        <plugin>
          <artifactId>maven-resources-plugin</artifactId>
          <version>3.0.2</version>
        </plugin>
        <plugin>
          <artifactId>maven-compiler-plugin</artifactId>
          <version>3.8.0</version>
        </plugin>
        <plugin>
          <artifactId>maven-surefire-plugin</artifactId>
          <version>2.22.1</version>
        </plugin>
        <plugin>
          <artifactId>maven-jar-plugin</artifactId>
          <version>3.0.2</version>
        </plugin>
        <plugin>
          <artifactId>maven-install-plugin</artifactId>
          <version>2.5.2</version>
        </plugin>
        <plugin>
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.8.2</version>
        </plugin>
        <!-- site lifecycle, see https://maven.apache.org/ref/current/maven-core/lifecycles.html#site_Lifecycle -->
        <plugin>
          <artifactId>maven-site-plugin</artifactId>
          <version>3.7.1</version>
        </plugin>
        <plugin>
          <artifactId>maven-project-info-reports-plugin</artifactId>
          <version>3.0.0</version>
        </plugin>
      </plugins>
    </pluginManagement>
  </build>
</project>
```


planner/languages/java/java_planner.go
```diff
@@ -17,9 +35,68 @@ func (p *Planner) Name() string {
...
+	} else {
+		return defaultJava
+	}
+}
+
+func parseJavaVersion(pomXMLPath string) int {
+	parsedVersion, err := parseXML(pomXMLPath)
+	if err != nil {
+		fmt.Printf("error parsing java version from pom file: %v", err)
+		return 0
```
LucilleH, line 69: Should this return an error instead?

planner/languages/java/java_planner.go
```diff
@@ -17,9 +35,68 @@ func (p *Planner) Name() string {
...
+	}
+	return parsedVersion
+}
+
+func fileExists(path string) bool {
```
LucilleH, line 74: use `plansdk.FileExists` instead?

planner/languages/java/java_planner.go
```diff
@@ -17,9 +35,68 @@ func (p *Planner) Name() string {
...
+	if err != nil {
+		return 0, errors.WithStack(err)
+	}
+
+	var project mvnparser.MavenProject
+	// unmarshaling byteArray which contains our pom file content into 'project'
+	err = xml.Unmarshal(byteArray, &project)
+	if err != nil {
+		return 0, errors.WithStack(err)
+	}
```
savil, line 91: Not blocking, but once #120 lands, these lines could become:
```
var project mvnparse.MavenProject
err = cuecfg.ParseFile(pomXMLPath, &project)
```

That PR adds support for xml parsing to `cuecfg` package.

savil (commented):
could you add testcases to `testdata/java/`?

We could start with just adding the files from your Test Plan into a sub-folder therein.
Eventually, we should have a test case each for the versions in `jVersionMap`...

planner/languages/java/java_planner.go
mohsenari: I copied the same changes in `cuecfg` package that you made in #120 . So if either of these PRs merge not conflict will happen (hopefully)

planner/languages/java/java_planner.go
```diff
@@ -17,9 +33,64 @@ func (p *Planner) Name() string {
...
+
+func parseJavaVersion(pomXMLPath string) (int, error) {
+	parsedVersion, err := parseXML(pomXMLPath)
+	if err != nil {
+		return 0, errors.WithMessage(err, "error parsing java version from pom file")
+	}
+	return parsedVersion, nil
+}
+
+func parseXML(pomXMLPath string) (int, error) {
```
savil, line 78: these two functions can now be collapsed, i think

savil (approved):
Awesome!

Two suggestions:

1. Add a README.md in `testdata/java` to explain how to generate the code for a test case for Java. For example, see the README in `testdata/csharp`.
2. (Can be in followup PR) Could we improve the testcases to check for the specific java version being used and panic or throw exception if the version is wrong? This is what we've been trying to add for golang, csharp and rust.

planner/languages/java/java_planner.go
mohsenari: @savil added the `README.md`. I'll add the specific java version check in the PR for build.

PR #126, opened 2022-09-14
make Config.Shell be struct pointer so omitempty works

savil:
## Summary

`devbox init` produces files like:
```
➜ cat devbox.json
{
  "packages": [],
  "shell": {}
}
```

This PR changes this to produce:
```
➜ cat devbox.json
{
  "packages": []
}%
```

## How was it tested?

didn't test - rely on Github Actions tests?

### DIFF
--- a/config.go
+++ b/config.go
@@ -8,6 +8,11 @@ import (
 	"go.jetpack.io/devbox/cuecfg"
 )
 
+type ConfigShell struct {
+	// InitHook contains commands that will run at shell startup.
+	InitHook string `json:"init_hook,omitempty"`
+}
+
 // Config defines a devbox environment as JSON.
 type Config struct {
 	// Packages is the slice of Nix packages that devbox makes available in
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #126** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/126" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/126?utm_source=stack-comment).

loreto:
Change looks fine, two requests:
1. You have some failing tests, fix :)
2. Could you add to `testdata/` an "empty" project? (a directory with a `devbox.json` and a corresponding `plan.json` and nothing else). In the past that case could trigger a nil pointer exception when rendering empty pointers in the templates, and I want to make sure we're catching those.

gcurtis:
This looks good to me, but I also have an upcoming PR which is going to set the default shell prompt, so `shell` will never be nil in new configs.

PR #131, opened 2022-09-14
devbox: allow shell.init_hook to be a JSON array

gcurtis:
## Summary

Add a `ConfigShellCmds` type that supports marshalling/unmarshalling shell commands from devbox.json as either a string or an array of strings. This makes writing multi-line shell hooks a bit easier.

Example:

	{
	  "packages": [
	    "go"
	  ],
	  "shell": {
	    "init_hook": [
	      "echo 'line 1'",
	      "echo 'line 2'"
	    ]
	  }
	}

The original format is preserved so that if the hook is a string, it will marshal back to a string. Converting from a string to an array results in an array where the only element is the original string.

## How was it tested?

Unit tests and manually creating shell hooks in both formats.

### DIFF
--- a/config.go
+++ b/config.go
@@ -4,6 +4,10 @@
 package devbox
 
 import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/cuecfg"
 )
@@ -26,7 +30,7 @@ type Config struct {
 	// Shell configures the devbox shell environment.
 	Shell struct {
 		// InitHook contains commands that will run at shell startup.
-		InitHook string `json:"init_hook,omitempty"`
+		InitHook ConfigShellCmds `json:"init_hook,omitempty"`
 	} `json:"shell,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -49,3 +53,84 @@ func ReadConfig(path string) (*Config, error) {
...
+	Cmds      []string
+}
+
+// MarshalJSON marshals shell commands according to s.MarshalAs. It marshals
+// commands to a string by joining s.Cmds with newlines.
+func (s ConfigShellCmds) MarshalJSON() ([]byte, error) {
+	switch s.MarshalAs {
+	case CmdArray:
+		return json.Marshal(s.Cmds)
+	case CmdString:
```
LucilleH, line 102:
```
case Array:
case String:
```
?

config.go
```diff
@@ -49,3 +53,84 @@ func ReadConfig(path string) (*Config, error) {
...
+	// CmdString formats shell commands as a single string.
+	CmdString
+)
+
+// CmdFormat defines a way of formatting shell commands in a devbox config.
+type CmdFormat int
+
+func (c CmdFormat) String() string {
+	switch c {
+	case CmdArray:
```
LucilleH, line 72: Can we generalize this to any fields?

config.go
```diff
@@ -49,3 +53,84 @@ func ReadConfig(path string) (*Config, error) {
...
+// MarshalJSON marshals shell commands according to s.MarshalAs. It marshals
+// commands to a string by joining s.Cmds with newlines.
+func (s ConfigShellCmds) MarshalJSON() ([]byte, error) {
+	switch s.MarshalAs {
+	case CmdArray:
+		return json.Marshal(s.Cmds)
+	case CmdString:
+		return json.Marshal(s.String())
+	default:
+		panic(fmt.Sprintf("invalid command format: %s", s.MarshalAs))
```
LucilleH, line 105: I don't know whether we should do `panic` or return error. What's your take?

LucilleH (approved):
I need this!

config.go
gcurtis: I made it a panic because it's guaranteed to be a bug if it ever happens. Similar to panicking when a template fails to parse.

gcurtis:
@LucilleH I'm going to merge this for now, but let me know if there's anything from your comments we should follow up on.

PR #132, opened 2022-09-15
[RFC][Rust Planner] Reduce image size by using glibc, and copying built binary

savil:
## Summary

Reduce image size by

1. Replace `gcc` with `glibc` in the Runtime image. 
   - `gcc` is needed in the DevPackages for the Build Stage. Replacing it with `glibc` doesn't work because `glibc` nix pkg is not available on mac OS (linux only). We want universal packages (defined as linux and mac, lol) since we share DevPackages with `devbox shell`.

2. Copy just the specific built binary to the Runtime image.

## How was it tested?

Using the basic `testdata/rust/rust-stable-hello-world` project:
- before, size of image = 320MB
- after, size of image = 64MB

### DIFF
--- a/devbox_test.go
+++ b/devbox_test.go
@@ -68,9 +68,21 @@ func assertPlansMatch(t *testing.T, expected *plansdk.Plan, actual *plansdk.Plan
 	// Check that input files are the same for all stages.
 	// Depending on where the test command is invoked, the input file paths can be different.
 	// We will compare the file name only.
-	assert.ElementsMatch(expected.InstallStage.GetInputFiles(), getFileNames(actual.InstallStage.GetInputFiles()), "InstallStage.InputFiles should match")
-	assert.ElementsMatch(expected.BuildStage.GetInputFiles(), getFileNames(actual.BuildStage.GetInputFiles()), "BuildStage.InputFiles should match")
-	assert.ElementsMatch(expected.StartStage.GetInputFiles(), getFileNames(actual.StartStage.GetInputFiles()), "StartStage.InputFiles should match")
+	assert.ElementsMatch(
+		expected.InstallStage.GetInputFiles(),
+		getFileNames(actual.InstallStage.GetInputFiles()),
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #132** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/132" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/132?utm_source=stack-comment).

savil:
Tests are failing due to this `getFileNames` logic in devbox_test, which I don't quite understand the motivation for: https://github.com/jetpack-io/devbox/blame/26e7e8297aa5e4e31f9926328d98fbf62fc375a2/devbox_test.go#L58-L64

devbox_test.go
```diff
@@ -68,9 +68,21 @@ func assertPlansMatch(t *testing.T, expected *plansdk.Plan, actual *plansdk.Plan
...
+	assert.ElementsMatch(
+		expected.BuildStage.GetInputFiles(),
+		getFileNames(actual.BuildStage.GetInputFiles()),
+		"BuildStage.InputFiles should match",
+	)
+	assert.ElementsMatch(
+		expected.StartStage.GetInputFiles(),
+		actual.StartStage.GetInputFiles(),
+		"StartStage.InputFiles should match",
+	)
```
savil, line 85: verified with @LucilleH that it makes sense to drop `getFileNames` for StartStage.

Because:
1. The motivation to introduce `getFileNames` was to normalize the filepaths used. The filepaths would differ based on which directory the `go test` command is invoked from.
2. This doesn't apply to StartStage, since the InputFiles are relative to the BuildStage's container. And _not_ relative to the local laptop/CICD invoking `go test`.

savil (commented):
@loreto @gcurtis PTAL. Updated to use `glibc`, as per our discussion IRL.

loreto (approved):
LGTM

PR #133, opened 2022-09-15
[WIP] Add platform flag to devbox build

LucilleH:
## Summary

Add platform flag to devbox build

Example: `devbox build --platform linux/amd64`

Had to move this issue https://github.com/jetpack-io/devbox/issues/73 up to this cycle since we need it internally.

Currently running into the following error:
```
error: unable to load seccomp BPF program: Invalid argument
```

## How was it tested?
`devbox build --platform linux/amd64`


mikeland86:
https://github.com/NixOS/nix/issues/5258

I tried `--option filter-syscalls false` in dockerfile but that didn't help

PR #134, opened 2022-09-16
[CSharpPlanner] implement build functionality, albeit bloated

savil:
## Summary

Submitting an initial PR to unblock `devbox build` for C#. 

The image size is rather bloated  at 840MB for `testdata/csharp/csharp_9-dotnet_6` project.
I tried to avoid including the `dotnet-sdk` nix package in the runtime image, but 
ran into some obstacles (see notes in the code comments). I'll make another pass
at it but we can land this now so this functionality begins working.

## How was it tested?

in `testdata/csharp/csharp_9-dotnet_6`:
`devbox build && docker run devbox`
This printed the correct output (of which .Net version it is using).

PR #135, opened 2022-09-16
[DotNetPlanner] shell and build for C# and F#

savil:
## Summary

- rename csharp package to dotnet
- Add support for `.fsproj` extension to enable support for F# language.

## How was it tested?

Added a testdata case for fsharp-hello-world that does version checking....


savil:
* main
  * **PR #134** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/134" target="_blank">Graphite</a> 
    * **PR #135** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/135" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/135?utm_source=stack-comment).

cuecfg/cuecfg.go
```diff
@@ -52,7 +52,7 @@ func Unmarshal(data []byte, extension string, valuePtr any) error {
 			return errors.WithStack(err)
 		}
 		return nil
-	case ".xml", ".csproj":
+	case ".xml", ".csproj", ".fsproj":
```
savil, line 55: I don't love needing to hard code `.csproj` and `.fsproj` in this package. It is needed because the `cuecfg.ParseFiles` function derives the extension from the filename, and so there is no means of indicating what extension to use.

An alternative idea is to introduce `cuecfg.ParseFilesForExtension(filepath string, ext string)` and I can call it as `cuecfg.ParseFiles(projectPath, ".xml")`

LucilleH (approved):
🚢

PR #136, opened 2022-09-16
[Bug fix] Pick one plan if no plans are buildable

LucilleH:
## Summary
Right now in the plans merging logic we pick a buildable plan.
In this case, if all plans have error messages, we should return a plan with error messages instead of returning an empty plan.

Is this the desired behavior?

## How was it tested?
go test ./...

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -82,16 +82,13 @@ func (d *Devbox) Remove(pkgs ...string) error {
 
 // Build creates a Docker image containing a shell with the devbox environment.
 func (d *Devbox) Build(flags *docker.BuildFlags) error {
-	if ok, err := planner.IsBuildable(d.srcDir); !ok {
-		return err
-	}
 	defaultFlags := &docker.BuildFlags{
 		Name:           flags.Name,
 		DockerfilePath: filepath.Join(d.srcDir, ".devbox/gen", "Dockerfile"),
 	}
 	opts := append([]docker.BuildOptions{docker.WithFlags(defaultFlags)}, docker.WithFlags(flags))
 
-	err := d.Generate()
+	err := d.generateBuildFiles()
 	if err != nil {
 		return errors.WithStack(err)
 	}
@@ -99,7 +96,7 @@ func (d *Devbox) Build(flags *docker.BuildFlags) error {
 }
 
 // Plan creates a plan of the actions that devbox will take to generate its
-// environment.
+// shell environment.
 func (d *Devbox) Plan() (*plansdk.Plan, error) {
 	userPlan := d.convertToPlan()
 	automatedPlan, err := planner.GetPlan(d.srcDir)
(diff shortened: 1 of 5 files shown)


savil (commented):
Since PR summary asks if this is right product behavior: I don't think so, because we are just randomly picking some plan -> the idea behind a nix build system is that its deterministic so I feel we'd be violating user expectations here.

Could we collect the errors from all the relevant plans in this scenario and print them?

Alternatively, what do you think about the following:

The `Plan` interface has:
```
IsRelevant() bool
GetPlan() Plan
```
So if `IsRelevant` is true, then `GetPlan` having an error should stop the process.

loreto:
An error can't stop the process unless we more clearly separate the shell and build behavior. We had an example before were a "failing" planner, was preventing users from even shelling, even though the only thing the error prohibited was building a container.

LucilleH:
@savil I think the code is deceiving in that, when we have two or more auto-detected planners we return an error, and the build process exits. So this part of the logic will only occur when there's one and only one planner, and it has an error.

Thoughts on how to restructure the code a bit better so we can make it more clear?

savil (approved):
Oh interesting. I hadn't seen `IsBuilldable` before. Makes sense.

🤔 not ideal that we have to reason about this behavior with these two functions (which may or may not be called together). 

Should we move that IsBuildable error checking in here? What is the value in doing it separately and calling that function up front?

LucilleH:
> 
> Should we move that IsBuildable error checking in here? What is the value in doing it separately and calling that function up front?

Because we don't want to error on `devbox shell`, and we still want to merge all plans even though some of them have an error. I don't like how I had to reason about this either. Maybe I can create a wrapper function on top of this just for `devbox build` to consolidate them into one place? I'll try that

LucilleH:
@savil @loreto I separated out the build and the shell plan. Let me know if this is clearer.

devbox.go
```diff
@@ -109,30 +106,36 @@ func (d *Devbox) Plan() (*plansdk.Plan, error) {
 	return plansdk.MergeUserPlan(userPlan, automatedPlan)
 }
 
+// Plan creates a plan of the actions that devbox will take to generate its
+// shell environment.
+func (d *Devbox) BuildPlan() (*plansdk.Plan, error) {
+	userPlan := d.convertToPlan()
+	buildPlan, err := planner.FindBuildablePlan(d.srcDir)
```
savil, line 113: What do you think of the API if we were to rename as:
- `planner.GetPlan` -> `planner.GetShellPlan`
- `planner.FindBuildablePlan` -> `planner.GetBuildPlan`

This would make it analogous to `Devbox.ShellPlan()` (if you were to rename it from current `Devbox.Plan()` and `Devbox.BuildPlan()`

savil (approved):
I find this easier to reason about. Thanks!

devbox.go
```diff
@@ -82,24 +82,21 @@ func (d *Devbox) Remove(pkgs ...string) error {
...
+	err := d.generateBuildFiles()
 	if err != nil {
 		return errors.WithStack(err)
 	}
 	return docker.Build(d.srcDir, opts...)
 }
 
 // Plan creates a plan of the actions that devbox will take to generate its
-// environment.
+// shell environment.
```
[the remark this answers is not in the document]
LucilleH: Didn't want to change `devbox plan` to `devbox shell plan` or `devbox plan --shell` so I left it as is. I think by default `devbox plan` means shell plan is acceptable?

devbox.go
LucilleH: I'll do `planner.FindBuildablePlan` -> `planner.GetBuildPlan`

Didn't change `GetPlan` for the same reason as noted above

PR #140, opened 2022-09-16
[bug] Change BUILDKIT to DOCKER_BUILDKIT

LucilleH:
## Summary
Not 100% why `BUILDKIT=1` worked locally, but not in CICD. 

Looking at all the documentations, it seems everywhere it uses `DOCKER_BUILDKIT`, hence the change.

Reference: https://docs.docker.com/develop/develop-images/build_enhancements/

## How was it tested?
devbox build


mikeland86 (approved):
I think locally we already all have it enabled

PR #141, opened 2022-09-16
[ZigPlanner] Add support for shell and build

savil:
## Summary

Adds basic support for shell and build.

Build Size optimization:
- If I include the `zig` nix package in the runtimePackages, then the image size 
is really large (700MB)
- But I believe the zig executable is standalone https://github.com/ziglang/zig/issues/3660#issuecomment-552610783
- So, I get the zig-executable's name by:
  - parsing the `build.zig` file (so unfortunate) to find the `addExecutable` line
  - fallback: if I fail to find exactly one `addExecutable`, then I fallback to installing the `zig` nix package and invoking `zig build run`
- the image size is 22MB 

## How was it tested?

shell:
in the testcase folder:
```
> devbox shell

(devbox)> zig build run
# see output
info: All your codebase are belong to us.
```
Don't be alarmed! Just a printout.

build:
```
> devbox build
> docker run devbox
info: All your codebase are belong to us.
```

### DIFF
--- a/planner/languages/zig/zig_planner.go
+++ b/planner/languages/zig/zig_planner.go
@@ -38,7 +38,7 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 	if err != nil {
 		runtimePkgs = []string{"zig"}
 		startStage = &plansdk.Stage{
-			InputFiles: []string{"."},
+			InputFiles: plansdk.AllFiles(),
 			Command:    "zig build run",
 		}
 	} else {
@@ -53,7 +53,7 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 		DevPackages:     []string{"zig"},
 		RuntimePackages: runtimePkgs,
 		BuildStage: &plansdk.Stage{
-			InputFiles: []string{"."},
+			InputFiles: plansdk.AllFiles(),
 			Command:    "zig build install",
 		},
 		StartStage: startStage,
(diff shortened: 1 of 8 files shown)


savil:
* main
  * **PR #141** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/141" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/141?utm_source=stack-comment).

testdata/zig/zig-hello-world/src/main.zig
```diff
@@ -0,0 +1,9 @@
+const std = @import("std");
+
+pub fn main() anyerror!void {
+    std.log.info("All your codebase are belong to us.", .{});
```
mohsenari, line 4: Can we change this message. It's a bit grim 😄

savil: Its the default zig hello world - I think their sense of humour is  very Adams Family-esque, which I love :)

But yeah, I'll edit it to be less alarming.

planner/languages/zig/zig_planner.go
```diff
@@ -17,9 +22,65 @@ func (p *Planner) Name() string {
...
+			InputFiles: []string{"./zig-out/bin/"},
+			Command:    fmt.Sprintf("./%s", exeName),
+		}
+	}
+
+	return &plansdk.Plan{
+		DevPackages:     []string{"zig"},
+		RuntimePackages: runtimePkgs,
+		BuildStage: &plansdk.Stage{
+			InputFiles: []string{"."},
```
LucilleH, line 56: Can use `plansdk.AllFiles()` if you want

LucilleH (approved):

PR #142, opened 2022-09-16
[DotNetPlanner] add installing packages to InstallStage

savil:
## Summary

In .Net, the package manager is called NuGet. https://www.nuget.org

`dotnet restore` is the equivalent of `pip install`. I add this to the InstallStage.

And ensure we do not do `restore` during `dotnet publish` and `dotnet run` (which it implicitly does otherwise)

## How was it tested?

addded testcase that adds a package for json: https://www.nuget.org/packages/Newtonsoft.Json/
```
> devbox build
> docker run devbox
serialized json for csharp_10_dotnet_6_with_package.Product is {"Name":"Apple","Expiry":"2008-12-28T00:00:00","Sizes":["Small"]}
```


savil:
* main
  * **PR #142** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/142" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/142?utm_source=stack-comment).

PR #144, opened 2022-09-19
Allow setting name and tag in devbox build

LucilleH:
## Summary
Allow setting name and tag in devbox build

Example: 
```
devbox build --name devbox-test --tags latest
```

## How was it tested?
`devbox build`

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -81,15 +81,15 @@ func (d *Devbox) Remove(pkgs ...string) error {
 }
 
 // Build creates a Docker image containing a shell with the devbox environment.
-func (d *Devbox) Build(name string, opts ...docker.BuildOptions) error {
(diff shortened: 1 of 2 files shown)


devbox.go
```diff
@@ -81,12 +81,12 @@ func (d *Devbox) Remove(pkgs ...string) error {
 }
 
 // Build creates a Docker image containing a shell with the devbox environment.
-func (d *Devbox) Build(opts ...docker.BuildOptions) error {
+func (d *Devbox) Build(name string, opts ...docker.BuildOptions) error {
```
loreto, line 84: Instead of adding name as an argument, can we just pass it as part of `opts`?

loreto (approved):
LGTM

PR #146, opened 2022-09-19
[Java planner] part3 reducing docker image size

mohsenari:
## Summary
- Added custom JRE for runtime instead of installing JDK + maven
- Updated testdata plan
- Added build command to java planner

## How was it tested?
- devbox plan
- devbox build
- docker run devbox
- run `docker images | grep devbox` to check image size

see screenshot:

Screen Shot 2022-09-19 at 5 14 24 PM

### DIFF
--- a/planner/languages/java/java_planner.go
+++ b/planner/languages/java/java_planner.go
@@ -26,6 +26,8 @@ var jVersionMap = map[int]string{
 const defaultJava = "jdk"
 const defaultMaven = "maven"
 
+const binUtils = "binutils"
+
 // Implements interface Planner (compile-time check)
 var _ plansdk.Planner = (*Planner)(nil)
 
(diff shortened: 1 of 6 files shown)


savil (approved):
nice!

planner/languages/java/java_planner.go
```diff
@@ -56,19 +58,24 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
...
 	return &plansdk.Plan{
 		DevPackages: []string{
 			defaultMaven,
 			javaPkg,
+			binUtils,
 		},
 		RuntimePackages: []string{
-			defaultMaven,
-			javaPkg,
+			binUtils,
```
savil, line 69: just curious, does `glibc` package work as a replacement?

For Rust's image, that had a smaller size than binutils

mohsenari: @savil I tried but I got the same error as before adding `binutils`
> Error: java.io.IOException: Cannot run program "objcopy": error=2, No such file or directory

savil: could you try `jlink` with `--strip-java-debug-attributes` instead of `--strip-debug`?

Looks like that may fix this issue.
https://github.com/docker-library/openjdk/issues/351#issuecomment-524470228

PR #147, opened 2022-09-19
[pip] Add pip support (via venv) DEV-1146

mikeland86:
## Summary

This is stacked on https://github.com/jetpack-io/devbox/pull/105. Please review that one first.

Adds basic pip support. 

* Uses requirements.txt for detection.
* Uses venv for shell
* Uses venv + pex for build

Does not yet libraries with native extensions like `pandas` but there's a plan for that.

## How was it tested?

* `devbox shell`. 
* `devbox build`
* Unit tests


planner/languages/python/python_pip_planner.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return plan
+}
+
+func (p *PIPPlanner) isBuildable(srcDir string) error {
+	if plansdk.FileExists(filepath.Join(srcDir, "setup.py")) {
+		return nil
+	}
+
+	return usererr.New(
```
savil, line 62: must **be** case-insen....

savil (commented):
TIL about pex. Nice work!

lgtm, but will defer to others who are more familiar with python environments.

devbox_test.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 func TestDevbox(t *testing.T) {
+	os.Setenv("TMPDIR", "/tmp")
```
gcurtis, line 15: If you use [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) it'll automatically restore the original value and guard against changing the value in parallel tests.

It might also be worth checking that `/tmp` exists before using it so we can provide a helpful error.

gcurtis (approved):

PR #149, opened 2022-09-19
devbox: add AppendScript method to ConfigShellCmds

gcurtis:
## Summary

This method takes a script as a string and appends it to the slice of shell commands in the config. From the docs:

	AppendScript appends each line of a script to s.Cmds. It also
	applies the following formatting rules:

	  - Trim leading newlines from the script.
	  - Trim trailing whitespace from the script.
	  - If the first line of the script begins with one or more tabs ('\t'), then unindent each line by that same number of tabs.
	  - Remove trailing whitespace from each line.

	Note that unindenting only happens when a line starts with at
	least as many tabs as the first line. If it starts with
	fewer tabs, then it is not unindented at all.

See `config_test.go` for an example.

## How was it tested?

Added tests to `config_test.go`.


mohsenari (commented):
`AppendScript()` is not used anywhere yet. Do you plan to use it in a sample use case?

config_test.go
```diff
@@ -145,3 +146,106 @@ func TestConfigShellCmdsString(t *testing.T) {
...
+		},
+		{
+			name:     "LeadingNewline",
+			script:   "\necho test",
+			wantCmds: []string{"echo test"},
+		},
+		{
+			name:     "LeadingNewlineAndSpace",
+			script:   "\n    echo test",
+			wantCmds: []string{"    echo test"},
```
ipince, line 206: why trim leading tabs but not leading spaces?

ipince (approved):

gcurtis:
> `AppendScript()` is not used anywhere yet. Do you plan to use it in a sample use case?

I plan on replacing things like the prompt commands here - https://github.com/jetpack-io/devbox/pull/145/files#diff-93be3f6a535ea60a829a423584b2d743d499267a0d6ab1ec0ed8be028c0e9866R153-R162. I think it just makes things a little easier to read.

config_test.go
gcurtis: Updated to also trim leading spaces from the script.

gcurtis: It only unindents by tabs because Go source is always indented with tabs. I figured there wouldn't really be a case where we'd want to indent using tabs.

PR #152, opened 2022-09-20
boxcli: increase shell test timeout to 3 mins

gcurtis:
## Summary

This is a workaround for some test runs that are failing due to how long it takes to install Nix packages. We need to figure out how to cache these instead.

## How was it tested?

CI


mikeland86 (approved):

loreto (approved):
LGTM

PR #153, opened 2022-09-20
[planner] Add typescript support to nodejs build

LucilleH:
## Summary
Add typescript support to nodejs build. I don't think we need a separate typescript planner, as most typescript projects need nodejs and `package.json` to function properly.

## How was it tested?
`go test ./...`


savil (approved):
sweet!

PR #155, opened 2022-09-20
[HaskellPlanner] enable shell and build for Stack framework

savil:
## Summary

This PR enables **basic** Haskell support using the Stack framework.
Both `devbox shell` and `devbox build` should work.

Limitations:
- haskell (ghc) version: We currently support the latest versions of haskell. Have not implemented
  parsing `stack.yaml` or `package.yaml` or `<project-name>.cabal` for the haskell
  version and installing just that.

- image size: The Runtime docker image is quite large (2 GB). I added a comment
inline on a way to possibly reduce this size, but haven't fully verified it will
work in all circumstances.

## How was it tested?

Added a basic hello world example using `stack new hello-world` generated code.

did:
- `devbox shell`  and then `stack exec hello-world-exe
- `devbox build` -> ran with `docker run devbox`


savil:
* main
  * **PR #155** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/155" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/155?utm_source=stack-comment).

testdata/haskell/hello-world/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Author name here (c) 2022
```
LucilleH, line 1: ✂️

LucilleH (approved):
This is beyond my knowledge so take my approval with a grain of salt

PR #156, opened 2022-09-21
Expose nix package path as environment variables

LucilleH:
## Summary
Expose nix package path as environment variables
This is valuable if a shell hook wants to set default values that points to a nix package path.

The environment variable will have the pattern of `DEVBOX_<pkg_name>`
eg. 
```
MYSQL_BASEDIR=${DEVBOX_mariadb}
```

Example variables:
Screen Shot 2022-09-20 at 6 02 36 PM

Example values
Screen Shot 2022-09-20 at 6 02 45 PM

## How was it tested?
`devbox shell`


LucilleH:
Closing this draft as we don't need this for now.

PR #158, opened 2022-09-21
devbox shell -- cmd

loreto:
## Summary
Simple implementation of `devbox shell -- <cmd>`. I struggled integrating this with all the other stuff that is happening with `devbox shell`. Would like thoughts on whether this is a good enough for an MVP of the feature, and later we improve it, or whether we think integrating with the other `devbox shell` logic is a must have.

## How was it tested?
Ran a few commands with `devbox shell --`

### DIFF
--- a/boxcli/shell.go
+++ b/boxcli/shell.go
@@ -15,9 +15,9 @@ import (
 
 func ShellCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:               "shell [<dir>]",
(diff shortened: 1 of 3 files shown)


gcurtis (approved):
I don't think integrating with the other shell stuff is super important right away.

- You're not using `--pure` so the locale and other env vars will carry over.
- It's not an interactive shell, so the user's shellrc won't run.
- The shell init hook won't run, but that's probably desired when running a single script/command.

PR #167, opened 2022-09-22
Elixir Planner

Lagoja:
## Summary

Add Build support and basic shell support for Elixir projects. Added this mostly as a learning exercise so I can write a short guide to creating planners in the future. 

Right now, it detects Elixir projects by searching for a `mix.exs` file, and then creates a standalone release of the project. Resulting image size is 200MB (which is mostly Nix Packages, and seems larger than expected...)

Future Updates: 
* Update the shell so that packages are installed and managed locally with Hex + Rebar
* [Phoenix Framework](https://phoenixframework.org/) support
* Make the image size smaller

## How was it tested?

Used the test Elixir project in `/testdata`


planner/languages/elixir/elixir_planner.go
```diff
@@ -3,10 +3,47 @@
...
+var versionMap = map[string]string{
+	"1.9":  "elixir_1_9",
+	"1.10": "elixir_1_10",
+	"1.11": "elixir_1_11",
+	"1.12": "elixir_1_12",
+	"1.13": "elixir",
+	"1.14": "elixir_1_14",
+}
+
+type ElixirProject struct {
```
savil, line 32: does this need to be visible outside this `elixir` package? if not, lets rename to `elixirProject` to keep it private.

planner/languages/elixir/elixir_planner.go
```diff
@@ -3,10 +3,47 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"go.jetpack.io/devbox/planner/plansdk"
+)
 
 type Planner struct{}
 
+var NoElixirVersionSetErr = errors.New("No version set in mix.exs")
```
savil, line 21: rename to `noElixirVersionSetErr` to keep private to this package?

savil (approved):
nice!

planner/languages/elixir/elixir_planner.go
Lagoja: done

planner/languages/elixir/elixir_planner.go
```diff
@@ -15,9 +52,96 @@ func (p *Planner) Name() string {
...
-	return false
+	mixPath := filepath.Join(srcDir, "mix.exs")
+	return plansdk.FileExists(mixPath)
 }
 
 func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
-	return &plansdk.Plan{}
+	elixirProject, err := getElixirProject(srcDir)
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
Lagoja: Does returning nil cancel the build at the top level? Would be good to document how to eject if something goes wrong in the Planner

PR #171, opened 2022-09-23
[plan] Plan.Definitions add omitempty

savil:
## Summary

I'm adding omitempty to Plan.Definitions.

I'm not adding it to Plan.DevPackages and Plan.RuntimePackages. Because it explains to a reader that these are empty. These fields are so core and commonly used that I think we want to make an exception for them.

## How was it tested?

compiles

### DIFF
--- a/planner/plansdk/plansdk.go
+++ b/planner/plansdk/plansdk.go
@@ -51,7 +51,7 @@ type Plan struct {
 	// Ex: python main.py
 	StartStage *Stage `json:"start_stage,omitempty"`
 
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #171** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/171?utm_source=stack-comment).

savil:
I'll update all the testcases separately.

gcurtis (approved):

PR #172, opened 2022-09-23
Change welcome message to shell init hook

LucilleH:
## Summary
This PR changes the following:
- Change welcome message to shell init hook
- Change `Definitions` to `packageExtensions`

## How was it tested?
go build
go test ./...
devbox shell

### DIFF
--- a/planner/languages/nginx/nginx_planner.go
+++ b/planner/languages/nginx/nginx_planner.go
@@ -29,10 +29,8 @@ func (p *Planner) IsRelevant(srcDir string) bool {
 
 func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 	return &plansdk.Plan{
-		ShellWelcomeMessage: fmt.Sprintf(welcomeMessage, p.shellConfig(srcDir)),
 		DevPackages: []string{
 			"nginx",
-			"shell-nginx",
 		},
 		RuntimePackages: []string{
 			"nginx",
@@ -45,11 +43,12 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 			Command:    fmt.Sprintf(startCommand, p.buildConfig(srcDir)),
 			InputFiles: plansdk.AllFiles(),
 		},
-		Definitions: []string{
-			fmt.Sprintf(nginxShellStartScript, srcDir, p.shellConfig(srcDir)),
-		},
 		GeneratedFiles: map[string]string{
 			"shell-helper-nginx.conf": fmt.Sprintf(shellHelperNginxConfig, os.TempDir()),
+			"shell-nginx.sh":          fmt.Sprintf(nginxShellStartScript, srcDir, p.shellConfig(srcDir)),
+		},
+		ShellInitHook: []string{
+			fmt.Sprintf(". %s", filepath.Join(srcDir, ".devbox/gen/shell-nginx.sh")),
 		},
 	}
 }
@@ -68,18 +67,6 @@ func (p *Planner) buildConfig(srcDir string) string {
 	return "shell-nginx.conf"
 }
 
-const welcomeMessage = `
-##### WARNING: nginx planner is experimental #####
-
-You may need to add
-
-\"include ./.devbox/gen/shell-helper-nginx.conf;\"
-
-to your %s file to ensure the server can start in the nix shell.
-
-Use \"shell-nginx\" to start the server
-`
-
 var startCommand = strings.TrimSpace(`
 	addgroup --system --gid 101 nginx && \
 	adduser --system --ingroup nginx --no-create-home --home /nonexistent --gecos "nginx user" --shell /bin/false --uid 101 nginx && \
@@ -92,12 +79,35 @@ var startCommand = strings.TrimSpace(`
 `)
 
 const nginxShellStartScript = `
-shell-nginx = pkgs.writeShellScriptBin "shell-nginx" ''
+#!/bin/bash
 
-echo "Starting nginx with command:"
-echo "nginx -p %[1]s -c %[2]s -e /tmp/error.log -g \"pid /tmp/mynginx.pid;daemon off;\""
-nginx -p %[1]s -c %[2]s -e /tmp/error.log -g "pid /tmp/shell-nginx.pid;daemon off;"
-'';`
+welcome() {
+	echo "##### WARNING: nginx planner is experimental #####
+
+	You may need to add
+
+	\"include shell-helper-nginx.conf;\"
+
+	to your %[2]s file to ensure the server can start in the nix shell.
+
+	Use \"shell-nginx\" to start the server"
+}
+
+pkg_path=$(readlink -f $(which nginx) | sed -r "s/\/bin\/nginx//g")
+conf_path=$pkg_path/conf
+
+mkdir -p %[1]s/.devbox/gen/nginx
+ln -sf $conf_path/* %[1]s/.devbox/gen/nginx/
+ln -sf $(pwd)/%[1]s/.devbox/gen/shell-helper-nginx.conf %[1]s/.devbox/gen/nginx/shell-helper-nginx.conf
+for file in %[1]s/*; do if [[ ! $file = .devbox ]]; then ln -sf $(pwd)/%[1]s/$file .devbox/gen/nginx/$file; fi; done
+
+shell-nginx() {
+	echo "Starting nginx with command:"
+	echo "nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g \"pid /tmp/mynginx.pid;daemon off;\""
+	nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g "pid /tmp/shell-nginx.pid;daemon off;"
+}
+welcome
+`
 
 const shellHelperNginxConfig = `access_log %[1]s/access.log;
 client_body_temp_path %[1]s/client_body;
--- a/planner/languages/php/php_planner.go
+++ b/planner/languages/php/php_planner.go
@@ -52,7 +52,7 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 			fmt.Sprintf("php%s", v.MajorMinorConcatenated()),
 			fmt.Sprintf("php%sPackages.composer", v.MajorMinorConcatenated()),
 		},
-		Definitions: p.definitions(srcDir, v),
+		PackageExtensions: p.pkgExtensions(srcDir, v),
 	}
 	if !plansdk.FileExists(filepath.Join(srcDir, "public/index.php")) {
 		return plan.WithError(usererr.New("Can't build. No public/index.php found."))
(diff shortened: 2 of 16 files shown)


planner/plansdk/plansdk.go
```diff
@@ -26,18 +26,17 @@ type PlanError struct {
...
 	// its development environment. They are also available in shell.
 	DevPackages []string `cue:"[...string]" json:"dev_packages"`
-
 	// RuntimePackages is the slice of Nix packages that devbox makes available in
 	// in both the development environment and the final container that runs the
 	// application.
 	RuntimePackages []string `cue:"[...string]" json:"runtime_packages"`
+	// packageExtensions is the slice of Nix packages that devbox wants to extend
+	// to include extra packages that needs global installation.
+	PackageExtensions []string `cue:"[...string]" json:"package_extensions,omitempty"`
```
LucilleH, line 39: @mikeland86 for php, I wonder if we can use the same mechanism I did for nginx, in that we can create a folder `.devbox/gen/php/` and add it to PATH, symlink the content it to the nix store, the folder is not immutable, thus they can install packages there instead?

mikeland86:
FYI, this PR already replaces welcome messages with a planner shell hook https://github.com/jetpack-io/devbox/pull/147 

I had put the user shell hook and the planner shell hook separate, but I don't feel strongly on either one. Single shell hook may be simpler which is usually better.

planner/languages/nginx/nginx_planner.go
```diff
@@ -45,11 +43,12 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
...
 		},
-		Definitions: []string{
-			fmt.Sprintf(nginxShellStartScript, srcDir, p.shellConfig(srcDir)),
-		},
 		GeneratedFiles: map[string]string{
 			"shell-helper-nginx.conf": fmt.Sprintf(shellHelperNginxConfig, os.TempDir()),
+			"shell-nginx.sh":          fmt.Sprintf(nginxShellStartScript, srcDir, p.shellConfig(srcDir)),
+		},
+		ShellInitHook: []string{
+			fmt.Sprintf(". %s", filepath.Join(srcDir, ".devbox/gen/shell-nginx.sh")),
```
mikeland86, line 51: I'm a bit unsure about splitting this out in this way. It feels a bit fragile. I almost want an option that does both (create a script and either calls it or sources it in the init hook)

mikeland86: Also, this should probably not be called `shell-nginx.sh` it it's an init script. Maybe just `init.sh` ?

planner/plansdk/plansdk.go
mikeland86: I think you may be misunderstanding what these do in php. They are not installing packages, they are recompiling PHP with the appropriate extensions.  Some extensions are included by default but need to be "enabled". Other extensions are not included at all and require php to be built from source. 


There's no way to just add an extension into a folder. The actual source of PHP needs to change.

LucilleH:
@mikeland86 oh nice. Do you want to merge yours first and I can rebase? This PR also includes removing the need for `definitions` in nginx planner

planner/plansdk/plansdk.go
LucilleH: We wouldn't need to recompile PHP if the store is mutable?

planner/languages/php/php_planner.go
```diff
@@ -52,7 +52,7 @@ func (p *Planner) GetPlan(srcDir string) *plansdk.Plan {
 			fmt.Sprintf("php%s", v.MajorMinorConcatenated()),
 			fmt.Sprintf("php%sPackages.composer", v.MajorMinorConcatenated()),
 		},
-		Definitions: p.definitions(srcDir, v),
+		PackageExtensions: p.pkgExtensions(srcDir, v),
```
mikeland86, line 55: Should we call these NixLocalVariables instead? extending packages is just one use case. 

You could argue that we should avoid adding stuff here because it makes it harder to do nix free version, but would probably prefer a name that more closely resembles what it represents in the nix file.

planner/languages/nginx/nginx_planner.go
```diff
@@ -92,12 +79,35 @@ var startCommand = strings.TrimSpace(`
...
+ln -sf $conf_path/* %[1]s/.devbox/gen/nginx/
+ln -sf $(pwd)/%[1]s/.devbox/gen/shell-helper-nginx.conf %[1]s/.devbox/gen/nginx/shell-helper-nginx.conf
+for file in %[1]s/*; do if [[ ! $file = .devbox ]]; then ln -sf $(pwd)/%[1]s/$file .devbox/gen/nginx/$file; fi; done
+
+shell-nginx() {
+	echo "Starting nginx with command:"
+	echo "nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g \"pid /tmp/mynginx.pid;daemon off;\""
+	nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g "pid /tmp/shell-nginx.pid;daemon off;"
+}
+welcome
```
mikeland86, line 109: This is long enough that is could benefit from being in own `sh` file and we can embed it.

mikeland86 (approved):
Merging plan hook and user hook seems fine to me, but we should rename the final hook to be more generic (e.g. ShellHook or ShellInitHook). Right now it is called `UserHook` which can be a bit misleading. We should also change the generated comments to indicate that as well.

planner/languages/php/php_planner.go
LucilleH: `NixLocalVariables` would work 👍

planner/languages/nginx/nginx_planner.go
```diff
@@ -92,12 +79,35 @@ var startCommand = strings.TrimSpace(`
...
+	You may need to add
+
+	\"include shell-helper-nginx.conf;\"
+
+	to your %[2]s file to ensure the server can start in the nix shell.
+
+	Use \"shell-nginx\" to start the server"
+}
+
+pkg_path=$(readlink -f $(which nginx) | sed -r "s/\/bin\/nginx//g")
```
gcurtis, line 96:
```suggestion
pkg_path=$(readlink -f "$(which nginx)" | sed -r "s/\/bin\/nginx//g")
```

planner/languages/nginx/nginx_planner.go
```diff
@@ -92,12 +79,35 @@ var startCommand = strings.TrimSpace(`
...
+conf_path=$pkg_path/conf
+
+mkdir -p %[1]s/.devbox/gen/nginx
+ln -sf $conf_path/* %[1]s/.devbox/gen/nginx/
+ln -sf $(pwd)/%[1]s/.devbox/gen/shell-helper-nginx.conf %[1]s/.devbox/gen/nginx/shell-helper-nginx.conf
+for file in %[1]s/*; do if [[ ! $file = .devbox ]]; then ln -sf $(pwd)/%[1]s/$file .devbox/gen/nginx/$file; fi; done
```
gcurtis, line 102:
```suggestion
mkdir -p "%[1]s/.devbox/gen/nginx"
ln -sf "$conf_path/*" "%[1]s/.devbox/gen/nginx/"
ln -sf "$(pwd)/%[1]s/.devbox/gen/shell-helper-nginx.conf" "%[1]s/.devbox/gen/nginx/shell-helper-nginx.conf"
for file in "%[1]s"/*; do if [[ ! "$file" = .devbox ]]; then ln -sf "$(pwd)/%[1]s/$file" ".devbox/gen/nginx/$file"; fi; done
```

planner/languages/nginx/nginx_planner.go
```diff
@@ -92,12 +79,35 @@ var startCommand = strings.TrimSpace(`
...
+
+shell-nginx() {
+	echo "Starting nginx with command:"
+	echo "nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g \"pid /tmp/mynginx.pid;daemon off;\""
+	nginx -p %[1]s -c %[1]s/.devbox/gen/nginx/%[2]s -e /tmp/error.log -g "pid /tmp/shell-nginx.pid;daemon off;"
```
gcurtis, line 107:
```suggestion
	nginx -p "%[1]s" -c "%[1]s/.devbox/gen/nginx/%[2]s" -e /tmp/error.log -g "pid /tmp/shell-nginx.pid;daemon off;"
```

planner/plansdk/plansdk.go
```diff
@@ -50,15 +49,15 @@ type Plan struct {
...
 	// Errors from plan generation. This usually means
 	// the user application may not be buildable.
 	Errors []PlanError `json:"errors,omitempty"`
-
 	// GeneratedFiles is a map of name => content for files that should be generated
 	// in the .devbox/gen directory. (Use string to make it marshalled version nicer.)
 	GeneratedFiles map[string]string `json:"generated_files,omitempty"`
+	// This is an array of shell init hook that gets appended in front of
+	// the shell { initHook } defined by users.
+	ShellInitHook []string `cue:"[...string]" json:"shell_init_hook"`
```
gcurtis, line 60:
```suggestion

	// ShellInitHook is a slice of shell commands to execute before the
	// user's shell init hook. They only run in interactive devbox shells.
	ShellInitHook []string `cue:"[...string]" json:"shell_init_hook"`
```

planner/plansdk/plansdk.go
```diff
@@ -144,7 +144,9 @@ func findBuildablePlan(plans ...*Plan) *Plan {
 	for _, p := range plans {
 		// For now, pick the first buildable plan.
 		if p.Buildable() {
-			return p
+			// Make a copy so that the original plan is not modified.
+			newP := *p
+			return &newP
```
gcurtis, line 149: I don't know if this matters for this change, but `ShellInitHook` looks like it's the only field that will be copied. You'd need to do a deep copy if you want the pointers/slices/maps to be copied as well.

gcurtis (approved):
This is probably just because of merge conflicts, but I think it would be nice to keep the user's shell commands and our commands separate with `planInitHook`. Otherwise LGTM.

LucilleH:
> This is probably just because of merge conflicts, but I think it would be nice to keep the user's shell commands and our commands separate with `planInitHook`. Otherwise LGTM.

Yeah, I'm actually rethinking this - I might abandon this entirely, but, I feel like so much of the script is because of the immutability of the nix store. I wonder if I can leave the global nix store as is (immutable), but create a local mutable store per project (let's say `.devbox/store`) and symlink the bins to `/nix/store`.

I might give that a try.

loreto:
> I wonder if I can leave the global nix store as is (immutable), but create a local mutable store per project (let's say `.devbox/store`) and symlink the bins to `/nix/store`. And the planners decide whether a mutable store is needed.
> 
> I might give that a try.

I believe that if we re-implement `devbox add` using nix profiles, that will happen automatically.

LucilleH:
> 
> I believe that if we re-implement `devbox add` using nix profiles, that will happen automatically.

@loreto so does that mean each nix profile will be mapped to a devbox project? 🤔

loreto:
> > I believe that if we re-implement `devbox add` using nix profiles, that will happen automatically.
> 
> @loreto so does that mean each nix profile will be mapped to a devbox project? 🤔 Or each devbox.json picks the right nix profile?

Each devbox project would have it's own nix profile, living in `.devbox/profile`

LucilleH:
> 
> Each devbox project would have it's own nix profile, living in `.devbox/profile`

Oh perfect!!

LucilleH:
closing this - `.devbox/profile` is a better way to solve it.

PR #176, opened 2022-09-25
nix: fix shellrc PATH word splitting in bash

### LINKED ISSUE #175: $PATH is broken in shell
Trying out devbox for the first time, but I cannot get it to work. The PATH environment variable in the shell is broken:
```
[roysten@arch RedditReader]$ echo $PATH
/home/roysten/.nix-profile/bin:/nix/var/nix/profiles/default/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
[roysten@arch RedditReader]$ devbox shell
Starting a devbox shell...
(devbox) [roysten@arch RedditReader]$ echo $PATH
/nix/store/0h73sj1n8hzc6fs36cjvsvcvz3av7n47-bash-interactive-5.1-p16/bin /nix/store/s6isdp5wxcz0hfcj72wfb9wvpkkh7ppg-nodejs-16.16.0/bin /nix/store/njzl18r4mc1b82jd172nlg1gs45z3g1h-patchelf-0.14.5/bin /nix/store/kia8xygy2r9iblwbjrl163j4bpz84wz7-gcc-wrapper-11.3.0/bin /nix/store/42izybv2fz3ps5pcg3dvcmn85pc9406y-gcc-11.3.0/bin /nix/store/fg85dscd4408xg9wc12paxxnkp69k8ix-glibc-2.35-163-bin/bin /nix/store/xp5z3k851fs7haqbcwqax1hh4pynzla9-coreutils-9.1/bin /nix/store/klgmnpyyan0gg6ra6d40qmjqs39cj1cf-binutils-wrapper-2.38/bin /nix/store/dpxy5cr7cm0nch0qvq2qygs9jwnmhiy5-binutils-2.38/bin /nix/store/xp5z3k851fs7haqbcwqax1hh4pynzla9-coreutils-9.1/bin /nix/store/b1c0270m7fcppqpga49fqy7n8hj3vbw2-findutils-4.9.0/bin /nix/store/73rp8gam5bpihafxbk3x39dza2iw3lzi-diffutils-3.8/bin /nix/store/cr41h6cg093d7z6w0nnaiqpsfc2fxd54-gnused-4.8/bin /nix/store/74lg75l5wdxkiab73pc7x3983lcppk86-gnugrep-3.7/bin /nix/store/d7nppmazji1gcg1nmp6nvsp59ihxadyn-gawk-5.1.1/bin /nix/store/j0h50va2scdc60gd20ib41awvl6dvfcp-gnutar-1.34/bin /nix/store/0b07qazapyf3vpaikkmnirxxd1dvjb3b-gzip-1.12/bin /nix/store/ir6kvyhfrpm20pqz3d7wxrbyrqd6ylkc-bzip2-1.0.8-bin/bin /nix/store/xwg1p2s1mcl3qj193yw1c3sv0cpqlz5s-gnumake-4.3/bin /nix/store/2r9n7fz1rxq088j6mi5s7izxdria6d5f-bash-5.1-p16/bin /nix/store/jgj30km7hyvmwrlq4yqdf74dx6apqypj-patch-2.7.6/bin /nix/store/r75z4y813x4924czbnkhmj66f4m5l0y7-xz-5.2.5-bin/bin /nix/store/01xvwqinks2bm4y8pxw7y6cqrp36bb7s-file-5.42/bin /usr/local/sbin /usr/local/bin /usr/bin /usr/lib/jvm/default/bin /usr/bin/site_perl /usr/bin/vendor_perl /usr/bin/core_perl:
(devbox) [roysten@arch RedditReader]$ ls
bash: ls: command not found
[roysten@arch RedditReader]$ devbox version
0.0.6
```
It seems to use spaces to separate paths but it should use semicolons? My devbox.json:
```json
{
  "packages": [
    "nodejs-16_x"
  ],
  "shell": {
    "init_hook": null
  }
}
```

gcurtis:
## Summary

Fix the loop in `shellrc.tmpl` so that `devbox shell` doesn't end up with a broken `PATH` delimited by spaces instead of colons. This was due to a subtle bug related to how bash and zsh do word splitting differently. Read on for an in-depth explanation.

We fix up the `PATH` in `shellrc.tmpl` by looping over each `PATH` directory. To get each dir, we set `IFS=:` and rely on the shell's word splitting to loop over `PATH`. A caveat to this approach is that zsh doesn't follow the POSIX behavior for word splitting. Here's an example showing the difference:

bash:

	bash$ export x=a:b:c
	bash$ export IFS=:
	bash$ echo $x
	a b c

zsh:

	zsh$ export x=a:b:c
	zsh$ export IFS=:
	zsh$ echo $x
	a:b:c

To workaround this, we use command substitution to force zsh to word split:

	zsh$ export x=a:b:c
	zsh$ export IFS=:
	zsh$ echo $(echo $x)
	a b c

There's a subtle bug here that causes the splitting to break in bash because the splitting happens too soon (note the single quotes around the first echo output):

	bash$ export x=a:b:c
	bash$ export IFS=:
	bash$ set -x; echo $(echo $x)
	++ echo a b c
	+ echo 'a b c' a b c

The subshell splits `$x`, resulting in 'a b c'. The parent shell doesn't split 'a b c' because it has already been split and doesn't have any colons. This doesn't happen in zsh, because zsh doesn't split unquoted variables. Quoting `$x` stops bash from splitting in the subshell, fixing this for both shells:

	bash$ export x=a:b:c
	bash$ export IFS=:
	bash$ set -x; echo $(echo "$x")
	++ echo a:b:c
	+ echo a b c a b c

Fixes #175.

## How was it tested?

Launched a `devbox shell` in bash and zsh and made sure both `PATH`s were correct.

### DIFF
--- a/nix/shellrc.tmpl
+++ b/nix/shellrc.tmpl
@@ -51,7 +51,7 @@ content readable.
 # 	  different shells.
 PATH="$(
 	IFS=:
-	for path_dir in $(echo $PATH); do
+	for path_dir in $(echo "$PATH"); do
 		case "$path_dir" in
 			$NIX_STORE/*) nix_path="${nix_path:+$nix_path:}${path_dir}" ;;
 			*) non_nix_path="${non_nix_path:+$non_nix_path:}${path_dir}" ;;
--- a/nix/testdata/shellrc/basic/shellrc.golden
+++ b/nix/testdata/shellrc/basic/shellrc.golden
@@ -65,7 +65,7 @@ zstyle ':completion:*:kill:*' command 'ps -u $USER -o pid,%cpu,tty,cputime,cmd'
 # 	  different shells.
 PATH="$(
 	IFS=:
-	for path_dir in $(echo $PATH); do
+	for path_dir in $(echo "$PATH"); do
 		case "$path_dir" in
 			$NIX_STORE/*) nix_path="${nix_path:+$nix_path:}${path_dir}" ;;
 			*) non_nix_path="${non_nix_path:+$non_nix_path:}${path_dir}" ;;
--- a/nix/testdata/shellrc/nohook/shellrc.golden
+++ b/nix/testdata/shellrc/nohook/shellrc.golden
@@ -65,7 +65,7 @@ zstyle ':completion:*:kill:*' command 'ps -u $USER -o pid,%cpu,tty,cputime,cmd'
 # 	  different shells.
 PATH="$(
 	IFS=:
-	for path_dir in $(echo $PATH); do
+	for path_dir in $(echo "$PATH"); do
 		case "$path_dir" in
 			$NIX_STORE/*) nix_path="${nix_path:+$nix_path:}${path_dir}" ;;
 			*) non_nix_path="${non_nix_path:+$non_nix_path:}${path_dir}" ;;
--- a/nix/testdata/shellrc/noshellrc/shellrc.golden
+++ b/nix/testdata/shellrc/noshellrc/shellrc.golden
@@ -23,7 +23,7 @@
 # 	  different shells.
 PATH="$(
 	IFS=:
-	for path_dir in $(echo $PATH); do
+	for path_dir in $(echo "$PATH"); do
 		case "$path_dir" in
 			$NIX_STORE/*) nix_path="${nix_path:+$nix_path:}${path_dir}" ;;
 			*) non_nix_path="${non_nix_path:+$non_nix_path:}${path_dir}" ;;


gcurtis:
@loreto this is probably hotfix worthy.

loreto (approved):
LGTM

ipince (approved):
Thanks for the thorough explanation!

loreto:
Yeah, we should do another release today so it includes this PR.

loreto:
@gcurtis As a separate PR, is there a test we could add that would have caught this?

gcurtis:
We should add some unit tests that exercise the shell script(s). The existing tests didn't catch it because the shellrc isn't run in non-interactive terminals.

PR #183, opened 2022-09-28
Support Rails in Ruby planner

ipince:
## Summary

Opted for bundling dependencies in "deployment" mode, which puts them in `/vendor`, kind of like `node_modules` and makes installation easy.

The files in `testdata/ruby/rails` are a sample project generated with `rails new`.

## How was it tested?
Tested both `shell` and `build` in both `ruby-3.1` and `rails` test directories.


planner/plansdk/plansdk.go
```diff
@@ -113,6 +114,10 @@ func (p *Plan) WithError(err error) *Plan {
 	return p
 }
 
+// MergePlans merges multiple Plans into one. The merged plan's packages, definitions,
+// and overlays is the union of the packages, definitions, and overlays of the input plans,
+// respectively. The install/build/start stages of the merged plans are taken from the _first_
+// buildable plan (order matters!). If no plan is buildable, returns a non-buildable plan.
```
LucilleH, line 120: `If no plan is buildable, returns a non-buildable plan.` Should be `If no plan is buildable, returns an empty plan.`

LucilleH (approved):
So many files 😆

planner/plansdk/plansdk.go
ipince: Well.. no. It returns a plan with dependencies, but with no "stages". And "no stages" == "not buildable"...

right?

PR #187, opened 2022-09-29
Update Python Docs

Lagoja:
## Summary
Add details in the Python docs on PIP support

## How was it tested?
Locally


loreto (approved):
LGTM

PR #188, opened 2022-09-29
[shell] add packages within shell, leveraging nix-profile

savil:
## Summary

**Motivation:**
We can define a custom nix-profile, and install packages scoped to this profile.

The goal is to make it easier to add packages when within devbox shell. Currently,
users need to exit the shell and restart it for the nix package to be installed - clearly
not ideal. With a nix-profile, if we install the package using that profile, then the
package's binary should be downloaded and visible within the shell.

Within `devbox shell`:
- previously, we would manually add the path to each package to `PATH`.
- now, we can add the path to the nix-profile to `PATH` and skip manually adding each package.

**Limitations:**
1. For `devbox add` within a `devbox shell`, the user needs to manually execute `hash -r` to ensure the latest binaries are visible. 
2. ~For `devbox rm` within a `devbox shell`, the user still needs to restart the shell. I'm not quite sure why the `nix-env --profile <profile dir> --uninstall <pkg>` isn't working as intended.~ Fixed via reapplying `nix-env --profile <profile> -if development.nix`
3. `devbox add` and `rm` when in a shell started from a parent directory will not work. 
    - A possible fix is to set `DEVBOX_CONFIG_DIR` env-var when inside a shell, and use that value in `devbox add` and `rm`.
    - Example:
```
> cd testdata/rust/
> devbox shell rust-stable
(devbox)> devbox add go_1_18
```
  

## How was it tested?

- [x] in `testdata/rust/rust-stable`:
  - open `devbox shell`
  - verify openssl not installed via `which openssl` and getting `/usr/bin/openssl`
  - do `devbox add openssl`
     - note that `which openssl` still reflects the non nix-store location. It has NOT yet been updated.
  - do `hash -r` and then `which openssl` and see it point to a nix-store location
  - do `devbox rm openssl`  
    - note that `which openssl` still reflects the nix-store location. It has NOT yet been updated.   
  - restart `devbox shell` and do `which openssl` to verify it has been removed

### DIFF
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -4,6 +4,9 @@
 package boxcli
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
@@ -27,6 +30,31 @@ func addCmdFunc() runFunc {
 		if err != nil {
 			return errors.WithStack(err)
 		}
-		return box.Add(args...)
+
+		if err = box.Add(args...); err != nil {
+			return err
+		}
+
+		if err := box.Generate(); err != nil {
+			return err
+		}
+
+		fmt.Print("Installing nix packages. This may take a while...")
+		if err = installDevPackages(box.SourceDir()); err != nil {
+			fmt.Println()
+			return err
+		}
+		fmt.Println("done.")
+
+		if isDevboxShellEnabled() {
+			successMsg := fmt.Sprintf("%s is now installed.", args[0])
+			if len(args) > 1 {
+				successMsg = fmt.Sprintf("%s are now installed.", strings.Join(args, ", "))
+			}
+			fmt.Print(successMsg)
+			fmt.Println(" Run `hash -r` to ensure your shell is updated.")
+		}
+
+		return nil
 	}
 }
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -4,8 +4,7 @@
 package boxcli
 
 import (
-	"fmt"
-	"strings"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -26,35 +25,11 @@ func AddCmd() *cobra.Command {
 
 func addCmdFunc() runFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		box, err := devbox.Open(".")
+		box, err := devbox.Open(".", os.Stdout)
 		if err != nil {
 			return errors.WithStack(err)
 		}
 
-		if err = box.Add(args...); err != nil {
-			return err
-		}
-
-		if err := box.Generate(); err != nil {
-			return err
-		}
-
-		fmt.Print("Installing nix packages. This may take a while...")
-		if err = installDevPackages(box.SourceDir()); err != nil {
-			fmt.Println()
-			return err
-		}
-		fmt.Println("done.")
-
-		if isDevboxShellEnabled() {
-			successMsg := fmt.Sprintf("%s is now installed.", args[0])
-			if len(args) > 1 {
-				successMsg = fmt.Sprintf("%s are now installed.", strings.Join(args, ", "))
-			}
-			fmt.Print(successMsg)
-			fmt.Println(" Run `hash -r` to ensure your shell is updated.")
-		}
-
-		return nil
+		return box.Add(args...)
 	}
 }
--- a/boxcli/rm.go
+++ b/boxcli/rm.go
@@ -4,8 +4,7 @@
 package boxcli
 
 import (
-	"fmt"
-	"strings"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -23,35 +22,10 @@ func RemoveCmd() *cobra.Command {
 }
 
 func runRemoveCmd(cmd *cobra.Command, args []string) error {
-	box, err := devbox.Open(".")
+	box, err := devbox.Open(".", os.Stdout)
 	if err != nil {
 		return errors.WithStack(err)
 	}
 
-	if err = box.Remove(args...); err != nil {
-		return err
-	}
-
-	if err := box.Generate(); err != nil {
-		return err
-	}
-
-	fmt.Print("Uninstalling nix packages. This may take a while...")
-	// We need to reinstall the packages
-	if err = installDevPackages(box.SourceDir()); err != nil {
-		fmt.Println()
-		return err
-	}
-	fmt.Println("done.")
-
-	if isDevboxShellEnabled() {
-		successMsg := fmt.Sprintf("%s is now removed.", args[0])
-		if len(args) > 1 {
-			successMsg = fmt.Sprintf("%s are now removed.", strings.Join(args, ", "))
-		}
-		fmt.Print(successMsg)
-		fmt.Println(" Run `hash -r` to ensure your shell is updated.")
-	}
-
-	return nil
+	return box.Remove(args...)
 }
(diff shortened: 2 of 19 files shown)


savil:
* main
  * **PR #188** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/188" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/188?utm_source=stack-comment).

boxcli/rm.go
```diff
@@ -25,5 +28,31 @@ func runRemoveCmd(cmd *cobra.Command, args []string) error {
...
+
+	if isDevboxShellEnabled() {
+		successMsg := fmt.Sprintf("%s is now removed.", args[0])
+		if len(args) > 1 {
+			successMsg = fmt.Sprintf("%s are now removed.", strings.Join(args, ", "))
+		}
+		fmt.Print(successMsg)
+
+		// Sadface. This doesn't seem to work within devbox shell for now.
+		fmt.Println(" You may need to restart `devbox shell` for this to take effect.")
```
LucilleH, line 54: I think that's reasonable.

devbox.go
```diff
@@ -49,13 +49,26 @@ func Open(dir string) (*Devbox, error) {
 		return nil, errors.WithStack(err)
 	}
 
+	// if dir is current directory, then get the full path
+	if dir == "." {
```
LucilleH, line 53: Why?

nix/shell.go
```diff
@@ -250,11 +258,13 @@ func (s *Shell) writeDevboxShellrc() (path string, err error) {
 		OriginalInitPath string
 		UserHook         string
 		PlanInitHook     string
+		NixProfileDir    string
```
LucilleH, line 261: nit: `ProfileBinDir`

nix/shellrc.tmpl
```diff
@@ -57,7 +61,7 @@ PATH="$(
 			*) non_nix_path="${non_nix_path:+$non_nix_path:}${path_dir}" ;;
 		esac
 	done
-	echo "${nix_path:+$nix_path:}${non_nix_path}"
+	echo "{{ .NixProfileDir }}:${non_nix_path}"
```
LucilleH, line 64: nice!

LucilleH (approved):

devbox.go
```diff
@@ -49,13 +49,26 @@ func Open(dir string) (*Devbox, error) {
 		return nil, errors.WithStack(err)
 	}
 
+	// if dir is current directory, then get the full path
```
loreto, line 52: Instead of only resolving `.`, why not call `filpath.Abs` on any directory, to ensure everything is an absolute path?

boxcli/add.go
```diff
@@ -27,6 +30,31 @@ func addCmdFunc() runFunc {
 		if err != nil {
 			return errors.WithStack(err)
 		}
-		return box.Add(args...)
+
+		if err = box.Add(args...); err != nil {
```
loreto, line 34: I think you have too much logic in here for `add`, `rm` and `shell`.

The goal of the `devbox` _library_ is that it can be used as an SDK that is _equivalent_ to using the CLI. In other words, `box.Add("package")` should be roughly the same as calling `devbox add package`; similarly `box.Remove("package")` should be roughly the same as calling `devbox rm package`. But since you've added all this logic inside of the CLI, instead of inside of the library, that is no longer true.

nix/shell.go
```diff
@@ -107,7 +110,7 @@ func rcfilePath(basename string) string {
 	return filepath.Join(home, basename)
 }
 
-func (s *Shell) Run(nixPath string) error {
+func (s *Shell) Run(nixShellFilePath string, srcDir string) error {
```
loreto, line 113: I find this new interface for `Run` slightly confusing. Can you elaborate on why we now need both of these to be passed in?

devbox.go
```diff
@@ -49,13 +49,26 @@ func Open(dir string) (*Devbox, error) {
...
+	}
+
 	box := &Devbox{
 		cfg:    cfg,
 		srcDir: dir,
 	}
 	return box, nil
 }
 
+func (d *Devbox) SourceDir() string {
```
loreto, line 68: I'm not sure this needs to be public – I think you have to make it public right now because you implemented a lot of the logic outside of the Devbox SDK – if you move that logic to be internal, then you don't need to expose this.

nix/shell.go
savil: I use it to construct the directory-for-profile-binaries i.e. `srcDir + .devbox/profile/bin`

It seems that the intention is for `nix` package to simulate a go-sdk. In that case, this is maybe not the best API.

devbox.go
savil: solved via #200. Removed this code.

savil:
@loreto @LucilleH PTAL. Made a number of changes:

1. Pushed logic from CLI into `devbox` package.
2. Fixed `shell.Exec()` functionality
3. added `nix.WithProfile` api to handle the profileBinDir that is used in the shellrc file.

boxcli/shell.go
```diff
@@ -29,18 +29,15 @@ func runShellCmd(cmd *cobra.Command, args []string) error {
...
 		return errors.WithStack(err)
 	}
 
-	inDevboxShell := os.Getenv("DEVBOX_SHELL_ENABLED")
-	if inDevboxShell != "" && inDevboxShell != "0" && inDevboxShell != "false" {
+	if devbox.IsDevboxShellEnabled() {
 		return errors.New("You are already in an active devbox shell.\nRun 'exit' before calling devbox shell again. Shell inception is not supported.")
 	}
 
-	fmt.Println("Installing nix packages. This may take a while...")
```
LucilleH, line 42: Why delete this?

devbox.go
```diff
@@ -266,3 +295,79 @@ func findConfigDir(dir string) (string, error) {
...
+func (d *Devbox) ensurePackagesAreInstalled(mode installMode) error {
+	if err := d.Generate(); err != nil {
+		return err
+	}
+
+	installingVerb := "Installing"
+	if mode == uninstall {
+		installingVerb = "Uninstalling"
+	}
+	fmt.Fprintf(d.writer, "%s nix packages. This may take a while...", installingVerb)
```
LucilleH, line 316: Can we just say, "updating nix packages"?

devbox.go
```diff
@@ -266,3 +295,79 @@ func findConfigDir(dir string) (string, error) {
...
+	// (Only when in devbox shell) Prompt the user to run `hash -r` to ensure their
+	// shell can access the most recently installed binaries, or ensure their
+	// recently uninstalled binaries are not accidentally still available.
+	if len(pkgs) > 0 && IsDevboxShellEnabled() {
+		installedVerb := "installed"
+		if mode == uninstall {
+			installedVerb = "removed"
+		}
+
+		successMsg := fmt.Sprintf("%s is now %s.", pkgs[0], installedVerb)
```
LucilleH, line 338: Just say "packages updated?"

devbox.go
```diff
@@ -266,3 +295,79 @@ func findConfigDir(dir string) (string, error) {
...
+}
+
+// ApplyDevNixDerivation ensures the local profile has exactly the packages in the development.nix file
+//
+// Will move to a store interface/package
+func (d *Devbox) ApplyDevNixDerivation() error {
+
+	cmdStr := fmt.Sprintf(
+		"--profile %s --install -f %s/.devbox/gen/development.nix",
+		d.srcDir+"/"+profileDir,
```
LucilleH, line 355: `filepath.join`?

LucilleH (commented):
👍

devbox.go
savil: I had considered that, but "updated package" has another meaning i.e. <pkg> has been updated from version 1.0.1 to 1.0.2 so felt it better to be super explicit

devbox.go
savil: oh yes of course :)

savil: I'll audit other places in this PR too!

devbox.go
LucilleH: oh right, haven't thought about that hehe

Issue #137, opened 2022-09-16
Go planner prints obscure error when install stage fails


gcurtis:
We should print a more helpful error message if the `install_stage` for a Go project (and possibly other languages/stages) exits with an error.

This is the error right now:

```
executor failed running [/bin/sh -eux -o pipefail -c go get]: exit code: 1
segment 2022/09/16 12:46:34 ERROR: sending request - Post "https://api.segment.io/v1/batch": dial tcp: lookup api.segment.io on 10.0.254.2:53: no such host
segment 2022/09/16 12:46:34 ERROR: 1 messages dropped because they failed to be sent and the client was closed
```

The segment error should be suppressed and we should show the `go get` combined output along with the exit code. We should also include the steps on how to fix it.

Lagoja:
Related to #70

LucilleH:
@loreto I think we can close this? Should be addressed by both your segment PR and my golang ones?

loreto:
Yes. The segment one is #197. Closing.

PR #199, opened 2022-10-03
[planner] Golang planner return error when main.go is not found

LucilleH:
## Summary
Golang planner return error when main is not found. It looks for `main()` and if not found, print the following error as warnings:

```
Error: Cannot find main() file in the directory. If you wish to specify a different import path, add `build_stage` in your devbox.json
```

Warnings will only get printed out if the stage is not overridden by devbox.json.

related to https://github.com/jetpack-io/devbox/issues/137

## How was it tested?
devbox build

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -5,6 +5,7 @@
 package devbox
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


gcurtis (approved):

loreto:
I haven't read the code yet: but I wanted to confirm that the error does not affect the ability for `devbox shell` to start even if the builder fails. Correct?

LucilleH:
> I haven't read the code yet: but I wanted to confirm that the error does not affect the ability for `devbox shell` to start even if the builder fails. Correct?

This does not affect `devbox shell`

PR #200, opened 2022-10-04
[config] look for the devbox.json file in parent directories as well

### LINKED ISSUE #185: Feature request: Subfolder search of the devbox.json
Hello,
I'm using devbox inside a monorepo and I want to exec `devbox shell` while I'm in a nested folder.

```
.
├── a
│   └── content
├── b
│   └── content
└── devbox.json
```

While I'm inside `a` or `b`, running `devbox shell` result in `Error: open devbox.json: no such file or directory` and I want devbox to lookup if there is some parent folders that uses devbox.

savil:
## Summary

**Motivation**
Previously, a user would need to specify the directory of the devbox.json when
invoking `devbox shell` or `devbox build` and similar commands. In the absence,
of an explicit directory argument, we would assume the user is saying that
the devbox.json file is in the current directory.

This PR changes this to look for the devbox.json file in the specified-directory, or current directory if none specified,
AND any parent directory. 

This is desirable for a few reasons:
1. Convenience: it is nice especially for mono-repos to be located in a sub-directory
   and be able to invoke `devbox shell` or `devbox build`, in a manner similar to
   how `git` works.
2. Enable `devbox add` and `devbox rm` to run from any sub-directory. These commands
   currently _require_ the user to be located in the "directory of the devbox.json".
   With this PR, this requirement is relaxed so that the commands can be run more naturally
   in any sub-directory as well.

Fixes #185

## How was it tested?

- `devbox shell` and `devbox build` inside a sub-directory
```
# this directory has a devbox.json
> mkdir fake-dir
> cd fake-dir

# verify shell is using the rustc from the nix store
> devbox shell
(devbox) > which rustc

> devbox build
> docker run devbox

> devbox add openssl
> git diff ../devbox.json
diff --git a/testdata/rust/rust-stable/devbox.json b/testdata/rust/rust-stable/devbox.json
index 288dade..66982f9 100644
--- a/testdata/rust/rust-stable/devbox.json
+++ b/testdata/rust/rust-stable/devbox.json
@@ -1,3 +1,9 @@
 {
-  "packages": ["rust-bin.stable.latest.default"]
-}
+  "packages": [
+    "rust-bin.stable.latest.default",
+    "openssl"
+  ],
+  "shell": {
+    "init_hook": null
+  }
+}
\ No newline at end of file

> devbox rm openssl
> git diff
# none! as expected (not quite, I'm lying, there's been some unrelated changes which still show up)
```

- `devbox shell` and `devbox build` in a directory containing a devbox.json
- did these commands in `testdata/rust/rust-stable`

- explicitly specify the folder of the devbox.json
```
> cd testdata/rust
> devbox shell rust-stable
```

- Error scenario
```
> cd testdata/rust/
> devbox shell

> devbox build

```

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/boxcli/usererr"
 	"go.jetpack.io/devbox/cuecfg"
+	"go.jetpack.io/devbox/debug"
 	"go.jetpack.io/devbox/docker"
 	"go.jetpack.io/devbox/nix"
 	"go.jetpack.io/devbox/pkgslice"
@@ -39,11 +40,12 @@ type Devbox struct {
 
 // Open opens a devbox by reading the config file in dir.
 func Open(dir string) (*Devbox, error) {
-	cfgPath := filepath.Join(dir, configFilename)
 
-	if !plansdk.FileExists(cfgPath) {
-		return nil, missingDevboxJSONError(dir)
+	cfgDir, err := findConfigDir(dir)
+	if err != nil {
+		return nil, err
 	}
+	cfgPath := filepath.Join(cfgDir, configFilename)
 
 	cfg, err := ReadConfig(cfgPath)
 	if err != nil {
@@ -52,7 +54,7 @@ func Open(dir string) (*Devbox, error) {
 
 	box := &Devbox{
 		cfg:    cfg,
-		srcDir: dir,
+		srcDir: cfgDir,
 	}
 	return box, nil
 }
@@ -226,7 +228,7 @@ func (d *Devbox) generateBuildFiles() error {
 func missingDevboxJSONError(dir string) error {
 
 	// We try to prettify the `dir` before printing
-	if dir == "." {
+	if dir == "." || dir == "" {
 		dir = "this directory"
 	} else {
 		// Instead of a long absolute directory, print the relative directory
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #200** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/200" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/200?utm_source=stack-comment).

devbox.go
```diff
@@ -236,5 +238,29 @@ func missingDevboxJSONError(dir string) error {
...
+func findConfigDir(dir string) (string, error) {
+
+	cur := dir
+	if cur == "" || cur == "." {
+		var err error
+		cur, err = os.Getwd()
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+	}
```
gcurtis, line 253: I think you need to call `filepath.Abs` to handle all the relative path cases. For example, if the config is in `./repo` and the user runs `devbox ..` from `./repo/a`.

savil: The `filepath.Abs` was being called inside `pathArgs`: https://github.com/jetpack-io/devbox/blob/7ae5af019d7f568bb50e164401fa6b293a22e7c5/boxcli/args.go#L18
which is then passed in here.

But I do like the suggestion to sanitize the `dir` here as well via calling `filepath.Abs`

gcurtis: Yeah, it might be a good idea to sanitize it here too. Otherwise it's not super obvious that this method requires it.

gcurtis (approved):

LucilleH (approved):
Handy 👍

savil:
will re-submit in the hopes that helps with the auto-checks running properly.

PR #204, opened 2022-10-05
[shell] preserve history

savil:
## Summary

This PR aims to preserve shell command history across sessions, where sessions is defined as starting and exiting a `devbox shell`.

## How was it tested?

- started a `devbox shell`
- invoked `go version` and `which go` and `exit`
- restarted the `devbox shell`
- used up arrow key to get the previous commands used.

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -25,11 +25,16 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-// profileDir contains the contents of the profile generated via `nix-env --profile profileDir <command>`
-const profileDir = ".devbox/profile"
+const (
+	// configFilename is name of the JSON file that defines a devbox environment.
+	configFilename = "devbox.json"
 
-// configFilename is name of the JSON file that defines a devbox environment.
-const configFilename = "devbox.json"
+	// profileDir contains the contents of the profile generated via `nix-env --profile profileDir <command>`
+	profileDir = ".devbox/profile"
+
+	// shellHistoryFile keeps the history of commands invoked inside devbox shell
+	shellHistoryFile = ".devbox/shell_history"
+)
 
 // InitConfig creates a default devbox config file if one doesn't already
 // exist.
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #200** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/200" target="_blank">Graphite</a> 
    * **PR #188** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/188" target="_blank">Graphite</a> 
      * **PR #204** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/204" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/204?utm_source=stack-comment).

devbox.go
```diff
@@ -175,7 +180,11 @@ func (d *Devbox) Shell() error {
 		return errors.WithStack(err)
 	}
 	nixShellFilePath := filepath.Join(d.srcDir, ".devbox/gen/shell.nix")
-	sh, err := nix.DetectShell(nix.WithPlanInitHook(plan.ShellInitHook), nix.WithProfile(d.profileDir()))
+	sh, err := nix.DetectShell(
+		nix.WithPlanInitHook(plan.ShellInitHook),
+		nix.WithProfile(d.profileDir()),
+		nix.WithHistoryFile(filepath.Join(d.srcDir, shellHistoryFile)),
```
savil, line 186: The current design maintains a strict separation between the outside shell history and the devbox shell history.

Should we be respecting any existing shell history file?

That would involve a small change like so:

```
nix.WithHistoryFile(d.shellHistoryFile()),
```

and
```
func (d *Devbox) shellHistoryFile() string {
   envHistFile := os.Getenv("HISTFILE")
   if envHistFile != "" {
       return envHistFile
   }
   return filepath.Join(d.srcDir, shellHistoryFile)
}
```

WDYT? 
I'm inclined to keep the current implementation (i.e. maintaining the strict separation of histories), but wanted to have the discussion.

gcurtis: I think keeping them separate makes sense, but this is also probably one of those UX things where we won't know for sure until we use it for a while.

gcurtis (approved):

devbox.go
LucilleH: My intuition is that each devbox project has its own history is 👍 👍 👍

mjgallag: I found this convo while considering logging a feature request to use (and append to) existing shell history. Before devbox I made a decision to go with asdf over dev containers or nix at my company for the primary reason that us devs are persnickety about our shells. I'd like devbox to fade into the background as much as possible and this feels like a rough edge to me.

savil: @mjgallag thanks for raising this topic. I filed an issue to track your request and have a discussion: https://github.com/jetpack-io/devbox/issues/939

PR #206, opened 2022-10-05
[shell] enable devbox add/rm to work for shells started in child directories

savil:
## Summary

When doing `devbox shell <dir>`, we want `devbox add` and `rm` to continue working.

This PR enables this by setting an environment variable when within devbox shell that
contains the directory of devbox.json (config).

In `devbox add` and `devbox rm` we check if this environment-variable is set and  use that as the starting point for the config-directory search.

## How was it tested?

```
> cd testdata/rust
> devbox shell rust-stable
(devbox)> go version
1.19

(devbox)> devbox add go_1_18
(devbox)> hash -r
(devbox)> go version
1.18

(devbox)> devbox rm go_1_18
(devbox)> go version
1.19

(devbox)> git status
# no changes in devbox.json


```

### DIFF
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -32,7 +32,13 @@ func AddCmd() *cobra.Command {
 
 func addCmdFunc(flags *addCmdFlags) runFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		box, err := devbox.Open(flags.config.path, os.Stdout)
+		dir := ""
+		if devbox.IsDevboxShellEnabled() {
+			if envdir := os.Getenv("DEVBOX_CONFIG_DIR"); envdir != "" {
+				dir = envdir
+			}
+		}
+		box, err := devbox.Open(dir, os.Stdout)
 		if err != nil {
 			return errors.WithStack(err)
 		}
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -32,9 +32,9 @@ func AddCmd() *cobra.Command {
 
 func addCmdFunc(flags *addCmdFlags) runFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		dir := ""
-		if devbox.IsDevboxShellEnabled() {
-			if envdir := os.Getenv("DEVBOX_CONFIG_DIR"); envdir != "" {
+		dir := flags.config.path
+		if dir == "" && devbox.IsDevboxShellEnabled() {
+			if envdir := os.Getenv(shellConfigEnvVar); envdir != "" {
 				dir = envdir
 			}
 		}
--- a/generate.go
+++ b/generate.go
@@ -27,7 +27,7 @@ func generate(rootPath string, plan *plansdk.Plan, files []string) error {
 	outPath := filepath.Join(rootPath, ".devbox/gen")
 
 	for _, file := range files {
-		err := writeFromTemplate(outPath, plan, file)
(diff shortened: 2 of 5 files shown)


savil:
* main
  * **PR #206** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/206" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/206?utm_source=stack-comment).

savil:
Converted to Draft. Will send another related PR based on discussion with @loreto and then circle back to this one.

savil:
um, still working on this...

savil:
We had some discussion about a generic way of overriding flags with env-vars but I'd prefer keeping this PR scoped down. We can always do that separately.

boxcli/add.go
```diff
@@ -32,7 +32,13 @@ func AddCmd() *cobra.Command {
 
 func addCmdFunc(flags *addCmdFlags) runFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		box, err := devbox.Open(flags.config.path, os.Stdout)
+		dir := flags.config.path
+		if dir == "" && devbox.IsDevboxShellEnabled() {
```
gcurtis, line 36: I think I missed some of the discussion around this feature, so apologies if I'm rehashing some stuff.

I'm trying to think through what the expected behavior would be here.

- If I'm in a devbox shell and cd into another devbox directory, is it obvious that add/rm will be operating on the shell's config and not the current directory's config?
- Similarly, it's weird that add and remove will affect a different config than all of the other commands. Or are we going to leave all the other commands disabled?
- What happens if my devbox directory is renamed or moved while I'm in a shell?
- Does this feature provide a big enough benefit now that we search parent directories for the config?

I want to say that if we're adding this feature, then we should disallow setting the config flag when in a shell and print a warning when the env var is overriding another devbox.json. Also, if the path in the env var becomes invalid, then we should tell the user to restart their shell. Thoughts?

generate.go
```diff
@@ -78,8 +78,12 @@ func toJSON(a any) string {
...
-	"json":     toJSON,
-	"contains": strings.Contains,
-	"debug":    debug.IsEnabled,
+func templateFuncs(rootPath string) template.FuncMap {
+
+	return template.FuncMap{
+		"json":      toJSON,
+		"contains":  strings.Contains,
+		"debug":     debug.IsEnabled,
+		"configDir": func() string { return rootPath },
```
gcurtis, line 87: Can we pass this in as part of the template data instead of a function?

tmpl/shell.nix.tmpl
```diff
@@ -26,6 +26,8 @@ mkShell {
       # exec a devbox shell.
       export IN_NIX_SHELL=0
       export DEVBOX_SHELL_ENABLED=1
+      # Keep this env-var name same as boxcli.shellConfigEnvVar
+      export DEVBOX_SHELL_CONFIG={{ configDir }}
```
gcurtis, line 30: This needs to be quoted to handle spaces.

boxcli/add.go
savil: oh ho 🤦🏾 I was on auto-pilot and rebased this PR onto the `--config` flag parent PR, without thinking it through.

With the new `--config` flag functionality, this PR is not needed! yay!

Abandoning.

PR #208, opened 2022-10-05
boxcli: skip shell tests due to timeout

gcurtis:
## Summary

The shell tests are taking even longer now, presumably due to how long it takes to install the Nix packages. Skip them in CI until we can figure out a way to make them run faster and more reliably.

## How was it tested?

CI.


loreto (approved):
LGTM

PR #207, opened 2022-10-05
[Bug] Fix bash regression for PATH

LucilleH:
## Summary
Fixing zsh shell regression as the following:
```
bash: command substitution: line 37: syntax error near unexpected token `newline'
bash: command substitution: line 37: `			$NIX_STORE/*'
```

Closes #196 
Closes #186 

## How was it tested?
devbox shell
go test ./...

PR #210, opened 2022-10-05
[shell] move nix-profile into a specific directory inside .devbox directory

savil:
## Summary

To avoid the nix-profile symlinks crowding the `.devbox/` directory, I push them
inside a `.devbox/nix/profile/` sub-directory. The symlink is renamed to `default` inside `.devbox/nix/profile` to be consistent with [nix naming](https://nixos.org/manual/nix/stable/package-management/profiles.html).

I also want to rename `profileDir` to `profilePath`. That is more accurate since 
it is not a directory but a symlink. However, will do in a separate PR since
it has some conflicts with concurrent work like #207.

## How was it tested?

```
> devbox shell
> which rustc
# see path in nix store

> devbox add openssl
> hash -r
> which openssl
# see path in nix store

# inspect directory and symlinks:
> ls -al /Users/savil/code/jetpack/devbox/testdata/rust/rust-stable/.devbox/nix/profile/
total 0
drwxr-xr-x 5 savil staff 160 Oct  6 11:55 .
drwxr-xr-x 3 savil staff  96 Oct  6 11:55 ..
lrwxr-xr-x 1 savil staff  14 Oct  6 11:55 default -> default-2-link
lrwxr-xr-x 1 savil staff  60 Oct  6 11:55 default-1-link -> /nix/store/qn9px7zzx500d0l78nhz0b39v8nm1siq-user-environment
lrwxr-xr-x 1 savil staff  60 Oct  6 11:55 default-2-link -> /nix/store/6q654ka2lcdvamxxvnsa9vp9dcj9b6yg-user-environment
```

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -30,7 +30,8 @@ const (
 	configFilename = "devbox.json"
 
 	// profileDir contains the contents of the profile generated via `nix-env --profile profileDir <command>`
-	profileDir = ".devbox/profile"
+	// Instead of using directory, prefer using the devbox.profileDir() function that ensures the directory exists.
+	profileDir = ".devbox/nix-profile/profile"
 
 	// shellHistoryFile keeps the history of commands invoked inside devbox shell
 	shellHistoryFile = ".devbox/shell_history"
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #210** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/210" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/210?utm_source=stack-comment).

devbox.go
```diff
@@ -30,7 +30,8 @@ const (
 	configFilename = "devbox.json"
 
 	// profileDir contains the contents of the profile generated via `nix-env --profile profileDir <command>`
-	profileDir = ".devbox/profile"
+	// Instead of using directory, prefer using the devbox.profileDir() function that ensures the directory exists.
+	profileDir = ".devbox/nix-profile/profile"
```
LucilleH, line 34: move it under `/gen`?

savil: I initially felt we should reserve `gen` for generated files....maybe we should discuss the internal structure of `.devbox` folder.

savil:
test failure is regression from #218. Tests pass locally if I delete `testdata/python/pipenv`.

PR #212, opened 2022-10-06
[build] use relative paths in devbox-build

savil:
## Summary

This should fix a consequence of a change introduced in #200 to make the devbox.srcDir be an absolute path.
In the BuildPlanners, we need the paths to be relative so they are inside the "docker context" and
can be used in Dockerfiles.

## How was it tested?

Still testing and verifying this works.


savil:
* main
  * **PR #212** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/212" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/212?utm_source=stack-comment).

LucilleH (approved):
👍 
Just to confirm, this works with `devbox build <absolute_path>`, `devbox build <relative_child_path>`, `devbox build <relative_parent_path>`, and `devbox build`?

savil:
> 👍
> Just to confirm, this works with `devbox build <absolute_path>`, `devbox build <relative_child_path>`, `devbox build <relative_parent_path>`, and `devbox build`?

I'll test these to verify, but conceptually these should work because we normalize all of these by using the absolute-path in `devbox.Open()` when initializing `devbox.srcDir`.

savil:
This has a bug: using `devbox build <path>` can sometimes fail. Will re-submit for review after I fix it.

LucilleH (approved):
Thanks for fixing!

loreto:
For a follow up PR: What can we do to catch these automatically? (and avoid a regression)

savil:
> For a follow up PR: What can we do to catch these automatically? (and avoid a regression)

Good question. I was thinking about this as well. One idea is to have a `validatePlan` function in `planner`, which can have validation rules. We can then check if the inputFiles are not absolute paths. 

We could run this `validatePlan` in `devbox_test` or even every time `planner.BuildPlan()` and `planner.ShellPlan()` is called. My preference is for doing it in `devbox_test` and ensuring we have good test coverage of various scenarios for the planners.

PR #215, opened 2022-10-06
[RFC] move gitignore file to .devbox directory and ignore gen/profile/shell_history

savil:
## Summary

With the recent use of nix-profile, the `profile` symlinks are showing up in git history.
This is regrettable.

This PR adds a `.gitignore` file to `.devbox` directory and ignoring the `gen/`, `profile*`
, `nix-profile` and `shell-history` symlinks.

Going forward, we should avoid editing this file. Each edit will need to be checked
in by our users. So, this PR is a breaking change, in a sense.

New features we add that generate files in `.devbox` should be added inside `gen/` 
if they want to be gitignored.

New features that we do want users to check in (like a lock file) can be placed
in `.devbox/` (and not `.devbox/gen`).

## How was it tested?

in a new devbox project, with `git init` and initial files already committed:
- did `devbox shell` to generate the gitignore and other files 
- did `git status` to inspect changes and found no new file changes were reported.
- did `ls -al .devbox` and saw the other files (like gen/ and profile* were generated) as well as `.devbox/.gitignore`.

### DIFF
--- a/tmpl/.gitignore.tmpl
+++ b/tmpl/.gitignore.tmpl
@@ -1,2 +1,4 @@
-*
-.*
+gen/
+profile*
+nix-profile/
+shell_history
--- a/tmpl/.gitignore.tmpl
+++ b/tmpl/.gitignore.tmpl
@@ -1,4 +1,2 @@
-gen/
-profile*
-nix-profile/
-shell_history
+*
+.*
--- a/generate.go
+++ b/generate.go
@@ -20,8 +20,8 @@ import (
 //go:embed tmpl/* tmpl/.*
 var tmplFS embed.FS
 
-var shellFiles = []string{".gitignore", "development.nix", "shell.nix"}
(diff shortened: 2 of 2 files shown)


savil:
* main
  * **PR #215** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/215" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/215?utm_source=stack-comment).

tmpl/.gitignore.tmpl
```diff
@@ -1,2 +1,4 @@
-*
-.*
\ No newline at end of file
+gen/
+profile*
+nix-profile/
+shell_history
```
savil, line 4: alternatively, this could be changed to:
```
*
.*
```
and later if adding features we want checked in, then we can modify this file as:
```
*
.*
!<feature>
```
https://stackoverflow.com/a/5250314

savil: discussed IRL, and we'll move to this way.

loreto (approved):
LGTM

Issue #216, opened 2022-10-06
ModuleNotFoundError: No module named 'pkg_resources' error after upgrading to Devbox v0.0.9


alehuo:
## Description

After Devbox automatically upgraded itself to `v0.0.9`, my Devbox environment broke. This was not happening in Devbox `v0.0.7`.

When running `python -m pipenv` inside the Devbox instance, the following error is shown:

```
Traceback (most recent call last):
  File "/nix/store/rs23kn5r3qm96cq26b6bzy2w7rs5r39d-python3-3.10.5/lib/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/nix/store/rs23kn5r3qm96cq26b6bzy2w7rs5r39d-python3-3.10.5/lib/python3.10/runpy.py", line 146, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/nix/store/rs23kn5r3qm96cq26b6bzy2w7rs5r39d-python3-3.10.5/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/Users//*****/.local/lib/python3.10/site-packages/pipenv/__init__.py", line 58, in <module>
    from .cli import cli
  File "/Users//*****/.local/lib/python3.10/site-packages/pipenv/cli/__init__.py", line 1, in <module>
    from .command import cli  # noqa
  File "/Users//*****/.local/lib/python3.10/site-packages/pipenv/cli/command.py", line 7, in <module>
    from pipenv.cli.options import (
  File "/Users//*****/.local/lib/python3.10/site-packages/pipenv/cli/options.py", line 3, in <module>
    from pipenv.project import Project
  File "/Users//*****/.local/lib/python3.10/site-packages/pipenv/project.py", line 23, in <module>
    from pipenv.environment import Environment
  File "/Users/*****/.local/lib/python3.10/site-packages/pipenv/environment.py", line 14, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

Running `pipenv` alone does not cause the error, only `python -m pipenv` does now.

My `devbox.json`:

```
{
  "packages": [
    "python310",
    "pipenv"
  ]
}
```

## How to reproduce

1. Use Devbox `v0.0.9` (I suspect `v0.0.8` is the version, where this regression started to occur)
2. Use the `devbox.json` shown above
[...]

PR #217, opened 2022-10-06
boxcli: reenable shell tests

gcurtis:
## Summary

Increase the CI test timeout and reenable the end-to-end shell tests. Also print some extra information when the workflow is rerun with debugging enabled.

## How was it tested?

CI


loreto:
So ... one of the test seems to be failing. Is it a real regression? (we're working on a new release, so if it is we should fix before). Or is this what #221 fixed?

@savil ^

gcurtis:
@loreto they were broken for the same reason as #216. Looks like #218 should fix it again.

LucilleH (approved):

gcurtis:
@savil it's failing for a different reason now (it hangs in the bash test). Still trying to debug why.

PR #219, opened 2022-10-06
fix pipenv test

savil:
## Summary

Fixes this test by making the python pip planner run for it.

did:
```
touch setup.py
touch requirements.txt
```
Needed this because `devbox_test` requires buildable plans.

NOTE: I had to manually edit the generated `plan.json` to get rid of the absolute paths.
Lets discuss if the planner needs to use these absolute paths, or if relative paths suffice.

## How was it tested?

`go test ./...`

PR #220, opened 2022-10-06
nix: fix removing user Nix profiles from PATH

gcurtis:
## Summary

When launching a devbox shell, we filter out any Nix user profiles from the PATH (i.e., ~/.nix-profile). This fixes a couple of bugs with how we determine if a PATH directory matches a Nix profile:

1. The `strings.HasPrefix("/Users", profileDir)` call is backwards.
2. Checking for `/Users` is macOS-specific and won't work on Linux. Instead, check if the directory starts with `/nix` to determine if it's the default system profile.

## How was it tested?

Added a test.

### DIFF
--- a/nix/shell.go
+++ b/nix/shell.go
@@ -370,7 +370,8 @@ func splitNixList(s string) []string {
 //
 //  1. Applies filepath.Clean.
 //  2. Removes the path if it's relative (must begin with '/' and not be '.').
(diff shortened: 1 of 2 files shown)


savil (approved):
ugh 🤦🏾 

Thank you!

gcurtis:
This is just waiting on the test fix in #219.

PR #221, opened 2022-10-06
devbox: handle paths with spaces

gcurtis:
## Summary

The current command for nix-env starts as a string and then gets split by spaces to form the args slice. This doesn't handle paths with spaces, so change it to just create a slice directly instead.

Also improve the error message when `nix-env` fails to give more info instead of just the exit code.

Finally, unexport `ApplyDevNixDerivation` because isn't used anywhere else.

## How was it tested?

Ran `devbox shell` from a directory with spaces in the name.


savil (approved):
thanks!

loreto (approved):
LGTM

PR #222, opened 2022-10-06
Turn off auto-detection for language on devbox shell

LucilleH:
## Summary
Turn off auto-detection for language on devbox shell. This means:
1. `devbox plan` outputs plan for build only
2. `devbox shell` does not automatically add packages based on the project
3. `devbox shell` no longer automatically sets init hooks (python can potentially be the biggest one being affected by this change, but as far as I know, it wasn't being used before) 

I also added a `devbox.json` to the root of this project, because I need go 1.19 and I have go 1.18 on my personal machine.

## How was it tested?
devbox shell
devbox build
go test ./...

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -138,6 +138,7 @@ func (d *Devbox) Build(flags *docker.BuildFlags) error {
 // Plan creates a plan of the actions that devbox will take to generate its
 // shell environment.
 func (d *Devbox) ShellPlan() *plansdk.Plan {
+	// TODO: Move shell plan to a separate struct from build plan.
 	return d.convertToPlan()
 }
 
--- a/devbox.go
+++ b/devbox.go
@@ -138,12 +138,7 @@ func (d *Devbox) Build(flags *docker.BuildFlags) error {
 // Plan creates a plan of the actions that devbox will take to generate its
 // shell environment.
 func (d *Devbox) ShellPlan() (*plansdk.Plan, error) {
-	userPlan := d.convertToPlan()
-	shellPlan, err := planner.GetShellPlan(d.srcDir)
-	if err != nil {
-		return nil, err
-	}
-	return plansdk.MergeUserPlan(userPlan, shellPlan)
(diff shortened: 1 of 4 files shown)


devbox_test.go
```diff
@@ -48,15 +48,16 @@ func testExample(t *testing.T, testPath string) {
 		box.srcDir, err = filepath.Rel(currentDir, box.srcDir)
 		assert.NoErrorf(err, "expect to construct relative path from %s relative to base %s", box.srcDir, currentDir)
 
-		plan, err := box.ShellPlan()
+		plan, err := box.BuildPlan()
+		buildPlanErrorFile := filepath.Join(baseDir, "has_build_error")
+		hasBuildErrorFile := fileExists(buildPlanErrorFile)
+		if hasBuildErrorFile {
+			return
```
loreto, line 55: Instead of just returning, shouldn't we be asserting that there _is_ an error?

devbox.go
```diff
@@ -137,13 +137,8 @@ func (d *Devbox) Build(flags *docker.BuildFlags) error {
...
 // shell environment.
-func (d *Devbox) ShellPlan() (*plansdk.Plan, error) {
-	userPlan := d.convertToPlan()
-	shellPlan, err := planner.GetShellPlan(d.srcDir)
-	if err != nil {
-		return nil, err
-	}
-	return plansdk.MergeUserPlan(userPlan, shellPlan)
+func (d *Devbox) ShellPlan() *plansdk.Plan {
+	return d.convertToPlan()
```
loreto, line 141: Could you add a TODO to make this a different struct at a later point? (it makes no sense for the "shell plan" to still include things like build stage)

loreto (approved):
LGTM

Make sure to test manually too against several examples just to be safe.

devbox_test.go
LucilleH: oh ummm ok I'll rename the file. It should be `build_error_expected` or `build_should_error_and_it_should_not_succeed`

PR #225, opened 2022-10-07
[CLI] add --config flag

savil:
## Summary

**Motivation**
We want to introduce a `--config` flag and deprecate the use of config as an argument.

1. **ephemeral shells**. By enabling a way to (eventually) do `devbox shell --config path/to/devbox/json  nodejs-18_x`, or even more simply `devbox shell nodejs-18_x`.
2. **using devbox add/rm from shells started in parent directories**: today, if I have `my_sub_directory/devbox.json` and start `devbox shell my_sub_directory/devbox.json` from the root of the repository, then doing `devbox add` or `devbox rm` inside this shell will fail. 
    - A first attempt at fixing this was in #206. This PR (#225) is a generalization of that idea. I'll be stacking #206 on top of this PR, and enabling the `--config` flag to override the env-var.

<insert motivation>

## How was it tested?

**preliminary steps**
- i added `nodejs-18_x` to `testdata/nodejs/nodejs-18`
- I (temporarily) deleted the `devbox.json` at the root of this repo. This is to ensure I verify that the CLI still complains if no devbox.json is found.

```
# giving a path
> cd testdata/nodejs
> devbox shell nodejs-18
Warning: devbox shell <path> is deprecated, use devbox shell --config <path> instead
Installing nix packages. This may take a while...done.
Starting a devbox shell...
> which node
# get nix store path
> exit

# using --config
> devbox shell -c nodejs-18
> which node
# get nix store path
> exit

# using neither path nor --config
> cd testdata/nodejs/node-18
> devbox shell
> which node
# get nix path
> exit

# using neither path nor --config, where devbox.json resides in a parent directory
> cd testdata/nodejs/nodejs-18/fake-dir
> devbox shell
which node
# get nix path
> exit

# using both path and --config
> cd testdata/nodejs
> devbox shell -c nodejs-18 nodejs-18

Error: Cannot specify devbox.json's path via both --config and the command arguments. Please use --config only.

# No devbox.json in parent dirs
> cd testdata/nodejs
> devbox shell

# giving a path where no devbox.json exists
> devbox shell nodejs-18/fake-dir
Warning: devbox shell <path> is deprecated, use devbox shell --config <path> instead

Error: No devbox.json found in nodejs-18/fake-dir. Did you run `devbox init` yet?
```

### DIFF
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -11,21 +11,28 @@ import (
 	"go.jetpack.io/devbox"
 )
 
+type addCmdFlags struct {
+	config configFlags
+}
+
 func AddCmd() *cobra.Command {
+	flags := &addCmdFlags{}
+
 	command := &cobra.Command{
 		Use:               "add <pkg>...",
 		Short:             "Add a new package to your devbox",
 		Args:              cobra.MinimumNArgs(1),
 		PersistentPreRunE: nixShellPersistentPreRunE,
-		RunE:              addCmdFunc(),
+		RunE:              addCmdFunc(flags),
 	}
 
+	registerConfigFlags(command, &flags.config)
 	return command
 }
 
-func addCmdFunc() runFunc {
+func addCmdFunc(flags *addCmdFlags) runFunc {
 	return func(cmd *cobra.Command, args []string) error {
-		box, err := devbox.Open(".", os.Stdout)
+		box, err := devbox.Open(flags.config.path, os.Stdout)
 		if err != nil {
 			return errors.WithStack(err)
 		}
--- a/boxcli/add.go
+++ b/boxcli/add.go
@@ -16,27 +16,27 @@ type addCmdFlags struct {
 }
 
 func AddCmd() *cobra.Command {
-	flags := &addCmdFlags{}
+	flags := addCmdFlags{}
 
 	command := &cobra.Command{
 		Use:               "add <pkg>...",
 		Short:             "Add a new package to your devbox",
 		Args:              cobra.MinimumNArgs(1),
 		PersistentPreRunE: nixShellPersistentPreRunE,
-		RunE:              addCmdFunc(flags),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return addCmdFunc(cmd, args, flags)
+		},
 	}
 
-	registerConfigFlags(command, &flags.config)
+	flags.config.register(command)
 	return command
 }
 
-func addCmdFunc(flags *addCmdFlags) runFunc {
-	return func(cmd *cobra.Command, args []string) error {
-		box, err := devbox.Open(flags.config.path, os.Stdout)
-		if err != nil {
-			return errors.WithStack(err)
-		}
-
-		return box.Add(args...)
+func addCmdFunc(_ *cobra.Command, args []string, flags addCmdFlags) error {
+	box, err := devbox.Open(flags.config.path, os.Stdout)
+	if err != nil {
+		return errors.WithStack(err)
 	}
+
+	return box.Add(args...)
 }
--- a/boxcli/args.go
+++ b/boxcli/args.go
@@ -9,37 +9,42 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/boxcli/usererr"
 )
 
 // Functions that help parse arguments
 
 // If args empty, defaults to the current directory
 // Otherwise grabs the path from the first argument
-func pathArg(args []string, flags *configFlags) string {
+func configPathFromUser(args []string, flags *configFlags) (string, error) {
 
-	if flags.path != currentDir {
-		if len(args) > 0 {
-			// Choose the --config flag because config argument is being deprecated
-			fmt.Printf(
-				"%s You are specifying the config path as an argument and using the --config flag. "+
-					"Choosing to ignore the argument and use the flag.\n",
-				color.HiYellowString("Warning:"),
-			)
-		}
-		return flags.path
+	if flags.path != "" && len(args) > 0 {
+		return "", usererr.New(
+			"Cannot specify devbox.json's path via both --config and the command arguments. " +
+				"Please use --config only.",
+		)
+	}
+
+	if flags.path != "" {
+		return flags.path, nil
 	}
 
 	if len(args) > 0 {
 		fmt.Printf(
-			"%s please use the --config or -c flag to specify the path to the devbox.json config. "+
-				"We are deprecating the previous way of specifying this path as an argument to the command.\n",
+			"%s devbox <command> <path> is deprecated, use devbox <command> --config <path> instead\n",
 			color.HiYellowString("Warning:"),
 		)
+	}
+	return pathArg(args), nil
+}
+
+func pathArg(args []string) string {
+	if len(args) > 0 {
 		p, err := filepath.Abs(args[0])
 		if err != nil {
 			panic(errors.WithStack(err)) // What even triggers this?
 		}
 		return p
 	}
-	return "."
+	return ""
 }
--- a/boxcli/args.go
+++ b/boxcli/args.go
@@ -4,8 +4,10 @@
 package boxcli
 
 import (
+	"fmt"
 	"path/filepath"
 
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
 )
 
@@ -13,8 +15,26 @@ import (
 
 // If args empty, defaults to the current directory
 // Otherwise grabs the path from the first argument
-func pathArg(args []string) string {
+func pathArg(args []string, flags *configFlags) string {
+
+	if flags.path != currentDir {
+		if len(args) > 0 {
+			// Choose the --config flag because config argument is being deprecated
+			fmt.Printf(
+				"%s You are specifying the config path as an argument and using the --config flag. "+
+					"Choosing to ignore the argument and use the flag.\n",
+				color.HiYellowString("Warning:"),
+			)
+		}
+		return flags.path
+	}
+
 	if len(args) > 0 {
+		fmt.Printf(
+			"%s please use the --config or -c flag to specify the path to the devbox.json config. "+
+				"We are deprecating the previous way of specifying this path as an argument to the command.\n",
+			color.HiYellowString("Warning:"),
+		)
 		p, err := filepath.Abs(args[0])
 		if err != nil {
 			panic(errors.WithStack(err)) // What even triggers this?
--- a/boxcli/build.go
+++ b/boxcli/build.go
@@ -12,8 +12,13 @@ import (
 	"go.jetpack.io/devbox/docker"
 )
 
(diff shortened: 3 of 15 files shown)


savil:
* main
  * **PR #225** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/225" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/225?utm_source=stack-comment).

savil:
I'll fix the linter errors (minor), but this is ready for review.

boxcli/build.go
```diff
@@ -12,8 +12,13 @@ import (
...
+	config configFlags
+	docker *docker.BuildFlags
+}
+
 func BuildCmd() *cobra.Command {
-	flags := &docker.BuildFlags{}
+	flags := newBuildFlags()
 
 	command := &cobra.Command{
 		Use:   "build [<dir>]",
```
loreto, line 24: Update `Use` to not have `[<dir>]` – same for any other commands where this is changing

boxcli/generate.go
```diff
@@ -11,18 +11,29 @@ import (
...
 
+type generateCmdFlags struct {
+	config configFlags
+}
+
 func GenerateCmd() *cobra.Command {
+	flags := &generateCmdFlags{}
+
 	command := &cobra.Command{
 		Use:    "generate [<dir>]",
```
loreto, line 22: Update `Use`

boxcli/init.go
```diff
@@ -9,19 +9,29 @@ import (
 	"go.jetpack.io/devbox"
 )
 
+type initCmdFlags struct {
```
loreto, line 12: So ... `init` is the one command I was still imagining would work as `devbox init my/project` instead of `devbox init -c my/project`

Thoughts on whether we should keep init working as before?

boxcli/args.go
```diff
@@ -4,17 +4,37 @@
...
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
 )
 
 // Functions that help parse arguments
 
 // If args empty, defaults to the current directory
 // Otherwise grabs the path from the first argument
-func pathArg(args []string) string {
+func pathArg(args []string, flags *configFlags) string {
```
loreto, line 18: ### 1
If you want to keep the logic simpler, I think it'd be ok to error if both flag and path as an argument are specified at the same time:
+ path-as-arg: warning
+ flag: no warning
+ both: error
### 2
I think you somehow need to distinguish between the flag being explicitly set and not set. At least I was assuming that:
- `devbox add foo` looks for `devbox.json` in a parent directory if it's not in the current directory. 
- `devbox add foo -c /my/project` looks for a `devbox.json` in `/my/project` _only_ (and not in any of it's parent directories). By extension that means `devbox add foo -c "."` only looks in the current directory and not in the parents.
- `devbox add foo -c /my/project/my-file.json` treats `my-file.json` as the config file.

That said, open to a different approach. How were you thinking it should work?

boxcli/config.go
```diff
@@ -0,0 +1,18 @@
...
+const currentDir = "."
+
+// to be composed into xyzCmdFlags structs
+type configFlags struct {
+	path string
+}
+
+func registerConfigFlags(cmd *cobra.Command, flags *configFlags) {
+	cmd.Flags().StringVarP(
+		&flags.path, "config", "c", currentDir, "path to directory containing a devbox.json config file",
```
loreto, line 16: If you need to distinguish between the flag being set and not set, you might want the default value to actually be the empty string.

loreto:
Two more thoughts:
1. The warning message feels too long – it should be more succinct.  Something like `WARNING: devbox shell <path> is deprecated, use devbox shell --config <path> instead`
2. While I'm happy enough with `--config`, I'm not sure I love `-c` when seeing it in the examples. I'm not entirely sure people will be able to immediately guess that in `devbox shell -c nodejs-18` the `-c` stands for config. Thoughts?

boxcli/args.go
savil: > I think it'd be ok to error if both flag and path as an argument are specified at the same time:

👍🏾 

> That said, open to a different approach. How were you thinking it should work?

I'd been assuming we'd match the current behavior i.e. `devbox add foo -c /my/project` will treat `/my/project` as the starting point of the search and traverse up to the parent directories.

Happy to change that logic to your proposal. It _does_  feel more clear to not do the magic search if the user is specifying the config-path directly. 

How do you feel about doing that as a followup PR? I fear a bigger refactoring is needed because the search-parent-dirs logic is inside `devbox.go`. If its simple, I'll fold the change into this PR.

boxcli/build.go
savil: ah, yeah, I overlooked that

boxcli/config.go
savil: cobra has an api something like `if cmd.Flags().Changed(flagName) {`

I suppose it depends on where I need to distinguish this.

boxcli/init.go
savil: I don't feel strongly, I can change it back ....

savil (commented):
@loreto Thanks for the review! I'll fix up early tomorrow.

> Two more thoughts:
The warning message feels too long – it should be more succinct. Something like WARNING: devbox shell <path> is deprecated, use devbox shell --config <path> instead

sounds good, easy change.

> While I'm happy enough with --config, I'm not sure I love -c when seeing it in the examples. I'm not entirely sure people will be able to immediately guess that in devbox shell -c nodejs-18 the -c stands for config. Thoughts?

Not sure I follow the concern here. We do still have the `--config` flag. The `-c` is just a shorthand for that flag for users who get familiar with devbox and don't wanna type as much. Could you elaborate a bit more?

boxcli/build.go
savil: Fixed

boxcli/init.go
savil: How do you feel about keeping it for consistency?

devbox.go
```diff
@@ -313,17 +313,50 @@ func missingDevboxJSONError(dir string) error {
...
+			}
+			return absDir, nil
+		case mode.IsRegular(): // regular means 'file'
+			if !plansdk.FileExists(filepath.Clean(absDir)) {
+				return "", missingDevboxJSONError(dir, false /*didCheckParents*/)
+			}
+			// we return a directory from this function
+			return filepath.Dir(absDir), nil
+		default:
+			return "", errors.Errorf("unhandled mode %v", mode)
```
gcurtis, line 354: What if we just assume that everything that isn't a directory is a file? If for some reason it is unsupported, then we'll get a read error later.

devbox.go
```diff
@@ -313,17 +313,50 @@ func missingDevboxJSONError(dir string) error {
...
+	parentDirCheckAddendum := ""
+	if didCheckParents {
+		parentDirCheckAddendum = ", or any parent directories"
+	}
+
+	return usererr.New("No devbox.json found in %s%s. Did you run `devbox init` yet?", dir, parentDirCheckAddendum)
 }
 
+// findConfigDir
 func findConfigDir(dir string) (string, error) {
```
gcurtis, line 326: Should this go in `config.go`?

gcurtis (approved):
LGTM! Can we just add some tests for the config searching behavior?

Another minor, optional change is to pass around the flag structs as values if they're not being modified.

boxcli/add.go
```diff
@@ -11,21 +11,28 @@ import (
...
-		RunE:              addCmdFunc(),
+		RunE:              addCmdFunc(flags),
 	}
 
+	registerConfigFlags(command, &flags.config)
 	return command
 }
 
-func addCmdFunc() runFunc {
+func addCmdFunc(flags *addCmdFlags) runFunc {
```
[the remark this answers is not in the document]
savil: hmm okay I can do that. I generally avoid using values because its a copy-operation but the perf impact is likely negligible.

devbox.go
```diff
@@ -313,17 +313,50 @@ func missingDevboxJSONError(dir string) error {
...
+	// Otherwise, we search the parent directories.
+	if dir != "" {
+		fi, err := os.Stat(dir)
+		if err != nil {
+			return "", err
+		}
+		switch mode := fi.Mode(); {
+		case mode.IsDir():
+			if !plansdk.FileExists(filepath.Join(absDir, configFilename)) {
+				return "", missingDevboxJSONError(dir, false /*didCheckParents*/)
```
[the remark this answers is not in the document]
savil: well, it won't quite be:
> return "", &missingConfigError{dir: dir, checkedParents: false}

because we need to run some logic in `missingDevboxJSONError` to make the error message read nicely.

devbox.go
savil: sure :)

savil (commented):
> Can we just add some tests for the config searching behavior?

will do!

devbox.go
savil: its added inside devbox.go to keep the CLI minimal. Moving this logic inside here enables `devbox` as a package to encapsulate all the logic and functionality (enabling reuse outside this CLI scenario).

devbox.go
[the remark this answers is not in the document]
LucilleH: 🤔  if the struct implements the `Error` interface with the same logic in `missingDevboxJSONError`, would that work?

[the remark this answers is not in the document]
gcurtis: That's what I was thinking. Just make the `Error` method format the message however we want.

devbox.go
gcurtis: I agree with keeping it in `devbox`. I was wondering if it should go in `config.go` within the same package.

gcurtis:
@savil there are a couple of options for testing `findConfig`.

1. Use `t.TempDir()` to create a test directory layout and give its path to `findConfig`.
2. Refactor `findConfig` to use an `fs.FS`. The CLI would pass in `os.DirFS("/")` and tests would pass in an `fstest.MapFS`. This might be a bit of work.

With either approach I think `findConfig` will need to be updated to take in a root path and a subpath. For example:

```go
// or: findConfig(root fs.FS, dir string)
func findConfig(root, subdir string) (string, error) {
	root, _ := filepath.Abs(root)
	for subdir != "." {
		rooted := filepath.Join(root, cur)
		debug.Log("finding %s in dir: %s\n", configFilename, rooted)
		if plansdk.FileExists(filepath.Join(rooted, configFilename)) {
			return rooted, nil
		}
		subdir = filepath.Dir(subdir)
	}
}

// CLI
wd, _ := os.Getwd()
findConfig("/", wd)

// Tests
tmpDir := t.TempDir()
os.MkdirAll(filepath.Join(tmpDir, "a/b/c"))
os.WriteFile(filepath.Join(tmpDir, "a", "devbox.json"), []byte(`{}`), 0666)
findConfig(tmpDir, "a/b/c")
```

That will let you control where the search stops (`cur` will equal `.`) so the test doesn't go outside of the test directory.

PR #232, opened 2022-10-13
Make IDE docs visible via sidebar.js

Lagoja:
## Summary
IDE Section of the docs were not visible because they were not added to sidebar.js. This should fix this on the next deploy

## How was it tested?

localhost

### DIFF
--- a/docs/app/sidebars.js
+++ b/docs/app/sidebars.js
@@ -65,6 +65,21 @@ const sidebars = {
         id: 'configuration'
     }, {
         type: 'category',
+        label: "IDE Configuration",
+        collapsed: false,
+        items: [{
+            type: 'doc',
+            id: 'ide_configuration/direnv',
+        }, {
+            type: 'doc',
+            id: 'ide_configuration/eclipse',
+        }, {
+            type: 'doc',
+            id: 'ide_configuration/vscode'
+        }]
+
+    }, {
+        type: 'category',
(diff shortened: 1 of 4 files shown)


Lagoja:
@Mohsenari -- your docs were not visible in the navigation because you didn't add them to sidebar.js. This PR should fix it

savil (approved):

docs/app/sidebars.js
```diff
@@ -63,6 +63,21 @@ const sidebars = {
...
+        items: [{
+            type: 'doc',
+            id: 'ide_configuration/direnv',
+        }, {
+            type: 'doc',
+            id: 'ide_configuration/eclipse',
+        }, {
+            type: 'doc',
+            id: 'ide_configuration/vscode'
+        }]
```
savil, line 79: should we omit these empty pages?

Lagoja: They aren't empty. They look empty in the diff because all I changed was the title. You can see the full file here: https://github.com/jetpack-io/devbox/blob/d3ce7a36afaba679ccba25c1951d28b862030453/docs/app/docs/ide_configuration/eclipse.md

mohsenari: they're not empty there are eclipse.md and vscode.md files in `ide_configuration/`

mohsenari (approved):

PR #234, opened 2022-10-17
[RFC] add config packages to BuildPlan's runtimePackages

savil:
## Summary

Previously, the `devbox.json` config's packages were being added to the Runtime packages
of `devbox build`. #227 changed that. Was it intentional (RFC)?

This PR fixes that to make the rust build work again

## How was it tested?

```
> devbox build && docker run devbox
Installed version is 1.64.0
```


savil:
Oh - I see why you did it:
> Build plan is only applied when user defined packages includes all of the dev packages the build plan requires

PR #235, opened 2022-10-18
[Rust Shell] change to use rustup nix package to avoid overlay

savil:
## Summary

In furtherance of our efforts to remove "magic" and simplify the plans, this PR
removes the NixOverlay from the Rust Planner.

We replace it with the `rustup` toolchain. This should work, although the
performance for `devbox build` is bad, and the image is bloated now.

The `devbox shell` experience is good. I'll add an example in `devbox-example`
repo for others to reference to add the shell init hooks.

- [ ] send followup PR to remove NixOverlay from the core plansdk.

## How was it tested?

did `devbox shell` and `devbox build && docker run devbox` in the rust folders

### DIFF
--- a/planner/languages/rust/rust_planner.go
+++ b/planner/languages/rust/rust_planner.go
@@ -13,9 +13,6 @@ import (
 	"go.jetpack.io/devbox/planner/plansdk"
 )
 
-// Source and reference: https://github.com/oxalica/rust-overlay
-const RustOxalicaOverlay = "https://github.com/oxalica/rust-overlay/archive/stable.tar.gz"
-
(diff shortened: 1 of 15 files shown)


savil:
* main
  * **PR #235** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/235" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/235?utm_source=stack-comment).

LucilleH (approved):
💯

LucilleH:
Can we remove `Overlay` from `ShellPlan` entirely?

Issue #226, opened 2022-10-07
Proposal: Add a devcontainer config for VS Code / Github Codespaces


pomdtr:
I did not find a way to effectively integrate devbox with vscode. As the broad majority of devs are using this editor these days, this is a concern for adoption in large teams.

You can start a devbox shell in the vscode terminal, but it does not set the env of all other vscode process (tasks, debugger, extensions...)

I feel like writing a [devcontainer](https://code.visualstudio.com/docs/remote/containers) could be a great way to fix this (and free user for the required nix dependency), but I was not successful in implementing it.

Lagoja:
Hey @pomdtr, we'll look into this feature. Is this something you use to develop locally on your own machine as well? Or is it for remote development?

pomdtr:
I use this workflow both on my local machine and my remote workspaces

mohsenari:
@pomdtr We recently published the first version of [devbox VSCode extension](https://marketplace.visualstudio.com/items?itemName=jetpack-io.devbox&ssr=false#overview). We'll try to see if we can have the environment for the whole IDE (debugger & path to binaries and dependencies) be updated by the extension. Devcontainer seems like a possible solution as well. We'll update under this issue once we have a solution to move forward on this.

pomdtr:
Cool ! Here is an interesting read on the devcontainer / nix interaction: https://levelup.gitconnected.com/vs-code-remote-containers-with-nix-2a6f230d1e4e

pomdtr:
I've got a draft version working for devbox. Please take a look at it: https://github.com/pomdtr/devbox-devcontainer

mohsenari:
@pomdtr Thanks for sharing this. I don't have codespaces enabled on my github account but I tested it locally with Docker desktop and Dev Containers extension on VSCode and it woks great. Although, it runs vscode from inside a Docker container with a shared volume between the host machine and Docker. That means in a non-Linux OS there might be a performance hit.

That said, I really liked the Dev Containers extension when it detects a `.devcontainer` it prompts with "Reopen this in devcontainers". Maybe we can add a similar functionality to Devbox's extension. Because if we open vscode from inside a devbox shell, the language binaries and interpreter settings in vscode will be set to devbox's shell environment.

pomdtr:
I think the performance hit should be quite minimal since the bind mount only contains the workspace folder. There are way to mitigate the performance penalty (see : https://code.visualstudio.com/remote/advancedcontainers/improve-performance)

[...]

Issue #209, opened 2022-10-05
Proposal: add direnv config


pomdtr:
Here is what I came up with:

```bash
#!/usr/bin/env bash

direnv_load devbox shell -- direnv dump
watch_file devbox.json
```

Thanks to this, as soon as I cd inside my project folder, I'm inside the devbox shell. Newly added package are also automatically loaded in the current shell.

It might be worth to add it somewhere, but I'm not sure where.

pomdtr:
Sadly it is not working really well in vscode (the direnv extension is not really good at overriding the user path). I think writing  a devcontainer instead would be better :)

mohsenari:
@pomdtr That's a great suggestion. We've been looking into integrating with `direnv` for the past couple weeks. You suggestion technically works but it also makes `devbox` think it's in a shell subprocess. So if you go into your direnv allowed directory and run `devbox shell` it will output:
>Error: You are already in an active devbox shell.
Run 'exit' before calling devbox shell again. Shell inception is not supported 
and when you run `exit` you're thrown out of your terminal, not the devbox shell subprocess which is not the intended behavior and can be confusing to some users.

This can be fixed, of course, but for the time being we're working on #213 which you put `devbox shell --print-env` in your `.envrc` and it updates your PATH with all the binaries you would get if you had been in a devbox shell.
We're also working on improving this command to make your shell's environment as close to a devbox shell environment as possible while making sure the confusions like the error above doesn't happen.

mohsenari:
@methyl We're looking into VSCode integration with `direnv + devbox` as well. If we can't get the extension to work with a devbox environment as we intend, we might have to write our own extension.
Either way, we'll follow up on this issue once we have a solid solution for VSCode setup.

pomdtr:
For direnv to work with vscode, it looks like you will need to patch the whole `code` process. See https://rastertail.net/blog/hacking-vscode/ for more details.

mohsenari:
@methyl We published a [devbox extension for VSCode](https://marketplace.visualstudio.com/items?itemName=jetpack-io.devbox&ssr=false#overview). It detects if you have devbox.json and when you open terminal in the IDE, it automatically goes to devbox shell.
Feel free to give it a try and let us know if you have any suggestions.

methyl:
@mohsenari thanks for a heads up! I think it's solving less problematic issue of having terminals in vscode open devbox shell, what's more important I think is to make sure extensions can use devbox to run the binaries.

[...]

PR #236, opened 2022-10-20
VSCode extension for devbox

mohsenari:
## Summary
Added vscode extension for devbox under `vscode-extension/`.
It shows 15 file changes but most of the implementation is in `extension.ts` and `package.json` files.

## How was it tested?
- cd vscode-extension/
- yarn install
- open vscode in this directory
- press F5 or press run in the run & debug section from the sidebar in vscode
- in the new opened vscode window, open a devbox project then open the vscode integrated Terminal
- confirm terminal runs devbox shell automatically


vscode-extension/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
```
LucilleH, line 5: Should this be the same license as Jetpack License in devbox?

vscode-extension/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
```
LucilleH, line 3: Is this by convention for vscode extension? Generally I use es6 instead of commonjs. But it doesn't matter much really.

vscode-extension/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+		"lib": [
+			"ES2020"
+		],
+		"sourceMap": true,
+		"rootDir": "src",
+		"strict": true   /* enable all strict type-checking options */
+		/* Additional Checks */
+		// "noImplicitReturns": true, /* Report error when not all code paths in function return a value. */
+		// "noFallthroughCasesInSwitch": true, /* Report errors for fallthrough cases in switch statement. */
+		// "noUnusedParameters": true,  /* Report errors on unused parameters. */
```
LucilleH, line 15: Don't want those checks?

LucilleH (approved):
Cool!

vscode-extension/tsconfig.json
mohsenari: No, for now I left it as default values as the official guide for vscode extensions.

vscode-extension/tsconfig.json
mohsenari: This was preset as I followed the steps on the extension guide. So I'd rather leave it as is.

vscode-extension/LICENSE
mohsenari: It is, but in the extension marketplace, there needs to be a license file at the root level of the extension project. That's why I made a separate file here.

mohsenari:
Linking somewhat related issues:
#209 
#226

PR #239, opened 2022-10-24
Define and Run Tasks in Devbox Shells

Lagoja:
## Summary

_This PR is a Draft, and not a final feature._

Adds the ability to specify and run Tasks in your Devbox Shell. A Task is a command that will run in a shell until it completes or is terminated, after which it will exit the shell. 

The motivation is to allow users to spin up a shell for each service in their Devbox project. For example, in a LAMP project, you may want to spin up a shell to run and monitor MySQL, another to run Apache, and another to run PHP + develop your app. 

You can specify Tasks in your `devbox.json` using a tasks map. For example: 

```json
"tasks": {
    "start_db": {
        "init_hook": [
	      	"source conf/mysql/set-exit.sh",
	      	"conf/mysql/mysql.sh"
	 ],
        "command": "tail -f conf/mysql/mysql.log"
     }
}
```
You can then run a task using `devbox run <task-name>`, e.g., `devbox run start_db`. It will run your Shell's `init_hook`, then the Task `init_hook`, then the `Task` command in the foreground. The shell will exit if your task completes, or if you terminate the shell/task manually. 

## How was it tested?

Tested using the LAMP stack in [Devbox Examples](https://github.com/jetpack-io/devbox-examples)

### Questions: 

1. Do we need an `init_hook` for tasks?
2. Are there other fields we need? (e.g., cleanup?)
3. How do we automate starting multiple tasks at once?


LucilleH:
`It will run your Shell's init_hook, then the Task init_hook, then the Task command in the foreground.`

I fear that there will be conflicts between hooks that is hard to debug for users, as we are moving away from the declarative model and things can become increasingly convoluted for users.

I do like the idea of `devbox run`. I really don't know if devbox need to handle the init hook, or exiting a service. For example, if I start something with yarn, a mongodb let's say, I stop it myself outside of yarn. 

Do we want to do a user test of simply changing `devbox shell -- <cmd>` to `devbox run -- <cmd>` and put that out there first? `devbox run` in this case would run no shell init hooks. It simply runs whatever follows `--`. (which I really prefer that over running shell init hooks during `devbox shell -- <cmd>`)

I want to know how necessary it is for devbox to handle the rest of the stuff defined in `task` before committing to it. My fear is that we can always evolve the devbox.json later, but we can't roll back. 

Thoughts?

Lagoja:
> I do like the idea of devbox run. I really don't know if devbox need to handle the init hook, or exiting a service. For example, if I start something with yarn, a mongodb let's say, I stop it myself outside of yarn.

Setting an init hook or exit hook could be optional, but I think it's tedious to stop or restart services manually every time you exit a shell. 

When working on the LAMP and LEMP examples, I had to frequently run `pkill <service_name>` to cleanup between sessions, and if I forgot to run the cleanup, my packages would block ports that kept other shells from running. It's a better user experience if I can automate having the shell clean up after itself. 

> Do we want to do a user test of simply changing devbox shell -- <cmd> to devbox run -- <cmd> and put that out there first? 

I've tested a similar workflow while working on the examples, and some of the commands can be pretty long and tedious to type. Having them passed via the command line also means that any scripts or steps needed to run my shell are undocumented. Having the tasks specified in the `devbox.json` makes them easier for developers to discover and run.

PR #240, opened 2022-10-25
[RFC][config dir] enable init-shell scripts to run in shell started with --config flag

savil:
## Summary

When I changed the `rust` examples in `testdata/rust`  to use `rustup` nix package instead of the nix overlay,
I noticed that I could no longer start a devbox shell as:
```
> cd testdata/rust
> devbox shell --config rust-stable
```
The error would point to the `init-shell.sh` script not being found in the current directory.

This PR takes an approach of introducing a `DEVBOX_CONFIG` env-var, which can be referenced
in `devbox.json` for the init script as:
```
➜ cat rust-stable/devbox.json
{
...
  "shell": {
    "init_hook": [
      "source $DEVBOX_CONFIG/init-shell.sh"
    ]
  }
}
```

I can also use this env-var inside the init-script to create the `.rustup` directory,
and other purposes.

RFC:
An alternative approach is to ensure we invoke the init-script as if the current
directory is the config-dir.

A third option is to do both (introduce `DEVBOX_CONFIG` env-var as well as invoke
the init-script as if the current directory is the config-dir).

## How was it tested?

started a `devbox shell` in the following scenarios:
1. from within a child-directory 
2. from the config dir directory
3. from the parent directory using the --config flag.

Did this with `testdata/rust/rust-stable`, and within the shell ran:
1. `echo $DEVBOX_CONFIG`
2. `cargo run`. To ensure the init-shell.sh script worked.


savil:
* main
  * **PR #240** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/240" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/240?utm_source=stack-comment).

savil:
ping

loreto:
Sorry, must have missed this one before.

Could we do option 2: init-hooks run as if the current directory is the config-dir?

savil:
moving to draft until I implement: init-hooks run as if the current directory is the config-dir

savil:
done in #302

PR #247, opened 2022-10-27
[WIP] nix-flakes etude

savil:
## Summary

Generates a basic `flake.nix` file that can be used with `nix develop`.

- [ ] understand how flakes and nix-profile work together
- [ ] figure out story for adding to git (or not)
- [ ] shell environment and prompt doesn't seem to be getting set

## How was it tested?

```

# See TODO around managing this git add
> git add flake.nix

> nix develop

> ➜ nix develop
Starting a devbox shell...
Savil-Srivastavas-MacBook-Pro:rust-stable savil$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/hello-rust`
Installed version is 1.61.0
thread 'main' panicked at 'Version 1.61.0 is less than expected version of 1.63.0', src/main.rs:11:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This failure is somewhat expected. It arises because we are (likely) getting the stable
nix channel. That likely has the older version of `rustup`.

### DIFF
--- a/generate.go
+++ b/generate.go
@@ -41,6 +41,11 @@ func generateForShell(rootPath string, plan *plansdk.ShellPlan) error {
 		return errors.WithStack(err)
 	}
 
+	err = writeFromTemplate(rootPath, plan, "flake.nix")
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
 	for name, content := range plan.GeneratedFiles {
 		filePath := filepath.Join(outPath, name)
 		if err := os.WriteFile(filePath, []byte(content), 0644); err != nil {
--- a/testdata/rust/rust-stable/flake.lock
+++ b/testdata/rust/rust-stable/flake.lock
@@ -1,0 +1,25 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1666753130,
+        "narHash": "sha256-Wff1dGPFSneXJLI2c0kkdWTgxnQ416KE6X4KnFkgPYQ=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f540aeda6f677354f1e7144ab04352f61aaa0118",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/testdata/rust/rust-stable/flake.nix
+++ b/testdata/rust/rust-stable/flake.nix
@@ -1,0 +1,40 @@
+{
+    description = "A devbox shell";
+    outputs = { self, nixpkgs }:
+        # TODO generalize system using flake-util
+        let pkgs = nixpkgs.legacyPackages.x86_64-darwin;
+
+        in {
+            devShell.x86_64-darwin = pkgs.mkShell {
+              shellHook =
+                ''
+                  echo "Starting a devbox shell..."
+
+                  # We're technically no longer in a Nix shell after this hook because we
+                  # exec a devbox shell.
+                  export IN_NIX_SHELL=0
+                  export DEVBOX_SHELL_ENABLED=1
+
+                  # Undo the effects of `nix-shell --pure` on SSL certs.
+                  # See https://github.com/NixOS/nixpkgs/blob/dae204faa0243b4d0c0234a5f5f83a2549ecb5b7/pkgs/stdenv/generic/setup.sh#L677-L685
+                  if [ "$NIX_SSL_CERT_FILE" == "/no-cert-file.crt" ]; then
+                     unset NIX_SSL_CERT_FILE
+                  fi
+                  if [ "$SSL_CERT_FILE" == "/no-cert-file.crt" ]; then
+                     unset SSL_CERT_FILE
+                  fi
+
+                  # Append the parent shell's PATH so that we retain access to
+                  # non-Nix programs, while still preferring the Nix ones.
+                  export "PATH=$PATH:$PARENT_PATH"
+
+                  
+                '';
+              buildInputs = with pkgs; [
+                    rustup
+                  
+                    libiconv
+                  ];
+            };
+        };
+}
--- a/tmpl/flake.nix.tmpl
+++ b/tmpl/flake.nix.tmpl
@@ -1,0 +1,42 @@
+{
+    description = "A devbox shell";
+    outputs = { self, nixpkgs }:
+        # TODO generalize system using flake-util
+        let pkgs = nixpkgs.legacyPackages.x86_64-darwin;
+
+        in {
+            devShell.x86_64-darwin = pkgs.mkShell {
+              shellHook =
+                ''
+                  echo "Starting a devbox shell..."
+
+                  # We're technically no longer in a Nix shell after this hook because we
+                  # exec a devbox shell.
+                  export IN_NIX_SHELL=0
+                  export DEVBOX_SHELL_ENABLED=1
+
+                  # Undo the effects of `nix-shell --pure` on SSL certs.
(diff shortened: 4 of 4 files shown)


savil:
* main
  * **PR #247** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/247" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/247?utm_source=stack-comment).

internal/impl/flake.go
```diff
@@ -0,0 +1,78 @@
...
+func (d *Devbox) installNixProfileFlakeCommand(profileDir string) *exec.Cmd {
+	cmd := exec.Command(
+		"nix", "profile", "install",
+		"--profile", profileDir,
+		"--extra-experimental-features", "nix-command flakes",
+		filepath.Join(d.configDir, ".devbox/gen/flake/"), // installables
+	)
+	if d.hasDevboxLockFile() {
+		_ = d.copyDevboxLockToFlakeLock()
+		cmd.Args = append(cmd.Args, "--no-write-lock-file")
```
LucilleH, line 21: Help me understand this: if the user already has a lock file, we don't want the `nix profile install` command to create a lock file?

internal/impl/tmpl/flake.nix.tmpl
```diff
@@ -0,0 +1,68 @@
+{
+    description = "A devbox shell";
+
+    inputs = {
+        # master branch can be slow because not all binaries are built and populated in the cache.
+        # using the nixpkgs-unstable tag ensures everything has been built and populated in the cache.
+        # https://www.reddit.com/r/NixOS/comments/ydr4po/comment/itv6oqo/?utm_source=reddit&utm_medium=web2x&context=3
+        nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
```
LucilleH, line 8: Can we use the `nixpkgs` commit hash in the devbox.json? or is it supposed to download the latest unstable and ignore that commit hash?

internal/impl/tmpl/flake.nix.tmpl
```diff
@@ -0,0 +1,68 @@
...
+                  echo "PARENT_PATH=$PARENT_PATH"
+                  echo "PATH=$PATH"
+                  {{- end }}
+
+                  # exec env "SHELL=/bin/zsh" "ZDOTDIR=/var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/devbox3840985630" /bin/zsh
+                '';
+
+              # We do not need to install these because we do `nix profile install`
+              # using the defaultPackages above.
+              buildInputs = [
```
LucilleH, line 60: ✂️ ?

LucilleH (approved):
Some QQ, approve to unblock

internal/impl/flake.go
savil: the second one!

internal/impl/tmpl/flake.nix.tmpl
savil: Ah, good point. Yes, I should use the nixpkgs-commit-hash from devbox.json

Longer term, I think we would deprecate the nixpkgs-commit-hash from devbox.json, and instead rely on the devbox.lock file (which will contain the commit-hash of nixpkgs as well)

internal/impl/tmpl/flake.nix.tmpl
savil: will remove

internal/impl/flake.go
LucilleH: But what if the packages field has changed? Does `nix profile` overwrites the old?

savil: when one does `devbox add`, then the `nix profile --install` is run.

btw, right now there's some issue with this. I need to iterate on this PR (hence I was anxious to get it in).
```
❯ DEVBOX_FEATURE_FLAKES=1 devbox add ripgrep
Reinitialized existing Git repository in /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/gen/flake/.git/
Error: apply Nix derivation: running command /nix/var/nix/profiles/default/bin/nix profile install --profile /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/nix/profile/default --extra-experimental-features nix-command flakes /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/gen/flake --no-write-lock-file: exit status 1 with command stderr: warning: Git tree '/Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/gen/flake' is dirty
this derivation will be built:
  /nix/store/jp2yvrsx3iy7izmp297yh56nz9knw2y6-devbox-shell-env.drv
these 2 paths will be fetched (1.65 MiB download, 6.77 MiB unpacked):
  /nix/store/hisnvr8mcci3si2fsjs6r7s3h0r9yh5i-ripgrep-13.0.0
  /nix/store/qs2pncr051xkd9kq9l8mjalj7dqww6k8-pcre2-10.40
copying path '/nix/store/qs2pncr051xkd9kq9l8mjalj7dqww6k8-pcre2-10.40' from 'https://cache.nixos.org'...
copying path '/nix/store/hisnvr8mcci3si2fsjs6r7s3h0r9yh5i-ripgrep-13.0.0' from 'https://cache.nixos.org'...
building '/nix/store/jp2yvrsx3iy7izmp297yh56nz9knw2y6-devbox-shell-env.drv'...
error: files '/nix/store/xi7frhzjyxi7862wj54pxsw55xyzdg8y-devbox-shell-env/bin/go' and '/nix/store/za0pnll14dv61b124n6xxnjapd150lcr-go-1.19.4/share/go/bin/go' have the same priority 5; use 'nix-env --set-flag priority NUMBER INSTALLED_PKGNAME' or type 'nix profile install --help' if using 'nix profile' to find out howto change the priority of one of the conflicting packages (0 being the highest priority)

```

internal/impl/tmpl/flake.nix.tmpl
savil: actually, I want to revisit this in light of today's slack conversation about the environment. Added TODO to revisit.

internal/impl/generate.go
```diff
@@ -92,3 +98,36 @@ var templateFuncs = template.FuncMap{
...
+	err := writeFromTemplate(flakeDir, plan, "flake.nix")
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	// make an empty git repo
+	// Alternatively consider: git add intent-to-add path/to/flake.nix, and
+	// git update-index --assume-unchanged path/to/flake.nix
+	// https://nixos.wiki/wiki/Flakes#How_to_add_a_file_locally_in_git_but_not_include_it_in_commits
+	cmd := exec.Command("git", "-C", flakeDir, "init")
```
gcurtis, line 114: Why does the flake need to be added to git? What if the project isn't using git?

internal/impl/flake.go
gcurtis: I'm also a little confused about this. What's the difference between a devbox lockfile and the flake lockfile? Is it just to hide the flakes implementation?

gcurtis (approved):
LGTM. I'm assuming we can figure out the git dependency stuff later.

internal/impl/flake.go
savil: there's no difference. We copy flake.lock to devbox.lock, because the flake lockfile is generated in an inner git repository. We use an inner git repository to workaround the concern in this [issue](https://github.com/NixOS/nix/issues/3121).

internal/impl/generate.go
savil: to workaround this concern: https://github.com/NixOS/nix/issues/3121

That's okay, if the project is not using `git`, as long as their is a local `git` installed in the system.

savil: I'll revisit this. I think what we can do is:
- if in `git` repository, then do this step.
- if not in `git` repository, then can skip this step.

PR #249, opened 2022-10-31
nix: fix PkgExists panic for some missing packages

gcurtis:
## Summary

nix-env returns an empty JSON object instead of an error for some missing packages. This leads to a panic because we check that unmarshalled nix-env JSON output only has 1 entry. Remove that check since it seems unnecessary.

## How was it tested?

Added a test that looks for the "rust" package, which triggers this panic for some reason.


loreto (approved):
LGTM

PR #250, opened 2022-11-01
[devbox shell] ensure all zsh startup files are linked to ZDOTDIR

savil:
## Summary

User `mtunski` in discord helpfully pointed out that their shell setup using
zsh4humans (https://github.com/romkatv/zsh4humans) wasn't working with devbox shell.

Specifically, they get an error saying:
> Installing nix packages. This may take a while...done.
Starting a devbox shell...
/tmp/devbox3134336657/.zshrc:62: command not found: z4h

We currently just copy over the `.zshrc` file into the new `ZDOTDIR` of devbox shell. The fix is:
1. to link more zsh startup files (specifically, .zshenv, .zprofile, .zlogin)
2. manually link other zsh-plugin specific files in the .zshrc. This is not great, but a workaround until we introduce a more formal architecture for custom shells to be hooked up (i.e. with a provider or plugin model).

## How was it tested?

1. installed zsh4humans on my laptop
2. 
The advanced config docs point to three files to manage for [Backup and Restore](https://github.com/romkatv/zsh4humans/blob/master/tips.md#backup-and-restore) for zsh4humans:
- zshenv
- zshrc
- p10k*.zsh

The first two are handled by devbox CLI as part of zsh startup scripts. The last one needs to be manually handled.

Added the following to my .zshrc:
```
# within my .zshrc file:
...
#### my addition ###
if [ -n "${ZDOTDIR+1}" ] && [[ $HOME != $ZDOTDIR ]] then
  ln $HOME/.p10k*.zsh $ZDOTDIR
fi
#### end of my addition ###

# Install or update core components (fzf, zsh-autosuggestions, etc.) and
# initialize Zsh. After this point console I/O is unavailable until Zsh
# is fully initialized. Everything that requires user interaction or can
# perform network I/O must be done above. Everything else is best done below.
z4h init || return
```
3. opened `devbox shell` in `testdata/zig/zig-hello-world`. Did `which zig`
to verify that the nix paths are active (i.e. devbox shell is active).

NOTE: in the absence of step (2), the `z4h init` step will start a setup wizard.

4. Opened a new zsh shell (new terminal window) to ensure the .zshrc changes are okay.

4. Deleted zsh4humans (instructions https://github.com/romkatv/zsh4humans#uninstalling), and re-did step 3 without issues.

### DIFF
--- a/nix/shell.go
+++ b/nix/shell.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	_ "embed"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -202,6 +203,9 @@ func (s *Shell) execCommand() string {
 		return strings.Join(append(args, s.binPath), " ")
 	}
 
+	// Link or Copy other files that affect the shell settings and environments.
+	s.copyShellSettings(filepath.Dir(shellrc))
+
 	// Shells have different ways of overriding the shellrc, so we need to
 	// look at the name to know which env vars or args to set.
 	var (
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #250** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/250" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/250?utm_source=stack-comment).

nix/shell.go
```diff
@@ -286,6 +290,34 @@ func (s *Shell) writeDevboxShellrc() (path string, err error) {
...
+		// TODO savil: make this a glob search `.p10k*.zsh`
+		filenames := []string{".zshenv", ".p10k.zsh"}
+		for _, filename := range filenames {
+			fileOld := filepath.Join(filepath.Dir(s.userShellrcPath), filename)
+			if _, err := os.Stat(fileOld); errors.Is(err, fs.ErrExist) {
+				// this file may not be relevant for the user's setup.
+				continue
+			}
+
+			fileNew := filepath.Join(shellSettingsDir, filename)
```
LucilleH, line 309: Can we somehow solve this in a more generic way, or in a way that the users can configure this themselves? Something along the lines of exporting an environment variable for devbox shellrc dir and have the users copy any files over?

savil: Darn. Moving `.p10k.zsh` to be linked from the shell-init-script doesn't seem to work. 

I think the reason is:
- the devbox shell's `.zshrc` first runs the user's `.zshrc` and afterwards runs the shell init hooks.
- the user's `.zshrc` has init logic for zsh4humans, but at this point the `.p10k.zsh` is missing and so the zsh4humans-wizard starts.

savil: okay, an ugly workaround for now is to hard-link the `.p10k.zsh` files from within the `.zshrc` prior to calling `z4h init`.

savil: for example, I did:
```
# within my .zshrc file:
...
if (( ${+ZDOTDIR} )); then
  ln ~/.p10k*.zsh $ZDOTDIR
fi


# Install or update core components (fzf, zsh-autosuggestions, etc.) and
# initialize Zsh. After this point console I/O is unavailable until Zsh
# is fully initialized. Everything that requires user interaction or can
# perform network I/O must be done above. Everything else is best done below.
z4h init || return
```

LucilleH: oh well

LucilleH (approved):

PR #251, opened 2022-11-01
[feature-flag] Add simple feature flag package

mikeland86:
## Summary

Add very simple feature flag package that uses hard coded boolean to switch on and off. Can be overridden on by setting environment variable. 

## How was it tested?

unit tests

### DIFF
--- a/boxcli/featureflag/feature.go
+++ b/boxcli/featureflag/feature.go
@@ -8,11 +8,8 @@ type feature struct {
 	name    string
 	enabled bool
 }
-type Feature interface {
-	Enabled() bool
-}
 
-var features = map[string]Feature{}
+var features = map[string]*feature{}
 
 func disabled(name string) {
 	features[name] = &feature{name: name}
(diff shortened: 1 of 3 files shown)


boxcli/featureflag/feature.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"os"
+)
+
+type feature struct {
+	name    string
+	enabled bool
+}
+type Feature interface {
```
gcurtis, line 11: Why an interface?

gcurtis (approved):

boxcli/featureflag/feature.go
mikeland86: You are right, this is overkill, I don't imagine us passing feature flag objects outside this package. Will rm

boxcli/featureflag/feature.go
```diff
@@ -0,0 +1,34 @@
...
+func enabled(name string) {
+	features[name] = &feature{name: name, enabled: true}
+}
+
+func Get(name string) Feature {
+	return features[name]
+}
+
+func (f *feature) Enabled() bool {
+	if os.Getenv("DEVBOX_FEATURE_"+f.name) == "1" {
```
savil, line 30: prefer using `strconv.ParseBool` so that its more forgiving of users

PR #254, opened 2022-11-03
First pass at devbox scripts implementation

ipince:
## Summary

Devbox scripts allows you to define scripts to run inside a devbox shell. You can add them to your devbox.json like so:
```
{
  "packages": [],
  "shell": {
    "init_hook": "echo hook",
    "scripts": {
      "a": "echo hello",
      "b": [
        "echo good",
        "echo bye"
      ]
    }
  }
}
```

And then you can run with `devbox run <script>`. For example:
```
> devbox run b
Starting a devbox shell...
hook
good
bye
```
Note that the shell's init hook will always run.

## How was it tested?
With a devbox.json as written above and running `devbox run <script>` with no script, invalid script, valid script (single command and multi-command).

### DIFF
--- a/boxcli/run.go
+++ b/boxcli/run.go
@@ -25,6 +25,7 @@ func RunCmd() *cobra.Command {
 		Hidden:            true,
 		Short:             "Starts a new devbox shell and runs the target script",
 		Long:              "Starts a new interactive shell and runs your target script in it. The shell will exit once your target script is completed or when it is terminated via CTRL-C. Scripts can be defined in your `devbox.json`",
+		Args:              cobra.MaximumNArgs(1),
 		PersistentPreRunE: nixShellPersistentPreRunE,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runScriptCmd(args, flags)
@@ -77,8 +78,6 @@ func parseScriptArgs(args []string, flags runCmdFlags) (string, string, error) {
 	script := ""
 	if len(args) == 1 {
 		script = args[0]
-	} else if len(args) > 1 {
-		return "", "", errors.Errorf("too many arguments (expected 1, got %d)", len(args))
 	}
 
 	return path, script, nil
--- a/boxcli/run.go
+++ b/boxcli/run.go
@@ -6,7 +6,7 @@ package boxcli
 import (
 	"os"
 	"os/exec"
-	"strings"
+	"sort"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -51,9 +51,10 @@ func runScriptCmd(args []string, flags runCmdFlags) error {
 
 	// Validate script exists.
 	scripts := box.ListScripts()
+	sort.Slice(scripts, func(i, j int) bool { return scripts[i] < scripts[j] })
 	if script == "" || !slices.Contains(scripts, script) {
-		return errors.Errorf("no script found with name \"%s\". Here's a list of the existing scripts in devbox.json: %s",
-			script, strings.Join(box.ListScripts(), ", "))
+		return errors.Errorf("no script found with name \"%s\"."+
+			"Here's a list of the existing scripts in devbox.json: %v", script, scripts)
 	}
 
 	if devbox.IsDevboxShellEnabled() {
--- a/boxcli/run.go
+++ b/boxcli/run.go
@@ -1,0 +1,85 @@
+// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package boxcli
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"go.jetpack.io/devbox"
+	"golang.org/x/exp/slices"
+)
+
+type runCmdFlags struct {
+	config configFlags
+}
+
+func RunCmd() *cobra.Command {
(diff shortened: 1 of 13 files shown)


config.go
```diff
@@ -37,7 +37,8 @@ type Config struct {
 	// Shell configures the devbox shell environment.
 	Shell struct {
 		// InitHook contains commands that will run at shell startup.
-		InitHook ConfigShellCmds `json:"init_hook,omitempty"`
+		InitHook ConfigShellCmds             `json:"init_hook,omitempty"`
+		Scripts  map[string]*ConfigShellCmds `json:"scripts,omitempty"`
```
savil, line 41: in `launchpad.yaml`, we had a similar undeterministic-ordering issue with the Services. We handled it by:
1. defining it as `Scripts []ConfigShellCmds` and adding the "name" as a field in `ConfigShellCmds`
2. adding custom marshal and unmarshal functions to have it print as a map with the key

Would you wanna try that? Could do as a followup. I feel it would be really nice to ensure determinism. With determinism, we won't get random diffs of devbox.json when we write to it for (other unrelated) reasons.

config.go
```diff
@@ -77,9 +78,22 @@ func upgradeConfig(cfg *Config, absFilePath string) error {
...
 	return cuecfg.WriteFile(path, cfg)
 }
 
+func (c *Config) validate() error {
+	for k := range c.Shell.Scripts {
+		if k == "" {
+			return errors.New("cannot have script with empty name in devbox.json")
+		}
+	}
+	return nil
```
savil, line 94: oh boy, I ended up adding another validate function today, so will need to merge it. It should be fairly straightforward.

@loreto had requested me to see if we can use the `cue` validation instead, which I am yet to do.

devbox.go
```diff
@@ -238,6 +238,50 @@ func (d *Devbox) Shell() error {
...
+		nix.WithHistoryFile(filepath.Join(d.srcDir, shellHistoryFile)),
+		nix.WithUserScript(scriptName, script.String()))
+
+	if err != nil {
+		fmt.Print(err)
+		shell = &nix.Shell{}
+	}
+
+	shell.UserInitHook = d.cfg.Shell.InitHook.String()
+	return shell.Run(nixShellFilePath)
```
savil, line 272: oh boyo boy. I thought this would reuse the implementation of `box.Exec`....
but I agree with your implementation more (as in, we should strive to emulate the `devbox shell` environment which `box.Exec` doesn't seem to).

So, there are three different implementations now:
- this `RunScript`
- `box.Exec`
- the `box.Shell`  (except this doesn't run a specific command, but starts an interactive shell)

I feel:
1. `box.Exec` and `RunScript` should use the exact same implementation
2. `box.Shell` should also be refactored to use the same core implementation as `box.Exec` and `RunScript` (which is your TODO above)

savil (approved):
approving to unblock but interested in discussion on the comments below

boxcli/run.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	script := ""
+	if len(args) == 1 {
+		script = args[0]
+	} else if len(args) > 1 {
+		return "", "", errors.Errorf("too many arguments (expected 1, got %d)", len(args))
+	}
```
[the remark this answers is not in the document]
ipince: Yeah I guess.. the error messages from the cobra validators are pretty bad though, but I'll see if this is reasonable.

boxcli/run.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Validate script exists.
+	scripts := box.ListScripts()
+	if script == "" || !slices.Contains(scripts, script) {
+		return errors.Errorf("no script found with name \"%s\". Here's a list of the existing scripts in devbox.json: %s",
+			script, strings.Join(box.ListScripts(), ", "))
+	}
+
+	if devbox.IsDevboxShellEnabled() {
+		return errors.New("You are already in an active devbox shell.\nRun 'exit' before calling devbox run again. Shell inception is not supported yet.")
```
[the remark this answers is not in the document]
ipince: lol yeah

config.go
ipince: Yeah, I actually attempted to use `cue` for validation, but I couldn't find what I needed quickly enough so I gave up and added the validation function. I can merge it with yours

devbox.go
ipince: Hmmm ok I'm not too familiar with `box.Exec`, I can look into it. I'm new to this code, so this type of feedback is really useful.

boxcli/run.go
```diff
@@ -0,0 +1,85 @@
...
+	config configFlags
+}
+
+func RunCmd() *cobra.Command {
+	flags := runCmdFlags{}
+	command := &cobra.Command{
+		Use:               "run <script>",
+		Hidden:            true,
+		Short:             "Starts a new devbox shell and runs the target script",
+		Long:              "Starts a new interactive shell and runs your target script in it. The shell will exit once your target script is completed or when it is terminated via CTRL-C. Scripts can be defined in your `devbox.json`",
```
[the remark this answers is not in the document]
ipince: I guess the point is that you can start a long-running process, like a server, and it's stays in the foreground so you can Ctrl+C out of it when you want.

config.go
ipince: oh interesting, i'll see about it!

ipince: I'll do as a follow-up

config.go
ipince: done

PR #256, opened 2022-11-04
[nixpkgs version] change field from version to commit; add validation

savil:
## Summary

Changing the field name inside `Config.Nixpkgs` from `version` to `commit`.
The motivation stems from the realization that even `YY.MM` stable channels of
nixpkgs have periodic updates (mostly for security), which means the commit-hash
tracked by the `nixos-stable` branch changes.

To simplify the product design for now, and to avoid introducing a lockfile, 
we change `version` to be the unambiguous `commit`. For users who would like
to track the nixpkgs commit from `search.nixos.org` webpage, they can refer
to `status.nixos.org` and use the commit for the nixpkgs channel they want.

## How was it tested?

`go test ./...` b/c added a testcase for the validation function.

```
> cd testdata/nodejs/nodejs-18
> DEVBOX_FEATURE_NIXPKG_VERSION=1 DEVBOX_DEBUG=0 devbox shell -- node --version
Installing nix packages. This may take a while...done.
Starting a devbox shell...
v18.7.0
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -42,7 +42,7 @@ type Config struct {
 
 	// Nixpkgs specifies the repository to pull packages from
 	Nixpkgs struct {
-		Version string `json:"version,omitempty"`
+		Commit string `json:"commit,omitempty"`
 	} `json:"nixpkgs,omitempty"`
 }
 
@@ -51,8 +51,7 @@ type Stage struct {
 	Command string `cue:"string" json:"command"`
 }
 
-// ReadConfig reads a devbox config file.
-func ReadConfig(path string) (*Config, error) {
+func readConfig(path string) (*Config, error) {
 	cfg := &Config{}
 	err := cuecfg.ParseFile(path, cfg)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #256** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/256" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/256?utm_source=stack-comment).

loreto (approved):
LGTM

loreto (approved):
LGTM

Although, are you sure you need to implement custom validation? We've been using `cue` annotations in the config struct so that cue itself does the validation. I'm wondering if it'd be better to specify the validation for this field through `cue` as well (assuming it's simple enough that it can be done through there)

savil:
@loreto good point. I'll look into `cue` and do a followup to switch validation to that if it works.

PR #257, opened 2022-11-04
[pkg-config] Add bin sym links and add to path

mikeland86:
## Summary

This PR does two things:

* Adds `.devbox/conf/bin` to path
* Creates sym links for any files in `.devbox/conf/[pkg]/bin`

## How was it tested?

Using dummy package config

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -214,7 +214,11 @@ func (d *Devbox) Shell() error {
 		if err != nil {
 			return err
 		}
-		opts = append(opts, nix.WithEnvVariables(env))
+		opts = append(
+			opts,
+			nix.WithEnvVariables(env),
+			nix.WithPath(filepath.Join(d.srcDir, ".devbox/conf/bin")),
+		)
 	}
 
 	shell, err := nix.DetectShell(opts...)
--- a/devbox.go
+++ b/devbox.go
@@ -217,7 +217,7 @@ func (d *Devbox) Shell() error {
 		opts = append(
 			opts,
 			nix.WithEnvVariables(env),
-			nix.WithPath(filepath.Join(d.srcDir, ".devbox/conf/bin")),
+			nix.WithPKGCOnfigDir(filepath.Join(d.srcDir, ".devbox/conf/bin")),
 		)
 	}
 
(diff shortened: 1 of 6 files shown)


devbox.go
```diff
@@ -214,7 +214,11 @@ func (d *Devbox) Shell() error {
 		if err != nil {
 			return err
 		}
-		opts = append(opts, nix.WithEnvVariables(env))
+		opts = append(
+			opts,
+			nix.WithEnvVariables(env),
+			nix.WithPath(filepath.Join(d.srcDir, ".devbox/conf/bin")),
```
gcurtis, line 220: The name `WithPath` made me think this was setting the path to Nix. What do you think of having the caller update the `PATH` and then pass it in via `WithEnvVars`? That would also fix the inconsistency with how we currently bypass the "with-ed" env vars when reading the `PATH`.

pkgcfg/pkgcfg.go
```diff
@@ -51,16 +26,22 @@ func CreateFiles(pkg, basePath string) error {
 	}
 	for name, contentPath := range cfg.CreateFiles {
 		filePath := filepath.Join(basePath, name)
-		if _, err := os.Stat(filePath); err == nil {
-			continue
-		}
-		content, err := os.ReadFile(filepath.Join(basePath, contentPath))
+		// if _, err := os.Stat(filePath); err == nil {
+		// 	continue
+		// }
```
gcurtis, line 31: Can this be removed entirely?

pkgcfg/pkgcfg.go
```diff
@@ -51,16 +26,22 @@ func CreateFiles(pkg, basePath string) error {
...
+		content, err := os.ReadFile(filepath.Join(cfg.localConfigPath, contentPath))
 		if err != nil {
 			return errors.WithStack(err)
 		}
+		if err = createDir(filepath.Dir(filePath)); err != nil {
+			return err
+		}
 		if err := os.WriteFile(filePath, content, 0744); err != nil {
```
gcurtis, line 39: Not for this PR but something to think about - it sounds like we'll eventually need a way to specify the file's permissions. I'm guessing not all of them should be executable.

gcurtis (approved):

devbox.go
mikeland86: Actually, this wont work. We need to explicitly prepend the conf bin path in a few places so I think it has to be its own argument. If we just send an updated PATH then we'll end up prepending the original nix binaries (via the profile path). 

I do think we can clean stuff up by maybe just adding the config bin path by default (the way we do with profile bin dir). I think I'll do that instead.

PR #260, opened 2022-11-04
Use <picture> for logo light/dark mode

gcurtis:
Using the `#gh-light/dark-mode-only` fragments is deprecated and doesn't render correctly on other sites (like pkgs.go.dev).


loreto (approved):
LGTM

PR #262, opened 2022-11-04
vscode extension settings + devcontainer setup + devbox commands

mohsenari:
## Summary
This doesn't fully implement the behaviour we discussed. But it does update the python interpreter and go path for the user.
The dev container setup is also possible via command palette (cmd + shift + P) in vscode.
- Added settings to the extension
- Added auto generating dev container config files
- cleaned up the code for more readability
- Added devbox commands (`devbox init | add | shell | install`) to the extension

## How was it tested?
- Open vscode-extensions/ in vscode
- press Run+debug
- check the prompt for update settings from devbox extension
- press cmd+shift+p and type devbox > choose "Devbox: Generate Dev Containers config files"
- check the generated files in .devcontainer/

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -25,6 +25,18 @@
       {
         "command": "devbox.add",
         "title": "Devbox: Add - add packages to your devbox project"
+      },
+      {
+        "command": "devbox.shell",
+        "title": "Devbox: Shell - Go to devbox shell in the terminal"
+      },
+      {
+        "command": "devbox.install",
+        "title": "Devbox: Install - Install packages specified in devbox.json"
+      },
+      {
+        "command": "devbox.init",
+        "title": "Devbox: Init - Initiate a devbox project"
       }
     ],
     "menus": {
@@ -36,6 +48,18 @@
         {
           "command": "devbox.add",
           "when": "devbox.configFileExists == true"
+        },
+        {
+          "command": "devbox.shell",
+          "when": "devbox.configFileExists == true"
+        },
+        {
+          "command": "devbox.install",
+          "when": "devbox.configFileExists == true"
+        },
+        {
+          "command": "devbox.init",
+          "when": "devbox.configFileExists == false"
         }
       ]
     },
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -39,10 +39,6 @@
         "title": "Devbox: Shell - Go to devbox shell in the terminal"
       },
       {
-        "command": "devbox.install",
-        "title": "Devbox: Install - Install packages specified in devbox.json"
-      },
-      {
         "command": "devbox.init",
         "title": "Devbox: Init - Initiate a devbox project"
       }
@@ -67,10 +63,6 @@
         },
         {
           "command": "devbox.shell",
-          "when": "devbox.configFileExists == true"
-        },
-        {
-          "command": "devbox.install",
           "when": "devbox.configFileExists == true"
         },
         {
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -21,8 +21,24 @@
       {
         "command": "devbox.setupDevContainer",
         "title": "Devbox: Generate Dev Containers config files"
+      },
+      {
+        "command": "devbox.add",
+        "title": "Devbox: Add - add packages to your devbox project"
       }
     ],
+    "menus": {
+      "commandPalette": [
+        {
+          "command": "devbox.setupDevContainer",
+          "when": "devbox.configFileExists == true"
+        },
+        {
+          "command": "devbox.add",
+          "when": "devbox.configFileExists == true"
+        }
+      ]
+    },
     "configuration": {
       "title": "devbox",
       "properties": {
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -27,6 +27,14 @@
         "title": "Devbox: Add - add packages to your devbox project"
       },
       {
+        "command": "devbox.remove",
+        "title": "Devbox: Remove - remove packages from your devbox project"
+      },
+      {
+        "command": "devbox.run",
+        "title": "Devbox: Run - execute scripts specified in devbox.json"
+      },
+      {
         "command": "devbox.shell",
         "title": "Devbox: Shell - Go to devbox shell in the terminal"
       },
@@ -50,6 +58,14 @@
           "when": "devbox.configFileExists == true"
         },
         {
+          "command": "devbox.remove",
+          "when": "devbox.configFileExists == true"
+        },
+        {
+          "command": "devbox.run",
+          "when": "devbox.configFileExists == true"
+        },
+        {
           "command": "devbox.shell",
           "when": "devbox.configFileExists == true"
         },
@@ -66,11 +82,6 @@
     "configuration": {
       "title": "devbox",
       "properties": {
-        "devbox.promptUpdateSettings": {
-          "type": "boolean",
-          "default": true,
-          "description": "Prompt to update workspace language settings to match devbox shell."
-        },
         "devbox.autoShellOnTerminal": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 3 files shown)


vscode-extension/package.json
```diff
@@ -13,10 +13,72 @@
...
+        "command": "devbox.shell",
+        "title": "Devbox: Shell - Go to devbox shell in the terminal"
+      },
+      {
+        "command": "devbox.install",
+        "title": "Devbox: Install - Install packages specified in devbox.json"
+      },
+      {
+        "command": "devbox.init",
+        "title": "Devbox: Init - Initiate a devbox project"
```
LucilleH, line 39: Can we add `devbox run` command when it is out? (which should be tomorrow?)

vscode-extension/package.json
```diff
@@ -13,10 +13,72 @@
...
+        }
+      ]
+    },
+    "configuration": {
+      "title": "devbox",
+      "properties": {
+        "devbox.promptUpdateSettings": {
+          "type": "boolean",
+          "default": true,
+          "description": "Prompt to update workspace language settings to match devbox shell."
```
LucilleH, line 72: Do we still want this?

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
-
-	// Use the console to output diagnostic information (console.log) and errors (console.error)
+export function activate(context: ExtensionContext) {
 	// This line of code will only be executed once when your extension is activated
-	vscode.window.onDidOpenTerminal(async (event) => {
-		runDevboxShell();
+	initialCheckDevboxJSON();
+	// Creating file watchers to watch for events on devbox.json
+	const fswatcher = workspace.createFileSystemWatcher("**/devbox.json", false, false, false);
+	fswatcher.onDidDelete(e => {
```
LucilleH, line 16: Don't need the curly brakets?

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
+		const { stdout, stderr } = await exec("uname -m");
+		let cpuArch = stdout;
+		if (stderr) {
+			console.log(stderr);
+			const response = await window.showErrorMessage(
+				"Could not determine the CPU architecture type. Is your architecture type Apple M1/arm64?",
+				"Yes",
+				"No",
+			);
+			cpuArch = response === "Yes" ? "arm64" : "something else";
```
LucilleH, line 50: something else?

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
+		await commands.executeCommand('workbench.action.terminal.sendSequence', {
+			'text': 'devbox shell\r\n'
+		});
+	});
+
+	const devboxInstall = commands.registerCommand('devbox.install', async () => {
+		// todo: add support for --config path to devbox.json
+		ensureTerminalExists();
+		await commands.executeCommand('workbench.action.terminal.sendSequence', {
+			'text': 'devbox shell -- echo "installed packages!""\r\n'
```
LucilleH, line 79: nit: packages installed

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
+			}
+		}
+		if (pkg.startsWith("go_1_") || pkg === "go") {
+			if (extensions.getExtension("golang.go")?.isActive) {
+				workspace.getConfiguration("go").update("gopath", workspacePath + "/.devbox/nix/profile/default/bin/go");
+			}
+		}
+		if (pkg.startsWith("nodejs-") || pkg === "nodejs") {
+			if (extensions.getExtension("eslint")?.isActive) {
+				workspace.getConfiguration("eslint").update("nodepath", workspacePath + "/.devbox/nix/profile/default/bin/node");
```
LucilleH, line 149: Thoughts: This part of the logic makes me nervous. I'd rather avoid it and ask the user to manually configure for now, or expose an API so that other extensions can invoke

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
+		terminal.show();
 	}
 }
 
+function getDockerfileContent(): String {
+	return `
+	# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/debian/.devcontainer/base.Dockerfile
+
+	# [Choice] Debian version (use bullseye on local arm64/Apple Silicon): bullseye, buster
+	ARG VARIANT="buster"
```
LucilleH, line 220: Why not alpine?

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,290 @@
...
+
+	devboxJson["packages"].forEach((pkg: String) => {
+		if (pkg.includes("python3")) {
+			devcontainerObject.customizations.vscode.settings["python.defaultInterpreterPath"] = "/devbox/.devbox/nix/profile/default/bin/python3";
+			devcontainerObject.customizations.vscode.extensions.push("ms-python.python");
+		}
+		if (pkg.includes("go_1_") || pkg === "go") {
+			devcontainerObject.customizations.vscode.extensions.push("golang.go");
+		}
+		//TODO: add support for other common languages
```
LucilleH, line 283: Ditto - not sure if we should handle this at all

LucilleH:
Is there an opportunity to consolidate the Dockerfile with https://github.com/jetpack-io/axiom/pull/2831 and just overwrite the entrypoint?

vscode-extension/package.json
mohsenari: Just added it in the last commit

vscode-extension/package.json
mohsenari: No, I removed it

vscode-extension/src/extension.ts
mohsenari: fair point 😄  updated
I used to work with a senior JS dev who enforced the use if curly brackets at all times.

vscode-extension/src/extension.ts
mohsenari: the value of "something else doesn't matter as long as it's not `arm64` but I change the phrase to `"undefined"`

vscode-extension/src/extension.ts
mohsenari: updated

vscode-extension/src/extension.ts
mohsenari: this is removed along with others parts for updating other extension settings

mohsenari:
> Is there an opportunity to consolidate the Dockerfile with [jetpack-io/axiom#2831](https://github.com/jetpack-io/axiom/pull/2831) and just overwrite the entrypoint? Assuming the other Dockerfile will be published and you can use it as the base image?

@LucilleH I'm not sure since I haven't tested the devcontainer feature with alpine. Regardless, I'd rather not be blocked by the container publish for now. We can do our due diligent testing with alpine and once the base container is published in dockerhub I can quickly update this.

vscode-extension/src/extension.ts
LucilleH: oh I see!

vscode-extension/src/extension.ts
```diff
@@ -1,27 +1,162 @@
...
+			const result = await window.showQuickPick(items);
+			await runInTerminal(`devbox rm ${result}`);
+		} else {
+			window.showInformationMessage("No packages found in devbox.json");
+		}
+	});
+
+	const devboxInstall = commands.registerCommand('devbox.install', async () => {
+		// todo: add support for --config path to devbox.json
+		await runInTerminal('devbox shell -- echo "Packages installed""');
```
LucilleH, line 63: Not sure if we should have `devbox install` command at all. It doesn't do what a user might expect it to do, since `devbox shell -- echo "Packages installed"` installs the nix packages but does not expose them in the current PATH. So from a user's perspective, nothing is installed. Thoughts?

LucilleH (approved):

vscode-extension/src/extension.ts
mohsenari: yeah that's fair I'll remove it 👍

PR #263, opened 2022-11-04
[pkg-config] Implement basic mariadb config

mikeland86:
## Summary

* Adds golang template interpolation to the config and any files it creates
* Creates a basic v0.0.1 config for mariafb (can be greatly improved)

## How was it tested?

```bash
devbox add mariadb
devbox shell
mysqld
mysql
# success!
```


pkgcfg/package-configuration/mariadb.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "0.0.1",
+  "env": {
+    "MYSQL_BASEDIR": "{{ .DevboxProfileDefault }}",
+    "MYSQL_HOME": "{{ .DevboxRoot }}/conf/mysql",
+    "MYSQL_DATADIR": "{{ .DevboxRoot }}/conf/mysql/data",
+    "MYSQL_UNIX_PORT": "{{ .DevboxRoot }}/conf/mysql/mysql.sock",
+    "MYSQL_PID_FILE": "{{ .DevboxRoot }}/conf/mysql/mysql.pid"
+  },
+  "create_files": {
+    ".devbox/conf/mariadb/bin/mysql": "/mariadb/mysql",
```
savil, line 12: should this be `"{{ .DevboxRoot }}/conf/mariadb/bin/mysql": ... ` ?

pkgcfg/pkgcfg.go
```diff
@@ -19,34 +21,47 @@ type config struct {
...
 		}
-		if err := os.WriteFile(filePath, content, 0744); err != nil {
+		t, err := template.New(name + "-template").Parse(string(content))
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		var buf bytes.Buffer
+		if err = t.Execute(&buf, map[string]string{
+			"DevboxRoot":           filepath.Join(rootDir, ".devbox"),
+			"DevboxProfileDefault": filepath.Join(rootDir, ".devbox/nix/profile/default"),
```
savil, line 47: instead of hardcoding this here, lets reference `devbox.profileDir` (may have to refactor to make it publicly visible)

pkgcfg/pkgcfg.go
```diff
@@ -77,11 +92,23 @@ func getLocalConfig(configPath, pkg string) (*config, error) {
...
 	cfg := &config{localConfigPath: configPath}
-	if err = json.Unmarshal(content, cfg); err != nil {
+	t, err := template.New(pkg + "-template").Parse(string(content))
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	var buf bytes.Buffer
+	if err = t.Execute(&buf, map[string]string{
+		"DevboxRoot":           filepath.Join(rootDir, ".devbox"),
+		"DevboxProfileDefault": filepath.Join(rootDir, ".devbox/nix/profile/default"),
```
savil, line 102: ditto

savil (approved):

pkgcfg/package-configuration/mariadb.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "mariadb",
+  "version": "0.0.1",
+  "env": {
+    "MYSQL_BASEDIR": "{{ .DevboxProfileDefault }}",
```
gcurtis, line 5: It might be a good idea to remove the quotes and then `json.Marshal` the string values:

```suggestion
    "MYSQL_BASEDIR": {{ .DevboxProfileDefault }},
```

That way characters get escaped properly.

gcurtis (approved):

pkgcfg/package-configuration/mariadb.json
mikeland86: The key is a relative path so it's not needed. I think allowing relative paths makes writing the configs a bit nicer

pkgcfg/package-configuration/mariadb.json
mikeland86: These strings can be concatenated sometimes so I don't think to can do that. We could register a function so we could do:

`{{ escape .DevboxProfileDefault }}`

?

PR #272, opened 2022-11-10
ci: truncate long k8s namespace names

gcurtis:
They can't be longer than 63 characters.

### DIFF
--- a/.github/workflows/previews.yaml
+++ b/.github/workflows/previews.yaml
@@ -61,6 +61,15 @@ jobs:
         if: steps.filter.outputs.docs == 'true'
         run: |
           curl https://get.jetpack.io -fsSL | bash -s -- -y
+
+          # Make the GitHub branch name safe for Kubernetes namespaces names:
+          #
+          #   * `sed "s/[^0-9a-zA-Z]/-/g"`: replace any invalid characters in
+          #      the branch name with dashes.
+          #   * `tr -s -`: squash repeated consecutive dashes into one.
+          #   * `cut -b -63`: truncate the name to 63 characters.
+          safe_github_ref="$(echo "${GITHUB_REF_NAME}" | sed "s/[^0-9a-zA-Z]/-/g" | tr -s -)"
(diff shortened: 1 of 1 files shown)


.github/workflows/previews.yaml
```diff
@@ -61,6 +61,15 @@ jobs:
...
           curl https://get.jetpack.io -fsSL | bash -s -- -y
+
+          # Make the GitHub branch name safe for Kubernetes namespaces names:
+          #
+          #   * `sed "s/[^0-9a-zA-Z]/-/g"`: replace any invalid characters in
+          #      the branch name with dashes.
+          #   * `tr -s -`: squash repeated consecutive dashes into one.
+          #   * `cut -b -63`: truncate the name to 63 characters.
+          safe_github_ref="$(echo "${GITHUB_REF_NAME}" | sed "s/[^0-9a-zA-Z]/-/g" | tr -s -)"
+          namespace="$(echo "jetpack-io-preview-${safe_github_ref}-ref" | cut -b -63 | tr -s -)"
```
loreto, line 72: Is this a preview for the devbox docs?

If yes, could we make the namespace `"$(echo "devbox-docs-preview-${safe_github_ref}"`

loreto (approved):
LGTM

.github/workflows/previews.yaml
LucilleH: Wouldn't truncating this way means it could truncate to a number at the end again?

LucilleH:
Wouldn't truncating this way means it could truncate to a number at the end again?

loreto:
> Wouldn't truncating this way means it could truncate to a number at the end again?

That's a good point. Greg maybe the format should be `devbox-docs-${safe_github_ref}-ref", and you first truncate ``devbox-docs-${safe_github_ref}` to 59 characters and then add `-ref` at the end

gcurtis:
Ugh, good catch! I think everything should work now.

PR #273, opened 2022-11-10
Add golangci-lint to devbox.json

ipince:
## Summary

So that we can run the linter locally and not only when pushing to Github.

## How was it tested?
```
golangci-lint version
golangci-lint run
```


savil (approved):

savil:
thanks!

PR #277, opened 2022-11-11
[config] error if devbox.json has unknown fields

savil:
## Summary

A user reports that they added the following to their `devbox.json` but the scripts
got blown away when the config was upgraded to have the new `nixpkgs.commit` field.

```

"shell": {
        "init_hook": [ 
            // omitted
        ],
    }
"scripts": {
    "start" : "cargo run",
}
```

The error is that `scripts` should be a key inside `shell`, as in `shell.scripts`

## How was it tested?

```
devbox/testdata/go/go-1.19 on  main [!] via 🐹 v1.19 at ☸️  jetpack-cloud (savil-srivastava-jetpack-io)
➜ cat devbox.json
{
  "packages": ["go_1_19"],
  "scripts": "foo"
}
```
and did `devbox shell`, which gave this error:

```
Error: Failed to parse file at /Users/savil/code/jetpack/devbox/testdata/go/go-1.19/devbox.json: json: unknown field "scripts"
```


savil:
* main
  * **PR #277** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/277" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/277?utm_source=stack-comment).

gcurtis:
I think this will block us from adding new fields to the config. Users will get an error if they're running an older version of devbox and someone else added a newer field.

This is gross, but would this work?

1. Unmarshal the original config to a `map[string]json.RawMessage`
2. Marshal the updated config to JSON.
3. Unmarshal the updated config to the same `map[string]json.RawMessage` from step 1, overwriting the known fields.
4. Marshal the `map[string]json.RawMessage` to JSON and save.

savil:
> I think this will block us from adding new fields to the config. Users will get an error if they're running an older version of devbox and someone else added a newer field.

@gcurtis I don't think so. Unless I am misinterpreting what you said.

The `dec.DisallowUnknownFields()` prevents unknown fields from being in the devbox.json, but doesn't insist that all the fields defined in our struct `Config` must be present. So, if we add new fields to the `Config` struct, then older `devbox.json` will continue to work as before.

gcurtis:
@savil things will still work for the latest version of devbox, but not for older versions. For example:

1. User A and User B both work on a devbox project together. They're both running Devbox v1.
2. User A upgrades to Devbox v1.1, and runs `devbox add` which happens to add a new field called `new_option` to the `devbox.json`.
3. User B hasn't upgraded yet, so when they run Devbox v1 they get an error saying the field `new_option` is unrecognized.

savil:
@gcurtis Yep, you're right about this scenario breaking with this PR.

On the other hand, more realistically, what will happen is:
1. Devbox auto-updates to latest version. In this case, both teammates are on the latest version.
2. Both teammates choose to pin the devbox version, so it doesn't auto-update. In this case, they are responsible for coordinating together when they want to bump the pinned devbox version to a newer version.

The scenario where one teammate updates and the other doesn't shouldn't be supported IMO. The reason is that the user A's new devbox.json field could make meaningful changes to the shell environment, which they would want reflected in how their teammate user B exercises their project's code.

gcurtis:
I still think that requiring all teammates to immediately have the latest devbox isn't a great experience. What if the user installed devbox via Nix or by downloading the binary? Having devbox break because they ran `git pull` or switched branches would be confusing and frustrating.

I'd be less concerned if a) devbox could lock its own version in `devbox.json`, and b) it could still work but warn if the version doesn't match.

savil:
Discussed in team meeting. Follow up steps:

1. Print warning if a field is not recognized. This should be a head's up to a user who has placed a field incorrectly.
2. Use a golang library that can preserve unknown-fields and comments in `devbox.json`. This prevents the jarring experience outlined above by @gcurtis, and also lets our current version-pinning work so that users who experience some product regression can go back to an older version to unblock themselves.

PR #285, opened 2022-11-12
Tweak docs details on Docker builds

Lagoja:
## Summary
* Clarify that Docker is not required to run a shell
* Remove requirement to install Docker from install steps
* Emphasize exporting environment

## How was it tested?

Localhost


LucilleH (approved):
Cool!

PR #287, opened 2022-11-16
vscode extension terminal behavior fixes + readme update

mohsenari:
## Summary
Based on feedback from @Lagoja I added the following fixes:
- terminal only pops in and shows for devbox run and shell
- autoshell to a terminal doesn't happen when the opened terminal is from running devbox run/shell from extension's command palette (example: preventing duplicate devbox shell bug)
- updated readme to include features for version 0.0.2

## How was it tested?
- running the extension in vscode

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -111,7 +111,8 @@ async function initialCheckDevboxJSON() {
 async function runInTerminal(cmd: string, showTerminal: boolean) {
 	// check if a terminal is open
 	if ((<any>window).terminals.length === 0) {
-		const terminal = window.createTerminal({ name: 'DevboxTerminal' });
+		const terminalName = 'DevboxTerminal';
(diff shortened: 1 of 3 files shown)


vscode-extension/src/extension.ts
```diff
@@ -107,11 +108,13 @@ async function initialCheckDevboxJSON() {
 	}
 }
 
-async function runInTerminal(cmd: string) {
+async function runInTerminal(cmd: string, showTerminal: boolean) {
 	// check if a terminal is open
 	if ((<any>window).terminals.length === 0) {
-		const terminal = window.createTerminal({ name: `Terminal` });
-		terminal.show();
+		const terminal = window.createTerminal({ name: 'DevboxTerminal' });
```
LucilleH, line 114: Const for `DevboxTerminal`?

LucilleH (approved):

PR #289, opened 2022-11-16
referencing vscode token from env variable

mohsenari:
## Summary
changed how we reference vscode marketplace token

## How was it tested?
will have to be tested in github actions


loreto (approved):
LGTM

PR #288, opened 2022-11-16
Fix Example: A Rust Devbox

dritter:
## Summary

Fix Example: A Rust Devbox. This Example had some syntax errors in it.

PR #290, opened 2022-11-16
[github actions] add workflow_dispatch and pull_request_target events to previews and tests Actions

savil:
## Summary

The PR at #288 has pending workflows. However, the workflows are not running because
they are triggered by the `push` event, which doesn't get triggered for public forks.
This PR adds `pull_request_target` event for them.

It also adds `workflow_dispatch` for manually triggering the workflow.

## How was it tested?

not tested.

### DIFF
--- a/.github/workflows/previews.yaml
+++ b/.github/workflows/previews.yaml
@@ -1,10 +1,11 @@
 name: docs-preview
 
 on:
-  workflow_dispatch:
+  pull_request_target:
   push:
     branches-ignore:
       - main
+  workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #290** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/290" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/290?utm_source=stack-comment).

gcurtis (approved):

PR #291, opened 2022-11-16
[github actions] s/pull_request_target/pull_request for previews and tests Actions

savil:
## Summary

Reading more about pull_request_target, it looks like it may not be what we want. 
> Due to the dangers inherent to automatic processing of PRs, GitHub’s standard pull_requestworkflow trigger by default prevents write permissions and secrets access to the target repository. However, in some scenarios such access is needed to properly process the PR. To this end the pull_request_target workflow trigger was introduced.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

I think we want pull_request type:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

## How was it tested?

not tested

### DIFF
--- a/.github/workflows/previews.yaml
+++ b/.github/workflows/previews.yaml
@@ -1,7 +1,7 @@
 name: docs-preview
 
 on:
-  pull_request_target:
+  pull_request:
   push:
     branches-ignore:
       - main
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #291** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/291" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/291?utm_source=stack-comment).

gcurtis (approved):

PR #292, opened 2022-11-16
improve installing devbox docs

savil:
## Summary

I think clarifying that `nix` is required, while `docker` is optional would be good.

## How was it tested?

will view the preview docs from CICD.


LucilleH:
While you are at it, do you want to update the readme as well?

LucilleH (approved):
TY!

PR #294, opened 2022-11-17
Fixed github action workflow publishing vscode extension

mohsenari:
## Summary
Title says it.
The fix was to use `vsce publish -p <token>` rather than `echo <token> vsce login` then `vsce publish`.
For some reason github actions didn't like the 2 step method. 

## How was it tested?
Manually ran the github action workflow and it published the extension.

### DIFF
--- a/.github/workflows/release-extension.yaml
+++ b/.github/workflows/release-extension.yaml
@@ -20,8 +20,10 @@ jobs:
           yarn install
         working-directory: vscode-extension
       - name: publish
+        env:
+          VS_MARKETPLACE_TOKEN: ${{ secrets.VS_MARKETPLACE_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release-extension.yaml
```diff
@@ -23,7 +23,5 @@ jobs:
         env:
           VS_MARKETPLACE_TOKEN: ${{ secrets.VS_MARKETPLACE_TOKEN }}
         run: |
-          echo $VS_MARKETPLACE_TOKEN | vsce login jetpack-io
-          vsce package --yarn
-          vsce publish --yarn
+          vsce publish -p ${{ secrets.VS_MARKETPLACE_TOKEN }} --yarn
```
LucilleH, line 26: You can use the env variable, or get rid of the env definition.

LucilleH (approved):

PR #295, opened 2022-11-17
[pkgcfg] Add postgresql config

mikeland86:
## Summary

* postgres is actually pretty simple, does not require shim.
* running initdb will actually tell you how to start it (but uses logfile in root)
* readme text is not currently used

## How was it tested?

```
devbox add postgresql
devbox shell
initdb
pg_ctl start
pg_ctl stop
```


mikeland86:
FYI, I'm aware that the postgresql (and other) package may have version in the name (e.g. postgresql_14)

I plan to implement solution in different PR.

loreto (approved):
LGTM

PR #297, opened 2022-11-17
[pkgcfg] Allow for wildcards in package config filenames

mikeland86:
## Summary

This allows for wildcards in package configuration names. For example:

`postgresql*.json` will match:

* postgressql
* postgressql_14

## How was it tested?

```
devbox add postgresql
devbox add postgresql_14
```


loreto (approved):
LGTM to unblock, but ... are you sure you want to have the `wildcard` as part of the filename? I worry it could be error prone.

For example, you currently match against `postgresql_jdbc` and `postgresqlTestHook` and `postgresql14Packages.rum` to name a few.

mikeland86:
@loreto yeah that's fair. This might over-match. An alternative could be to have a match regex in the json:

`pkgMatchRegex: "postgresql(_[0-9]+)?"`

PR #298, opened 2022-11-17
[pkgcfg] Add ruby config

mikeland86:
## Summary

Very simple ruby config

## How was it tested?

```
devbox add ruby
devbox shell
gem install rainbow 
# inspected gem directory
ruby hello.rb # file that uses rainbow gem
```

image


pkgcfg/github.go
```diff
@@ -0,0 +1,89 @@
...
+
+var baseConfigURL = "https://raw.githubusercontent.com/jetpack-io/devbox/main/pkgcfg/package-configuration"
+
+const githubDirContentAPI = "https://api.github.com/repos/jetpack-io/devbox/contents/pkgcfg/package-configuration"
+
+func getConfig(pkg, rootDir string) (*config, error) {
+	confURL, err := getBestConfigPath(pkg)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	} else if confURL == "" {
```
savil, line 24: what scenarios would lead to an empty confURL?

pkgcfg/github_test.go
```diff
@@ -0,0 +1,30 @@
...
+	result := wildcardMatch("pkg*.json", "pkg.json")
+	assert.True(t, result)
+
+	result = wildcardMatch("pkg*.json", "pkg1.json")
+	assert.True(t, result)
+
+	result = wildcardMatch("pkg*.json", "pkg12.json")
+	assert.True(t, result)
+
+	result = wildcardMatch("pkg*.json", "pkg.json")
```
savil, line 19: repeated from line 10?

pkgcfg/package-configuration/ruby*.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "ruby",
+  "version": "0.0.1",
+  "env": {
+    "RUBY_CONFDIR": "{{ .DevboxRoot }}/conf/ruby",
+    "GEMRC": "{{ .DevboxRoot }}/conf/ruby/.gemrc",
+    "GEM_HOME": "{{ .DevboxRoot }}/conf/ruby"
+  }
```
savil, line 8: @ipince you've had recent ruby experience, so could you double check this? Thanks!

savil (approved):

pkgcfg/github.go
mikeland86: if there's no json for that package. I guess there's technically a third error case where there may be a json but we errored out (possibly because of a network failure)

mikeland86: edit: error case already handled, so all 3 cases handled correctly.

pkgcfg/package-configuration/ruby*.json
ipince: Hmm... well I didn't really mess with setting these vars, so I'm not sure. Seems good to me though :)

ipince (approved):

PR #299, opened 2022-11-18
[pkgcfg] Add nginx config

mikeland86:
## Summary

Adds basic nginx config. Creates a simple config file and a shim that uses it. Can be customized with environment variable.

Note that the readme is currently not implemented but I'm adding so it will be available once it is.

## How was it tested?

```
devbox add nginx
devbox shell
nginx
```

### DIFF
--- a/pkgcfg/package-configuration/nginx.json
+++ b/pkgcfg/package-configuration/nginx.json
@@ -1,0 +1,14 @@
+{
+  "name": "nginx",
+  "version": "0.0.1",
+  "readme": "To start nginx use command \"nginx\". nginx is configured to use .devbox/conf/nginx.conf. Use $NGINX_CONFDIR to change the configuration directory. Use $NGINX_LOGDIR to change the log directory. Use $NGINX_PIDDIR to change the pid directory. Use $NGINX_RUNDIR to change the run directory. Use $NGINX_SITESDIR to change the sites directory. Use $NGINX_TMPDIR to change the tmp directory. Use $NGINX_USER to change the user. Use $NGINX_GROUP to customize.",
+  "env": {
+    "NGINX_CONFDIR": "{{ .DevboxRoot }}/conf/nginx"
+  },
+  "create_files": {
+    ".devbox/conf/nginx/temp": "",
+    ".devbox/conf/nginx/nginx.conf": "nginx/nginx.conf",
+    ".devbox/conf/nginx/fastcgi.conf": "nginx/fastcgi.conf",
+    ".devbox/conf/nginx/bin/nginx": "nginx/nginx"
+  }
+}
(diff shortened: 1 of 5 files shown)


pkgcfg/package-configuration/nginx.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "nginx",
+  "version": "0.0.1",
+  "readme": "To start nginx use command \"nginx\". nginx is configured to use .devbox/conf/nginx.conf. Use $NGINX_CONFDIR to change the configuration directory. Use $NGINX_LOGDIR to change the log directory. Use $NGINX_PIDDIR to change the pid directory. Use $NGINX_RUNDIR to change the run directory. Use $NGINX_SITESDIR to change the sites directory. Use $NGINX_TMPDIR to change the tmp directory. Use $NGINX_USER to change the user. Use $NGINX_GROUP to customize.",
```
savil, line 4: Does this Readme get rendered somewhere or is it for reading within this file?

Would suggest adding some new line chars inside this

pkgcfg/package-configuration/nginx/nginx.conf
```diff
@@ -0,0 +1,20 @@
+events {}
+http{
+server {
+         listen       80;
+         listen       [::]:80;
```
savil, line 5: Why the two separate listen calls?

pkgcfg/pkgcfg.go
```diff
@@ -55,6 +55,7 @@ func CreateFiles(pkg, rootDir string) error {
 		}
 		var buf bytes.Buffer
 		if err = t.Execute(&buf, map[string]string{
+			"UserRoot":             rootDir,
```
savil, line 58: Is root dir the absolute path to project directory?

savil (approved):
Nice!

pkgcfg/package-configuration/nginx.json
mikeland86: Currently readme is not used at all. I'm planning on implementing it and improving them pending a conversation with @loreto

pkgcfg/package-configuration/nginx/nginx.conf
mikeland86: Not sure, copy pasted this from @Lagoja 's template

pkgcfg/pkgcfg.go
mikeland86: yep. I named it UserRoot because we also have DevboxRoot which is the path to `.devbox`

savil: we also have the concept of `configDir` which is the path to where the project's `devbox.json` is kept. 

ugh, although this `configDir` is written as `Devbox.srcDir` in `devbox.go`. I can send a PR to update that.

PR #300, opened 2022-11-18
[pkgcfg] Add Apache httpd config

mikeland86:
## Summary

Adds config for apache httpd

## How was it tested?

image

### DIFF
--- a/pkgcfg/package-configuration/apache/httpd.conf
+++ b/pkgcfg/package-configuration/apache/httpd.conf
@@ -1,0 +1,63 @@
+ServerAdmin             "root@localhost"
+ServerName              "devbox-apache"
+Listen                  "${HTTPD_PORT}"
+PidFile                 "${HTTPD_CONFDIR}/apache.pid"
+
+LoadModule mpm_event_module modules/mod_mpm_event.so
+LoadModule authz_host_module modules/mod_authz_host.so
+LoadModule authz_core_module modules/mod_authz_core.so
+LoadModule auth_basic_module modules/mod_auth_basic.so
+LoadModule mime_module modules/mod_mime.so
+LoadModule headers_module modules/mod_headers.so
+LoadModule unixd_module modules/mod_unixd.so
+LoadModule status_module modules/mod_status.so
+LoadModule proxy_module modules/mod_proxy.so
+LoadModule proxy_fcgi_module modules/mod_proxy_fcgi.so
+LoadModule dir_module modules/mod_dir.so
+LoadModule alias_module modules/mod_alias.so
+
+<IfModule unixd_module>
+    User daemon
+    Group daemon
+</IfModule>
+
+<Directory />
+    AllowOverride none
+    Require all denied
+</Directory>
+
+DocumentRoot  "${PWD}/web"
+<Directory "${PWD}/web">
+    Options Indexes FollowSymLinks
+    AllowOverride None
+    Require all granted
+</Directory>
+
+<Files ".ht*">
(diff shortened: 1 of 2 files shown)


pkgcfg/package-configuration/apache/httpd.conf
```diff
@@ -0,0 +1,63 @@
...
+
+<IfModule unixd_module>
+    User daemon
+    Group daemon
+</IfModule>
+
+<Directory />
+    AllowOverride none
+    Require all denied
+</Directory>
```
savil, line 27: what does this do?

pkgcfg/package-configuration/apache/httpd.conf
```diff
@@ -0,0 +1,63 @@
...
+        <IfModule mod_authz_host.c>
+            Require all granted
+        </IfModule>
+    </Directory>
+
+    ## Added for php-fpm
+    # ProxyPassMatch ^/(.*\.php(/.*)?)$ fcgi://{LOCALHOST}:{PHPFPM_PORT}/<?php echo $root ?>/$1
+    DirectoryIndex index.html 
+
+</VirtualHost>
```
savil, line 63: what resource are you referencing when writing this? Is there a recommended conf on the apache website? It's hard to code review without more knowledge since I'm not an apache expert.

pkgcfg/package-configuration/apache/httpd.conf
mikeland86: it defaults all directories to access denied (and later whitelists just `web`)

pkgcfg/package-configuration/apache/httpd.conf
mikeland86: @Lagoja this may be a question for you. Where did you get these? we may want to point to the source.

pkgcfg/package-configuration/apacheHttpd.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "apache",
+  "version": "0.0.1",
+  "readme": "To start use \"apachectl start -f $HTTPD_CONFDIR/httpd.conf\".",
+  "env": {
+    "HTTPD_CONFDIR": "{{ .DevboxRoot }}/conf/apache"
+  },
+  "create_files": {
+    ".devbox/conf/apache/httpd.conf": "apache/httpd.conf"
```
LucilleH, line 9: Similar to nginx, I feel like these default files can be directly copied from the `/nix/store`, instead of copying them from this repo. It seems like we need a way to specify files that we need to copy over from the immutable nix store to `.devbox/`

Almost like
```
".devbox/conf/apache/httpd.conf": "{{ .packagePath }}/lib/httpd.conf"
```

mikeland86: @LucilleH that's reasonable (@loreto @Lagoja and I chatted about it a bit). From what I can tell though the default apache config is not 100% ready to use. So while it makes sense to try to use it as a base, it would require some editing if we want it to be ready to use without changes.

We chatted about potentially bringing over all of the nix store files (some as symlinks and others as simple rewrites so their internal paths point to right place) but we're punting on it for now because of complexity.

Anyway, since these are versioned, we can always improve them without being afraid of breaking people's setup.

savil (approved):
approving to unblock, but would still appreciate some pointers on the apache conf code....thanks!

mikeland86:
@savil gonna merge. All of this is behind feature flag so we can iterate

PR #303, opened 2022-11-18
[dependabot] Fix vulnerability

mikeland86:
## Summary

* Upgrade dependencies to ensure we get a patched loader-utils
* Add a few packages to docs devbox (as recommended by devbox shell)

## How was it tested?

`yarn start`


docs/app/devbox.json
```diff
@@ -1,6 +1,12 @@
 {
-  "packages": [],
+  "packages": [
+    "nodejs-16_x",
```
LucilleH, line 3: Can just do `nodejs`

LucilleH (approved):
TY!

docs/app/devbox.json
mikeland86: FYI, reason I chose this:

image

PR #305, opened 2022-11-18
rename srcDir to configDir for clarity

savil:
## Summary

This is more technically accurate since we set the erstwhile `srcDir` from the `--config` flag.
Or if the `--config` is not set, then we search in parent-directories for the `devbox.json`
and set it to that directory.

## How was it tested?

compiles

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -53,10 +53,9 @@ func InitConfig(dir string) (created bool, err error) {
 // Nix packages.
 type Devbox struct {
 	cfg *Config
-	// srcDir is the directory where the config file (devbox.json) resides
-	// TODO savil. Rename to configDir.
-	srcDir string
-	writer io.Writer
+	// configDir is the directory where the config file (devbox.json) resides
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #305** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/305" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/305?utm_source=stack-comment).

ipince (approved):

mikeland86 (approved):
Don't feel strongly, but I've also seen us use `rootDir` ? config is just always overloaded but src is not terribly descriptive either

savil:
yeah, I hear ya. The idea is that plain `config` without qualifiers refers to the core `devbox.json` config, whereas other configs will need qualifiers and prefixes like `pkgconfig`.

PR #306, opened 2022-11-19
[pkgcfg] Show readme when adding packages. Add info command

mikeland86:
## Summary

This shows notes whenever a package configuration has a non-empty "readme" field. It also adds a new (hidden) command `devbox info <pkg>` that displays said info.

I also updated a few readme fields to make them look better.

## How was it tested?

`devbox add mariadb`
`devbox add nginx`
`devbox info mariadb`

image

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -111,6 +111,13 @@ func (d *Devbox) Add(pkgs ...string) error {
 	if err := d.ensurePackagesAreInstalled(install); err != nil {
 		return err
 	}
+	if featureflag.Get(featureflag.PKGConfig).Enabled() {
+		for _, pkg := range pkgs {
+			if _, err := pkgcfg.PrintReadme(pkg, d.srcDir, d.writer); err != nil {
+				return err
+			}
+		}
+	}
 	return d.printPackageUpdateMessage(install, pkgs)
 }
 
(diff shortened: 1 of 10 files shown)


mikeland86:
I do realize that having a config flag for the info command is kinda weird. It's also not really needed (it gets used to build the package config struct but we only really need the name and the "readme" field so the config path is not used). This requires a slight change in `pkgcfg` architecture so I might do it in a follow up.

devbox.go
```diff
@@ -340,6 +347,17 @@ func (d *Devbox) PrintShellEnv() error {
...
 
+func (d *Devbox) Info(pkg string) error {
+	hasReadme, err := pkgcfg.PrintReadme(pkg, d.configDir, d.writer)
+	if err != nil {
+		return err
+	}
+	if !hasReadme {
+		_, err = fmt.Fprintf(d.writer, "No info available for %s\n", pkg)
+	}
+	return errors.WithStack(err)
```
savil, line 358: hmm the `devbox info <pkg>` command is a bit confusing. I'd expect the `info` command to detail information about any nix package, and additionally any devbox-package (what are we calling it?).

But it does only the latter, so it would feel confusing or incomplete to the user.

Is there a plan to add info about nix packages as well?

savil: @Lagoja @loreto what do you think?

loreto:
Yeah, I think:
+ `devbox info <pkg>` should accept any package
+ It should print the metadata about the package. We already have `PkgInfo` in `nix.go`, so it should be trivial to print that.
+ For packages that have a README / additional configuration info, we can *also* print that.

mikeland86:
I'll add `PkgInfo` but I have a suspicion it is not useful.

savil (approved):

PR #309, opened 2022-11-22
[pkgcfg] Add php extension support

mikeland86:
## Summary

This links php extensions by using nix definitions. We might be able to do this without nix, but this is most straight forwardway.

RFC: I'm reusing part of the planner interface because it already had definitions, but this could be its own thing.

## How was it tested?

```
devbox add php
devbox add phpExtensions.memcached
devbox shell
php -m | grep memecached
```


pkgcfg/package-configuration/php*.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "php",
+  "version": "0.0.1",
+  "readme": "PHP is compiled with default extensions. If you would like to use non-default extensions you can add them them with `devbox add php81Extensions.memcached`"
```
savil, line 4: double `them`

savil: suggestion:
> If you would like to use non-default extensions you can add them with `devbox add php81Extensions.{extension}` . For example, for the memcache extension you can do `devbox add php81Extensions.memcached`.

planner/languages/php/php_v2_planner.go
```diff
@@ -0,0 +1,70 @@
...
+	"strings"
+
+	"go.jetpack.io/devbox/planner/plansdk"
+)
+
+type V2Planner struct {
+	userPackages []string
+}
+
+// PHPV2Planner implements interface Planner (compile-time check)
```
savil, line 18: `implements interface PlannerForPackages`

savil (approved):
I assume some of this will evolve with @mohsenari's work to remove build and planners.

mikeland86:
@savil I think planners are staying for shell package hints but we may rename them?

I can change this once we decide in what direction that is going.

cc: @mohsenari @LucilleH

PR #311, opened 2022-11-23
Add Github Badges

Lagoja:
## Summary

Add SVGs to the static folder that can be used as Github Badges.

## How was it tested?

Visually verified


LucilleH (approved):

loreto:
LGTM.

For what it's worth: I'm tempted to kill the white version as it's less "on brand" than the other two.

PR #320, opened 2022-11-29
[pkgcfg] Embed config instead of using GitHub

mikeland86:
## Summary

ugh, so it turns out GitHub API limits are very low. I thought it was 60 per minute but it was 60 per hour per IP. So reading configs from GitHub doesn't work unless we want to authenticate, but I didn't want to do that. We could also cache or use a proxy but both seemed like more work than it was worth.

Instead, this changes package configuration to use embedded configs instead of GitHub. It means that configurations are now tied to releases. In the future if we want to revive the GitHub code to make the packages more configurable by allowing custom URLs we can do that, but it doesn't feel like a priority right now.

Other improvements:
* Added a new `env` file that allows users to manually source env variables if they run `devbox add` inside a devbox shell.
* Added a new `match` field that allows us to be more precise when deciding to use a config. If it is not provided then the config is used if it is contained in a package name.

## How was it tested?

```
devbox add nginx
devbox add ruby
devbox shell
```


loreto (approved):
LGTM

PR #321, opened 2022-11-29
[cloud] ensure devbox.json exists and pass devbox struct to cloud.Shell functions

savil:
## Summary

This does some preliminary work to ensure that `devbox cloud shell`:
1. a devbox.json exists and we handle the `--config` flag like all the other commands.
2. can use the devbox.ConfigDir for syncing from
3. can use the name of the devbox.ConfigDir to eventually sync to in the vm. Not hooked up since we need VM changes for this.

## How was it tested?

TODO

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -8,20 +8,21 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/fatih/color"
+	"go.jetpack.io/devbox"
 	"go.jetpack.io/devbox/cloud/mutagen"
 	"go.jetpack.io/devbox/cloud/sshclient"
 	"go.jetpack.io/devbox/cloud/sshconfig"
 	"go.jetpack.io/devbox/cloud/stepper"
+	"go.jetpack.io/devbox/debug"
 )
 
-func Shell() error {
-	// TODO: check if `devbox.json` exists.
-	// TODO: find project's "root" directory based on `devbox.json` location.
+func Shell(box *devbox.Devbox) error {
 	setupSSHConfig()
 
 	c := color.New(color.FgMagenta).Add(color.Bold)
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #321** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/321" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/321?utm_source=stack-comment).

cloud/cloud.go
```diff
@@ -93,21 +94,24 @@ func getVirtualMachine(username string) string {
...
 		Name:        fmt.Sprintf("devbox-%s", id),
-		AlphaPath:   ".", // We should use location of `devbox.json`
+		AlphaPath:   box.GetConfigDir(),
 		BetaAddress: fmt.Sprintf("%s@%s", username, hostname),
 		// It's important that the beta path is a "clean" directory that will contain *only*
 		// the projects files. If we pick a pre-existing directories with other files, those
 		// files will be synced back to the local directory (due to two-way-sync) and pollute
 		// the user's local project
-		BetaPath:  "~/Code/",
+		BetaPath:  "~/Code/", // Use ~/Code/{dirName}, where dirName is the variable above
```
savil, line 114: We'll sync to `~/{dirName}`. No need for `Code/` prefix.

devbox.go
```diff
@@ -85,6 +85,10 @@ func Open(dir string, writer io.Writer) (*Devbox, error) {
 	return box, nil
 }
 
+func (d *Devbox) GetConfigDir() string {
```
gcurtis, line 88: In Go you usually leave out the Get prefix:

```suggestion
func (d *Devbox) ConfigDir() string {
```

gcurtis (approved):

cloud/cloud.go
loreto: I kinda like:
`~/code/<project>` or `~/project/<project>` in case we ever want to sync up multiple projects to the VM, and keep your home directory "uncluttered"

FWIW, this is what stackblitz does: 
`~/projects/nuxt-starter-klkvsl`

savil: okay okay, I don't feel strongly either ways, so lets go with `~/code/<project>`

gcurtis (approved):

PR #323, opened 2022-11-29
 add gateway subdomains to sshconfig to avoid StrictHostKeyChecking

savil:
## Summary

`devbox cloud shell` would hang since ssh would want user confirmation about connecting
to a previously unknown host.

For the gateway subdomains, we should not make this a strict requirement.

## How was it tested?

did `devbox cloud shell` and could connect successfully to the vm/agent.


savil:
* main
  * **PR #323** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/323" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/323?utm_source=stack-comment).

loreto (approved):
LGTM

PR #324, opened 2022-11-30
[pkgcfg] Remove config if package is removed

mikeland86:
## Summary

Removes config and broken symlinks when a package is removed. We could also clean up configs as part of `ensurePackagesAreInstalled` (which makes it more reliable) but this requires a bit of refactoring I didn't want to do as part of this PR.

Small unrelated change: made `installMode` type an int using iota (which is a bit more inline with golang quasi-enum pattern)

## How was it tested?

Manual testing:

* Added nginx and mariadb. 
* Inspected conf directory
* Removed one by one
* Inspected conf directory at each step.

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -143,6 +143,10 @@ func (d *Devbox) Remove(pkgs ...string) error {
 		return err
 	}
 
+	if featureflag.Get(featureflag.PKGConfig).Enabled() {
+		pkgcfg.Remove(d.configDir, uninstalledPackages)
+	}
+
 	if err := d.ensurePackagesAreInstalled(uninstall); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


devbox.go
```diff
@@ -423,11 +429,11 @@ func (d *Devbox) profileBinDir() (string, error) {
...
 
 // installMode is an enum for helping with ensurePackagesAreInstalled implementation
-type installMode string
+type installMode int
 
 const (
-	install   installMode = "install"
-	uninstall installMode = "uninstall"
+	install   installMode = iota
+	uninstall installMode = iota
```
savil, line 436: Not a fan of this change TBH:

I prefer to type enums as strings because they are easier to debug (example: if we need to print or log the value).
The int enums are more space efficient and make more sense where that is a concern

savil (approved):

devbox.go
mikeland86: I'll revert because this is unrelated to PR

PR #325, opened 2022-11-30
[featureflags] Simplify feature flags

mikeland86:
## Summary

* Simplifies feature flag lib by using a single const as the name and the feature itself. 
* Removes the need for `init`
* Makes callsites more concise
* Remove `NixpkgVersion` flag

## How was it tested?

unit tests


savil (approved):
nice, thanks for the clean up 👍🏾

PR #327, opened 2022-11-30
Point the cli to the production gateway

loreto:
## Summary
Point the CLI to the production gateway (`gateway.devbox.sh`) by default.

Note that this means that if you're working on the gateway, you'll need to re-compile the devbox binary to test against your development version. Let me know if you prefer I add an (internal) flag to select which gateway to talk to.

## How was it tested?
Didn't.

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -76,9 +76,15 @@ type authResponse struct {
 }
 
 func getVirtualMachine(username string) string {
+	// When developing we can use this env variable to point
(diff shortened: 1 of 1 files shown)


savil (approved):
would be nice to have an env-var we could override the gateway destination with

gcurtis (approved):
Agree with having an environment variable.

loreto:
Added an env variable.

PR #329, opened 2022-11-30
[cloud] client directly calls start_devbox_server.sh <project dir>

savil:
## Summary

This PR updates the cloud client to call `ssh -t <server> "bash -l -c \"start_devbox_server.sh <project-name>\""`.
This informs the server about which project to start the `devbox shell` for.

## How was it tested?

setup:
```
fly machines run --port 22:2222/tcp -e SSH_USER=savil --name=savil-test --region=sjc .
```

ran:
```
DEVBOX_DEBUG=1 DEVBOX_VM=savil@0e286003ad5486.vm.devbox-vms.internal devbox cloud shell
```
and got a `devbox shell` and could see the project's files and execute it.

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -34,6 +34,7 @@ func Shell(box *devbox.Devbox) error {
 	s1 := stepper.Start("Creating a virtual machine on the cloud...")
 	vmHostname := getVirtualMachine(username)
 	s1.Success("Created virtual machine")
+	debug.Log("vmHostname: %s", vmHostname)
 
 	s2 := stepper.Start("Starting file syncing...")
 	err := syncFiles(username, vmHostname, box)
@@ -90,6 +91,7 @@ func getVirtualMachine(username string) string {
 	if err != nil {
 		log.Fatal(err)
 	}
+	debug.Log("gateway.devbox.sh auth response: %s", string(bytes))
 	resp := &authResponse{}
 	err = json.Unmarshal(bytes, resp)
 	if err != nil {
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -29,7 +29,7 @@ func Shell(configDir string) error {
 	fmt.Println("Blazingly fast remote development that feels local")
 	fmt.Print("\n")
 
-	username, vmHostname := parseVmEnvVar()
+	username, vmHostname := parseVMEnvVar()
 	if username == "" {
 		username = promptUsername()
 	}
@@ -155,7 +155,7 @@ func projectDirName(configDir string) string {
 	return name
 }
 
-func parseVmEnvVar() (username string, vmHostname string) {
+func parseVMEnvVar() (username string, vmHostname string) {
 	vmEnvVar := os.Getenv("DEVBOX_VM")
 	if vmEnvVar == "" {
 		return "", ""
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #329** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/329" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/329?utm_source=stack-comment).

cloud/cloud.go
```diff
@@ -127,15 +127,16 @@ func syncFiles(username string, hostname string, box *devbox.Devbox) error {
 	return nil
 }
 
-func shell(username string, hostname string) error {
+func shell(username string, hostname string, box *devbox.Devbox) error {
```
loreto, line 130: I don't love passing box as an argument here, but it's fine for a quick implementation.

loreto (approved):
LGTM

cloud/cloud.go
savil: what is your concern?

For now, we use `box.ConfigDir()` so another alternative is to pass the `configDir string` as an argument.

cloud/sshclient/sshclient.go
```diff
@@ -38,7 +40,9 @@ func (c *Client) Exec(remoteCmd string) ([]byte, error) {
 }
 
 func (c *Client) cmd() *exec.Cmd {
-	cmd := exec.Command("ssh", destination(c.Username, c.Hostname))
+
+	remoteCmd := fmt.Sprintf("bash -l -c \"start_devbox_shell.sh %s\"", c.ProjectDirName)
```
savil, line 44: no no no, this is wrong :(

`cmd()` is called from multiple spots. I need to pass the `bash -l ...` remote command into the `exec()` function above.

cloud/sshclient/sshclient.go
```diff
@@ -2,19 +2,26 @@ package sshclient
...
+	Username       string
+	Hostname       string
+	Port           int
+	ProjectDirName string
 }
 
 func (c *Client) Shell() error {
-	cmd := c.cmd()
+	cmd := c.cmd("-t")
+	remoteCmd := fmt.Sprintf("bash -l -c \"start_devbox_shell.sh %s 300\"", c.ProjectDirName)
```
gcurtis, line 22: The project name needs quotes in case of spaces:

```suggestion
	remoteCmd := fmt.Sprintf(`bash -l -c "start_devbox_shell.sh \"%s\" 300"`, c.ProjectDirName)
```

gcurtis (approved):

cloud/cloud.go
```diff
@@ -90,6 +94,7 @@ func getVirtualMachine(username string) string {
 	if err != nil {
 		log.Fatal(err)
 	}
+	debug.Log("gateway.devbox.sh auth response: %s", string(bytes))
```
gcurtis, line 97: I dunno if this would've helped when debugging, but you can also log the abbreviated stderr output inside of `client.Exec` with `debug.Log(string(exerr.Stderr))`.

cloud/sshclient/sshclient.go
savil: good catch!

PR #331, opened 2022-12-01
devbox run inside shell

mohsenari:
## Summary
Added support for running devbox run when inside a devbox shell
## How was it tested?
- Compile devbox
- `./devbox shell`
- `./devbox run <script_name>`
example output:
devbox.json
```json
{
  "packages": [
    "python38"
  ],
  "shell": {
    "init_hook": null,
    "scripts": {
      "test": [
        "echo helloworldtest",
        "ls -l"
      ]
    }
  },
  "nixpkgs": {
    "commit": "af9e00071d0971eb292fd5abef334e66eda3cb69"
  }
}
```

output:
```sh
(devbox) ➜  sample1 ./devbox run test
helloworldtest
total 23700
-rwxr-xr-x 1 mohsenansari 21500226 Nov 30 18:41 devbox
-rw-r--r-- 1 mohsenansari      250 Nov 30 22:26 devbox.json
-rw-r--r-- 1 mohsenansari        0 Oct  5 18:27 requirements.txt
-rw-r--r-- 1 mohsenansari        0 Oct  5 18:26 setup.py
```

### DIFF
--- a/boxcli/run.go
+++ b/boxcli/run.go
@@ -57,10 +57,10 @@ func runScriptCmd(args []string, flags runCmdFlags) error {
 	}
 
 	if devbox.IsDevboxShellEnabled() {
-		return errors.New("You are already in an active devbox shell.\nRun 'exit' before calling devbox run again. Shell inception is not supported yet.")
+		err = box.RunScriptInShell(script)
+	} else {
+		err = box.RunScript(script)
 	}
-
-	err = box.RunScript(script)
 
 	var exitErr *exec.ExitError
 	if errors.As(err, &exitErr) {
--- a/devbox.go
+++ b/devbox.go
@@ -304,7 +304,6 @@ func (d *Devbox) RunScriptInShell(scriptName string) error {
 		shell = &nix.Shell{}
 	}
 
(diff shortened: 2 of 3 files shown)


ipince (approved):
Sorry for the late review @mohsenari .

I'm approving to unblock, because this implementation works (as evidenced by your demo). But I think maybe there's a simpler implementation?

A suggested alternative is:
* Always add the script commands in the generated rcfile, not only when calling `RunScript`. That is, also pass `nix.WithUserScript(scriptName, script.String()),` in `Shell()`. That way, we know that the scripts are always in our rc file.
* When you call `devbox run <script>` inside an existing devbox shell, then just execute `<cmd>`, instead of (a) setting up the environment and (b) executing `<shell> -ic <cmd>`.

I think that'd be better... Unless, of course, we have a good reason why we want to always execute the command in a new shell, instead of the existing shell. I don't know.

cc @savil  to follow-on while I'm gone.

mohsenari:
@ipince @savil As discussed on today's meeting. I'll merge this to have the fix in main branch. We'll do more testing and discuss a better approach in future cycles.

PR #334, opened 2022-12-01
[pkgcfg] Ensure php-fpm works out of the box

mikeland86:
## Summary

This makes php-fpm work pretty much out of the box. It creates an php-fpm config (deamonized). Adds default ports and always includes `ProxyPassMatch` in the apache config (because it should not break anything if even if not used)

## How was it tested?

```
devbox add php apacheHttpd
devbox shell
php-from
apachectl start -f $HTTPD_CONFDIR/httpd.conf
```


Lagoja (approved):
Lgtm

PR #336, opened 2022-12-01
deprecate devbox build + hide devbox plan

mohsenari:
## Summary
Title says it.

## How was it tested?
- compile
- try ./devbox build
- try ./devbox plan

### DIFF
--- a/boxcli/build.go
+++ b/boxcli/build.go
@@ -4,11 +4,7 @@
 package boxcli
 
 import (
-	"os"
-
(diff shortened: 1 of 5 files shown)


LucilleH (approved):
Should we mark `devbox build` hidden?

mohsenari:
> Should we mark `devbox build` hidden?

I marked it as `deprecated: true` which both removes it from `devbox help` output and shows a message when it is used.
So in a sense it does both hiding and deprecating.

PR #337, opened 2022-12-02
[nix] Add nix install command

mikeland86:
## Summary

This is a first step in having devbox install nix for you. This command can help us iterate faster. It's currently hidden.

This adds `devbox nix install` which does exactly that, it attempts to install nix. Right now all output/input is redirected to stdout/stdin which means the user will control how the installation proceeds (nix explains, prompts and requests sudo). 

In follow ups we can improve this by:

* Requiring sudo up front
* Hiding prompts and answering them automatically.
* Detecting if nix is not installed and offering to install it when user takes any step that requires it.

## How was it tested?

`devbox nix install`


loreto (approved):
LGTM, but two requests:
1. Could we call the command `devbox setup nix` – I'm thinking we might end up with others like `devbox setup mutagen` and I don't want to create a top-level command for each tool.

2. Could we make the command hidden for now?

PR #340, opened 2022-12-05
cloud: use unique SSH keys per VM

gcurtis:
## Summary

Apologies for the giant diff. A lot of it was from just moving `sshclient` and `sshconfig` into `openssh`.

When the gateway creates a new VM for a user, it also generates an ed25519 key pair that the client can use to login. This change updates the client to save the key pair in `~/.config/devbox/ssh/keys` and configures SSH to use it when connecting to the VM.

- Consolidate `cloud/sshclient` and `cloud/openssh` into `cloud/openssh` to make it clearer that they manipulate the OpenSSH command.
- Remove `openssh.Client.Port` and instead parse the port from the hostname. This allows the port to be overridden via `DEVBOX_GATEWAY`.
- Add `openssh.AddVMKey(host, key)` that adds a new key to `~/.config/devbox/ssh/keys`.
- Atomically edit `~/.ssh/config` and `~/.config/devbox/ssh/*` files so that partial edits don't break the user's SSH.
- Add a bunch of tests that check that the client correctly modifies the user's `~/.ssh/config` and `~/.config/devbox` files.
- Misc. bug fixes related to SSH configs:
	- Quote paths in ~/.ssh/config to handle paths with spaces.
	- Make the devbox include regex case-insensitive and detect when the include is commented out.
	- Set recommended permissions on the various SSH directories and files.

## How was it tested?

Added some unit tests.


cloud/cloud.go
```diff
@@ -77,33 +74,49 @@ func promptUsername() string {
...
+	if envGateway := os.Getenv("DEVBOX_GATEWAY"); envGateway != "" {
+		client.Hostname = envGateway
 	}
 	bytes, err := client.Exec("auth")
 	if err != nil {
-		log.Fatal(err)
+		log.Println(err)
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			log.Println("ssh gateway.devbox.sh stderr:\n", string(exitErr.Stderr))
```
loreto, line 103: Since we can change the hostname of the gateway via an env variable, do you want to pass in `client.Hostname` instead of hardcoding it?
```suggestion
			log.Println("ssh %s stderr:\n", client.Hostname, string(exitErr.Stderr))
```

loreto (approved):
LGTM

cloud/cloud.go
gcurtis: Good idea. I parameterized it in the debug log for stdout too.

PR #343, opened 2022-12-06
[nix-installer] auto-install nix if we detect it's missing

mikeland86:
## Summary

This change makes devbox attempt to install nix if it is missing and the command the user ran requires it. It shows yellow text alerting the user before it starts (otherwise it is really jarring). One improvement could be to combine the sudo warning into the yellow text.

If the installation fails we show the error that caused it. If the installation succeeds we show a user warning which is yellow text. This warning is an error which causes the CLI execution to end. We need this because nix requires restarting the terminal. This is another area where having a pkg installer may be an improvement. 

cc: @Lagoja 
## How was it tested?

Ran `devbox add mariadb` without having nix installed. Got:

image

Pressing enter started the installation.

After installation we show:

image

### DIFF
--- a/nix/install.go
+++ b/nix/install.go
@@ -1,7 +1,6 @@
 package nix
 
 import (
-	"bufio"
 	_ "embed"
 	"fmt"
 	"io"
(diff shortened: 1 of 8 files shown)


loreto:
Do you feel like the installation process is reliable enough to run it automatically already?

I think I lean towards:
+ Make `devbox setup nix` public
+ If nix is not installed print an error saying nix is required, please run `devbox setup nix` to install

That way we can track the reliability of `devbox setup nix` before we start doing it automatically. Thoughts?

Separately we could change our install script to call `devbox setup nix` at the end (that way if you run the installer, nix gets installed as part of that process)

nix/install.go
```diff
@@ -43,3 +48,20 @@ func Install() error {
 
 	return errors.WithStack(cmd.Wait())
 }
+
+func EnsureInstalled(cmd *cobra.Command, args []string) error {
+	_, err := exec.LookPath("nix-shell")
```
gcurtis, line 53: Do you think we should also check to see if `/nix` exists? That way we don't try to install it again if the user forgot to restart their session or a macOS update stomped on `/etc/zshrc`.

nix/install.go
```diff
@@ -43,3 +48,20 @@ func Install() error {
...
+	if err == nil {
+		return nil
+	}
+
+	if featureflag.NixInstaller.Enabled() {
+		color.Yellow("\nNix is not installed. Devbox will attempt to install it. \n\nPress enter to continue.\n")
+		_, _ = bufio.NewReader(os.Stdin).ReadBytes('\n')
```
gcurtis, line 60:
```suggestion
		fmt.Scanln()
```

nix/install.go
```diff
@@ -43,3 +48,20 @@ func Install() error {
...
 }
+
+func EnsureInstalled(cmd *cobra.Command, args []string) error {
+	_, err := exec.LookPath("nix-shell")
+	if err == nil {
+		return nil
+	}
+
+	if featureflag.NixInstaller.Enabled() {
+		color.Yellow("\nNix is not installed. Devbox will attempt to install it. \n\nPress enter to continue.\n")
```
gcurtis, line 59: Can we also add something like "... or ctrl-c to exit." so the user knows how to bail out?

mikeland86:
@loreto sure, we can do that (I don't feel strongly)

I haven't really hardened the nix installation process. It does print out useful error messages if it fails, but covering all types of failures is hard.

gcurtis (approved):
@Daniel I vote for leaving this as-is instead of making `devbox setup nix` public. There's a feature flag and we can always let the user cancel at the prompt.

nix/install.go
mikeland86: oh good call

loreto:
> @daniel I vote for leaving this as-is instead of making `devbox setup nix` public. There's a feature flag and we can always let the user cancel at the prompt.

I'm not opposed to it – but I have worries on the UX besides the reliability.

If I type `devbox shell python310` – and it asks me to install nix, but then it immediately creates the shell and drops me in it, will I understand what happened? Does the installation take a while? How much output does it show?

Are we just saying that we'll hide it behind a feature flag, and we won't enable it until we've ironed that stuff out? (if that's what we're saying, that sounds fine, but my follow up question would be what's the smallest thing we can polish that we can put in front of users ASAP)

rogeliogonzalez3531-a11y:
gh pr checkout 343

PR #345, opened 2022-12-07
[cloud] set KeepAlive setting for ssh to vm

savil:
## Summary

The ssh session to my vm would auto-close with this output:

```
❯ time ssh savil@91859edc4e2783.vm.devbox-vms.internal
Welcome to Devbox Cloud!


(devbox) savil in 🌐 91859edc4e2783 in ~
❯ Connection to proxy.devbox.sh closed by remote host.
Connection to 91859edc4e2783.vm.devbox-vms.internal closed by remote host.
Connection to 91859edc4e2783.vm.devbox-vms.internal closed.
ssh savil@91859edc4e2783.vm.devbox-vms.internal  0.08s user 0.01s system 0% cpu 1:01.30 total
```

Setting this value in the config would result in the ssh connection staying active
for longer than a minute.

## How was it tested?

1. Set the config value, and could keep the connection alive.
2. As a sanity test, I removed the value and tried again: connection terminated at one minute again.

### DIFF
--- a/cloud/openssh/sshconfig.tmpl
+++ b/cloud/openssh/sshconfig.tmpl
@@ -13,3 +13,4 @@ Host *.devbox-vms.internal
 	StrictHostKeyChecking no
 	UserKnownHostsFile "{{ .ConfigDir }}/known_hosts"
 	ServerAliveInterval 50
+	ServerAliveCountMax 3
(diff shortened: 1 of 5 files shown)


savil:
* main
  * **PR #345** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/345" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/345?utm_source=stack-comment).

savil:
Finally got a consistent repro for this!

gcurtis (approved):
Should we explicitly set `ServerAliveCountMax` to 3 too? Just in case the user has changed the default somewhere else in their config.

savil:
> ServerAliveCountMax

I had that in my config when testing initially, but then removed it after reading the docs and seeing it had a default value.

I like your reason though. I'll add it.

Issue #279, opened 2022-11-11
Generate a Generic Container?


vsoch:
hey devbox! I was wondering - why not have a fallback container generation for "all other cases" that are outside python, go, etc? For example, if we use nixos/nix and then the best way to install devbox, and then initiative a shell with a command that exits (to install stuffs) it would be fairly straight forward to have that be the fallback - here is an example: https://github.com/rse-ops/devbox/blob/main/bases/devbox/Dockerfile. And maybe I'd even do:

```dockerfile
FROM nixos/nix
RUN nix-channel --update
RUN wget https://github.com/jetpack-io/devbox/releases/download/0.1.0/devbox_0.1.0_linux_amd64.tar.gz && \
    tar -xzvf devbox_0.1.0_linux_amd64.tar.gz && \
    chmod +x devbox && \
    mv devbox /usr/bin
ENV PATH=/usr/bin:$PATH
WORKDIR /code
COPY devbox.json devbox.json
RUN devbox shell -- echo "Installing packages"
ENTRYPOINT ["devbox"]
CMD ['shell']
```
So the default entrypoint is devbox shell right into the environment (not I haven't tried this, my container build above I default to a shell and then run `devbox shell` at my leisure!

LucilleH:
Wow this is awesome. In fact, instead of the fallback scenario, this should be the default. cc @Lagoja

vsoch:
Thanks! The one thing I'd like to improve upon is to figure out a way to have a one liner "install the latest devbox" that will work in the container. If we use something like the above, we'd minimally want to expose the version as an `ARG` I think. Let me know if I can help! I would absolutely love to be able to do like:

```bash
$ devbox build --name ghcr.io/rse-ops/openmpi-devbox
```
in a directory with a devbox.json and magically get a container with it!

Lagoja:
I agree this should be the default behavior. I think having a way to get a generic container that has your packages + configuration installed could help a lot with portability. We'll add this as a task.

Thanks for suggesting @vsoch

vsoch:
Heyo! I'm trying this out on a few computers (looking for issues) - right now on an older ubuntu and hit:

```
$ docker run -it ghcr.io/rse-ops/openmpi-devbox:clang-14 
bash-5.1# devbox shell
Installing nix packages. This may take a while...
[...]

PR #346, opened 2022-12-07
Added `devbox generate devcontainer` command

mohsenari:
## Summary
- Added `devbox generate devcontainer` command that generates the following files:
  -  `.devcontainer/Dockerfile` and a `.devcontainer/devcontainer.json`
- Added `devbox generate dockerfile` command that generates Dockerfile in same directory as `devbox.json`

## How was it tested?
- compile
- run `devbox generate devcontainer`
- run `ls -la` to confirm the directory (`.devcontainer`) and the files mentioned above are created.
- run `devbox generate dockerfile`
- run `ls -la` to confirm dockerfile is created.


boxcli/container/container_util.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	return nil
+}
+
+func getDockerfileContent() string {
+
+	return `
+	# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/debian/.devcontainer/base.Dockerfile
```
LucilleH, line 82: Put this in a .tmpl file?

mohsenari: Done

boxcli/container.go
```diff
@@ -0,0 +1,57 @@
...
+	command.AddCommand(exportCmd())
+
+	return command
+}
+
+func exportCmd() *cobra.Command {
+	flags := containerCmdFlags{}
+	command := &cobra.Command{
+		Use:   "export",
+		Short: "Generate Dockerfile to build a container.",
```
LucilleH, line 36: `build a dev container` - it is only for development purposes, not something people can use in production? Also, it seems like it is very vscode specific

tmpl/devcontainerDockerfile.tmpl
```diff
@@ -0,0 +1,27 @@
...
+USER vscode
+
+# Install nix
+ARG NIX_INSTALL_SCRIPT=https://nixos.org/nix/install
+RUN curl -fsSL \${NIX_INSTALL_SCRIPT} | sh -s -- --no-daemon
+RUN . ~/.nix-profile/etc/profile.d/nix.sh
+ENV PATH /home/vscode/.nix-profile/bin:\${PATH}
+
+# Install devbox
+RUN sudo mkdir /devbox && sudo chown vscode /devbox
```
LucilleH, line 20: Is there a reason why we don't do `WORKDIR /devbox` here?

tmpl/devcontainerDockerfile.tmpl
```diff
@@ -0,0 +1,27 @@
...
+RUN curl -fsSL https://get.jetpack.io/devbox | bash -s -- -f
+
+# Setup devbox environment
+COPY --chown=vscode ./devbox.json /devbox/devbox.json
+RUN devbox shell --config /devbox/devbox.json -- echo "Nix Store Populated"
+ENV PATH /devbox/.devbox/nix/profile/default/bin:\${PATH}
+ENTRYPOINT devbox shell
```
LucilleH, line 27: The previous line has `RUN devbox shell --config /devbox/devbox.json`. How does this `ENTRYPOINT` find the right config location?

tmpl/devcontainerDockerfile.tmpl
```diff
@@ -0,0 +1,27 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/debian/.devcontainer/base.Dockerfile
+
+# [Choice] Debian version (use bullseye on local arm64/Apple Silicon): bullseye, buster
+ARG VARIANT="buster"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-\${VARIANT}
```
LucilleH, line 5: I wonder if we really need to tie ourselves to vscode base image, or can we generate a generic image?

LucilleH (commented):
There's another feature request related to this https://github.com/jetpack-io/devbox/issues/279

I guess do we want to export a vscode specific image, or a generic one, or both? I lean towards generating a generic image, and leave the vscode specific one in the vscode extension. We can even publish the generic one in dockerhub, and people can do 
```
From devbox 
```
as their docker base image.

There's something uneasy about tying this CLI command & its generated image with vscode. Thoughts? @Lagoja @loreto this is more like a product question.

mohsenari:
Based on the last conversation I'm going to make the following changes:

- [ ] Simplify Dockerfile to use a nixos base image
- [ ] Rename the command to `devbox generate devcontainer`
- [ ] Add a force flag to override existing files/directories 
- [ ] Create another sub command called `devbox generate dockerfile` to generate only a dockerfile

devbox.go
```diff
@@ -379,6 +381,73 @@ func (d *Devbox) Info(pkg string) error {
...
+}
+
+// generates a Dockerfile that replicates the devbox shell
+func (d *Devbox) GenerateDockerfile(force bool) error {
+	dockerfilePath := filepath.Join(d.configDir, "Dockerfile")
+	filesExist := false
+	// check if Dockerfile doesn't exits
+	if plansdk.FileExists(dockerfilePath) {
+		filesExist = true
+	}
```
LucilleH, line 434: Can this be simplified as `filesExist := plansdk.FileExists(dockerfilePath)`?

devbox.go
```diff
@@ -379,6 +381,73 @@ func (d *Devbox) Info(pkg string) error {
...
+
+	filesExist := false
+	// check if devcontainer.json doesn't exits
+	if plansdk.FileExists(devContainerJSONPath) {
+		filesExist = true
+	}
+	// check if Dockerfile doesn't exits
+	if plansdk.FileExists(dockerfilePath) {
+		filesExist = true
+	}
```
LucilleH, line 400: Should we just say, if the directory `.devcontainer` exist, then files exist?

devbox.go
```diff
@@ -379,6 +381,73 @@ func (d *Devbox) Info(pkg string) error {
...
+	// construct path to devcontainer directory
+	devContainerPath := filepath.Join(d.configDir, ".devcontainer/")
+	devContainerJSONPath := filepath.Join(devContainerPath, "devcontainer.json")
+	dockerfilePath := filepath.Join(devContainerPath, "Dockerfile")
+
+	filesExist := false
+	// check if devcontainer.json doesn't exits
+	if plansdk.FileExists(devContainerJSONPath) {
+		filesExist = true
+	}
```
LucilleH, line 396: Can this be simplified as filesExist := plansdk.FileExists(devContainerJSONPath)?

boxcli/generate.go
```diff
@@ -46,3 +48,62 @@ func runGenerateCmd(_ *cobra.Command, args []string, flags *generateCmdFlags) er
 
 	return box.Generate()
 }
+
+func devcontainerCmd() *cobra.Command {
+	flags := &generateCmdFlags{}
+	command := &cobra.Command{
+		Use:   "devcontainer",
+		Short: "Generate Dockerfile and devcontainer.json files",
```
LucilleH, line 56: add "under `.devcontainer/` folder" ?

LucilleH (approved):

devbox.go
mohsenari: I initially did that, but I realized if users have other files in there, it will be annoying to have to delete the folder just to generate the devcontainer files.

boxcli/generate/devcontainer_util.go
```diff
@@ -0,0 +1,147 @@
...
+		return errors.WithStack(err)
+	}
+	return nil
+}
+
+func getDevcontainerContent(pkgs []string) *devcontainerObject {
+	// getting the correct container image variant based on whether it is running on Apple arm64 chip.
+	variant := "buster"
+	if getCPUArch() == "arm64" {
+		variant = "bullseye"
```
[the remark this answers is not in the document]
mohsenari: Good call removed this and the corresponding function

PR #342, opened 2022-12-06
[experimental] add ssh shim in devbox cloud for mutagen

savil:
## Summary

this introduces a command that is a shim for the `ssh` command.

The idea is that `devbox cloud ssh` will first verify that the machine is active
before invoking ssh on it. Mutagen says that it uses `MUTAGEN_SSH_PATH` to find
the ssh and scp binaries, so we intend to set this env-var before starting
the sync-sessions.

- [ ] generate the ssh and scp shim files
- [ ] test

## How was it tested?

**ssh shim**
Manually made two executable files as shims for ssh and scp:
```
❯ cat ~/.config/devbox/shims/ssh
#!/bin/bash

devbox cloud ssh "$@"

❯ cat ~/.config/devbox/shims/scp
#!/bin/bash

scp "$@"
```

Sanity-test for ssh shim. I could call this and get a remote shell:
```
> ~/.config/devbox/shims/ssh savil@<machine>.vm.devbox-vms.internal 
```

**Using the ssh shim with devbox cloud shell**
Close old mutagen daemon: `mutagen daemon stop`. NOTE: This was needed to ensure a new daemon is started. This will be a problem for users who already are running a mutagen daemon.

But this does NOT yet work, in terms of calling the ssh shim in `~/.config/devbox/shims`:
```
MUTAGEN_SSH_PATH=/Users/savil/.config/devbox/shims DEVBOX_DEBUG=1 DEVBOX_VM=savil@e286072fe29986.vm.devbox-vms.internal devbox cloud shell
...
✘ Starting file syncing [FAILED]
Connecting to agent (POSIX)...                                                  Creating session...
Error: unable to connect to beta: unable to connect to endpoint: unable to dial agent endpoint: unable to handshake with agent process: server magic number incorrect (error output: [DEBUG] 2022/12/05 20:30:33 /Users/savil/code/jetpack/devbox/boxcli/midcobra/debug.go:46: Debug mode enabled.
bash: line 1: .mutagen/agents/0.16.1/mutagen-agent: No such file or directory
```
This error is arising from mutagen invoking this command via `devbox cloud ssh`:
```
executing command: /usr/bin/ssh -oConnectTimeout=5 -oServerAliveInterval=10 -oServerAliveCountMax=1 savil@e148e446b51189.vm.devbox-vms.internal .mutagen/agents/0.16.1/mutagen-agent synchronizer --log-level=info
```
I've been trying to locate where `.mutagen/agents/` should be, or how it is configured on the Beta machine.

PR #348, opened 2022-12-07
[cloud] ssh shim part 1: generate shim files

savil:
## Summary

This PR generates the shim ~files~ symlinks in `~/.config/devbox/shims`.
It also sets the MUTAGEN_SSH_PATH env-var when `devbox cloud shell` is starting the sync.

Follow up PRs (break up #342):
- part 2: hook up `devbox` to do the ssh
- part 3: hook up `devbox cloud ssh` to check for inactive VM and terminate mutagen

## How was it tested?

control:
```
DEVBOX_DEBUG=1 DEVBOX_VM=savil@21781990f34e89.vm.devbox-vms.internal devbox cloud shell
```
could open a shell in the remote machine

test:
```
DEVBOX_FEATURE_SSH_SHIM=1 DEVBOX_DEBUG=1 DEVBOX_VM=savil@21781990f34e89.vm.devbox-vms.internal devbox cloud shell
```
could also open the same shell. Also, verified the generated files in the `.config` dir.


savil:
* main
  * **PR #348** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/348" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/348?utm_source=stack-comment).

cloud/mutagen/wrapper.go
```diff
@@ -123,6 +125,14 @@ func execMutagen(args []string) error {
 	return nil
 }
 
+func env(envVars map[string]string) []string {
+	customEnv := []string{}
```
gcurtis, line 129: You can start out with the environment to avoid the extra append.

```suggestion
	customEnv := os.Environ()
```

cloud/sshshim/sshshim.go
```diff
@@ -0,0 +1,57 @@
...
+	shimDir, err := Dir()
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if err := openssh.EnsureDirExists(shimDir, 0744, true /*chmod*/); err != nil {
+		return err
+	}
+
+	if err := os.WriteFile(filepath.Join(shimDir, "ssh"), []byte(sshShimContents), 0744); err != nil {
```
gcurtis, line 40: What about just making symlinks instead of scripts? It might simplify things a bit.

cloud/sshshim/sshshim.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+const (
+	configShimDir = ".config/devbox/shims"
+)
+
+const sshShimContents = `#!/bin/sh
+# Next PR:
+# devbox cloud ssh "$@"
+ssh "$@"
```
gcurtis, line 19: If you do keep the script, I'd make this an `exec` to replace the shell process.

boxcli/featureflag/feature.go
```diff
@@ -44,3 +44,7 @@ func (f *feature) Enabled() bool {
 	}
 	return f.enabled
 }
+
+func (f *feature) Disabled() bool {
```
gcurtis, line 48: Why both `Enabled()` and `Disabled()`?

gcurtis (approved):
What do you think of putting this in the `openssh` package? It looks like it would fit well with the other `SetupDevbox` stuff.

cloud/sshshim/sshshim.go
savil: how would a symlink work for calling `devbox cloud ssh` when `ssh` is invoked by mutagen?

boxcli/featureflag/feature.go
savil: I found it more clear in the caller code.

In my initial implementation, I missed the not operator in `!featureflag.<name>.Enabled()` and then switched to Disabled which is more clear.

savil:
I'm gonna further clean up in another PR the envVars being passed into the mutagen.Exec caller functions. Was hyper focused on just getting this working, so it's a bit incomplete with passing nil in some places.

We can also refactor mutagen.List to call execMutagen instead of the current duplication.

cloud/openssh/sshshim/generate.go
```diff
@@ -0,0 +1,67 @@
...
+	//if err != nil {
+	//	return errors.WithStack(err)
+	//}
+	sshExecutablePath, err := exec.LookPath("ssh")
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	if err := os.Symlink(sshExecutablePath, filepath.Join(shimDir, "ssh")); err != nil {
+		if !os.IsExist(err) {
```
gcurtis, line 42: Should the symlink always be overwritten just in case the devbox path changed?

gcurtis (approved):

cloud/openssh/sshshim/generate.go
savil: Yeah, I also ran into it because the ssh symlink needed to be updated from /usr/bin/ssh to point to the devbox binary (when testing Part 1 and then Part 2 PRs) 

I'll make this fix.

savil: fixed in part 2 PR: https://github.com/jetpack-io/devbox/pull/349/commits/077eeb274bac167f7ab6ac12050f928ee220314f

PR #349, opened 2022-12-07
[cloud] ssh shim part 2: invoke ssh

savil:
## Summary

This PR hooks up the ssh command invocation.

Two design choices:

1. Choose to have the `ssh` symlink in `~/.config/devbox/ssh/shims` point to `devbox` binary, and have devbox first check if the `args[0]` is `ssh` before starting the cobra logic for regular commands. This ensures that all flags are handled directly by ssh, instead of cobra. The alternate design would be implementing `devbox cloud ssh` command which relies on cobra parsing flags, and any global devbox flags would interfere.

2. Choose syscall.Exec instead of exec.Cmd so that we preserve the exit code and environment, and the current process is replaced by ssh. Without this, we'd see errors during mutagen sync: the mutagen binary would fail to be copied to Beta, the remote machine. 

Also make a small fix to ensure we always overwrite the ssh and scp symlinks to point to the most-recent devbox (for ssh) and scp (for scp) binary locations.

## How was it tested?

```
>  mutagen sync terminate -a && mutagen daemon stop
>  DEVBOX_FEATURE_SSH_SHIM=1 DEVBOX_DEBUG=1 DEVBOX_VM=savil@21781990f34e89.vm.devbox-vms.internal devbox cloud shell
```

NOTE: future work would require managing any running mutagen daemon so it is started with the MUTAGEN_SSH_PATH

### DIFF
--- a/boxcli/root.go
+++ b/boxcli/root.go
@@ -5,6 +5,7 @@ package boxcli
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


savil:
* main
    * **PR #349** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/349" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/349?utm_source=stack-comment).

gcurtis (approved):

savil:
@gcurtis could you make another sanity check? I updated to use `syscall.Exec` instead of `exec.Cmd`, and also implemented the symlink overwriting.

gcurtis (approved):

PR #350, opened 2022-12-08
[nix] Improve nix installer

mikeland86:
## Summary

A few small changes:

* No more sudo up-front. This breaks situations where user is already root but sudo is not available
* Default to `--daemon` (multi-user) install. This makes the macOS and linux installs the same
* pipe error to stdout. Helps debug issues

## How was it tested?

* Ran `devbox setup nix` on mac and ubuntu (via docker)

### DIFF
--- a/nix/install.go
+++ b/nix/install.go
@@ -25,13 +25,13 @@ func Install() error {
 	defer r.Close()
 	defer w.Close()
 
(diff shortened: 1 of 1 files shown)


loreto (approved):
LGTM

PR #353, opened 2022-12-09
[services] Add services to package configs (plugins)

mikeland86:
## Summary

This adds new `services` command (behind flag). It allows:

`devbox services ls`
`devbox services start|stop [service]`

Services are configured using package configuration (plugins)

Also modified `info` command to include services instructions.

## How was it tested?

```
devbox services ls
devbox services start nginx
devbox services start apache
devbox services start php-fpm
...
devbox services stop nginx
devbox services stop apache
devbox services stop php-fpm
```


loreto (approved):
LGTM

BUT, hide the command (so that's not publicly visible) until we make `devbox services start` work outside the shell too.

mikeland86:
Command is hidden with flag.

PR #357, opened 2022-12-12
[extension] Generate the same devcontainer files as CLI

mohsenari:
## Summary
CLI command `devbox generate devcontainer` generates a Dockerfile that's Alpine based. The vscode extension generates a microsoft-based image and requires finding out if the cpu architecture is arm64 based. 
This change aligns both generated files to be identical and removes the need for differentiating arm64 based image.

## How was it tested?
- Run extension via vscode (run extension action in run & debug)
- Choose "Devbox: Generate Dev Containers config files" from command palette (cmd + shift + p)
- Confirm the .devcontainer files are generated.
- Choose "Dev Containers: Rebuild and Reopen in Container" from command palette.
- Confirm the IDE opens successfully in a container.


vscode-extension/src/devcontainer.ts
```diff
@@ -42,37 +42,39 @@ export async function readDevboxJson(workspaceUri: Uri) {
...
-	RUN apt -y install --no-install-recommends curl xz-utils
-
-	USER vscode
-
-	# Install nix
-	ARG NIX_INSTALL_SCRIPT=https://nixos.org/nix/install
-	RUN curl -fsSL \${NIX_INSTALL_SCRIPT} | sh -s -- --no-daemon
+    FROM alpine:3
+
+    # Setting up devbox user
```
LucilleH, line 47: Reuse the same `.tmpl` file instead of copy pasting the string?

mohsenari: fair point! 
I updated the code to use the CLI similar to the rest of commands in the extension. It's very simple now.

LucilleH (approved):
Cool! But question: I thought we'd call `devbox generate devcontainer` instead of `devbox generate dockerfile`?

mohsenari:
> Cool! But question: I thought we'd call `devbox generate devcontainer` instead of `devbox generate dockerfile`?

there are 2 commands: `devbox generate devcontainer` creates devcontainer directory and files
`devbox generate dockerfile` just creates a dockerfile

PR #355, opened 2022-12-10
[cloud] ssh shim part 5: run own mutagen daemon

savil:
## Summary

This PR enables running devbox's own mutagen daemon by setting MUTAGEN_DATA_DIRECTORY.

Refactored code so that more of the devbox-specific logic is centralized in `mutagenbox`
package, which then calls the generic `mutagen` package.

## How was it tested?

```
# start a non-devbox mutagen daemon
> brew install mutagen
> which mutagen 
/usr/local/bin/mutagen
> mutagen daemon start
> ps aux | grep mutagen
# confirm the daemon is running: `mutagen daemon start` will be an active process
```

Then:
1. manually start a vm (command omitted due to closed source, but happy to share)
2. `DEVBOX_FEATURE_SSH_SHIM=1 DEVBOX_VM=username@vmaddress devbox cloud shell`
3. `cd ~/.config/devbox/ssh/shims` and `tail -f logs.txt` to monitor the shim.
4. exit the cloud shell 
5. `fly m stop <id>` to stop the vm
6. `touch foo.txt` in the devbox project folder to trigger a sync action
7. wait for the `logs.txt` to show the `mutagen sync terminate` command.
8. `MUTAGEN_DATA_DIR=~/.config/devbox/mutagen ~/.cache/mutagen/bin/mutagen sync list` to confirm the session is terminated.

To confirm two distinct daemons are running:
- `ps aux | grep mutagen` to see two running mutagen daemons (i.e. two processes listed as `mutagen daemon start`)
- `MUTAGEN_DATA_DIR=~/.config/devbox/mutagen ~/.cache/mutagen/bin/mutagen daemon stop`
- `ps aux | grep mutagen` to see that the non-devbox mutagen daemon is still running.

PR #351, opened 2022-12-09
[cloud] ssh shim part 3: invoke scp

savil:
## Summary

This PR generalizes part 2 to also shim scp

## How was it tested?

1. started a vm
2. `DEVBOX_FEATURE_SSH_SHIM=1 DEVBOX_DEBUG=1 DEVBOX_VM=savil@24d89664a95187.vm.devbox-vms.internal devbox cloud shell`
gave a devbox shell in the vm

PR #358, opened 2022-12-12
[midcobra] add executionID

savil:
## Summary

**Motivation**
The sshshim command path bypasses the existing cobra stack. We need to add some basic telemetry around error tracking so we have visibility into errors users may run into (due to our programming mistakes).

We would like to reuse the existing `midcobra` abstraction which has a `telemetry` middleware that does both sentry and segment tracking. However, for sshshim we desire just sentry, not segment. So we want to:
1. split telemetry middleware into sentry and segment middlewares
2. set a common sentryID for both sentry and segment to correlate the corresponding events/data.

**Implemention**
This PR adds an executionID to `midCobra.Middleware`.

We use this executionID in two ways:

1. We set executionID in both sentry.EventID and segment's sentry_event_id
2. For an error and stack trace reported when `DEVBOX_DEBUG=1`, we also print this executionID.
  If a user reports a reproducible error, then we can ask them to enable DEVBOX_DEBUG to reproduce
  it and report the executionID to us. We can then look up the corresponding events in both
  sentry and segment to construct an understanding of the problem the user is facing.

In the next PRs, I will:
1. split telemetry into sentry and segment middleware
2. call the sentry middleware for the sshshim command path

## How was it tested?

Set up:
1. via code changes, forced the `midcobra.telemetry` middleware to be enabled.
2. added a print statement where the segment.sentry_event_id is set to print
   both the executionID and the eventID returned by sentry's CaptureException. 
3. Manually inserted an error in `devbox.Open` and ran a CLI command to trigger it.

Observed:
1. executionID and eventID from sentry's CaptureException are the same.
2. the executionID is printed prior to the stack trace of the error.

### DIFF
--- a/boxcli/midcobra/debug.go
+++ b/boxcli/midcobra/debug.go
@@ -16,7 +16,8 @@ import (
 )
 
 type DebugMiddleware struct {
-	flag *pflag.Flag
+	executionID string // uuid
+	flag        *pflag.Flag
 }
 
 var _ Middleware = (*DebugMiddleware)(nil)
@@ -62,5 +63,10 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
 	}
 
 	st := debug.EarliestStackTrace(runErr)
-	debug.Log("Error: %v\n%+v", runErr, st)
+	debug.Log("Error: %v\nexecutionID:%s\n%+v\n", runErr, d.executionID, st)
+}
+
+func (d *DebugMiddleware) withExecutionID(execID string) Middleware {
+	d.executionID = execID
+	return d
 }
--- a/boxcli/midcobra/debug.go
+++ b/boxcli/midcobra/debug.go
@@ -63,7 +63,7 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
 	}
 
 	st := debug.EarliestStackTrace(runErr)
-	debug.Log("Error: %v\nexecutionID:%s\n%+v\n", runErr, d.executionID, st)
+	debug.Log("Error: %v\nExecutionID:%s\n%+v\n", runErr, d.executionID, st)
 }
 
 func (d *DebugMiddleware) withExecutionID(execID string) Middleware {
--- a/boxcli/midcobra/midcobra.go
+++ b/boxcli/midcobra/midcobra.go
@@ -5,7 +5,7 @@ package midcobra
 
 import (
 	"context"
-	"strings"
+	"encoding/hex"
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
@@ -85,5 +85,6 @@ func executionID() string {
 	// An EventID must be 32 characters long, lowercase and not have any dashes.
 	//
 	// so we pre-process to match sentry's requirements:
-	return strings.ToLower(strings.ReplaceAll(uuid.New().String(), "-", ""))
+	id := uuid.New()
+	return hex.EncodeToString(id[:])
 }
--- a/boxcli/midcobra/midcobra.go
+++ b/boxcli/midcobra/midcobra.go
@@ -5,7 +5,9 @@ package midcobra
 
 import (
 	"context"
+	"strings"
 
+	"github.com/google/uuid"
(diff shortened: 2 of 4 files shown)


savil:
* main
    * **PR #349** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/349" target="_blank">Graphite</a> 
      * **PR #351** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/351" target="_blank">Graphite</a> 
        * **PR #352** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/352" target="_blank">Graphite</a> 
          * **PR #355** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/355" target="_blank">Graphite</a> 
            * **PR #358** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/358" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/358?utm_source=stack-comment).

LucilleH (approved):
cool!

boxcli/midcobra/debug.go
```diff
@@ -62,5 +63,10 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
 	}
 
 	st := debug.EarliestStackTrace(runErr)
-	debug.Log("Error: %v\n%+v", runErr, st)
+	debug.Log("Error: %v\nexecutionID:%s\n%+v\n", runErr, d.executionID, st)
```
gcurtis, line 66:
```suggestion
	debug.Log("Error: %v\nExecutionID: %s\n%+v\n", runErr, d.executionID, st)
```

boxcli/midcobra/debug.go
```diff
@@ -62,5 +63,10 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
...
+}
+
+func (d *DebugMiddleware) withExecutionID(execID string) Middleware {
+	d.executionID = execID
+	return d
 }
```
gcurtis, line 72: Should this be optional? Why not just always generate an execution ID? If you do keep it optional, there should probably be a check to see if it's empty before printing it in `postRun`.

boxcli/midcobra/midcobra.go
```diff
@@ -63,3 +74,16 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
...
+	// google/uuid package's String() returns a value of the form:
+	// xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+	//
+	// but sentry's EventID specifies:
+	//
+	// > EventID is a hexadecimal string representing a unique uuid4 for an Event.
+	// An EventID must be 32 characters long, lowercase and not have any dashes.
+	//
+	// so we pre-process to match sentry's requirements:
+	return strings.ToLower(strings.ReplaceAll(uuid.New().String(), "-", ""))
```
gcurtis, line 88:
```suggestion
	id := uuid.New()
	return hex.EncodeToString(id[:])
```

boxcli/midcobra/sentry.go
```diff
@@ -0,0 +1 @@
+package midcobra
```
gcurtis, line 1: Should this file be deleted?

gcurtis (approved):

boxcli/midcobra/debug.go
savil: its optional in the sense that I can see some middleware not caring about handling this value (i.e. saving it and using it)

The midcobra framework is responsible for calling withExecutionID, so I don't think individual middlewares need to worry about it not being called. In debug middleware's case, if postRun would has an empty value, then it'll just print as empty-string....which is fine i guess?

boxcli/midcobra/midcobra.go
savil: ahhh good call

boxcli/midcobra/sentry.go
savil: oops, yes, it doesn't belong in this PR. I use this file in the next one.

PR #362, opened 2022-12-13
[plugins] Move conf files out of .devbox and more DEV-1259 DEV-1258

mikeland86:
## Summary

This changes how some plugins work (FKA package configuration) and makes a few related improvements. 

Specifically:

* conf files are moved out of `.devbox/conf` and moved to `./devbox.d`. (or `.DevboxDir`)
* Files created outside of `.devbox` directory are never overridden.
* Hidden files are now saved in `.devbox/virtenv` (or `.Virtenv`)
* mariadb plugin has been disabled (not yet ready for prime time, it's the only plugin that requires shims)
* Added service to postgresql (I forget it in previous PR)
* Only files stored in `bin` get file mode 755 and sym links. Everything else gets 644 and no link.
* Updated all plugins to use new variables.
* Added `web/index.html` example file to nginx and apache plugins.

Everything is still hidden behind flag so no user change.

TODO (in follow up):

* Only create non-`.devbox` files when doing `devbox add` (vs now they get created on `shell`, `run`, `generate` etc) (DEV-1260)

## How was it tested?

Added: `postgresql`, `nginx`, and `php`. 

* Inspected configs in `./devbox.d` directory. 
* Inspected .devbox/virtenv directory.
* Started all services. 
* Created simple `web/index.html` for nginx and curled it. All working.


pkgcfg/package-configuration/postgresql.json
```diff
@@ -2,11 +2,17 @@
...
+  "readme": "To initialize the database run `initdb`.",
   "env": {
     "PGDATA": "{{ .DevboxRoot }}/conf/postgresql/data"
   },
   "create_files": {
     ".devbox/conf/postgresql/data": ""
+  },
+  "services": {
+    "postgresql": {
+      "start": "pg_ctl -l .devbox/conf/postgresql/logfile start",
```
gcurtis, line 14: Should this be `conf/postgresql/logfile start`?

pkgcfg/pkgcfg.go
```diff
@@ -4,16 +4,21 @@ import (
...
 
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/debug"
 	"go.jetpack.io/devbox/nix"
 )
 
-const confPath = ".devbox/conf"
+const (
+	ConfBinPath = ".devbox/conf/bin"
+	confPath    = ".devbox/conf"
```
gcurtis, line 20: Same question - should these still have `.devbox/`?

gcurtis (approved):
Is `./config` just for plugin JSON configs? Or does it also include the configs for programs (e.g., nginx.conf)?

pkgcfg/package-configuration/postgresql.json
mikeland86: The idea was to put non-shared files (logs, data, etc) in `.devbox` and put configs in `./config` but I do see how the naming is confusing. I'll rethink the naming.

pkgcfg/pkgcfg.go
mikeland86: Same comment here, this is correct but confusing because now we have 2 conf directories (one hidden and one not). Maybe we do:

* `./config` is for configs  (or maybe `./plugins`) ? 
* `.devbox/plugin-data/`  for hidden non-permanent non-shared data.

mikeland86:
@loreto @gcurtis this changed enough that a new review is needed. PTAL

loreto (approved):
LGTM

gcurtis (approved):

PR #361, opened 2022-12-13
[midcobra] split Telemetry middleware into sentry and segment middlewares

savil:
## Summary

see previous PR for motivational context.

This PR just does a split of the code, but is a functional no-op

## How was it tested?

1. compiles
2. sanity check: opened `devbox shell`

but didn't really verify that the sentry or segment are being logged properly.

PR #363, opened 2022-12-13
[sshshim] use midcobra to execute command with sentry middleware

savil:
## Summary

This PR uses `midcobra` in the sshshim to:
1. drive the execution of the main command.
2. ensure the sentry middleware is run

Design choices:

1. I chose to use the existing `midcobraExecutable` instead of implementing another
struct conforming to the `Executable` interface. The reason is that `midcobraExecutable.Execute()`
function nicely wraps the preRun and postRun functions, in a manner that the middlewares
current expect.

2. Chose to pass `os.Args` into `Executable.Execute` instead of the prior `os.Args[1:]`.
Reason: the sshshim Execute requires the `os.Args[0]` value and I chose to standardize
both sshshim and the regular devbox `boxcli.Execute` to have the same meaning of the
`arg` variable in the `Executable.Execute` function.

3. Added a `Command` interface to `midcobra` so that regular devbox can use cobra.Command
in the middlewares, while sshshim can use a fake sshshimCommand that conforms to this
interface. 

## How was it tested?

This compiles.

Need to test. Will add details here.

### DIFF
--- a/boxcli/midcobra/debug.go
+++ b/boxcli/midcobra/debug.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 
 	"github.com/fatih/color"
-	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.jetpack.io/devbox/boxcli/usererr"
 	"go.jetpack.io/devbox/debug"
@@ -32,7 +31,7 @@ func (d *DebugMiddleware) AttachToFlag(flags *pflag.FlagSet, flagName string) {
 	d.flag.Hidden = true
 }
 
-func (d *DebugMiddleware) preRun(cmd *cobra.Command, args []string) {
+func (d *DebugMiddleware) preRun(cmd Command, args []string) {
 	if d == nil {
 		return
 	}
@@ -48,7 +47,7 @@ func (d *DebugMiddleware) preRun(cmd *cobra.Command, args []string) {
 	}
 }
 
-func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr error) {
+func (d *DebugMiddleware) postRun(cmd Command, args []string, runErr error) {
 	if runErr == nil {
 		return
 	}
--- a/boxcli/midcobra/midcobra.go
+++ b/boxcli/midcobra/midcobra.go
@@ -9,20 +9,46 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
 )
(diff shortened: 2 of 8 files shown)


savil:
* main
    * **PR #349** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/349" target="_blank">Graphite</a> 
      * **PR #351** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/351" target="_blank">Graphite</a> 
        * **PR #352** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/352" target="_blank">Graphite</a> 
          * **PR #355** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/355" target="_blank">Graphite</a> 
            * **PR #358** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/358" target="_blank">Graphite</a> 
              * **PR #361** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/361" target="_blank">Graphite</a> 
                * **PR #363** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/363" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/363?utm_source=stack-comment).

gcurtis (approved):
Approving to keep things moving, but I also wanted to throw out an idea that might simplify some things.

Instead of adapting the middleware to work with non-Cobra commands, what if we move the middleware logic into separate functions that can be called directly? For example, there would be code to do the tracing:

```go
package telemetry

type Trace struct

func StartTrace() *Trace
func (t *Trace) End(err error)
```

And then `midcobra` and `sshshim` can use that to trace errors/analytics:

```go
package midcobra

type telemetryMiddleware struct

func (m *telemetryMiddleware) preRun(cmd *cobra.Command, args []string) {
	m.trace = telemetry.StartTrace()
	m.trace.Command = ...
	m.trace.CommandArgs = ...
}

func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr error) {
	m.trace.End(runErr)
}
```

```go
package sshshim

func Execute(ctx context.Context, args []string) int {
	trace := telemetry.StartTrace()
	m.trace.Command = "ssh"
	trace.End(execute(args))
}
```

savil:
abandoning in favor of #366

@gcurtis thanks for your suggestion. I agree. I'd been thinking that as well when I wrote this PR. Ended up with a simpler approach in #366.

PR #364, opened 2022-12-13
[CLI] Changed devbox generate to devbox generate debug

mohsenari:
## Summary
Title says it

## How was it tested?
- compile
- `devbox generate` shows command help
- `devbox generate debug` runs what previously was generate


docs/app/docs/cli_reference/devbox_generate.md
```diff
@@ -0,0 +1,29 @@
...
+
+- ### dockerfile
+
+Generate a Dockerfile that replicates devbox shell environment in a container
+
+## Options
+
+```text
+  -h, --help   help for add
+  -f, --force  overwrite on existing files
```
savil, line 23: 🤔 this is a product behavior question. I know I've not been super involved in the devcontainer work, so wondering if I'm missing some context/reasoning.

For `.devbox` folder and files generated within, we always generate those files and overwrite new ones. The reason is that our generation code may evolve with improvements.

Should we always overwrite the existing files here as well?

savil (approved):

docs/app/docs/cli_reference/devbox_generate.md
mohsenari: The reasoning here was that these are 3rd party tools we are generating files for: (vscode devcontainer and Docker). So users might have already setup their own custom files for these. In that case, automatic overwrite is a bad user experience.

boxcli/generate.go
```diff
@@ -20,33 +20,28 @@ func GenerateCmd() *cobra.Command {
...
 
 	command := &cobra.Command{
-		Use:    "generate",
-		Args:   cobra.MaximumNArgs(1),
-		Hidden: true, // For debugging only
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return runGenerateCmd(cmd, args, flags)
-		},
+		Use:  "generate",
+		Args: cobra.MaximumNArgs(1),
```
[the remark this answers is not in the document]
mohsenari: I think it should be zero as well. I must've forgotten to change it after copying over the command from other examples.

docs/app/docs/cli_reference/devbox_generate.md
savil: makes sense.

Could we make a minor language fix: I believe it should be `overwrite existing files`, dropping the `on`?

PR #366, opened 2022-12-13
[sshshim] add sentry without using middleware

savil:
## Summary

This PR is an alternative to both #361 and #363.

This PR's approach is lighterweight:
1. refactor sentry code into its new `telemetry` package.
2. directly call that in the sshshim command execution function.

The advantages it has are:
1. Major: The implementation in #363 is rather gross because we need to introduce an interface equivalent of cobra.Command. This is really ugly, and not very maintainable. If new middleware uses other functions of cobra commands, then we'll have to add those to the interface and also handle them in the sshshim-command version.
2. Minor: We keep the `midcobra.telemetry` middleware for the regular devbox command.

Disadvantages are:
1. We may in the future need to incorporate new middleware into the sshshim command in a non-middleware manner.

## How was it tested?

compiles.

Tested via:
1. Look up sentryDSN in sentry dashboard > Settings > Project (devbox) > ClientKeys (DSN)
2. Inserted that in `internal/build/build.go`
3. Manually inserted errors in devbox.Open and invokeSSHOrSCPCommand. Ran devbox commands that triggered them.
4. verified in sentry dashboard that the errors were logged.

### DIFF
--- a/internal/boxcli/midcobra/midcobra.go
+++ b/internal/boxcli/midcobra/midcobra.go
@@ -25,7 +25,7 @@ type Middleware interface {
 func New(cmd *cobra.Command) Executable {
 	return &midcobraExecutable{
 		cmd:         cmd,
-		executionID: executionID(),
+		executionID: ExecutionID(),
 		middlewares: []Middleware{},
 	}
 }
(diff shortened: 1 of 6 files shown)


savil:
* main
  * **PR #366** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/366" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/366?utm_source=stack-comment).

LucilleH (approved):
This works too. I don't feel strongly one way or another

PR #369, opened 2022-12-14
[plugins] Only create devbox.d files on add

mikeland86:
## Summary

This changes plugins so that `devbox.d` files are only created on `add` and not on other commands.

Following discussion with @loreto I was going to return error if any of the files are missing, but I'm afraid this will backfire. Specifically there are plugin files that are non-essential to plugin. Two immediate examples:

* nginx plugin creates a `web/index.html` example page.
* php plugin creates a php-fpm config  which is not needed if you don't want to use php-fpm.

Currently this PR simply takes no action if a given file is missing. I think that's probably Ok for now.

Some possible solutions:

* We could mark certain files as "optional" but this would complicate the plugin json.
* We could save some data in `devbox.d/devbox.lock` to keep track of plugins we've already generated files for so we don't go it again.
* We could move optional files to `.devbox`. This makes sense for web/index.html but for some files like php-fpm.conf it doesn't.

## How was it tested?

```sh
devbox add nginx
rm devbox.d/nginx/nginx.conf
devbox add nginx
# verified it got recreated
rm devbox.d/nginx/nginx.conf
devbox shell
# verified it was not recreated
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -50,8 +50,9 @@ func InitConfig(dir string) (created bool, err error) {
 type Devbox struct {
 	cfg *Config
 	// configDir is the directory where the config file (devbox.json) resides
-	configDir string
-	writer    io.Writer
+	configDir     string
+	pluginManager *pkgcfg.Manager
+	writer        io.Writer
 }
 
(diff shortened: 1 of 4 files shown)


loreto (approved):
LGTM

PR #370, opened 2022-12-14
[add] Show better message if package is missing. Specify exact version after installing

mikeland86:
## Summary

Minor UX improvements:

* Using devbox add without specifying a package shows a helpful message.
* After installing a package we show the exact version that was installed.

## How was it tested?

```sh
➜ devbox add
Usage: devbox add <pkg>... [flags]

```

```sh
➜  devbox add curl

curl (curl-7.86.0) is now installed.
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -530,13 +530,14 @@ func (d *Devbox) ensurePackagesAreInstalled(mode installMode) error {
 
 func (d *Devbox) printPackageUpdateMessage(mode installMode, pkgs []string) error {
 	installedVerb := "installed"
+	info, _ := nix.PkgInfo(d.cfg.Nixpkgs.Commit, pkgs[0])
 	if mode == uninstall {
 		installedVerb = "removed"
 	}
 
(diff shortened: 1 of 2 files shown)


internal/impl/devbox.go
```diff
@@ -530,13 +530,14 @@ func (d *Devbox) ensurePackagesAreInstalled(mode installMode) error {
...
 	if mode == uninstall {
 		installedVerb = "removed"
 	}
 
 	if len(pkgs) > 0 {
 
-		successMsg := fmt.Sprintf("%s is now %s.", pkgs[0], installedVerb)
+		successMsg := fmt.Sprintf("%s (%s) is now %s.", pkgs[0], info, installedVerb)
 		if len(pkgs) > 1 {
 			successMsg = fmt.Sprintf("%s are now %s.", strings.Join(pkgs, ", "), installedVerb)
```
loreto, line 542: Are you also printing the version when you install multiple packages? (we should do it in that case too)

loreto (approved):
LGTM

PR #372, opened 2022-12-14
Track panics in sentry

LucilleH:
## Summary
I noticed that a lot of the shell-related functionality have `panics` and we have no idea if they are happening or not until some reports come in. This is to inform us on the panics and proactively fix issues

## How was it tested?
devbox run build
./devbox shell on invalid base shell


savil (approved):
nice!


Could we move sentry.Init to before command-execution in the sshshim code too: https://github.com/jetpack-io/devbox/blob/main/internal/cloud/openssh/sshshim/command.go#L17

(would have to remove sentry.Init from the logSentry function in that file)

PR #373, opened 2022-12-14
Default shellrc to profile for shells that are not commonly seen

LucilleH:
## Summary
Default shellrc file path to `.profile` for shells that are not commonly seen. `devbox shell` does not execute `init_hook` and `scripts` for an unrecognized shell (For example Fish).

Seems like `.profile` is the most commonly used for various shell. Is this a good idea? If not, is there a way to solve this more gracefully?

## How was it tested?
devbox shell

### DIFF
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -90,6 +90,7 @@ func DetectShell(opts ...ShellOption) (*Shell, error) {
 		}
 	default:
 		sh.name = shUnknown
+		sh.userShellrcPath = rcfilePath(".profile")
 	}
 
 	for _, opt := range opts {


Lagoja:
I don't think fish will read the `.profile`, so this won't help the fish case : \

LucilleH:
> I don't think fish will read the `.profile`, so this won't help the fish case : \

Setting the shellrc to `.profile`, or anything, will trigger generating our temporary `.shellrc` file which includes the inithook + scripts. @Lagoja mind confirming that that is not the case for fish?

gcurtis:
Unfortunately there's no common way to set a "startup" script for all shells (which is why devbox has to do so much work here). `fish` especially has its own unique way of doing things. I don't think we've tested against it at all, so I wouldn't be surprised if there are other random things that break. We should probably document what our supported shells are.

Sourcing `~/.profile` is also only supposed to happen for login shells, which is why we don't do anything with it right now.

PR #231, opened 2022-10-13
Added search command with sorted results

mohsenari:
## Summary
Added `devbox search <package name>` command that searches nix store for packages that contain the `<package name>` in their name or description.

## How was it tested?
1. compile devbox
2. run `./devbox search openjdk` or any other package
3. see results

PR #374, opened 2022-12-14
[search] Add nix package search

mikeland86:
## Summary

Adds basic nix package search

## How was it tested?

```sh
➜  devbox search python
* python2 (version: 2.7.18.5)
* python39 (version: 3.9.15)
* python38 (version: 3.8.16)
* python37 (version: 3.7.16)
* python311 (version: 3.11.1)
* python (version: 3.2)
* python3Full (version: 3.10.8)
* python2Full (version: 2.7.18.5)
* python39Full (version: 3.9.15)
* python38Full (version: 3.8.16)

➜  devbox search go
* go (version: 1.19.3)
* gox (version: 1.0.1)
* got (version: 0.79)
* gom (version: 0.4)
* goa (version: 3.10.0)
* govc (version: 0.29.0)
* gosu (version: 1.14)
* gost (version: 2.11.1)
* goss (version: 0.3.18)
* gosh (version: 2020523-7ccb068
```


LucilleH:
I feel like in general, adding a top-level command merits a longer convo. But this is certainly useful. Can we repurpose this and display a list of suggestions when `devbox add <invalid_package_name` happens, or `devbox add <empty_package_name>`? Print out the search result, and followed with `See more on https://search.nixos.org/` is more helpful than having another top-level command

loreto (commented):
Looks great. But a few quick questions:
- How much does it increase the size of the binary? (Is there an embedded browser or it's just a driver against the OSs browser?)
- Does it require Chrome to be installed? (What happens if user only has Safari or Internet Explorer)
- Does it work out of the box in Linux and WSL?

mikeland86:
@LucilleH sure! happy to chat. I had assumed we all wanted this command, we just didn't have the time to implement it so I didn't think it would controversial. I think adding as suggestions to `add` is a good idea! I do think search vs suggestions are a bit different use cases. If I search for python I want to know what python is available. If I do `add python` I will end up with a specific version.

I assume that even if we don't want search as a top level command we would want it somewhere right? Do you have any suggestions?

-----------

@loreto ugh, you are 100% right. I read no dependencies and assumed that extended to chrome, but it does needed. 

This makes the binary 6mb larger (22->28)

It should work on linux, but I had issues getting it to work on ubuntu docker, mostly getting the headless chrome to work.

Alternatives:

* Only enable this if chrome is in path (we can use same logic as the driver)
* Turn this into a simple web service (we could use an approach closer to https://github.com/jetpack-io/devbox/pull/231 in that case and cache the packages to speed it up)

loreto:
The easy solution (what I would probably do as a first pass), is open a tab with the right query in the user's browser: using https://pkg.go.dev/github.com/pkg/browser#OpenURL

The better solution:
+ Move the logic you just wrote to the server side and expose it as an API
+ Have the CLI call that API

There's a third solution, although I'm unsure how well it would work: I was able to output the JSON with the information for each package that is indexed by their search engine. We could load that in memory and search just against package name instead of against all the fields. I think I like the server side solution more anyways.

LucilleH:
Not that I'm against search itself, but more like if the goal is to reduce `devbox add` user errors (as we've seen in the error reporting), adding search functionality just for that is a little overkill in my opinion (unless this works without much work and can be easily done). If the hypothesis is that users don't know where to look for packages, then simply printing out a link to the nix search page is sufficient imo.

PR #375, opened 2022-12-15
[UX] Improve post-install readme, improve message if package not found

mikeland86:
## Summary

UX improvements:

* If user adds a package without a readme, don't show any readme. (previously we would show: "To show this information, run `devbox info %s")
* If user tries to add non-existent package show better message.

## How was it tested?

```sh
devbox add curl
curl (curl-7.86.0) is now installed.
```

```sh
devbox add foobar

Error: foobar: package not found

```

### DIFF
--- a/internal/boxcli/add.go
+++ b/internal/boxcli/add.go
@@ -10,8 +10,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/nix"
 )
+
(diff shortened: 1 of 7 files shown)


savil (approved):
I'd call it `getConfigIfAny` so that the caller knows to check for a `nil` return value even if `err == nil`

mikeland86:
> I'd call it getConfigIfAny so that the caller knows to check for a nil return value even if err == nil

yeah, I don't love the interface right now either. I'm planing on refactoring this stuff to change the name to `plugin`

so in reality it would be something like:

```
p, err := plugin.ForPackage(pkg)
// handle err
p.PrintReadme(writer)
```

and `PrintReadme` can work correctly even if p is nil.

PR #377, opened 2022-12-16
openssh: improve command errors and debug logging

gcurtis:
Consistently return a more informative error message when an SSH command fails. The errors now contain the pid, the quoted command, and exit code.

When debug logging is enabled, also write the full command when the process starts and when it exits. If the command fails, indent and log its stdout and stderr buffers. If it succeeds, only log stderr to avoid logging private VM keys. Instead, the caller will redact, format, and log a successful JSON response.

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -5,11 +5,9 @@ package cloud
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -90,6 +88,11 @@ type vm struct {
 	VMPrivateKey string `json:"vm_private_key"`
 }
 
+func (vm vm) redact() *vm {
+	vm.VMPrivateKey = "***"
+	return &vm
+}
+
 func getVirtualMachine(username string) string {
 	client := openssh.Client{
 		Username: username,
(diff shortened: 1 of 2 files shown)


internal/cloud/cloud.go
```diff
@@ -101,26 +104,24 @@ func getVirtualMachine(username string) string {
...
 	}
-	if resp.VMPrivateKey != "" {
-		err := openssh.AddVMKey(resp.VMHost, resp.VMPrivateKey)
-		if err != nil {
-			log.Fatal(err)
-		}
+	if redacted, err := json.MarshalIndent(resp.redact(), "\t", "  "); err == nil {
+		debug.Log("got gateway response:\n\t%s", redacted)
+	}
+	if resp.VMPrivateKey == "" {
```
savil, line 118: what are the scenarios for when VMPrivateKey is empty versus filled?

internal/cloud/openssh/client.go
```diff
@@ -73,3 +82,41 @@ func destination(username, hostname string) string {
...
 	return result
 }
+
+func logCmdRun(cmd *exec.Cmd) error {
+	// Use cmd.Start so we can log the pid. Don't bother writing errors to
+	// the debug log since those will be printed anyway.
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("openssh: start command %q: %w", cmd, err)
+	}
+	debug.Log("openssh: started process %d with command %q", cmd.Process.Pid, cmd)
```
savil, line 92: no objection, more curiosity: why did you feel the need to log the pid? what is the pid used for?

savil (approved):

internal/cloud/cloud.go
gcurtis: It's filled the first time it's created. After that, it throws away its copy of the private key and expects the client to already have it.

internal/cloud/openssh/client.go
gcurtis: Just to help me track down which instance of SSH was causing problems when I had multiple connections running.

PR #381, opened 2022-12-16
[services] Start/stop multiple services DEV-1270

mikeland86:
## Summary

This allows starting/stoping multiple services:

`devbox services start` starts all services
`devbox services start s1 s2` starts s1 and s2

If a service fails to start/stop we print and error and continue. If only a single service is meant to be started/stoped we return an error.

## How was it tested?

```sh
devbox services start
devbox services stop
devbox services start apacheHttpd php-fpm
devbox services stop apacheHttpd php-fpm
```


mikeland86:
@loreto agreed, I really like https://github.com/F1bonacc1/process-compose but it will require some modifications to work the way we want.

PR #382, opened 2022-12-16
[plugins] Use relative paths or env variables in configs DEV-1272

mikeland86:
## Summary

@Lagoja called out that we were using absolute paths in conf files. Since the conf files are checked-in, they need to be env variables or relative paths if the file doesn't support env variables.

I also added `services restart` which really helped me test this out.

## How was it tested?

```
devbox add nginx apacheHttpd php
devbox services start
curl localhost
curl localhost:8080
curl localhost:8080/index.php
```

### DIFF
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -49,8 +49,17 @@ func ServicesCmd() *cobra.Command {
 		},
 	}
 
+	restartCommand := &cobra.Command{
+		Use:   "restart [service]...",
+		Short: "Restarts service. If no service is specified, restarts all services",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return restartServices(cmd, args, flags)
+		},
+	}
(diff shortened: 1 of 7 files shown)


loreto (approved):
LGTM to get it working properly.

That said, I'm worried that this is easy to mess up by a new plugin writer. Maybe we need to do something like paths are always from the `virtenv` root and we rewrite them automatically? (so even if you write `/foo` that becomes `{{ .Virtenv }}/foo`.

Or something like that; the main point being that it should be hard for plugin writers to get the paths wrong.

mikeland86:
@loreto I agree, this is fragile. I made a bunch of mistakes doing it as well. The challenge is when to do the rewrite, it can't happen on `shell` in `devbox.d` because that code is checked-in. 

I think this is good for now, but I'll think of ways to make it less error prone.

PR #385, opened 2022-12-16
[planner] New framework for planners

mohsenari:
## Summary
- Removed all planners except phpv2
- Added `suggestors` package with the following interface
  - `IsRelevant() bool`
  - `Pakcages() [] string`
- Implemented suggestors for non-empty planners
- Suggestors suggest packages at `devbox init`

## How was it tested?
compile
run devbox init with different language files (e.g., package.json, requirements.txt, etc.)


internal/pkgsuggest/pkgsuggest.go
```diff
@@ -0,0 +1,40 @@
...
+
+func GetSuggestors(srcDir string) ([]string, error) {
+	result := []string{}
+	for _, sg := range SUGGESTORS {
+		if sg.IsRelevant(srcDir) {
+			result = append(result, sg.Packages(srcDir)...)
+		}
+	}
+
+	return result, nil
```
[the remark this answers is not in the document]
mikeland86: We should also probably avoid returning packages that are already installed?

internal/pkgsuggest/suggestors/interface.go
```diff
@@ -0,0 +1,6 @@
+package suggestors
+
+type Suggestor interface {
+	IsRelevant(srcDir string) bool
+	Packages(srcDir string) []string
```
mikeland86, line 5: I know this was modeled after planners, but we should take advantage and improve if we can. Specifically, passing srcDir on every function call (even when not needed) makes code a bit messier. An alternative is to add `srcDir` as a field to suggesters. For example:

```golang
package nginx

type Suggestor struct{
  SrcDir string
}
```

mikeland86 (approved):
Nice work! 

My comments about package naming and struct naming are not blocking. We can refactor that later if we think it makes sense. Similar about srcDir, food for thought but not blocking.

The other comments (copyright, dedup, Fprintf, renaming `GetSuggestors`) I think you should do before merging. pre-approving to unblock!

internal/planner/planner.go
```diff
@@ -5,67 +5,12 @@ package planner
...
-	&haskell.Planner{},
-	&java.Planner{},
-	&javascript.Planner{},
-	&kotlin.Planner{},
-	&lua.Planner{},
-	&nginx.Planner{},
-	&nim.Planner{},
-	&objectivec.Planner{},
-	&ocaml.Planner{},
-	&perl.Planner{},
```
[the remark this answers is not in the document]
mohsenari: I understand that it makes the PR bigger, but the changes are quite straightforward. Do you suggest I make 2 stacked PRs, one for deleting planners and one for devbox init recommendations?

internal/pkgsuggest/suggestors/interface.go
```diff
@@ -0,0 +1,6 @@
+package suggestors
+
+type Suggestor interface {
```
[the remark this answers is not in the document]
mohsenari: Sounds good, I can change the name to `initrec` my only concern is, if we use this outside of devbox init in the future, `initrec` will make less sense. But we can worry about that when the time comes.

internal/pkgsuggest/suggestors/interface.go
mohsenari: I did this but had to pass srcDir to the structs for each suggestor/recommender at initialization. So it looks like this:
```go
func getRecommenders(srcDir string) []recommenders.Recommender {

	return []recommenders.Recommender{
		&dotnet.Recommender{SrcDir: srcDir},
		&golang.Recommender{SrcDir: srcDir},
		&haskell.Recommender{SrcDir: srcDir},
		&java.Recommender{SrcDir: srcDir},
		&javascript.Recommender{SrcDir: srcDir},
		&nginx.Recommender{SrcDir: srcDir},
		&python.RecommenderPip{SrcDir: srcDir},
		&python.RecommenderPoetry{SrcDir: srcDir},
		&ruby.Recommender{SrcDir: srcDir},
		&rust.Recommender{SrcDir: srcDir},
		&zig.Recommender{SrcDir: srcDir},
	}
}
```

internal/pkgsuggest/pkgsuggest.go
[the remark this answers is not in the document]
mohsenari: Used a map[string]bool for packages and at the end I return the keys as a slice of strings.
For already installed packages we need a bit of discussion on what is considered installed. So I added a TODO to update.

mohsenari:
@mikeland86 I made the updates based on your feedback, but before I merge I'll wait for you to take a 2nd look since it's a relatively big PR.

internal/planner/planner.go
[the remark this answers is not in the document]
mikeland86: The straightforward-ness of the changes is the reason to keep them separate. That way, removing the planners would take hours (to code and review) and not be blocked by the bigger change of a new framework.

My hope was to try to get the planner removal into the previous release so the comment was about splitting up the work to ship some important stuff faster.

internal/pkgsuggest/suggestors/interface.go
mikeland86: This feels a bit better but agree not ideal (better because it replaces 2 params with a single field).

BTW, when initializing structs, if you pass the correct types in the correct order, you don't have to specify the field, this allows for less repetition:

```golang
&zig.Recommender{srcDir},
```

Another potentially overengineered solution:


```golang
// base recommender
type Recommender struct {
  srcDir string
}

func (h *Recommender) SetSourceDir(d string) { d.hasSourceDir = d }

type SourceDirSetter interface {
	SetSourceDir(d string)
}
```

You can embed `Recommender` in each recommender struct. And then do:

```golang 
for rec, _ := range recs {
  if r, ok := rec.(SourceDirSetter); ok {
    r.SetSourceDir(srcDir)
  }
}
```

internal/impl/devbox.go
```diff
@@ -36,14 +37,27 @@ const (
...
 
 	config := &Config{
 		Nixpkgs: NixpkgsConfig{
 			Commit: plansdk.DefaultNixpkgsCommit,
 		},
 	}
+
+	pkgsToSuggest, err := initrec.Get(dir)
+	if err != nil {
+		return true, err
```
mikeland86, line 51: Should this be false?

mikeland86 (approved):

internal/pkgsuggest/suggestors/interface.go
mohsenari: I thought of putting a setter but it seemed like it will unnecessarily complicate the logic. I'll do the shorthand {srcDir} at initialization. Even though it is repetitive, it's easier to read and understand later on.

mohsenari: Go linter doesn't like the unkeyed fields in struct initialization. Guess I'll just keep it as is 😄

PR #387, opened 2022-12-17
[cloud shell] respect gitignore file in --config directory to limit large data syncs

savil:
## Summary

I discovered that the rust-stable project was not syncing, because it stores compiled
artifacts in a child-directory: `.rustup`. This directory is in the `.gitignore` file
but mutagen was syncing it.

In the long run, we should not sync any files that are gitignored (agree?).
However, code for that is a bit more involved, so as a first pass I implement
looking for a .gitignore file in the `configDir` and passing its rules to mutagen
as `--ignore` flags. Thankfully, mutagen's ignore rules match gitignore format.

## How was it tested?

was able to start `devbox cloud shell` in the `rust-stable` example project

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/fatih/color"
+	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/cloud/mutagen"
 	"go.jetpack.io/devbox/internal/cloud/mutagenbox"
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #387** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/387" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/387?utm_source=stack-comment).

LucilleH (approved):
Oh this is important - imagine syncing the entire node_modules

PR #389, opened 2022-12-19
[plugins] Add init hook to plugins. Add ruby gem bin to path DEV-1271

mikeland86:
## Summary

* Adds init hooks to plugins using same structure as devbox.json 
* Ensures ruby gem bin path is in PATH. Adding in an init hook ensures we respect user provided `$GEM_HOME` (currently it's not possible to override plugin env vars, but once we allow env vars in `devbox.json` they will take precedence. 

cc: @mohsenari @Lagoja 

## How was it tested?

```
devbox add ruby nginx
devbox shell
gem install rails
which rails
```


internal/nix/shell.go
```diff
@@ -102,7 +102,9 @@ func DetectShell(opts ...ShellOption) (*Shell, error) {
 	return sh, nil
 }
 
-func WithPlanInitHook(hook string) ShellOption {
+// If/once we end up making plugins the same as devbox.json we probably want
+// to merge all init hooks into single field
+func WithPluginInitHook(hook string) ShellOption {
 	return func(s *Shell) {
 		s.planInitHook = hook
```
mohsenari, line 109: Change this to `s.pluginInitHook` too?

loreto:
Are hooks already run using bash only? (if not, this plugin will break people using ruby in shells that don't have the same syntax)

internal/nix/shell.go
mikeland86: yep, good catch

mikeland86:
@loreto `export` is POSIX. Is that not our target support? How would you do this otherwise?

Lagoja:
> @loreto `export` is POSIX. Is that not our target support? How would you do this otherwise?

I think we need to be explicit that we only support POSIX/Shell as our scripting language for now. For shells like Fish, we could run `exec fish` after sourcing our environment and running the init hooks.

loreto:
Can we just change the way `shell` works so that:
1. There's a list of supported shells (the posix shells)
2. There's a default shell we support (either `zsh` or `bash`)

If you have a shell that is not in the supported list (aka `fish`) then we'll start `devbox shell` using the default shell.

If we can do that, then everything will also work for `fish` users (instead of trying to run a script that will *not* work in fish and giving an error)

plugins/ruby.json
```diff
@@ -6,5 +6,10 @@
     "RUBY_CONFDIR": "{{ .Virtenv }}",
     "GEMRC": "{{ .Virtenv }}/.gemrc",
     "GEM_HOME": "{{ .Virtenv }}"
+  },
+  "shell": {
+    "init_hook": [
+      "export PATH={{ .Virtenv }}/bin:$PATH"
```
gcurtis, line 12:
```suggestion
      "export PATH=\"{{ .Virtenv }}/bin:$PATH\""
```

Lagoja:
> 1. There's a list of supported shells (the posix shells)
> 2. There's a default shell we support (either `zsh` or `bash`)

I think Nix installs bash by default, so we should just use bash if the shell isn't on our allowlist

loreto (approved):
LGTM

Shell changes can be a different PR

PR #390, opened 2022-12-19
[cloud shell polish] query Github to get username

savil:
## Summary

This PR makes an `ssh -T git@github.com` query that is rejected with a friendly error
message of the form:
```
Hi savil! You've successfully authenticated, but GitHub does not provide shell access.
```
We can parse the username (`savil`, for me) from this response.

In a future PR, we can cache this value in `~/.cache/devbox/user.yaml` to reduce
the number of network calls

## How was it tested?

Not the most extensive testing. In my case, `savil` is used on both my local
laptop and github, so the usernames match. 

1. Github auth is set up
`DEVBOX_DEBUG=1 devbox cloud shell` and saw the printed output matches the code paths
for successfully parsing the github username.

2. Github auth is not set up
(see comment below)

Used Stepper UI:

When we cannot resolve the username, we'll ask as before:
Screenshot 2022-12-20 at 11 38 37 AM

and when we can resolve the username, it'll appear like this
Screenshot 2022-12-20 at 11 40 22 AM

### DIFF
--- a/internal/cloud/user.go
+++ b/internal/cloud/user.go
@@ -16,30 +16,38 @@ var githubSSHRegexp = regexp.MustCompile("Hi (.+)! You've successfully authentic
 // with a friendly rejection message that contains the user's username (if they have github
 // credentials set up correctly). We parse the username from the error message.
 func queryGithubUsername() (string, error) {
-	cmd := exec.Command("ssh", "-T", "git@github.com")
+	cmd := exec.Command("ssh", "-T", "-o", "NumberOfPasswordPrompts=0", "git@github.com")
 	var bufOut, bufErr bytes.Buffer
+	cmd.Stdin = nil
 	cmd.Stdout = &bufOut
 	cmd.Stderr = &bufErr
 	err := cmd.Run()
 
-	var errorMessage string
 	if err != nil {
+
 		if e := (&exec.ExitError{}); errors.As(err, &e) && e.ExitCode() == 1 {
-			debug.Log("Received expected (this is good) error for cmd `%s` had exit code 1 with stderr: %v", cmd,
-				bufErr.String())
-			errorMessage = bufErr.String()
+			// This is the Happy case, and we can parse out the error message
+			debug.Log(
+				"Received expected (this is good) error for cmd `%s` had exit code 1 with stderr: %v", cmd,
+				bufErr.String(),
+			)
+			return parseUsernameFromErrorMessage(bufErr.String()), nil
 		} else {
+			// This is the sad case, and we should let the caller figure out how to proceed with the user
 			debug.Log("error from command `%s`: %v, out: %v, stderr: %v", cmd, err, bufOut.String(), bufErr.String())
 			return "", errors.WithStack(err)
 		}
 	}
 
-	// parse output
+	return "", nil
+}
+
+func parseUsernameFromErrorMessage(errorMessage string) string {
 	matchedUsernames := githubSSHRegexp.FindSubmatch([]byte(errorMessage))
 	if len(matchedUsernames) < 2 {
 		debug.Log("Did not find a username from github. Message is: %s", errorMessage)
-		return "", nil
+		return ""
 	}
 	debug.Log("matched username from github is: %s\n", matchedUsernames[1])
-	return string(matchedUsernames[1]), nil
+	return string(matchedUsernames[1])
 }
--- a/internal/cloud/user.go
+++ b/internal/cloud/user.go
@@ -1,0 +1,45 @@
+package cloud
+
+import (
+	"bytes"
+	"os/exec"
+	"regexp"
+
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #390** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/390" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/390?utm_source=stack-comment).

loreto (commented):
Interesting idea! I like it, but I'm a little nervous about the edge cases of parsing the response from github. Any chance we can add some test around that function? For example, do we handle the case when you do *not* have a key the successfully authenticates with github?

savil:
> Interesting idea! I like it, but I'm a little nervous about the edge cases of parsing the response from github. Any chance we can add some test around that function? For example, do we handle the case when you do _not_ have a key the successfully authenticates with github?

credit to @gcurtis for suggesting the idea!

Yup, can add a test around this.
I'll exercise the no-key case and reply.

savil:
> For example, do we handle the case when you do not have a key the successfully authenticates with github?

This works as expected.

I cleared my ssh key so that I can no longer authenticate with Github.
```
testdata/go/go-1.19 on  savil/default-to-github-username [$] via 🐹 v1.19.2 on ☁️  (us-east-2) on ☁️  savil.srivastava@jetpack.io
❯ ssh-add -D
All identities removed.

testdata/go/go-1.19 on  savil/default-to-github-username [$] via 🐹 v1.19.2 on ☁️  (us-east-2) on ☁️  savil.srivastava@jetpack.io
❯ ssh-add -l
The agent has no identities.

testdata/go/go-1.19 on  savil/default-to-github-username [$] via 🐹 v1.19.2 on ☁️  (us-east-2) on ☁️  savil.srivastava@jetpack.io
❯ ssh -T git@github.com
no such identity: /Users/savil/.ssh/id_ed25519: No such file or directory
git@github.com: Permission denied (publickey).
```
and then to exercise the code:
```
testdata/go/go-1.19 on  savil/default-to-github-username [$] via 🐹 v1.19.2 on ☁️  (us-east-2) on ☁️  savil.srivastava@jetpack.io took 9s
❯ DEVBOX_DEBUG=1 devbox cloud shell
[DEBUG] 2022/12/19 16:30:39 /Users/savil/code/jetpack/devbox/internal/boxcli/midcobra/debug.go:47: Debug mode enabled.
[DEBUG] 2022/12/19 16:30:39 /Users/savil/code/jetpack/devbox/internal/impl/config.go:207: findConfigDir: path is
[DEBUG] 2022/12/19 16:30:39 /Users/savil/code/jetpack/devbox/internal/impl/config.go:249: finding devbox.json in dir: /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19
[DEBUG] 2022/12/19 16:30:39 /Users/savil/code/jetpack/devbox/internal/cloud/user.go:32: error from command `/usr/bin/ssh -T git@github.com`: exit status 255, out: , stderr: no such identity: /Users/savil/.ssh/id_ed25519: No such file or directory
git@github.com: Permission denied (publickey).
[DEBUG] 2022/12/19 16:30:39 /Users/savil/code/jetpack/devbox/internal/cloud/cloud.go:86: Failed to get username from Github. Falling back to $USER: savil
? What is your github username? (savil)
```
this is as one would expect

gcurtis (approved):

savil:
@loreto updated to use stepper UI. Looks much better IMO. Also - added a very basic test. Mocking out the command and ExitError is non-trivial and not worth it given that we are very conservative about falling back to the current user prompt.

savil:
I now skip prompting the user, if we do manage to parse the username from Github.

savil:
....and we fail-fast if the user's ssh key has a passphrase!

internal/cloud/user.go
```diff
@@ -0,0 +1,53 @@
...
+	"bytes"
+	"os/exec"
+	"regexp"
+
+	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/debug"
+)
+
+var githubSSHRegexp = regexp.MustCompile("Hi (.+)! You've successfully authenticated, " +
+	"but GitHub does not provide shell access")
```
LucilleH, line 13: Does this break if github changes its error message lol. Should we shorten this and just match with `Hi (.+)!`?

LucilleH (approved):
I like it

internal/cloud/user.go
savil: It'll fallback to having the user prompt, so stuff should still be workable, but less frictionless.. 

I guess I was inclined to leave the longer message, so that we intentionally update this if GitHub ever changes their error message (not likely to be a frequent occurence).

PR #393, opened 2022-12-20
Updated docs for 0.1.3

Lagoja:
## Summary
Updated docs, which include

- Updated Examples to use Plugins
- Updated CLI docs with new commands
- Plugins Guide
- Services Guide

## How was it tested?

localhost, plus dev instance here: https://devbox-web-john-lago-jetpack-io.cloud.jetpack.dev/devbox/docs/devbox_examples/


loreto (approved):
LGTM

PR #399, opened 2022-12-22
[cloud VM] prototype updating mutagen status in starship prompt

savil:
## Summary

This is a quick hack prototype to update the starship prompt during the initial project sync.

In this design, the client polls for mutagen sync session's status.
- For now, we do it only during the initial project sync. We can later extend this to continuously happen in the background.
- We write the new status in a file in the `VM`. The starship prompt displays the contents of this file.

One downside of this implementation is that the client's code determines the text of the mutagen-sync-prompt-status that starship shows. This is not a clean separation of concerns. The client should be sending an enum's values, and the VM should be able to map that to specific messages to display in the prompt. We can change to that in a future step.

companion PR for the starship config is: https://github.com/jetpack-io/axiom/pull/2924

## How was it tested?

Screenshot 2022-12-22 at 3 48 45 PM


savil:
* main
  * **PR #399** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/399" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/399?utm_source=stack-comment).

LucilleH (approved):

PR #413, opened 2023-01-03
Add Redis Plugin

Lagoja:
## Summary

Adds a plugin for Redis, including environment variables, default conf file, and a service

## How was it tested?

Built locally, then ran `devbox services start redis` to validate configuration

### DIFF
--- a/plugins/redis.json
+++ b/plugins/redis.json
@@ -1,0 +1,19 @@
+{
+    "name": "redis",
+    "version": "0.0.1",
+    "match": "^redis$",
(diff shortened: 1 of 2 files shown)


plugins/redis.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "redis",
+    "version": "0.0.1",
+    "match": "^redis$",
+    "readme": "Running `devbox services start redis` will start redis as a daemon in the background. \n\nYou can manually start Redis in the foreground by running `redis-server $REDIS_CONF --port $REDIS_PORT`. \n\nLogs, pidfile, and data dumps are stored in `.devbox/virtenv/redis`. You can change this by modifying the `dir` directive in `devbox.d/redis/redis.conf`",
```
mikeland86, line 5: FYI, `devbox info` (also printed after installation) automatically includes services information. Ideally we don't add it twice, but maybe we could make the services section in `info` more readable.

mikeland86 (approved):

plugins/redis.json
Lagoja: I mainly added this comment to emphasize that it's run with the --daemonize flag. 

In the future I think we could show the full start command for the service, since that can provide useful information

Issue #406, opened 2022-12-26
Feature request: quiet mode


aacebedo:
Devbox is great and can be used with vscode easily. However the lack of a quiet mode triggers problems when you configure it with vscode (for instance for the formatter). In this case the log printed by devbox when the shell command is called appears in the formatted file.

It would be great if devbox featured a quiet mode in which it will not log anything else the output of the command it executed.
The quiet mode would be a simple `-q` flag to append to any devbox command and will remove the devbox logs.

mohsenari:
#414

mohsenari:
@aacebedo the pull request regarding this issue is merged. You'll have the ability to put `--quiet` or `-q` to devbox shell, devbox run, and devbox init commands in the next release which will likely be next week.

PR #414, opened 2023-01-03
[CLI] Added quiet mode to devbox run, init, and shell

mohsenari:
## Summary
- Wrote a custom IOWriter
- Added `--quiet`/`-q` flag to run, init, and shell commands
- Devbox init doesn't produce any output right now, but it will start suggesting packages after #385 is merged
- Addresses #406 
## How was it tested?
1. Compile
2. `devbox init -q`
3. Add scripts to devbox.json
4. `devbox run <script> -q`
5. `devbox shell -q`
6. Assert in all cases devbox logs aren't printed

### DIFF
--- a/internal/boxcli/init.go
+++ b/internal/boxcli/init.go
@@ -10,31 +10,24 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/writer"
 )
 
-type initCmdFlags struct {
-	quiet bool
-}
-
 func InitCmd() *cobra.Command {
-	flags := initCmdFlags{}
 	command := &cobra.Command{
 		Use:   "init [<dir>]",
 		Short: "Initialize a directory as a devbox project",
 		Long:  "Initialize a directory as a devbox project. This will create an empty devbox.json in the current directory. You can then add packages using `devbox add`",
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInitCmd(cmd, args, flags)
+			return runInitCmd(cmd, args)
 		},
 	}
-	command.Flags().BoolVarP(
-		&flags.quiet, "quiet", "q", false, "Quiet mode: Suppresses logs.")
 
 	return command
 }
 
-func runInitCmd(cmd *cobra.Command, args []string, flags initCmdFlags) error {
+func runInitCmd(cmd *cobra.Command, args []string) error {
 	path := pathArg(args)
 
-	w := &writer.DevboxIOWriter{W: cmd.OutOrStderr(), Quiet: flags.quiet}
+	w := writer.New(cmd)
 	_, err := devbox.InitConfig(path, w)
 	if err != nil {
 		return errors.WithStack(err)
--- a/internal/boxcli/init.go
+++ b/internal/boxcli/init.go
@@ -7,26 +7,35 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
+	"go.jetpack.io/devbox/internal/boxcli/writer"
 )
 
+type initCmdFlags struct {
+	quiet bool
+}
+
 func InitCmd() *cobra.Command {
+	flags := initCmdFlags{}
 	command := &cobra.Command{
 		Use:   "init [<dir>]",
 		Short: "Initialize a directory as a devbox project",
 		Long:  "Initialize a directory as a devbox project. This will create an empty devbox.json in the current directory. You can then add packages using `devbox add`",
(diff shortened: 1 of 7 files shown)


savil:
Need to review the code but from reading the Summary:

> Added --quiet/-q flag to run, init, and shell commands

I'd be in favor of adding this as a global flag that applies to all commands. It'll also simplify the way we write the code, so that we uniformly use the `io.Writer` instead of needing to consider which top-level command some function may be invoked from.

internal/boxcli/writer/writer.go
```diff
@@ -0,0 +1,23 @@
...
+// Use of this source code is governed by the license in the LICENSE file.
+package writer
+
+import "io"
+
+type DevboxIOWriter struct {
+	W     io.Writer
+	Quiet bool
+}
+
```
savil, line 11: I'd suggest introducing an API like:
```
func New(cmd *cobra.Command, quiet bool) *DevboxIOWriter {
   ....
}
```
That way we control how all users construct this, and can make `W` and `Quiet` be private fields

internal/boxcli/writer/writer.go
```diff
@@ -0,0 +1,23 @@
...
+func (d DevboxIOWriter) Write(p []byte) (int, error) {
+	if !d.Quiet {
+		n, err := d.W.Write(p)
+		if err != nil {
+			return n, err
+		}
+		if n != len(p) {
+			return n, io.ErrShortWrite
+		}
+	}
```
savil, line 21: what was the resolution of the discussion for having `--quiet` silence the output versus put it in Stderr?

internal/boxcli/shell.go
```diff
@@ -47,9 +50,9 @@ func runShellCmd(cmd *cobra.Command, args []string, flags shellCmdFlags) error {
 	if err != nil {
 		return err
 	}
-
+	w := &writer.DevboxIOWriter{W: cmd.OutOrStderr(), Quiet: flags.quiet}
```
savil, line 53: should this be strictly `stdout` instead of either `Out or Stderr`?

I think it depends on what we decide the behavior we want:
- `quiet` suppresses output
- `quiet` redirects output to stderr

internal/boxcli/init.go
```diff
@@ -7,26 +7,35 @@ import (
...
 		Short: "Initialize a directory as a devbox project",
 		Long:  "Initialize a directory as a devbox project. This will create an empty devbox.json in the current directory. You can then add packages using `devbox add`",
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInitCmd(cmd, args)
+			return runInitCmd(cmd, args, flags)
 		},
 	}
+	command.Flags().BoolVarP(
+		&flags.quiet, "quiet", "q", false, "Quiet mode: Suppresses logs.")
```
savil, line 29: if we decide to not make a global flag, then instead of copy-pasta this flag declaration could we move it to a function that all the commands call?

savil (commented):
Looks good!

Two questions to iterate on:
1. `quiet` behavior for silence versus stderr
2. having `--quiet` be a global flag

internal/boxcli/writer/writer.go
mohsenari: I think we can do both, I can pass in stderr as an io writer, so that devbox outputs are written on Stderr only. Also have -q silence the output.

internal/boxcli/shell.go
mohsenari: I'll change this to pass Stderr only

internal/boxcli/init.go
mohsenari: not all commands have to have `-q`. I prefer going with your 2nd suggestion.

internal/boxcli/writer/writer.go
mohsenari: good suggestions, makes sense, I'll make that change

internal/boxcli/init.go
mohsenari: I ended up adding `-q` to the root command. I think it makes the most sense there. This way we can implement the quiet behaviour for future commands with minimal change.

savil (approved):
nice!

mikeland86:
@mohsenari @savil I'm not sure we need to add a custom writer. You can add a `PersistentPreRun` to root command:

```golang
PersistentPreRun: func(cmd *cobra.Command, args []string) {
  if flags.quiet {
    cmd.SetOut(io.Discard)
  }
},
```

and then just make sure to use `cmd.OutOrStdout()` everywhere. That should be all.

PR #419, opened 2023-01-04
Fix typo

hezhizhen:
Signed-off-by: Zhizhen He <hezhizhen.yi@gmail.com>


LucilleH (approved):
Thank you!! 

Not sure why `go.mod` and `go.sum` would change. Do you mind unstaging those files?

hezhizhen:
> Thank you!!
> 
> Not sure why `go.mod` and `go.sum` would change. Do you mind unstaging those files?

Done. But every time I run `go mod tidy`, `go.mod` and `go.sum` will always change.

LucilleH:
> > Thank you!!
> > Not sure why `go.mod` and `go.sum` would change. Do you mind unstaging those files?
> 
> Done. But every time I run `go mod tidy`, `go.mod` and `go.sum` will always change.

I'll take a look 👀

PR #422, opened 2023-01-05
[cloud | polish] update intro text, and display VM region

savil:
## Summary

Two changes:
1. Update the introductory text
2. Update the stepper-success message to include the region the VM was created in.

For the second change, there are two scenarios:
1. New VM -> `Created a virtual machine in <location>`
2. Existing VM -> `Detected an existing virtual machine`. We don't include location information because we don't have it. This scenario occurs either when opening another cloud-shell in another terminal tab OR when starting a new cloud-shell within 300 seconds (`ControlPersist` time in sshconfig).

I copied the region names table from https://fly.io/docs/reference/regions/. I did not editorialize the names, but of course we can: for example, someone connecting to a region near them called `Sunnyvale, California` likely doesn't need the `(US)` part since they presumably know the California they are near.

## How was it tested?

opened cloud shells

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -17,6 +17,7 @@ import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/cloud/fly"
 	"go.jetpack.io/devbox/internal/cloud/mutagen"
 	"go.jetpack.io/devbox/internal/cloud/mutagenbox"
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #422** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/422" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/422?utm_source=stack-comment).

LucilleH (approved):

PR #424, opened 2023-01-05
Extract shell-specific init file injection into own func

ipince:
## Summary

While this seems like a no-op, please note that I removed the quotes around the formatted string in each of the cases. I was finding that the rc files weren't being executed when using the quotes (and while using nixless shell).

## How was it tested?
Add a var into `shellrc.tmpl`, build devbox, run `./devbox shell`, and check if the var is defined. Do this for bash, zsh, ksh.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 require golang.org/x/net v0.2.0 // indirect
 
 require (
+	github.com/alessio/shellescape v1.4.1 // indirect
 	github.com/andybalholm/brotli v1.0.4 // indirect
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
 	github.com/cockroachdb/apd/v2 v2.0.2 // indirect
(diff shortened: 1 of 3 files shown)


gcurtis (approved):
Does putting the quotes around just the value work? For example, `ENV="foo"` instead of `"ENV=foo"`?

I think this is okay as long as we can guarantee that the env var values and args don't contain spaces. Otherwise something like `exec env ENV=some file /bin/sh` would break.

ipince:
> I think this is okay as long as we can guarantee that the env var values and args don't contain spaces. Otherwise something like exec env ENV=some file /bin/sh would break.

Hmm.. well, the filepath is generated by us, and it is set to `os.MkdirTemp()/basename`, where basename is one of: .bashrc, .zshrc, .kshrc, .shinit, or... the value of `ENV` (for ash, dash, sh--what are those anyway?).

Let me see what happens when `ENV` points to a file with spaces.

PR #425, opened 2023-01-05
[CICD] add devbox install nix action in cicd, change release workflow

LucilleH:
## Summary
1. Add devbox install nix action as a test in cicd
2. Only update latest version in S3 on publishing a release

## How was it tested?
CICD

### DIFF
--- a/.github/workflows/cli-publish.yml
+++ b/.github/workflows/cli-publish.yml
@@ -1,0 +1,27 @@
+name: cli-publish
+# Publish the Devbox CLI
+
+on:
+  release:
+    types: [released]
+
+permissions:
+  contents: write
+  pull-requests: read
+  id-token: write # Needed for aws-actions/configure-aws-credentials@v1
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: release
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/cli-tests.yaml
```diff
@@ -77,3 +77,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: ./go.mod
+          cache: true
+      - name: Build devbox
+        run: go install ./cmd/devbox
+      - name: Install nix and devbox packages
```
loreto, line 91: Add comment that main reason we have this test is to ensure Devbox installs nix and works properly after installation.

.github/workflows/cli-publish.yml
```diff
@@ -0,0 +1,27 @@
+name: cli-publish
```
loreto, line 1: I'm tempted to call this `cli-post-release` (and the file)

.github/workflows/cli-publish.yml
```diff
@@ -0,0 +1,27 @@
...
+# Publish the Devbox CLI
```
loreto, line 2: `# Finalize and announce the release once its been published on Github.`

loreto (approved):
LGTM but see comments

PR #429, opened 2023-01-06
[auth] save Github username to local file

savil:
## Summary

During the `devbox cloud shell` command execution, we now save the user's 
Github username in `~/.config/devbox/ssh/github_username`. This should
improve the UX since they won't need to keep re-entering this information.

We wait for the `gateway auth` to successfully return a VM before saving the info
in the `github_username` file. Because the gateway verifies that the user's github username
is valid.

In case the user wishes to override the cached github username, they can use the new `--username` or `-u` flag in `devbox cloud shell`.

## How was it tested?

```
> devbox cloud shell
....

> cat ~/.config/devbox/ssh/github_username
savil
```

I then edited `github_username` to have an incorrect value like `savil2`.
And cleared the socker in `~/.config/devbox/ssh/sockets` to ensure the CLI client calls the gateway again.
This time `devbox cloud shell` gave an error about using a bad username `savil2`.

I then did: `devbox cloud shell -u savil` to fix the situation.


savil:
* main
  * **PR #429** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/429" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/429?utm_source=stack-comment).

internal/cloud/cloud.go
```diff
@@ -101,6 +97,28 @@ func Shell(configDir string) error {
 	return shell(username, vmHostname, configDir)
 }
 
+func enactGithubUsernameStep() string {
+	stepGithubUsername := stepper.Start("Detecting your Github username...")
```
gcurtis, line 101: What do you think of making the username detection silent? When it fails to detect my username it looks like something went wrong.

internal/cloud/cloud.go
```diff
@@ -101,6 +97,28 @@ func Shell(configDir string) error {
...
+		}
+
+		username, err = queryGithubUsername()
+		if err == nil && username != "" {
+			stepGithubUsername.Success("Username: %s", username)
+		} else {
+			stepGithubUsername.Fail("Unable to resolve username")
+			// The query for Github username is best effort, and if it fails to resolve
+			// we fallback to prompting the user, and suggesting the local computer username.
+			username = promptUsername()
```
gcurtis, line 116: This might be something for a different PR, but it'd be nice to also try the current username. The flow would be something like:

1. Query GitHub for username. If that fails, default to the current username.
2. Try to authenticate. If that fails, prompt for the username and try again.
3. Upon success save the GitHub username.

gcurtis (commented):
Saving the username looks good, but I'm not certain about adding the auth commands. It makes it sound like Devbox is logging into their GitHub account, but really it's just asking them to prove that they have a GitHub account.

IMO having a flag like `devbox cloud -u gcurtis shell` to set the username would make more sense for now. I would wait until we know what devbox cloud accounts will look like before adding an auth command.

internal/cloud/cloud.go
savil: We could try that...

savil:
> Saving the username looks good, but I'm not certain about adding the auth commands. It makes it sound like Devbox is logging into their GitHub account, but really it's just asking them to prove that they have a GitHub account.

We're using their Github username as an identifier. But I don't feel strongly. We can drop the auth commands for now.

The main issues are:
1. Communicate which Github username we are using, whether auto-resolved or explicitly entered.
2. Allow the user to either clear that state or override it.

Proposal for issue 1:
end with `[:check-mark:] Username: savil (override with --username flag)` whether auto-resolved or explicitly entered. That should resolve your concern that it looks like there is an error (which I agree with in the current implementation).

Proposal for issue 2: 
User can use the `--username` flag to override the previously saved one.

WDYT?

gcurtis:
Adding a `--username` flag sounds good to me.

Why do we need to communicate what GitHub username we used? I'm hoping that we can detect/guess the username most of the time and have everything magically work. I'd rather only make the user think about GitHub usernames if there's an auth failure.

savil:
hm -- kk so if we are downplaying the github username and not wanting to treat it as an auth identifier, then we can drop the github-username-stepper entirely. cc @loreto

loreto:
I'm fine with either.

We can think about the `auth` flows once we need "real" authentication

savil:
@gcurtis ping

gcurtis (approved):

PR #431, opened 2023-01-06
[CLI] Quiet mode for all commands

mohsenari:
## Summary
- Did a sweep for fmt.print and changed all instances of to to fmt.Fprint except in debug.
- Removed custom IO writer and used cobra command's cmd.SetErr(io.Discard) at root level.
- Changed all os.stdout writers passed to commands to cmd.ErrOrStderr() due to discussion around writing outputs to stderr rather than stdout.
- Changed `PersistentPreRunE` to 'PreRunE` in subcommands because it was causing the subcommands ignore the root command's `PersistentPreRunE` which has the quiet flag.

## How was it tested?
- Compile
- run devbox commands with -q and without -q
- confirm -q silences outputs


internal/boxcli/add.go
```diff
@@ -49,8 +48,8 @@ func AddCmd() *cobra.Command {
 	return command
 }
 
-func addCmdFunc(_ *cobra.Command, args []string, flags addCmdFlags) error {
-	box, err := devbox.Open(flags.config.path, os.Stdout)
+func addCmdFunc(cmd *cobra.Command, args []string, flags addCmdFlags) error {
+	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
```
mikeland86, line 52: These should be `OutOrStdout` no?

internal/boxcli/root.go
```diff
@@ -26,6 +27,11 @@ func RootCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "devbox",
 		Short: "Instant, easy, predictable development environments",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			if flags.quiet {
+				cmd.SetErr(io.Discard)
```
mikeland86, line 32: Should be `SetOut` ?

internal/boxcli/add.go
```diff
@@ -24,9 +23,9 @@ func AddCmd() *cobra.Command {
 	flags := addCmdFlags{}
 
 	command := &cobra.Command{
-		Use:               "add <pkg>...",
-		Short:             "Add a new package to your devbox",
-		PersistentPreRunE: nix.EnsureInstalled,
+		Use:     "add <pkg>...",
+		Short:   "Add a new package to your devbox",
+		PreRunE: nix.EnsureInstalled,
```
mikeland86, line 28: Just curious, why this change? can we not have 2 `PersistentPreRunE` ? (one in root and another in the subcommands?)

mikeland86: A found it https://github.com/spf13/cobra/issues/252

looks like others were confused by this as well. Good find!

mikeland86 (changes requested):
Change looks good, but unless I'm misunderstanding something everything should be stdout instead of stderr right?

Side note, we are currently sending errors to stdout in `DebugMiddleware`. That's another separate fix but I assume those should go to stderr instead and IMO those should not listen to quiet flag?

mikeland86 (approved):
Found Slack conversation where stderr vs out was discussed. IMakes sense

Also found this https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

internal/boxcli/add.go
mohsenari: This was causing the subcommands ignore the persistentPreRun at root level which contains `-q` flag. This was the bug I couldn't solve yesterday but eventually did 😄

PR #434, opened 2023-01-06
[plugins] Make postgresql use local socket

mikeland86:
## Summary

This changes postgresql plugin to use a virtenv based socket.

## Backwards compatibility

This change will cause psql to not work in the following edge case:

* User updates devbox to version that includes this plugin
* User had previously run `devbox services start postgresql` with old version and never turned the service off.
* User starts new shell with new devbox and tries to do `psql`

I think this edge case is small enough that we can merge this. Potential ideas for mitigation:

* Bump the version and add instructions that show up when there's a version missmatch.

## How was it tested?

```bash
devbox shell
initdb
devbox services start postgresql
psql -d template1
```


Lagoja:
Just to understand the edge case -- it would get resolved if the user restarts their service? What is the actual mitigation?

mikeland86:
yeah, restarting will fix it. And edge case only affects the case where a new shell on a new devbox is started and an old service is running so pretty minor.

Mitigation is for a later PR. Basically we can add comments to versions that tell users how to update. e.g.
```json
"version_notes": {
  "0.0.2": " This change defaults the socket dir to be inside the local virtenv directory. If you get a socket error please restart the service"
}
```

loreto (approved):
LGTM

PR #437, opened 2023-01-06
[Generate] Added devbox generate direnv command

mohsenari:
## Summary
title says it:
this is step 1 of 4 to make integration with direnv more convenient
## How was it tested?
- compile
- run `devbox generate direnv`
- confirm `.envrc` file is generated and works with direnv

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -21,6 +21,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/cuecfg"
 	"go.jetpack.io/devbox/internal/debug"
+	"go.jetpack.io/devbox/internal/fileutil"
 	"go.jetpack.io/devbox/internal/initrec"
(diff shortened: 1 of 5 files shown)


internal/impl/devbox.go
```diff
@@ -455,6 +455,26 @@ func (d *Devbox) GenerateDockerfile(force bool) error {
 	return nil
 }
 
+// generates a .envrc file that makes direnv integration convenient
+func (d *Devbox) GenerateEnvrc(force bool) error {
+	envrcfilePath := filepath.Join(d.configDir, ".envrc")
+	filesExist := plansdk.FileExists(envrcfilePath)
```
savil, line 461: `fileutil.Exists`?

savil (approved):

internal/impl/devbox.go
mohsenari: I didn't know about `fileutil.Exists` we use plansdk.FileExists in so many different places that I had to keep it in planner sdk code when refactoring. 
Will have to make a refactor on all instances later on

PR #438, opened 2023-01-06
Auto-generate .envrc at devbox init

mohsenari:
## Summary
Title says it
This is step 2 out of 4 for simplifying direnv integration with devbox

## How was it tested?
- compile
- run `devbox init`
- confirm a `.envrc` file is created

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -46,13 +46,6 @@ func InitConfig(dir string, writer io.Writer) (created bool, err error) {
 		},
 	}
 
-	// Should this be best-effort and continue forward if envrc file creation
-	// fails?
-	err = GenerateEnvrc(dir)
-	if err != nil {
-		return false, err
-	}
-
 	pkgsToSuggest, err := initrec.Get(dir)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


internal/impl/devbox.go
```diff
@@ -58,6 +58,16 @@ func InitConfig(dir string, writer io.Writer) (created bool, err error) {
 			color.HiYellowString(s),
 		)
 	}
+	// .envrc file creation
+	envrcfilePath := filepath.Join(dir, ".envrc")
+	filesExist := plansdk.FileExists(envrcfilePath)
+	if !filesExist { // don't overwrite an existing .envrc
+		err := generate.CreateEnvrc(tmplFS, dir)
+		if err != nil {
+			return false, errors.WithStack(err)
```
mohsenari, line 67: Should we ignore this? if `.envrc` creation fails, should we just move forward with devbox init flow and not interrupt?

savil: yeah, you can print a warning via `debug.Log` but probably not worth concerning a user about.

savil (approved):

PR #439, opened 2023-01-07
Added prompt for direnv allow after .envrc

mohsenari:
## Summary
Title says it
This is step 3 out of 4 for simplifying direnv integration
## How was it tested?
- compile
- run devbox init
- confirm it prompts for direnv integration if direnv is installed


internal/impl/devbox.go
```diff
@@ -59,12 +59,25 @@ func InitConfig(dir string, writer io.Writer) (created bool, err error) {
...
-	filesExist := plansdk.FileExists(envrcfilePath)
-	if !filesExist { // don't overwrite an existing .envrc
-		err := generate.CreateEnvrc(tmplFS, dir)
-		if err != nil {
-			return false, errors.WithStack(err)
+	if commandExists("direnv") {
+		var result string
+		// prompt for direnv allow
+		fmt.Println("Do you want to enable direnv integration for this devbox project?[y/N]")
+		fmt.Scanln(&result)
```
savil, line 66: what does this UX look like?

For UX consistency consider using the `survey` library, like we do elsewhere: https://github.com/jetpack-io/devbox/blob/d3494d605444711bdc5a9588c7ad7b9030cd9cea/internal/cloud/cloud.go#L109

savil (approved):

internal/impl/devbox.go
mohsenari: I think your suggestion makes perfect sense. Initially, I wanted to minimize the size and not add another dependency but the UX is much better with `survey` package

PR #440, opened 2023-01-07
[cloud] emulate user's local directory structure in VM

savil:
## Summary

**Problem**
A scenario we want to protect against is two distinct projects sharing the same directory name. For example, consider projects located at `~/customer1/my_project` and `~/customer2/my_project`. Prior to this PR, we would attempt to sync both projects to the same folder in the VM: `~/code/my_project`.

**Approach**
This PR attempts to solve this conflict by seeking to emulate the user's directory structure in the VM. So, for the aforementioned example projects, they will sync to `~/customer1/my_project` and `~/customer2/my_project` in the VM as well. This has the additional benefit that if a user has relative paths between two projects they are actively developing (for example, a golang library and golang application that uses that library), then these relative paths will continue working in the cloud-shell, in accordance with our desires to make the cloud-shell experience akin to local shells.

**Edge case: projects outside homedir**
An edge case is for projects that are located locally outside the user's homedir. For security reasons, in the VM, we do not want to give the user root access, or access to folders outside their homedir. We choose to sync to a special folder called: `~/outside-homedir-code/<absolute path>`. For example, a project located locally at `/my_company/my_project` will sync to `~/outside-homedir-code/my_company/my_project`. There are two downsides to this which are tolerable IMO:
1. If a user _also has_ a local folder called `~/devbox-cloud` and they want to sync those projects then there will be a conflict. 
2. Relative paths between a project in homedir and outside the homedir will not work. The user can inspect the paths in the cloud-shells via `pwd` and update the relative paths to match that.

## How was it tested?

for `testdata/go/go-1.19`: could sync to VM via `devbox cloud shell`, and inspect directory in VM with `pwd`.

for edge case: projects outside homedir:
1. copied `examples/testdata/rust/rust-stable` to `/usr/local/testy-nonhome-dir/rust-stable`
2. fixed permissions: `sudo chown savil:staff /usr/local/testy-nonhome-dir/rust-stable`
3. `devbox cloud shell` to VM
4. `pwd` in VM gave: `/home/savil/outside-homedir-code/usr/local/testy-nonhome-dir/rust-stable`
5. `devbox add gcc` and then `cargo run` ran successfully.

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -167,9 +167,10 @@ func syncFiles(username, hostname, projectDir string) error {
 	if err != nil {
 		return err
 	}
-	debug.Log("Will sync files to directory: ~/%s", projectPath)
+	pathInVM := projectPathInVM(username, projectPath)
+	debug.Log("pathInVM: %s", pathInVM)
 
-	err = copyConfigFileToVM(hostname, username, projectDir, projectPath)
+	err = copyConfigFileToVM(hostname, username, projectDir, pathInVM)
 	if err != nil {
 		return err
 	}
@@ -189,6 +190,7 @@ func syncFiles(username, hostname, projectDir string) error {
 	machineID, _, _ := strings.Cut(hostname, ".")
 	mutagenSessionName := mutagen.SanitizeSessionName(fmt.Sprintf("devbox-%s-%s", machineID,
 		hyphenatePath(projectPath)))
+
 	_, err = mutagen.Sync(&mutagen.SessionSpec{
 		// If multiple projects can sync to the same machine, we need the name to also include
 		// the project's id.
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #440** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/440" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/440?utm_source=stack-comment).

internal/cloud/cloud.go
```diff
@@ -200,7 +206,7 @@ func syncFiles(username, hostname, configDir string) error {
 	time.Sleep(1 * time.Second)
 
 	// In a background routine, update the sync status in the cloud VM
-	go updateSyncStatus(mutagenSessionName, username, hostname, projectName)
+	go updateSyncStatus(mutagenSessionName, username, hostname, filepath.Base(projectPath))
```
LucilleH, line 209: Should this be `hyphenatePath(projectPath)`? Would this status also have conflict?

LucilleH (approved):

internal/cloud/cloud.go
savil: oh yeah! good call. Updated.

savil:
1. Made an update that renames functions to `relativeProjectPathInVM` and `absoluteProjectPathInVM`. This feels more clear to me. I need to re-test after this change.
2. Added Lucille's suggestion to hyphenate the project path for the starship prompt updater.
3. This PR is dependent on the VM update in https://github.com/jetpack-io/axiom/pull/2934 (closed source, sorry). I need to make that change backwards compatible, and then deploy it prior to landing this code.

PR #441, opened 2023-01-08
Return exit code of the exec call

devholic:
## Summary

Currently, there is no way to figuring out if the shell / run command succeeded or not in the CI-like environment.

This commit fixes this issue by returning exit code of the exec call if errored.

## How was it tested?

- Running `shell` with the failing command

```sh
❯ ./devbox shell -- exit 1
Error: exit status 1

❯ echo $?
1
```
 
- Running `script` with the failing command

```json
{
  "shell": {
    "scripts": {
      "failing_script": "exit 1"
    }
  }
}
```

```sh
❯ ./devbox run failing_script
Starting a devbox shell...
Error: exit status 1

❯ echo $?
1
```

### DIFF
--- a/internal/boxcli/midcobra/midcobra.go
+++ b/internal/boxcli/midcobra/midcobra.go
@@ -6,6 +6,8 @@ package midcobra
 import (
 	"context"
 	"encoding/hex"
+	"errors"
+	"os/exec"
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
@@ -69,6 +71,11 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	}
 
 	if err != nil {
+		// If the error is from the exec call, return the exit code of the exec call.
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			return exitErr.ExitCode()
+		}
 		return 1 // Error exit code
 	} else {
 		return 0
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -4,7 +4,6 @@
 package boxcli
 
 import (
-	"os/exec"
 	"sort"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


LucilleH (approved):
Thank you! This is exactly what I would do! Just to confirm @loreto is this what you had in mind as well?

There might be some edge cases around different behaviors of `devbox shell -- <cmd>` vs `devbox run <script>`, which I'll do some testing. I might follow up with a PR around sentry to adapt to this change as well

go.sum
```diff
@@ -88,6 +85,7 @@ github.com/pelletier/go-toml/v2 v2.0.5 h1:ipoSadvV8oGUjnUbMub59IDPPwfxF694nG/jwb
 github.com/pelletier/go-toml/v2 v2.0.5/go.mod h1:OMHamSCAODeSsVrwwvcJOaoN0LIUIaFVNZzmWyNfXas=
 github.com/pierrec/lz4/v4 v4.1.14 h1:+fL8AQEZtz/ijeNnpduH0bROTu0O3NZAlPjQxGn8LwE=
 github.com/pierrec/lz4/v4 v4.1.14/go.mod h1:gZWDp/Ze/IJXGXf23ltt2EXimqmTUXEy0GFuRQyBid4=
+github.com/pingcap/errors v0.11.4 h1:lFuQV/oaUMGcD2tqt+01ROSmJs75VG1ToEOkZIZ4nE4=
```
LucilleH, line 88: Should this be `github.com/go-errors/errors`?

internal/boxcli/midcobra/midcobra.go
```diff
@@ -6,6 +6,8 @@ package midcobra
 import (
 	"context"
 	"encoding/hex"
+	"errors"
```
LucilleH, line 9: If you change this to `github.com/pkg/errors`, does that fix the go sum issues?

devholic: I'll check

devholic: @LucilleH I rebased changes onto the latest main, and seems resolved :)

PR #444, opened 2023-01-09
[cloud] Very basic port forwarding

mikeland86:
## Summary

In the spirit of fast incremental improvements, this is very basic port forwarding. It simply Execs an ssh command. This method has many caveats and is not meant to wide use. 

Caveats/TODOs:
* Uses ssh command instead of golang forwarding. Using ssh command directly is simpler so not sure we should change that.
* May not show an error if local port is not available. In my testing, only protected ports caused an error. Using `ExitOnForwardFailure=yes` did not seem to work for this. May need to manually ensure port is available.
* Does not stop port forwarding at all. There's several options to do this. We can attach stdin/out which would cause this command to block and we can add a clean up function that cancels the forwarding when user does ctrl-c.
* Better error message when forwarding does fail (currently just shows `Error: exit status 255`)

I plan follow up PRs that address these.

## How was it tested?

Create VM:
```bash
devbox cloud shell 
devbox add apacheHttpd
devbox services start apache
```

Locally:
```bash
devbox cloud port-forward 3000:8080
curl localhost:3000
```


loreto (approved):
LGTM

1. Make sure the command is hidden for now. I can't tell if it is.
2. I think you should allow for a single port specification too: `devbox cloud port-forward 8080` translates to `devbox cloud port-forward 8080:8080`
3. Check in this `ssh` based implementation, but after thinking a bit more about it, I think we should evolve it to use `mutagen` for the port forwarding. That will let us support multiple protocols (in addition to ssh), mutagen will be responsible for managing the ssh connection, and it'll make it trivial to implement a "list" of all the ports being forwarded.

loreto:
Example with mutagen:
```
mutagen forward create \
    --name "my-session" \
    tcp:127.0.0.1:8000 vm-hostname:127.0.0.1:80
 ```

mikeland86:
1. This is hidden because `cloud` is hidden but I can explicitly mark this as hidden as well.
2. Sounds good will do. Another UX improvement could be to allow `:8080` which automatically finds am unused local port
3. Sounds good. Need to catch up on mutagen stuff.

loreto:
> 1. This is hidden because `cloud` is hidden but I can explicitly mark this as hidden as well.
> 2. Sounds good will do. Another UX improvement could be to allow `:8080` which automatically finds am unused local port
I like this idea.

> 3. Sounds good. Need to catch up on mutagen stuff.
Before we get into the mutagen stuff, what do you think would be missing from the current implementation, to deem it "ready for open beta". Trying to think if it makes more sense to polish the ssh approach, and have it ready for beta or if it's best to get started on the mutagen approach as the way to make it ready for beta.

mikeland86:
> Before we get into the mutagen stuff, what do you think would be missing from the current implementation, to deem it "ready for open beta". Trying to think if it makes more sense to polish the ssh approach, and have it ready for beta or if it's best to get started on the mutagen approach as the way to make it ready for beta.

I think the 2 most important are:

1) Good error messages when forwarding fails for any reason
2) Keep devbox process alive and cancel forwarding when it stops (or some other way to stop forwarding)

mikeland86:
I can try to mutagen approach and if it's simple enough I'll move to that. If I hit any issues, I can always fallback

PR #449, opened 2023-01-10
Update sentry to avoid logging errors from os exec commands

LucilleH:
## Summary
Update sentry to avoid logging errors from os exec commands. After https://github.com/jetpack-io/devbox/pull/441 goes in, sentry starts to log user script errors which are not devbox errors. This is to exclude those errors. cc @devholic 

## How was it tested?
devbox run build
./dist/devbox shell -- exit 1
./dist/devbox run nonexistentcmd

### DIFF
--- a/internal/boxcli/midcobra/midcobra.go
+++ b/internal/boxcli/midcobra/midcobra.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
 )
 
 type Executable interface {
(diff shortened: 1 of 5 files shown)


savil (approved):
This code lgtm, thanks!  

I would prefer preserving the old behavior where none of the midcobra-post-run-handlers get a non-nil error for this scenario.

> We probably want to exclude ExitError errors from being passed into the midcobra-postRun handlers. But it has the edge-case that we may have some other internal use of exec.Command which returns ExitError which we do want to pass into midCobra-postRun handler, so sentry can log it. A fix for this edge-case is to wrap the devbox run and devbox shell -- cmd errors in a devbox defined UserExitError and exclude just those in midcobra. 

WDYT?

savil:
The scenario I'm a bit concerned about is that we do call `exec.Command` for a number of use-cases (e.g. `nix` commands, or `mutagen` commands). I think errors in those we would want to log in sentry because they would lead us to understanding some mis-configuration in how we set up these tools for our users to use.

LucilleH:
@savil good points. Updated. PTAL

internal/nix/shell.go
```diff
@@ -237,7 +238,7 @@ func (s *Shell) Run(nixShellFilePath string) error {
 	cmd.Stderr = os.Stderr
 
 	debug.Log("Executing nix-shell command: %v", cmd.Args)
-	return errors.WithStack(cmd.Run())
+	return errors.WithStack(usererr.NewExecError(cmd.Run()))
```
savil, line 241: This is a bit messy, but I believe this can be invoked for two scenarios:
1. regular `devbox shell`
2. running a devbox script.

I think we want `usererro.NewExecError` only for the scenario (2).

To fix that we could wrap this here by checking `if s.ScriptCommand != "" {`
(and for the if-flakes scenario above too)

internal/boxcli/midcobra/telemetry.go
```diff
@@ -91,6 +91,7 @@ func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr
 			"packages":   pkgs,
 		})
 	})
+
```
savil, line 94: kill?

internal/boxcli/midcobra/midcobra.go
```diff
@@ -62,17 +63,30 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
...
 	for _, m := range ex.middlewares {
-		m.postRun(ex.cmd, args, err)
+		m.postRun(ex.cmd, args, postRunErr)
 	}
 
 	if err != nil {
 		// If the error is from the exec call, return the exit code of the exec call.
+		// Note: order matters! Check if it is a user exec error before a generic exit error.
 		var exitErr *exec.ExitError
+		if errors.As(err, &userExecErr) {
```
savil, line 87: nit: can we move `var exitErr *exec.ExitError` to below this if-block?

savil (approved):
thanks!

One comment inline

internal/nix/shell.go
LucilleH: Will do. Ugh this logic is getting quite convoluted. cc @ipince please launch your print-dev-env stuff soon 😢

PR #454, opened 2023-01-11
[services] Ensure services have env vars + cleanup

mikeland86:
## Summary

Ensures services have correct environment when called in shell. This fixes a particular case where a plugin is installed and then a service is started without restarting the shell.

Cleanup:

* Removed feature flag
* Simplified code by having plugin.Env return a slice instead of a map
* Factored out service start/stop to make linter happy

## How was it tested?

* `devbox shell`
* `devbox add apacheHttpd && devbox service start apache`


loreto (approved):
LGTM

But could you add a todo next to the env var logic, that we would stop modifying the environment once we do the environment through nix itself?

mikeland86:
> But could you add a todo next to the env var logic, that we would stop modifying the environment once we do the environment through nix itself?

will do

PR #455, opened 2023-01-11
Refactor sentry logic and fix bug

loreto:
## Summary
- Refactor sentry logic into it's own function
- Ensure we do not call sentry when the `runErr` is `nil`.

## How was it tested?
I didn't ...
I'm thinking it's a small refactor, but let me know if you think I should test it more before landing.


LucilleH (approved):
Oh nice! And oops I guess my previous change made sentry logging nil error as well

LucilleH:
Pulled your branch and tested it. Works.

PR #457, opened 2023-01-11
Experimental impl of nixless shell

ipince:
## Summary
This is a simplistic implementation of `devbox shell` that launches a shell directly, instead of calling `nix-shell`. Instead it uses `nix print-dev-env` to construct the environment needed to replicate a nix shell, and launches the shell directly. As with `devbox shell`, it lets some parts of the user's environment "leak" into the shell, in the same ways that the current implementation of `devbox shell` does. The behavior is behind a feature flag, so this PR should be a no-op for any user that doesn't explicitly enable the feature.

The current implementation relies on exporting the variables in the shellrc file, which I'd like to change in the future (will probably just change it on this PR itself).

## How was it tested?
Tried it with a few shells and verified that the PATH and other vars looked reasonable, and that calls to nix packages resolved to the packages in the nix store.
```
DEVBOX_FEATURE_NIXLESS_SHELL=1 SHELL=[zsh bash sh] ./devbox shell

DEVBOX_FEATURE_NIXLESS_SHELL=1 ./devbox shell -- hello
```

### DIFF
--- a/internal/boxcli/featureflag/nixless_shell.go
+++ b/internal/boxcli/featureflag/nixless_shell.go
@@ -1,0 +1,6 @@
+package featureflag
+
+// NixlessShell controls the nixless shell feature. When enabled, `devbox shell`
+// creates a shell without relying on `nix-shell`, but by setting the required environment
+// variables and spawning a shell directly.
+var NixlessShell = disabled("NIXLESS_SHELL")
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -290,7 +290,7 @@ func (s *Shell) computeNixShellEnv(nixShellFilePath string) (map[string]string, 
 	ignoreList := []string{"HOME"} // do not overwrite the user's HOME.
 
 	vars := map[string]string{}
-	for name, v := range vaf.Variables {
+	for name, vrb := range vaf.Variables {
 		if slices.Contains(ignoreList, name) {
 			continue
 		}
@@ -300,8 +300,8 @@ func (s *Shell) computeNixShellEnv(nixShellFilePath string) (map[string]string, 
 		// var: export VAR=VAL
 		// exported: export VAR=VAL
 		// array: declare -a VAR=('VAL1' 'VAL2' )
-		if v.Type == "exported" {
-			vars[name] = shellescape.Quote(v.Value.(string))
+		if vrb.Type == "exported" {
+			vars[name] = shellescape.Quote(vrb.Value.(string))
 		}
 	}
 
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -19,6 +19,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/debug"
+	"golang.org/x/exp/slices"
 )
 
 //go:embed shellrc.tmpl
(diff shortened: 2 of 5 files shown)


internal/nix/shell.go
```diff
@@ -235,15 +236,44 @@ func (s *Shell) Run(nixShellFilePath string) error {
...
+
+		cmd = exec.Command(s.binPath)
+		cmd.Env = append(filterVars(env, buildAllowList(s.env)), extraEnv...)
+		cmd.Args = append(cmd.Args, extraArgs...)
+		debug.Log("Executing shell %s with args: %v", s.binPath, cmd.Args)
+	} else {
+		// Use nix-shell
+		cmd = exec.Command("nix-shell", "--command", s.execCommand(), "--pure")
+		keepArgs := toKeepArgs(env, buildAllowList(s.env))
+		fmt.Printf("kept args: %+v\n", keepArgs)
```
savil, line 267: remove

internal/nix/shell.go
```diff
@@ -235,15 +236,44 @@ func (s *Shell) Run(nixShellFilePath string) error {
...
+		if err != nil {
+			// We don't have a good fallback here, since all the variables we need for anything to work
+			// are in the shellrc file. For now let's fail. Later on, we should remove the vars from the
+			// shellrc file. That said, one of the variables we have to evaluate ($shellHook), so we need
+			// the shellrc file anyway (unless we remove the hook somehow).
+			debug.Log("Failed to write devbox shellrc: %s", err)
+			return errors.WithStack(err)
+		}
+		// Link other files that affect the shell settings and environments.
+		s.linkShellStartupFiles(filepath.Dir(shellrc))
```
savil, line 256: ah, this `linkShellStartupFiles` is what I was referring to in demo-day. If you look into the implementation, we have only filled it out for `zsh`. We should probably fill it out for other shells as well. WDYT?

internal/nix/shell.go
```diff
@@ -252,6 +282,32 @@ func (s *Shell) Run(nixShellFilePath string) error {
 	return errors.WithStack(err)
 }
 
+func (s *Shell) computeNixShellEnv(nixShellFilePath string) (map[string]string, error) {
+	vaf, err := PrintDevEnv(nixShellFilePath)
```
savil, line 286: what does `vaf` stand for?

internal/nix/shellrc.tmpl
```diff
@@ -16,6 +16,13 @@ content readable.
 
 */ -}}
 
+{{ range $name, $value := .NixEnv }}
+export {{ $name }}={{ $value }}
+{{ end }}
+
+# The shell shellHook defined in shell.nix.
+eval $shellHook
```
savil, line 24: I didn't follow this part. 

Is the shellHook, the same as the "init_hook" in devbox.json? I believe that is called `UserHook` in the struct that is passed into this shellrc.tmpl

If not, what is it?

savil (approved):

internal/nix/shellrc.tmpl
ipince: It's the hook defined in `shell.nix`

internal/nix/shell.go
ipince: variables and functions, but that's really long...

internal/nix/shell.go
ipince: ah i see.. well, yeah, it would make sense (but it should be separate from this)

internal/nix/shell.go
ipince: done

internal/boxcli/featureflag/nixless_shell.go
```diff
@@ -0,0 +1,6 @@
+package featureflag
+
+// NixlessShell controls the nixless shell feature. When enabled, `devbox shell`
+// creates a shell without relying on `nix-shell`, but by setting the required environment
+// variables and spawning a shell directly.
+var NixlessShell = disabled("NIXLESS_SHELL")
```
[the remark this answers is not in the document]
ipince: Hmmm if it's too confusing, we could do `ENV_SHELL` or something like that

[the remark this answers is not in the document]
savil: I like ENV_SHELL :)

PR #458, opened 2023-01-11
[direnv] Added env variables from plugins to direnv

mohsenari:
## Summary
Transformed `devbox shell --print-env` command to add env variables from plugins when cd-ing into a direnv directory.

## How was it tested?
- compile
- run `./devbox init` (requires direnv)
- respond yes to prompt for direnv integration
- Using nginx as an example for a package with plugin, run `./devbox add nginx`
- confirm nginx plugin's env vars are set 
- `echo $NGINX_CONFDIR `, `echo $NGINX_PATH_PREFIX`, and `echo $NGINX_TMPDIR`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -215,6 +215,7 @@ func (d *Devbox) Shell() error {
 		return err
 	}
 
+	// tes
 	env, err := plugin.Env(d.cfg.Packages, d.projectDir)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


internal/impl/devbox.go
```diff
@@ -348,14 +349,18 @@ func (d *Devbox) Exec(cmds ...string) error {
...
+	script := ""
+	for _, pluginEnv := range pluginEnvs {
+		script += fmt.Sprintf("export %s\n", pluginEnv)
 	}
-	// TODO: For now we just updated the PATH but this may need to evolve
-	// to essentially a parsed shellrc.tmpl
-	fmt.Fprintf(d.writer, "export PATH=\"%s:$PATH\"", profileBinDir)
+	// we have to use stdout here because d.writer is set to output to stderr
+	// which makes direnv unable to read the output
+	fmt.Fprintln(os.Stdout, script)
```
mikeland86, line 363: Instead of printing to stdout in a library, what if this returns the script? and in `runShellCmd` you can do `cmd.Print()` ?

internal/impl/devbox.go
```diff
@@ -348,14 +349,18 @@ func (d *Devbox) Exec(cmds ...string) error {
...
-		return errors.WithStack(err)
+		return err
+	}
+	script := ""
+	for _, pluginEnv := range pluginEnvs {
+		script += fmt.Sprintf("export %s\n", pluginEnv)
 	}
-	// TODO: For now we just updated the PATH but this may need to evolve
-	// to essentially a parsed shellrc.tmpl
-	fmt.Fprintf(d.writer, "export PATH=\"%s:$PATH\"", profileBinDir)
```
mikeland86, line 358: Is path no longer needed?

internal/impl/devbox.go
mikeland86: If we do this we should rename this function, maybe just `PluginEnv`

internal/impl/devbox.go
mohsenari: Correct. I'm using direnv's [use nix](https://direnv.net/man/direnv-stdlib.1.html#codeuse-nix-code) feature to get the nix environment. So no need to modify the path.

internal/impl/devbox.go
mohsenari: Good suggestion. I'll make the change and push it now.

mikeland86 (approved):

PR #460, opened 2023-01-11
[direnv] Fix a bug with devbox shell enabled & direnv

mohsenari:
## Summary
With recent changes to direnv integration. cd-ing into a direnv integration devbox project would prevent the user to run `devbox shell`. That is because the env variable `DEVBOX_SHELL_ENABLED` gets set alongside others with direnv integration even though the user may not be inside a devbox shell. 

This change keeps a back up of the value of `DEVBOX_SHELL_ENABLED` before letting direnv make any changes to env variables. Then puts the backed up value back for `DEVBOX_SHELL_ENABLED`.

## How was it tested?
- cd into a direnv directory
- confirm direnv has set devbox's env variables
- run `./devbox shell`
- confirm it goes into the shell successfully
- while in shell run `cd ..` and then `cd` back into the directory
- confirm `DEVBOX_SHELL_ENABLED` is still set to 1
- exit from devbox shell
- confirm `DEVBOX_SHELL_ENABLED` is either 0 or none

### DIFF
--- a/internal/impl/tmpl/envrc.tmpl
+++ b/internal/impl/tmpl/envrc.tmpl
@@ -4,13 +4,15 @@
 use_devbox() {
     watch_file devbox.json
     if [ -f .devbox/gen/shell.nix ]; then
+        DEVBOX_SHELL_ENABLED_BACKUP=$DEVBOX_SHELL_ENABLED
         use nix .devbox/gen/shell.nix
         eval $(devbox shell --print-env)
+        export DEVBOX_SHELL_ENABLED=$DEVBOX_SHELL_ENABLED_BACKUP
     fi
 }
 use devbox
 
 
 
-# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/ 
+# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
 # for more details
--- a/internal/impl/tmpl/envrc.tmpl
+++ b/internal/impl/tmpl/envrc.tmpl
@@ -12,7 +12,5 @@ use_devbox() {
 }
 use devbox
 
-
-
 # check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
 # for more details


internal/impl/devbox.go
```diff
@@ -215,6 +215,7 @@ func (d *Devbox) Shell() error {
 		return err
 	}
 
+	// tes
```
ipince, line 218: remove?

internal/impl/tmpl/envrc.tmpl
```diff
@@ -4,13 +4,13 @@
 use_devbox() {
     watch_file devbox.json
     if [ -f .devbox/gen/shell.nix ]; then
+        DEVBOX_SHELL_ENABLED_BACKUP=$DEVBOX_SHELL_ENABLED
```
ipince, line 7: I don't fully understand how this solves the shell problem? maybe because I'm not familiar with the direnv config language / model

internal/impl/tmpl/envrc.tmpl
```diff
@@ -4,13 +4,13 @@
...
         use nix .devbox/gen/shell.nix
         eval $(devbox shell --print-env)
```
ipince, line 9: do we even need this anymore?

internal/impl/tmpl/envrc.tmpl
savil: @mohsenari I'm also not following. I thought the reason `devbox shell` doesn't work is because `DEVBOX_SHELL_ENABLED` is true and we don't allow shell-in-shell so it fails, like below:
```
❯ devbox shell
Starting a devbox shell...

(devbox) ❯ devbox shell
Error: You are already in an active devbox shell.
Run 'exit' before calling devbox shell again. Shell inception is not supported.
```

So, I thought your fix would `unset DEVBOX_SHELL_ENABLED`  after `$(devbox shell --print-env)` prior to `eval` it. But then it seems `devbox shell --print-env` just prints `PATH`?
```
testdata/go/go-1.19 ❯ devbox shell --print-env
export PATH="/Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/nix/profile/default/bin:$PATH"%
```

so it seem that I don't understand either what was going wrong in the first place....

internal/impl/devbox.go
mohsenari: done

internal/impl/tmpl/envrc.tmpl
mohsenari: @savil  I changed the output of `devbox shell --print-env` to print env variables that are needed to be set for plugins. If you compile from main you'll see (https://github.com/jetpack-io/devbox/blob/main/internal/boxcli/shell.go#L53).

@ipince I added `use nix .devbox/gen/shell.nix` in `.envrc`. This line sets up the computed environment for devbox shell which includes `DEVBOX_SHELL_ENABLED`. That means if I cd into the directory direnv will set `DEVBOX_SHELL_ENABLED=1` without the user actually being in devbox shell. 
So I have to save the original value before the `use nix` line. Then afterwards I put the original value back. This way both in shell and outside of shell direnv works as expected.

internal/impl/tmpl/envrc.tmpl
mohsenari: yes, see explanation above.

savil (approved):

PR #468, opened 2023-01-13
Fixed plugin env not setting in direnv

mohsenari:
## Summary
cmd.print() was not printing in stdout. Therefore the output was not being picked up by `eval` in `.envrc`
## How was it tested?
- `devbox init`
- `devbox add nginx`
- `echo $NGINX_CONFDIR`

### DIFF
--- a/internal/boxcli/shell.go
+++ b/internal/boxcli/shell.go
@@ -5,7 +5,6 @@ package boxcli
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


internal/boxcli/shell.go
```diff
@@ -55,8 +56,8 @@ func runShellCmd(cmd *cobra.Command, args []string, flags shellCmdFlags) error {
 		if err != nil {
 			return err
 		}
-		// print to stdout instead of stderr so that direnv can read the output
-		cmd.Print(script)
+		// explicitly print to stdout instead of stderr so that direnv can read the output
+		fmt.Fprint(os.Stdout, script)
```
mikeland86, line 60: `fmt.Fprint(cmd.OutOrStdout(), script)`

mikeland86 (approved):
Good catch.

@loreto should we set cmd.Out to be Stdout? it starts of nil which causes cmd.Print to print to stderr which feels weird?

internal/boxcli/shell.go
mohsenari: done

PR #469, opened 2023-01-13
Prevent shell inception in cloud shell

savil:
## Summary

We want to prevent shell inception in both `devbox shell` and `devbox cloud shell`.

Specifically, within regular shell and cloud shell we want to prevent another
regular shell or cloud shell from starting.

- refactor into a common utility function, and parameterize by the attempted 
- change from vanilla golang error to `usererr`

## How was it tested?

tested in `devbox shell`:
```
(devbox)
❯ devbox shell

Error: You are already in an active devbox shell.
Run `exit` before calling `devbox shell` again. Shell inception is not supported.

(devbox)
❯ devbox cloud shell

Error: You are already in an active devbox shell.
Run `exit` before calling `devbox cloud shell` again. Shell inception is not supported.
```

I didn't test within a `devbox cloud shell`, but conceptually it should work the
same because in the VM the user will be within a `devbox shell` again.

### DIFF
--- a/internal/boxcli/cloud.go
+++ b/internal/boxcli/cloud.go
@@ -75,6 +75,10 @@ func cloudPortForwardCmd() *cobra.Command {
 }
 
 func runCloudShellCmd(cmd *cobra.Command, flags *cloudShellCmdFlags) error {
+	if devbox.IsDevboxShellEnabled() {
+		return shellInceptionErrorMsg("devbox cloud shell")
+	}
+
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #469** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/469" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/469?utm_source=stack-comment).

PR #471, opened 2023-01-13
[nix] Improve nix install, add --daemon flag

mikeland86:
## Summary

This improves nix install in a few ways:

* Downloads install script every time which ensures newest nix version
* Adds `--daemon` flag so users can control if they install single or multi user version (only on `setup nix` command, but we could extend this to other commands that ensure nix, not sure)
* If user tries to install single user as root (something nix does not support) we show a helpful error message in yellow

btw, `--daemon` is a confusing name, but it's what nix uses so I didn't want to change it.

## How was it tested?

On docker fresh ubuntu:

* `devbox setup nix` as root. Got error in yellow
* `devbox setup nix --daemon` worked as expected
* Tried all combinations of daemon flag (true, false, unset) and ensured `installScript` was what I expected


loreto (approved):
LGTM

PR #472, opened 2023-01-13
[cloud telemetry] part 1: derive userId from githubUsername, and identifyUser

savil:
## Summary

This PR will:
- derive a userId by hashing the github username
- call `segment.Identify` to connect the anonymousId to the userId

## How was it tested?

did a sanity test for starting a `devbox shell`

- [ ] will modify the compiled binary to have the build keys for telemetry so that 
it is logged for my github user and verify that it is shown in amplitude/segment.


savil:
* main
  * **PR #472** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/472" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/472?utm_source=stack-comment).

internal/boxcli/midcobra/telemetry.go
```diff
@@ -201,3 +209,35 @@ func (m *telemetryMiddleware) trackEvent(evt *event) {
...
+		Traits:      segment.NewTraits().Set("githubUsername", username),
+	})
+	return errors.WithStack(err)
+}
+
+// userIDFromGithubUsername hashes the github username and produces a 64-char string as userId
+func userIDFromGithubUsername(username string) string {
+
+	const salt = "d6134cd5-347d-4b7c-a2d0-295c0f677948"
+	mac := hmac.New(sha256.New, []byte(salt))
```
loreto, line 237: I'm not sure how much this matters, but the code we're using to hash the device ID has the salt in a different place than you do. See: https://github.com/denisbrodbeck/machineid/blob/f07179bb15a0bec252453468aa607c670f5c12b5/helper.go#L24 (`appID` is the salt)

internal/boxcli/midcobra/telemetry.go
```diff
@@ -176,6 +181,9 @@ func (m *telemetryMiddleware) trackEvent(evt *event) {
 		_ = segmentClient.Close()
 	}()
 
+	// deliberately ignore error
+	_ = identifyUser(segmentClient, evt.DeviceID)
```
loreto, line 185: Nit: Should we go ahead and rename the `DeviceID` field in the event object to `AnonymousID`?

internal/boxcli/midcobra/telemetry.go
```diff
@@ -176,6 +181,9 @@ func (m *telemetryMiddleware) trackEvent(evt *event) {
...
+
 	_ = segmentClient.Enqueue(segment.Track{ // Ignore errors, telemetry is best effort
```
loreto, line 187: So, when the `userID` is available, I think we need to set it in `Track` too.

I think you need to refactor your code a bit:
- Add the `userID` field to event when we first create the event object (btw, that whole logic of creating the event object could now belong in it's own function)
- If the field is set, use it in both the call to `segment.Identify` and the call to `segment.Track`

loreto (approved):
LGTM

PR #473, opened 2023-01-14
[cloud telemetry] part 2: add 'cloud_region' property to segment

savil:
## Summary

This PR adds two properties to each segment event from midcobra:
1. `cloud: bool`
2. if cloud is true, then `cloud.region: string`

## How was it tested?

- [x] opened a cloud shell and verified that `DEVBOX_REGION` env var is populated.
- [x] do an end to end test once the gateway change in https://github.com/jetpack-io/axiom/pull/2940 is deployed

### DIFF
--- a/internal/boxcli/midcobra/telemetry.go
+++ b/internal/boxcli/midcobra/telemetry.go
@@ -134,6 +134,7 @@ type event struct {
 	AnonymousID   string
 	AppName       string
 	AppVersion    string
+	CloudRegion   string
 	Command       string
 	CommandArgs   []string
 	CommandError  error
@@ -164,6 +165,7 @@ func (m *telemetryMiddleware) newEventIfValid(cmd *cobra.Command, args []string,
 		AnonymousID:  telemetry.DeviceID(),
 		AppName:      m.opts.AppName,
 		AppVersion:   m.opts.AppVersion,
+		CloudRegion:  os.Getenv("DEVBOX_REGION"),
 		Command:      subcmd.CommandPath(),
 		CommandArgs:  subargs,
 		CommandError: runErr,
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #472** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/472" target="_blank">Graphite</a> 
    * **PR #473** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/473" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/473?utm_source=stack-comment).

internal/boxcli/midcobra/telemetry.go
```diff
@@ -184,6 +184,22 @@ func (m *telemetryMiddleware) trackEvent(evt *event) {
...
+	isInDevboxCloud := cloudRegion != ""
+	properties := segment.NewProperties().
+		Set("command", evt.Command).
+		Set("command_args", evt.CommandArgs).
+		Set("failed", evt.Failed).
+		Set("duration", evt.Duration.Milliseconds()).
+		Set("packages", evt.Packages).
+		Set("sentry_event_id", evt.SentryEventID).
+		Set("shell", evt.Shell).
+		Set("cloud", isInDevboxCloud)
```
loreto, line 197: 1. Instead of computing these here, I'd prefer if you compute when creating the event. And here you're just copying fields from the even object.

2. I don't know if we need both `cloud` and `cloud.region`. Why don't we set a single `cloud_region` property and that's it? If the region is set, we know that it's in the cloud.

loreto (approved):
LGTM

PR #477, opened 2023-01-18
[cuecfg] Don't HTML escape json

mikeland86:
## Summary

This fixes issue reported by @Lagoja where we are HTML encoding the config.

Not sure if it makes sense to only expose `cuecfgMarshalJSON` but this was easiest/quickest fix.

## How was it tested?

* Added a script with "&&" and ">>" 
* Added a pkg to force a config write
* Inspected config.

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -12,6 +12,6 @@
     }
   },
   "nixpkgs": {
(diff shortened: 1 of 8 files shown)


loreto (approved):
LGTM

PR #479, opened 2023-01-19
[cloud] add shell time-to-interactive segment event

savil:
## Summary

This PR logs a `shell-time-to-interactive` segment event.

`devbox shell` approach:
1. Upon starting a command, we capture the `telemetry.CommandStartTime()`.
2. When `devbox shell` runs, in the custom shellrc we have a "Devbox Post Init Hooks" section. Here, we call `devbox log shell-tti <command-start-time>`.
3. `devbox log shell-tti <command-start-time>` logs to segment TrackEvent with the elapsed time.

`devbox cloud shell` approach:
1. invoke `ssh <username>@<vm-address> start_devbox_shell.sh project-path <command-start-time>`.
2. (TODO in a separate axiom PR): we will invoke `DEVBOX_SHELL_START_TIME=<command-start-time> devbox shell`
3. `devbox shell` will use the `DEVBOX_SHELL_START_TIME` instead of the `telemetry.CommandStartTime()` in step (2) in the previous `devbox shell` approach.
4. Added an `is_cloud` property in segment.TrackEvent.

Limitation: the elapsed time logged's accuracy is vulnerable to clock skew if the local laptop's clock is not accurate.

Implementation Note:
1. I tried to share as much as I could between `telemetry/segment.go` and `midcobra/telemetry.go`.
2. We could consider refactoring the current segment code in `midcobra/telemetry.go` to be moved to `telemetry/segment.go` as well.

## How was it tested?

1. Tested the local `devbox shell` case by printing out the elapsed-time value just prior to logging to segment.
2. `DEVBOX_DEBUG=1 devbox cloud shell` showed that the correct unix time stamp.
3.  I can verify end-to-end that this works.

### DIFF
--- a/internal/telemetry/segment.go
+++ b/internal/telemetry/segment.go
@@ -138,7 +138,9 @@ func UserIDFromGithubUsername() string {
 }
 
 // timeFromUnixTimestamp is a helper utility that converts the timestamp string
-// into a golang time.Time struct
+// into a golang time.Time struct.
+//
+// See UnixTimestampFromTime for the inverse function.
 func timeFromUnixTimestamp(timestamp string) (time.Time, error) {
 
 	i, err := strconv.ParseInt(timestamp, 10, 64)
@@ -146,4 +148,12 @@ func timeFromUnixTimestamp(timestamp string) (time.Time, error) {
 		return time.Time{}, errors.WithStack(err)
 	}
 	return time.Unix(i, 0), nil
+}
+
+// UnixTimestampFromTime is a helper utility that converts a golang time.Time struct
+// to a timestamp string.
+//
+// See timeFromUnixTimestamp for the inverse function.
+func UnixTimestampFromTime(t time.Time) string {
+	return strconv.FormatInt(t.Unix(), 10)
 }
--- a/internal/telemetry/segment.go
+++ b/internal/telemetry/segment.go
@@ -67,6 +67,12 @@ func CommandStartTime() time.Time {
 // LogShellDurationEvent logs the duration from start of the command
 // till the shell was ready to be interactive.
 func LogShellDurationEvent(eventName string, startTime string) error {
+	opts := InitOpts()
+	if !IsEnabled(opts) {
+		// disabled
+		return nil
+	}
+
 	start, err := timeFromUnixTimestamp(startTime)
 	if err != nil {
 		return errors.WithStack(err)
@@ -75,8 +81,8 @@ func LogShellDurationEvent(eventName string, startTime string) error {
 	evt := ttiEvent{
 		Event: Event{
 			AnonymousID: DeviceID(),
-			AppName:     "",
-			AppVersion:  "",
+			AppName:     opts.AppName,
+			AppVersion:  opts.AppVersion,
 			CloudRegion: os.Getenv("DEVBOX_REGION"),
 			OsName:      OS(),
 			UserID:      UserIDFromGithubUsername(),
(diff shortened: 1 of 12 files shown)


savil:
* main
  * **PR #479** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/479?utm_source=stack-comment).

internal/telemetry/segment.go
```diff
@@ -0,0 +1,147 @@
...
+func LogShellTimeToInteractiveEvent(startTime string) error {
+	start, err := timeFromUnixTimestamp(startTime)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	evt := ttiEvent{
+		Event: Event{
+			AnonymousID: DeviceID(),
+			AppName:     "",
```
LucilleH, line 77: Why `""`?

LucilleH (commented):
QQ: I'm not sure what we are measuring for the cloud shell part. When we do
```
ssh <username>@<vm-address> start_devbox_shell.sh project-path <command-start-time>
```
Why do we need `<command-start-time>` since it is just to tell the vm to start the `devbox shell`, and the devbox in the VM will time it correctly?

Are we trying to measure the time it takes from `devbox cloud shell` to the actual shell, or from `start_devbox_shell.sh` to the actual shell?

savil:
> QQ: I'm not sure what we are measuring for the cloud shell part. When we do
> 
> ```
> ssh <username>@<vm-address> start_devbox_shell.sh project-path <command-start-time>
> ```
> 
> Why do we need `<command-start-time>`? 🤔 Since the script is just to tell the vm to start the `devbox shell`, the devbox in the VM should be able to time it correctly?
> 
> Are we trying to measure the time it takes from `devbox cloud shell` to the actual shell, or from `start_devbox_shell.sh` to the actual shell?

That's a good question. The goal is to measure "the time taken from user typing `devbox cloud shell` to them getting a prompt`. Ideally, we'd wait for the prompt but I cut off the time just prior to before the user's own shell init-hooks run (since that is outside our control).

So, I do the following high level steps:
1. capture the timestamp when the `devbox cloud shell` command starts.
2. send that timestamp to the VM during the `ssh <vm-address> start_devbox_shell.sh` call.
3. inside `start_devbox_shell.sh`, I'll calll `DEVBOX_SHELL_START_TIME=shellStartTime devbox shell`.
4. in the vm, when `devbox shell` starts it'll call `devbox log shell-tti shellStartTime`.

Does that make sense?

internal/telemetry/segment.go
savil: oh oops, I need to hook up AppName and AppVersion. Will fix!

savil:
Chatted with @LucilleH. We're thinking of the following as next steps:

1. Continue with this PR's approach of using segment events. Its not ideal for tracing but low risk.
2. Do a follow up PR with sentry's distributed tracing - but this has more uncertainty due to the cross-process work we are doing.
3. In this PR, make two changes:
   a. change the start time to be _after_ the user completes their github username prompt.
   b. change to have two logging events: log once (event `shell-ready`) when devbox's work is done for the shell (we do this currently), and log again (event `shell-tti`) when user's init hooks etc. are done when the shell is truly interactive.

LucilleH (approved):
LGTM unblocking, pending app name change, and step 3 of the above

savil:
verified the logging works end-to-end by putting this PR's devbox binary in a VM image, and invoking cloud-shell in it. Saw logs in segment's development project.

PR #481, opened 2023-01-19
[test] New testing framework to easily test devbox commands and examples

mohsenari:
## Summary
Created a simple framework inside `examples/testdata` to help simplify writing test for devbox commands and examples. 
It still needs work, for example, we still can't get the output from inside devbox shell. But I prefer merging and iterating than be perfectionist.

The way it works is similar to devbox cli itself. The main components are `TestDevbox` interface in `examples/testdata/testframework/interface.go` and its implemented struct in `examples/testdata/testframework/impl/framework.go`.

To see how to utilize the framework, take a look at examples in `testdata/commands/`

## How was it tested?
Run each test like a unit test since it is implemented using `testing` go package


examples/testdata/testframework/impl/framework.go
```diff
@@ -0,0 +1,115 @@
...
+		Commit string `json:"commit,omitempty"`
+	} `json:"nixpkgs,omitempty"`
+}
+
+type TestDevbox struct {
+	devboxJsonPath string
+}
+
+func (td *TestDevbox) Info(pkg string, markdown bool) (string, error) {
+	cmd := exec.Command("devbox", "info", pkg)
```
savil, line 31: I may be missing something, but doesn't this rely on the `devbox` binary being compiled and being present in the path. How or where is that done when executing the golang tests....?

savil: I'm wondering if we can use the approach taken in this article?
https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra

excerpt:
```
func NewRootCmd() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "hugo",
		Short: "Hugo is a very fast static site generator",
			fmt.Fprintf(cmd.OutOrStdout(), in)
			return nil
		},
	}
	cmd.Flags().StringVar(&in, "in", "", "This is a very important input.")
	return cmd
}

func Test_ExecuteCommand(t *testing.T) {
	cmd := NewRootCmd()
	b := bytes.NewBufferString("")
	cmd.SetOut(b)
	cmd.SetArgs([]string{"--in", "testisawesome"})
	cmd.Execute()
	out, err := ioutil.ReadAll(b)
	if err != nil {
		t.Fatal(err)
	}
	if string(out) != "testisawesome" {
		t.Fatalf("expected \"%s\" got \"%s\"", "testisawesome", string(out))
	}
}
```

so that way, we can compile the command and also inspect the stdout and stderr?

savil (commented):
nice start!

So, the goal is to write end-to-end tests for each of the commands and inspect the output generated which the user may see? Makes sense.

I had a suggestion for changing the internals: specifically calling the cobra commands instead of exec.Command the devbox-binary. Curious to see your thoughts and whether that may work...

examples/testdata/testframework/impl/framework.go
```diff
@@ -0,0 +1,115 @@
...
+	Packages []string `cue:"[...string]" json:"packages"`
+	Shell    struct {
+		InitHook shellcmd.Commands             `json:"init_hook,omitempty"`
+		Scripts  map[string]*shellcmd.Commands `json:"scripts,omitempty"`
+	} `json:"shell,omitempty"`
+
+	Nixpkgs struct {
+		Commit string `json:"commit,omitempty"`
+	} `json:"nixpkgs,omitempty"`
+}
```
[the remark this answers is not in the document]
mohsenari: Done.

examples/testdata/testframework/impl/framework.go
mohsenari: tI changed the testing method from `exec.Command` to this. The only issue is that now, I can't read the shell output from the `Run` command. For example, if I have a script in my devbox.json that says: `echo 'hello'`, the `hello` will neither appear in stdout nor stderr of the command.
Not a big deal for now, we can just verify there was no error running `devbox run` command.

savil (approved):
nice! this looks better :)

I'm not totally sure how to manage the testing for scripts that would run inside the "devbox shell". One idea is to enable attaching a different stdout/stderr `io.Writers` to the bash/zsh/fish/etc. shell that we spawn, and exposing that `io.Writer` for the test-framework to hook into?

mohsenari:
@savil I'm not sure either. My suggestion to have an end-to-end test in different shell environments is to write shell scripts with [expect](https://likegeeks.com/expect-command)

PR #482, opened 2023-01-19
Implement stricter "devbox run" using "nix print-dev-env"

ipince:
## Summary

This is part of unifying everything (devbox run/shell) to use nix print-dev-env.

This implementation of `devbox run` is pretty simple:
* Write all the init hooks into a file named `.devbox/gen/scripts/.hooks.sh`. This includes plugin hooks.
* Write all the scripts into files named `.devbox/gen/scripts/<name>.sh.` The scripts have a call to source the hooks in the beginning.
* When calling `devbox run`: (1) compute the environment with `nix print-dev-env`, (2) merge any plugin environment vars, and (3) call `sh -c <script>` with the computed environment.

## How was it tested?
Tested by running all the scripts in this `devbox.json` and confirming their results as expected:
```
{
  "packages": [
    "go_1_19",
    "golangci-lint",
    "hello",
    "ps",
    "redis",
    "ruby"
  ],
  "shell": {
    "init_hook": [
      "export \"GOROOT=$(go env GOROOT)\"",
      "echo \"im the init hook and i ran\""
    ],
    "scripts": {
      "test_exit_code": "exit 1",
      "test_external_script": "./external.sh",
      "test_init_hook": [
        "echo \"inside script\"",
        "echo $GOROOT"
      ],
      "test_multi": [
        "go version",
        "hello"
      ],
      "test_pkg": [
        "echo $PATH",
        "hello"
      ],
      "test_plugin_env": "echo \"if plugin env works, you should see a port here (for redis): $REDIS_PORT\"",
      "test_plugin_hook": "echo \"if plugin hook works, you should see the ruby virt env here: $PATH\"",
      "test_single": "echo \"hello there\"",
      "test_touch": "which touch && touch test_touch",
      "test_vars": [
        "FOO=\"bar\"",
        "echo $FOO"
      ]
    }
  },
  "nixpkgs": {
    "commit": "aa7e3b940ad90ba58a5d8c0a2269ec557c9ecc70"
  }
}
```
as well as testing that `devbox run` works from a different directory, as well as a case with no hooks or plugins.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -38,6 +38,9 @@ const (
 
 	// shellHistoryFile keeps the history of commands invoked inside devbox shell
 	shellHistoryFile = ".devbox/shell_history"
+
+	scriptsDir    = ".devbox/gen/scripts"
+	hooksFilename = ".hooks"
 )
 
 func InitConfig(dir string, writer io.Writer) (created bool, err error) {
@@ -246,10 +249,9 @@ func (d *Devbox) Shell() error {
 	return shell.Run(nixShellFilePath)
 }
 
-func (d *Devbox) RunScriptInShell(scriptName string) error {
-	profileDir, err := d.profileDir()
-	if err != nil {
-		return err
+func (d *Devbox) RunScript(scriptName string) error {
+	if featureflag.NixDevEnvRun.Disabled() {
+		return d.RunScriptInNewNixShell(scriptName)
 	}
 
 	script := d.cfg.Shell.Scripts[scriptName]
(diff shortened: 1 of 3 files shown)


internal/impl/devbox.go
```diff
@@ -653,6 +685,63 @@ func (d *Devbox) installNixProfile() (err error) {
 	return
 }
 
+// writeScriptsToFiles writes scripts defined in devbox.json into files inside .devbox/gen/scripts.
+// Scripts (and hooks) are persisted so that we can easily call them from devbox run (inside or outside shell).
+func (d *Devbox) writeScriptsToFiles() error {
+	err := os.MkdirAll(filepath.Join(d.projectDir, scriptsDir), 0755) // Ensure directory exists.
+	if err != nil {
+		return err
```
savil, line 693: errors.WithStack?

savil (approved):
nice!

internal/impl/devbox.go
ipince: sure, i'll add, but.. does anyone ever use that stack anyway?  I question its usefulness :)

PR #483, opened 2023-01-20
removed unused params from Config

mohsenari:
## Summary
title says it.
Everything that's removed is unused and left over from devbox build which we have marked as deprecated now.

## How was it tested?
compile
run devbox commands such as init, shell, and run


LucilleH (approved):
awesome!!

PR #485, opened 2023-01-21
[telemetry] Don't log user errors to Sentry

mikeland86:
## Summary

We are currently logging user errors (without the message) to Sentry. These errors are mostly due to bad inputs or correctable situations that are not unexpected. We should continue counting them via Segment so we can identify bad/broken UX flows but there's no need to log them to Sentry.

I audited all user errors and confirmed they do not need to be logged.

14 of 68 current production errors are user errors.

## How was it tested?

* I triggered a user error by running `devbox add foobar`. Confirmed it was not logged. 
* Confirmed non-user error was logged.


internal/boxcli/midcobra/telemetry.go
```diff
@@ -63,7 +64,9 @@ func (m *telemetryMiddleware) postRun(cmd *cobra.Command, args []string, runErr
 		return
 	}
 
-	m.trackError(evt) // Sentry
+	if usererr.ShouldLogError(runErr) {
```
loreto, line 67: Can we move the if statement inside of TrackError? (there's other conditionals in there already, such as checking if the error is nil or not.)

loreto (approved):
LGTM

PR #489, opened 2023-01-23
[shell] do not return error if it was a user shell command or script

savil:
## Summary

For `devbox shell`, we do NOT return an exit error if a user's shell command had 
returned a non-zero exit code. `devbox cloud shell` inherits this behavior.

For `devbox run`, we do return an error for the above scenario (as before).

## How was it tested?

scenarios:
- verified local `devbox shell`
- cloud: custom built a binary and deployed a machine. Observed that we no longer
  return an error if the user's last command was errant:

cloud shell output:
```
(cloud) ❯ which foobar

(cloud) ❯ exit
exit
Shared connection to 9080557a662987.vm.devbox-vms.internal closed.

❯ echo "$?"
0
```

### DIFF
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -282,10 +282,25 @@ func (s *Shell) Run(nixShellFilePath string) error {
 	cmd.Stderr = os.Stderr
 
 	err := cmd.Run()
-	if err != nil && s.ScriptCommand != "" {
-		// Report error as exec error when executing shell -- <cmd> script.
-		err = usererr.NewExecError(err)
+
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #489** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/489" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/489?utm_source=stack-comment).

LucilleH (approved):
ah, now I understand why we return nil prior to these changes

PR #490, opened 2023-01-24
[cloud] Auto-forward ports when services start

mikeland86:
## Summary

This is a prototype for automatically doing port forwarding when services are started on devbox cloud. It works by keeping a pseudo-socket file in `.devbox.cloud/services.json` that shares state between the local machine and the cloud machine. If a given service is declared to have a port, then when that service is started remotely, devbox will start port forwarding automatically.

Currently auto port-forwarding is done with the command `devbox cloud port-forward auto` but this is mostly for debugging purposes, it should happen automatically whenever a cloud shell session is active. Another improvement would be for `devbox services start` to listen to socket changes as well so it can print out the correct local port.

TODOs (after this PR):

* Start auto-forwarding automatically whenever a cloud shell starts
* If a service is started in a cloud shell, listen to changes in `services.json` so we can print out appropriate ports.

## How was it tested?

Locally:
```bash
devbox cloud shell
```

In cloud:
```bash
devbox services start
# manually had to modify socket because devbox cloud doesn't currently write to it 
echo '{"hosts":{"3287367c036485":{"services":{"apache":{"name":"apache","port":"8080","running":true}}}}}' > .devbox.cloud/services.json 
```

Locally:
```bash
devbox cloud port-forward ls 
curl localhost:<port>
```


mikeland86:
@loreto @gcurtis this is rebased and ready for review. Would be nice to get this into next release if approach is reasonable.

gcurtis (approved):

PR #491, opened 2023-01-24
[install UX] show incremental progress

savil:
## Summary

I was playing around with making the install step show more incremental progress.

This PR explores it by doing `nix-env --install` for just the nixpkgs and then
each package individually. This lets us print installation output as we go along.

The implementation is rather hacky. I override `devbox.cfg.Packages` to re-generate
the `development.nix`. There's more likely a more direct `nix-env --install` 
command invocation we can use but I was failing to figure it out this evening. I don't think 
we should commit this code as-is, but sharing for product oriented feedback and suggestions
on the right `nix-env --install` command to use so we can improve the hacky code.

## How was it tested?

deployed a custom binary to a custom VM:
```
Devbox Cloud
Remote development environments powered by Nix

✓ File syncing started
→ Connecting to virtual machine


Ensure nixpkgs is downloaded, extracted and evaluated.
Ensure package installed: go_1_19
Ensure package installed: bazel
Done.
Starting a devbox shell...

(devbox ☁️ ) testdata/go/go-1.19 via 🐹 💫 watching for changes
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -688,11 +688,6 @@ func (d *Devbox) installNixProfile() (err error) {
 
 	// Non flakes below:
 	packages := append([]string{""}, d.cfg.Packages...)
-	defer func() {
-		// ensure the full development.nix is written back, even if we early return due to some error
-		d.cfg.Packages = packages
-		_ = d.generateShellFiles()
-	}()
 
 	total := len(packages)
 	for idx, pkg := range packages {
@@ -710,17 +705,21 @@ func (d *Devbox) installNixProfile() (err error) {
 
 		step := stepper.Start(msg)
 
-		d.cfg.Packages = []string{pkg}
-		if err = d.generateShellFiles(); err != nil {
-			return err
-		}
-
-		cmd = exec.Command(
+		cmd := exec.Command(
 			"nix-env",
 			"--profile", profileDir,
+			// TODO savil. hook this up to gcurtis's mirrorURL
+			"-f", fmt.Sprintf("https://github.com/nixos/nixpkgs/archive/%s.tar.gz", d.cfg.Nixpkgs.Commit),
 			"--install",
-			"-f", filepath.Join(d.projectDir, ".devbox/gen/development.nix"),
 		)
+		if pkg != "" {
+			cmd.Args = append(cmd.Args, "--attr", pkg)
+		} else {
+			// TODO savil. Figure out how to pre-install just the nixpkgs
+			step.Fail(msg)
+			continue
+		}
+
 		cmd.Env = nix.DefaultEnv()
 		_, err = cmd.Output()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #491** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/491" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/491?utm_source=stack-comment).

loreto:
I love this idea.

Before diving too much into the code, some thoughts:
1. Could you use the spinner UI implemented in `internal/cloud/stepper` (move out to a more general location)

2. You're using "old style" nix commands like `nix-env`, but if Rodrigo's changes are merged (are they?), `devbox shell` is implemented via `nix print-dev-env`. Would it make sense to start using `nix build` instead?

3. Could we make the output look like this:
```
Installing packages:
  [  ]  go_1_18 (1.18)
  [  ]  ruby (2.7.6)
  ...
```

Where the version in parenthesis is the version number.

savil:
@loreto output looks like this now:

I added the steps (step/total) because the first time installing some of these packages may take a while and it'll be helpful to know how many more there are. What do you think?

```
<img width="398" alt="Screenshot 2023-01-24 at 2 13 23 PM" src="https://user-images.githubusercontent.com/676452/214433045-8c4d4ccc-b268-4cc2-88ca-a70cb9e9f4d2.png">

File: internal/impl/devbox.go [hunk 1]
```diff
@@ -634,30 +635,68 @@ func (d *Devbox) installNixProfile() (err error) {
...
-
-	cmd.Env = nix.DefaultEnv()
-
-	debug.Log("Running command: %s\n", cmd.Args)
-	_, err = cmd.Output()
+		if pkg != "" {
+			cmd.Args = append(cmd.Args, "--attr", pkg)
+		} else {
+			// TODO savil. Figure out how to pre-install just the nixpkgs
+			step.Fail(msg)
```
gcurtis, line 686: If you're looking to make a step that just warms the nixpkgs cache and nothing else, you can try:

```
nix-instantiate --eval --expr path https://github.com/nixos/nixpkgs/archive/af9e00071d0971eb292fd5abef334e66eda3cb69.tar.gz
```

gcurtis (approved):
Nice! Approving to unblock for whenever you're ready.

internal/impl/devbox.go
savil: yes! thanks. Let me try this....

savil: @gcurtis hmm what does one put for `--expr path` above?

savil:
had to drop the specific version numbers for perf reasons. We can re-introduce them once we turn on flakes.

Screenshot 2023-01-24 at 6 11 25 PM

PR #492, opened 2023-01-24
Allow devbox run to accept script args

ipince:
## Summary
TSIA.

## How was it tested?
```
devbox run script arg1 arg2
devbox run script # no args
```


mohsenari (approved):
We should add some tests to cover edge cases that this introduces

ipince:
> We should add some tests to cover edge cases that this introduces

Yes! I have a LOT of tests in my own `devbox.json` and would love to add them to the new test framework!

PR #493, opened 2023-01-24
devbox run can run arbitrary commands

ipince:
## Summary
TSIA

## How was it tested?
```
devbox run echo "hi"
devbox run hello
devbox run cowsay -d hello      # fails because -d is interpreted as devbox flag
devbox run -- cowsay -d hello   # works!
```

### DIFF
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -62,7 +62,7 @@ func parseScriptArgs(args []string, flags runCmdFlags) (string, string, []string
 	}
 
 	script := ""
-	scriptArgs := []string{}
+	var scriptArgs []string
 	if len(args) >= 1 {
 		script = args[0]
 		scriptArgs = args[1:]
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -4,13 +4,10 @@
 package boxcli
 
 import (
-	"sort"
-
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
 	"go.jetpack.io/devbox/internal/debug"
-	"golang.org/x/exp/slices"
 )
 
 type runCmdFlags struct {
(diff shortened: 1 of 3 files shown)


internal/boxcli/run.go
```diff
@@ -50,14 +47,6 @@ func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
 		return errors.WithStack(err)
 	}
 
-	// Validate script exists.
-	scripts := box.ListScripts()
-	sort.Slice(scripts, func(i, j int) bool { return scripts[i] < scripts[j] })
-	if script == "" || !slices.Contains(scripts, script) {
-		return errors.Errorf("no script found with name \"%s\". "+
-			"Here's a list of the existing scripts in devbox.json: %v", script, scripts)
-	}
```
mohsenari, line 59: What happens when a user has a script in their devbox.json named "ls" and they run `devbox run ls` ?
Does the script alias take precedent over the command `ls`? or vice versa?

internal/nix/run.go
```diff
@@ -11,9 +11,9 @@ import (
 	"go.jetpack.io/devbox/internal/debug"
 )
 
-func RunScript(nixShellFilePath string, projectDir string, scriptPath string, additionalEnv []string) error {
-	if scriptPath == "" {
-		return errors.New("attempted to run script but did not specify script name")
+func RunScript(nixShellFilePath string, projectDir string, cmdWithArgs string, additionalEnv []string) error {
+	if cmdWithArgs == "" {
+		return errors.New("attempted to run a command but gave an empty command")
```
mohsenari, line 16: it can be both a command or a script right? so can we say "attempted to run a command or a script but gave an empty one?" or something like this?

mohsenari (changes requested):

internal/boxcli/run.go
ipince: The script takes precedence

ipince:
PTAL

mohsenari (approved):

PR #497, opened 2023-01-24
[forward] Rename port-forward to forward and simplify API

mikeland86:
## Summary

Renames:
* `port-forward` -> `forward`
* `terminate` -> `stop`

Simplifies:
* Removes single port option
* accepts 0 (zero) as a local port. If this is passed in, we determine the port.

## How was it tested?

```bash
devbox cloud forward :8080
devbox cloud forward list
devbox cloud forward stop
```

### DIFF
--- a/internal/boxcli/cloud.go
+++ b/internal/boxcli/cloud.go
@@ -53,21 +53,15 @@ func cloudShellCmd() *cobra.Command {
 
 func cloudPortForwardCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "port-forward <local-port>:<remote-port> | <port> | :<remote-port> | terminate",
+		Use:   "forward <local-port>:<remote-port> | :<remote-port> | stop | list",
 		Short: "Port forwards a local port to a remote devbox cloud port",
-		Long: "Port forwards a local port to a remote devbox cloud port. If a " +
-			"single port is specified, it is used for local and remote. If no local" +
-			" port is specified, we find a suitable local port. Use 'terminate' to " +
-			"terminate all port forwards.",
+		Long: "Port forwards a local port to a remote devbox cloud port. If 0 or " +
+			"no local port is specified, we find a suitable local port. Use 'stop' " +
+			"to stop all port forwards.",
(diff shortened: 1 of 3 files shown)


internal/cloud/mutagenbox/forward.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func ForwardCreate(host, localPort, remotePort string) (string, error) {
 	var err error
-	if localPort == "" {
+	if localPort == "" || localPort == "0" {
 		localPort, err = getFreePort()
```
loreto, line 17: `0` usually already means "find an open port" in linux networking, so you might be able to get rid of `getFreePort` and just pass `0` along. Worth trying.

savil (approved):
lgtm!

1. un-hide the command?
2. I don't quite understand the need for the `0` port....feels a bit odd to me.

internal/boxcli/cloud.go
```diff
@@ -53,21 +53,15 @@ func cloudShellCmd() *cobra.Command {
 
 func cloudPortForwardCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:   "port-forward <local-port>:<remote-port> | <port> | :<remote-port> | terminate",
+		Use:   "forward <local-port>:<remote-port> | :<remote-port> | stop | list",
```
loreto, line 56: `ls` instead of `list`?

loreto (approved):
LGTM

internal/boxcli/cloud.go
mikeland86: We accept both but the official command is `list` and `ls` is an alias. Do you want me to flip?

internal/cloud/mutagenbox/forward.go
mikeland86: will try

mikeland86: nope, mutagen just uses 0 and fails silently :(

mikeland86:
@savil see @loreto's comment about port 0.

PR #506, opened 2023-01-25
[install UX] incrementally show the output in-line

savil:
## Summary

Thanks to @mikeland86's suggestion we can have snappy incremental in-line updates
of what the `nix-env --install` is doing.

## How was it tested?

added `buf` to a devbox.json and did `devbox shell`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -5,6 +5,7 @@
 package impl
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #506** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/506" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/506?utm_source=stack-comment).

internal/impl/devbox.go
```diff
@@ -726,7 +727,45 @@ func (d *Devbox) installNixProfile() (err error) {
 		}
 
 		cmd.Env = nix.DefaultEnv()
-		_, err = cmd.Output()
+
+		// Get a pipe to read from standard out
+		r, _ := cmd.StdoutPipe()
```
mikeland86, line 732: I think this is an error we probably want to handle?

internal/impl/devbox.go
```diff
@@ -726,7 +727,45 @@ func (d *Devbox) installNixProfile() (err error) {
...
+
+		// Use the same pipe for standard error
+		cmd.Stderr = cmd.Stdout
```
mikeland86, line 735: This was from example I copy pasted (and probably correct) but just wanted to call out the previously we used `cmd.Output()` and not `CombinedOutput` so this changes it a bit.

internal/impl/devbox.go
```diff
@@ -726,7 +727,45 @@ func (d *Devbox) installNixProfile() (err error) {
...
+				step.Display(fmt.Sprintf("%s   %s", msg, line))
+			}
+
+			// We're all done, unblock the channel
+			done <- struct{}{}
+		}()
+
+		// Start the command and check for errors
+		err := cmd.Start()
+		if err != nil {
```
mikeland86, line 759: nit same line

mikeland86 (approved):
Nice!

internal/impl/devbox.go
savil: good call

internal/impl/devbox.go
savil: true, but we were ignoring the output returned from `cmd.Output()`, so I think this is okay for now.

PR #508, opened 2023-01-25
[bug fix] re-create nix-profile on each install of packages (workaround fix)

savil:
## Summary

Problem:
- Prior to #491: 
  - we'd install packages in the nix-profile using `nix-env --install development.nix`.
This was subsequently changed to install each package individually as in `nix-env --install --attr <package>`.
  - the nix-profile would have the package under `/nix/store/<hash>-devbox-development/path/to/binary`
which is now `/nix/store/<hash>-<package>/path/to/binary`.
- So, nix will complain about this conflict in the nix-profile.

Fix #1:
A brute-force fix is to `rm -rf .devbox` directory but this is bad UX

This PR's Fix:
The fix for now is to always re-create the profile. This is not great but tolerable
since we are not currently using the profile's generations. (We rely on the nix-profile
to enable us to install the package upon `devbox add` so it is present right away and when the
user next does `devbox shell`).

## How was it tested?

Step 1: built `devbox` from a commit prior to #491 to build the "old style" nix profile
Step 2: go to `main` and run `devbox shell` in the `go-1.19` example project. This fails.
Step 3: go to this branch, and run `devbox shell` again. This works.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -555,6 +555,12 @@ func (d *Devbox) generateShellFiles() error {
 	return generateForShell(d.projectDir, plan, d.pluginManager)
 }
 
+func (d *Devbox) recreateProfileDir() (string, error) {
+	absPath := filepath.Join(d.projectDir, nix.ProfilePath)
+	_ = os.Remove(absPath)
+	return d.profileDir()
+}
+
 func (d *Devbox) profileDir() (string, error) {
 	absPath := filepath.Join(d.projectDir, nix.ProfilePath)
 	if err := os.MkdirAll(filepath.Dir(absPath), 0755); err != nil {
@@ -656,7 +662,10 @@ func (d *Devbox) printPackageUpdateMessage(
 // installNixProfile installs or uninstalls packages to or from this
 // devbox's Nix profile so that it matches what's in development.nix or flake.nix
 func (d *Devbox) installNixProfile() (err error) {
-	profileDir, err := d.profileDir()
+	// Re-create the profile dir to avoid conflicts with previous versions
+	// of the profile that may have the same binary installed under a different
+	// `/nix/store` path
+	profileDir, err := d.recreateProfileDir()
 	if err != nil {
 		return err
 	}


savil:
* main
  * **PR #508** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/508" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/508?utm_source=stack-comment).

savil:
For underlying cause, posting raw output for posterity:
```
❯ devbox shell
✓ [1/2] nixpkgs
✘ [2/2] go_1_19

Error: apply Nix derivation: error running command /nix/var/nix/profiles/default/bin/nix-env --profile /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/nix/profile/default -f https://github.com/nixos/nixpkgs/archive/52e3e80afff4b16ccb7c52e9f0f5220552f03d04.tar.gz --install --attr go_1_19: exit status 1
```

I then run the nix-env --install command printed above:

```
testdata/go/go-1.19 on  main [$] via 🐹 v1.19.2 on ☁️  (us-east-2) on ☁️  savil.srivastava@jetpack.io 💫 watching
❯ nix-env --profile /Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/nix/profile/default -f https://github.com/nixos/nixpkgs/archive/52e3e80afff4b16ccb7c52e9f0f5220552f03d04.tar.gz --install --attr go_1_19
installing 'go-1.19.3'
building '/nix/store/rparx874qnyisdm73vw60zk23sm75pyz-user-environment.drv'...
error: files '/nix/store/w0lyimyyxxfl3gw40n46rpn1yjrl3q85-go-1.19.3/bin/go' and '/nix/store/v4qfigkkamxw9qnw94jv1vbwrjiqqym5-devbox-development/bin/go' have the same priority 5; use 'nix-env --set-flag priority NUMBER INSTALLED_PKGNAME' or type 'nix profile install --help' if using 'nix profile' to find out howto change the priority of one of the conflicting packages (0 being the highest priority)
error: builder for '/nix/store/rparx874qnyisdm73vw60zk23sm75pyz-user-environment.drv' failed with exit code 1;
       last 1 log lines:
       > error: files '/nix/store/w0lyimyyxxfl3gw40n46rpn1yjrl3q85-go-1.19.3/bin/go' and '/nix/store/v4qfigkkamxw9qnw94jv1vbwrjiqqym5-devbox-development/bin/go' have the same priority 5; use 'nix-env --set-flag priority NUMBER INSTALLED_PKGNAME' or type 'nix profile install --help' if using 'nix profile' to find out howto change the priority of one of the conflicting packages (0 being the highest priority)
       For full logs, run 'nix log /nix/store/rparx874qnyisdm73vw60zk23sm75pyz-user-environment.drv'.
```

ipince:
To be clear, we only need to do this during some "transition time", right? Once we can expect most users to have upgraded their devbox (within reason), then we can stop deleting the profile every time.

savil:
> To be clear, we only need to do this during some "transition time", right? Once we can expect most users to have upgraded their devbox (within reason), then we can stop deleting the profile every time.

Yes, I would expect so...

savil:
see https://github.com/jetpack-io/devbox/pull/512 for an alternative approach to fixing this.

savil:
closing in favor of #513

PR #510, opened 2023-01-25
Add script aliases (experimental)

ipince:
## Summary

This is just a quick and dirty experiment. It adds `alias <script>=<path-to-script>` for every script. This can unlock two potentially useful things:

1. A script can call another script. I tested this and it works just fine.
2. When a user is in `devbox shell`, they could potentially just type `<script>` instead of `devbox run <script>` and execute the script they want. That said, if you run it directly, the script would be executing within the `devbox shell` environment, so it may not behave _exactly_ the same as if it were run with `devbox run`, which is a bit more strict/pure. Of course, the user could still run with `devbox run` if they wanted the other behavior.

This is just a proof-of-concept. I don't know if `alias` will work with different shells--need to test. I'd also need to add a guard to `.hook.sh`, to prevent the hooks from being run multiple times, which isn't a problem today.

## How was it tested?
Tested a script can call another script, and that I can call a script inside a shell if I manually source `.hooks.sh`. The sourcing would have to be automatic. Haven't tested that or cross-shell support.


ipince:
Oh and, of course, if a script has the same name as some other command, then the script would shadow/take precedence.

mikeland86:
If you add a feature flag we can go ahead and merge it

mikeland86 (changes requested):
RC because this is old.

Thoughts on putting this behind a devbox.json setting? (so opt-in ?)

PR #513, opened 2023-01-25
[install UX] use --remove-all flag to avoid profile package collisions

savil:
## Summary

This avoids the issue in #508
but without manually re-creating the profile directory.

## How was it tested?

repeated the test plan of #508

- [x] I'll do a timing run to ensure it doesn't regress the timings too much


savil:
* main
  * **PR #513** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/513" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/513?utm_source=stack-comment).

PR #512, opened 2023-01-25
[install UX] go back to using development.nix

savil:
## Summary

With the recent install-UX improvement of #491, we may run into issues for
existing users. #508 summarizes the issue and has a workaround fix implemented.
However, it had me thinking whether this is introducing more risk than we want
for this week.

To summarize, the improvements made in #491 and later in #506 are:
1. #491: use stepper. Install each package individually, instead of using `development.nix`.
2. #506: stream the output of `nix-env --install`.

This PR makes it so we continue doing #506 to preserve some UX improvement, but revert #491 to minimize risk.

When I try it in `devbox cloud shell` with `bazel` and `buf`, the streaming output helps
preserve the feeling of progress being made. The viewer is no longer left pondering
"is it stuck? is it broken?".

## How was it tested?

- built a linux devbox binary and deployed a custom VM
- did `devbox cloud shell` to that VM:

The following output shows but is printed in a streaming fashion as opposed
to blank for a long time and then shows up in one-go.
```
Devbox Cloud
Remote development environments powered by Nix

✓ File syncing started
→ Connecting to virtual machine


Installing nix packages. This may take a while...
	installing 'devbox-development'
	these 2 derivations will be built:
	  /nix/store/9yk55y5ywvm4ihnllfa2kcz158bmbfbg-builder.pl.drv
	  /nix/store/xwcvqhd7fd49dndn5wvn2hx2vcxlrkp0-devbox-development.drv
	these 59 paths will be fetched (919.80 MiB download, 1815.57 MiB unpacked):
	  /nix/store/039g378vc3pc3dvi9dzdlrd0i4q93qwf-binutils-2.39
	  /nix/store/0zmylwn0sznzflf137hhqlnnyskr6cdk-openjdk-headless-11.0.15+10
	  /nix/store/1d6ian3r8kdzspw8hacjhl3xkp40g1lj-binutils-wrapper-2.39
	  /nix/store/1dgws25664p544znpc6f1nh9xmjf4ykc-pcre-8.45
	  /nix/store/1gf2flfqnpqbr1b4p4qz2f72y42bs56r-gcc-11.3.0
	  /nix/store/1i5ah27gxx3a3fyjyydfwwzqq8ni33i8-ncurses-6.3-p20220507

    ... // omitted

    /nix/store/y6aj732zm9m87c82fpvf103a1xb22blp-file-5.43
    /nix/store/zdba9frlxj2ba8ca095win3nphsiiqhb-python3-3.10.8
[...]

PR #515, opened 2023-01-25
[nix] Only install packages if needed

mikeland86:
## Summary

This checks if the current profile contains a package before installing it. If no packages need to be installed it skips printing a few lines so the UX looks cleaner.

This PR would require a revert of https://github.com/jetpack-io/devbox/pull/512


internal/impl/pkgs.go
```diff
@@ -254,3 +259,46 @@ type Package struct {
...
+	if err := json.Unmarshal(out, &installed); err != nil {
+		return nil, errors.WithStack(err)
+	}
+	for _, i := range installed {
+		// rekey by name
+		installed[i.Name] = i
+	}
+
+	for _, pkg := range d.cfg.Packages {
+		info, found := nix.PkgInfo(d.cfg.Nixpkgs.Commit, pkg)
```
savil, line 287: Unfortunately, this line will hurt perf. I had it in my original code as well, and needed to remove it to make the new UX perf-neutral. Because it relies on flakes-oriented nix apis.

The impact was pretty significant for the go.1.19 example project:
- local `devbox shell` went from 5 seconds to 8 seconds.
- `devbox cloud shell` went from 37 seconds to 55 seconds

mikeland86: @savil FYI, this is fixed. `nix.PkgInfo` can now do flakes or no flakes so it no longer incurs the flake download penalty if the flag is turned off. The non-flakes version will use the same version of nixpkgs as `nix-instantiate`.

gcurtis (approved):

mikeland86:
Closing this since it is obsoleted by @savil's new changes.

PR #516, opened 2023-01-25
[test] additional tests for devbox init and run

mohsenari:
## Summary
Title says it.

## How was it tested?
Ran the tests locally and it succeeded

### DIFF
--- a/examples/testdata/commands/init/init_test.go
+++ b/examples/testdata/commands/init/init_test.go
@@ -1,0 +1,33 @@
+// Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+package testinit
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.jetpack.io/devbox/examples/testdata/testframework"
+)
+
+func TestInit(t *testing.T) {
+	td := testframework.Open()
+	_, err := td.Init()
+	assert.NoError(t, err)
+	assert.FileExists(t, "devbox.json")
+}
+
+func TestInitRecommendation(t *testing.T) {
+	td := testframework.Open()
+	err := exec.Command("touch", "package.json").Run()
+	assert.NoError(t, err)
+	err = exec.Command("touch", "requirements.txt").Run()
+	assert.NoError(t, err)
+	output, err := td.Init()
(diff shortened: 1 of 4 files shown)


examples/testdata/testframework/impl/framework.go
```diff
@@ -18,27 +18,54 @@ type TestDevbox struct {
...
+func (td *TestDevbox) Info(pkg string, markdown bool) (string, error) {
+	cmd := boxcli.InfoCmd()
+	output, err := runCmd(cmd, []string{pkg})
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	return string(output), nil
+}
+
+func (td *TestDevbox) Init() (string, error) {
```
mohsenari, line 66: this is the only function I added. The rest of the changes in this file are due to re-organizing the order of funcs to the same order that they are listed in interface.go

examples/testdata/commands/init/init_test.go
```diff
@@ -0,0 +1,33 @@
...
+	err = exec.Command("touch", "requirements.txt").Run()
+	assert.NoError(t, err)
+	output, err := td.Init()
+	assert.NoError(t, err)
+	assert.FileExists(t, "devbox.json")
+	assert.Contains(t, output, "We detected extra packages you may need.")
+	assert.Contains(t, output, "devbox add")
+	assert.Contains(t, output, "nodejs")
+	assert.Contains(t, output, "python3")
+}
```
savil, line 33: nice!

examples/testdata/commands/run/run_test.go
```diff
@@ -16,3 +16,17 @@ func TestRun(t *testing.T) {
 	_, err = td.Run("test1")
 	assert.NoError(t, err)
 }
+
+func TestRunCommand(t *testing.T) {
+	td := testframework.Open()
+	err := td.SetDevboxJson("devbox.json")
+	assert.NoError(t, err)
+
+	_, err = td.Run("ls > test.txt")
```
savil, line 25: is this doing whatever the current functionality is to do what `devbox shell -- <command>` used to do?

savil (approved):

examples/testdata/commands/run/run_test.go
mohsenari: @savil basically yes. @ipince can give more context on the difference

PR #518, opened 2023-01-25
[nix] Filter out 'devbox-development' messages

mikeland86:
## Summary

Filters out

```bash
replacing old 'devbox-development'
installing 'devbox-development'
```

from every command that runs `ensurePackagesAreInstalled`

## How was it tested?

```bash
> devbox run echo hello
Done.
hello
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -29,6 +29,7 @@ import (
 	"go.jetpack.io/devbox/internal/planner/plansdk"
 	"go.jetpack.io/devbox/internal/plugin"
 	"go.jetpack.io/devbox/internal/telemetry"
+	"go.jetpack.io/devbox/internal/ux"
 	"golang.org/x/exp/slices"
 )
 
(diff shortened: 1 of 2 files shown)


internal/impl/devbox.go
```diff
@@ -677,8 +678,9 @@ func (d *Devbox) installNixProfile() (err error) {
 	}
 
 	cmd.Env = nix.DefaultEnv()
-	cmd.Stdout = d.writer
-	cmd.Stderr = d.writer
+	cmd.Stdout = ux.NewFilterWriter(d.writer, "devbox-development")
```
savil, line 681: can we make the filtering more specific to:
```
replacing old 'devbox-development'
installing 'devbox-development'
```
?

I think it could be done by having `NewFilterWriter` take in a `[]string` as the second parameter.

The name `devbox-development` is also sometimes used in the paths in the store `/nix/store/<hash>-devbox-development/path/to/derivation-or-binary` which the packages we're installing will have.  I think we want to preserve that output.

savil (approved):
thanks!

PR #521, opened 2023-01-26
plansdk: use nixpkgs mirror when in cloud VM 

gcurtis:
This change is a bit hacky. I'd like to eventually clean this up to work better with the various nix commands. It also reverts the previous env var change since it's no longer needed.

When generating the various *.nix files, check to see if the cloud mirror is reachable. If it is, download nixpkgs archives from there instead of GitHub.

Also update the default commit hash to be the current nixpkgs unstable, which is also in the cache. This guarantees that any new devbox projects created after this release will hit the cache in the cloud.

### DIFF
--- a/internal/planner/plansdk/plansdk.go
+++ b/internal/planner/plansdk/plansdk.go
@@ -5,7 +5,9 @@ package plansdk
 
 import (
 	"fmt"
+	"net/http"
 	"os"
+	"time"
 
 	"github.com/imdario/mergo"
 	"go.jetpack.io/devbox/internal/cuecfg"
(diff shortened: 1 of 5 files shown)


internal/planner/plansdk/plansdk.go
```diff
@@ -91,11 +93,34 @@ type NixpkgsInfo struct {
 	Sha256 string
 }
 
-// The commit hash for nixos-22.11 on 2022-12-06 from status.nixos.org
-const DefaultNixpkgsCommit = "52e3e80afff4b16ccb7c52e9f0f5220552f03d04"
+// The commit hash for nixpkgs-unstable on 2023-01-25 from status.nixos.org
```
savil, line 96: not convinced about changing from nixpkgs-stable to unstable....started an internal discussion in slack.

internal/planner/plansdk/plansdk.go
```diff
@@ -91,11 +93,34 @@ type NixpkgsInfo struct {
...
+func nixpkgsMirrorURL(commitHash string) string {
+	// Use DEVBOX_REGION as a hint to see if we're in Devbox Cloud.
+	if os.Getenv("DEVBOX_REGION") == "" {
+		return ""
+	}
+
+	// Check that the mirror is responsive and has the tar file. We can't
+	// leave this up to Nix because fetchTarball will retry indefinitely.
+	client := &http.Client{Timeout: 3 * time.Second}
+	mirrorURL := fmt.Sprintf("http://[fdaa:0:a780:0:1::2]:8081/nixos/nixpkgs/archive/%s.tar.gz", commitHash)
```
savil, line 120: where does `fdaa:0:a780:0:1::2` come from?

savil (approved):
Accepting to unblock, but not sure about changing to nixpkgs-unstable.

If the main benefit is that the nixed cache support nixpkgs-unstable, could we also have it support nixpkgs-stable?

internal/planner/plansdk/plansdk.go
LucilleH: Can we put this behind a domain? `cache.devbox.sh`?

gcurtis: It's a private fly IP. You can see get it with `fly ips list` (created with `fly ips allocate-v6)`. Unfortunately we can't put it behind a domain because fly explicitly doesn't support it. This is just a hack until we have a better way of running multiple fly processes.

internal/planner/plansdk/plansdk.go
gcurtis: Commented in Slack.

gcurtis:
@savil merging for now, but let's chat more about channels tomorrow.

PR #522, opened 2023-01-26
[devbox cloud] require git in home and fs-root dirs to use as devbox project dir

savil:
## Summary

Problem:
When running `devbox cloud shell` (and other commands) if we don't find 
a `devbox.json` in the current directory and no `--config` flag is specified,
then we walk up the directory tree to find a `devbox.json` file. If we do find one,
we use that directory as the `projectDir`.

This has the danger that we may sync a sensitive directory like a user's homedir
or the filesystem-root dir (i.e. `/`). The syncing may inadvertently delete something in the directories
under this sensitive dir. 

@Lagoja ran into this since he has a devbox.json in his homedir.

Solution:
This PR protects against this by requiring a `.git` repository to be present in 
the same dir. We presume this would protect against data loss. Without a `.git`
we display a user error.

## How was it tested?

```
❯ devbox cloud shell
Devbox Cloud
Remote development environments powered by Nix


Error: Found a config (devbox.json) file at /Users/savil, but since it is a sensitive directory we require it to be part of a git repository before we sync it to devbox cloud
```

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -33,6 +33,10 @@ func Shell(w io.Writer, projectDir string, githubUsername string) error {
 	fmt.Fprintln(w, "Remote development environments powered by Nix")
 	fmt.Fprint(w, "\n")
 
+	if err := ensureProjectDirIsNotSensitive(projectDir); err != nil {
+		return err
+	}
+
 	username, vmHostname := parseVMEnvVar()
 	// The flag for githubUsername overrides any env-var, since flags are a more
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #522** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/522" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/522?utm_source=stack-comment).

Lagoja (approved):

PR #524, opened 2023-01-26
[nix] Improve output in ensurePackagesAreInstalled

mikeland86:
## Summary

This PR does the following:

* If not explicitly doing add/remove we get rid of `installing nix packages.` text.
* Whem printing nix output add indentation

## How was it tested?

```bash
➜  devbox git:(main) ✗ devbox add python3
Installing nix packages.
	this derivation will be built:
	  /nix/store/frrcrw99ypkkxjid0y8b7rzrnv5s0f8h-devbox-development.drv
	building '/nix/store/frrcrw99ypkkxjid0y8b7rzrnv5s0f8h-devbox-development.drv'...
	created 30 symlinks in user environment
	building '/nix/store/9w1n2arl4k05nrf79c2k5590kljn5jvl-user-environment.drv'...

...
```

```bash
➜  devbox git:(landau/format-add) devbox shell
Starting a devbox shell...
(devbox) ➜  devbox git:(landau/format-add) exit
➜  devbox git:(landau/format-add) devbox run echo 5
5
➜  devbox git:(landau/format-add)
```


internal/impl/writer.go
```diff
@@ -0,0 +1,37 @@
...
+func (fw *nixPackageInstallWriter) Write(p []byte) (n int, err error) {
+	lines := strings.Split(string(p), "\n")
+	for _, line := range lines {
+		if line != "" && !fw.ignore(line) {
+			_, err = io.WriteString(fw.w, "\t"+line+"\n")
+			if err != nil {
+				return
+			}
+		}
+	}
```
savil, line 26: didn't know the indentation would be so involved!

savil (approved):
thanks!

internal/impl/writer.go
mikeland86: There may be a simpler way? But we need to make sure to write strings as they are being outputted, we can't really wait for the full string and add later.

PR #526, opened 2023-01-26
First pass at testing using testscripts

loreto:
## Summary
First pass at testing using testscripts: setup the testcript framework and write a dummy tests (that is just checking the environment is initialized correctly).

A future PR will start adding more meaningful devbox-related tests.

## How was it tested?
`go test -v`


gcurtis (approved):
Nice! testscript is great for this.

PR #539, opened 2023-01-27
[flakes] add flakes.nix to git only if we are in a git repo

savil:
## Summary

This PR looks if the devbox project is in a git repository before adding the
flake.nix to a "fake git repository". With this PR, `devbox` with flakes will no longer require `git` as a hard dependency.

When I tested with a simple flakes playground project, I observed:
1. if not in a git repository, then flakes.nix can be used by `nix develop`.
2. if in a git repository, then `nix develop` fails if flakes.nix is not tracked.
3. Surprisingly, if in a mercurial repository, then `nix develop` still works if flakes.nix is not tracked by mercurial. Surprising because docs online hint that flakes supports both git and mercurial, but I guess not enforced for this requirement.

## How was it tested?

Ran `DEVBOX_FEATURE_FLAKES=1 devbox shell` in devbox projects:
1. in a git repo 
2. outside a git repo

### DIFF
--- a/internal/impl/generate.go
+++ b/internal/impl/generate.go
@@ -36,8 +36,6 @@ func generateForShell(rootPath string, plan *plansdk.ShellPlan, pluginManager *p
 	}
 
 	// Gitignore file is added to the .devbox directory
-	// TODO savil. Remove this hardcode from here, so this function can be generically defined again
-	//    by accepting the files list parameter.
 	err := writeFromTemplate(filepath.Join(rootPath, ".devbox"), plan, ".gitignore")
 	if err != nil {
 		return errors.WithStack(err)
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #539** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/539" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/539?utm_source=stack-comment).

internal/impl/generate.go
```diff
@@ -136,3 +140,24 @@ func makeFlakeFile(outPath string, plan *plansdk.ShellPlan) error {
 	}
 	return nil
 }
+
+func isProjectInGitRepo(dir string) bool {
```
LucilleH, line 144: Didn't you implement something similar to check if the directory contains `.git` and only start devbox cloud shell when `.git` is present? Can we share the code between them or no?

LucilleH (approved):

internal/impl/generate.go
savil: That sounds similar, but in practice is fairly different:
https://github.com/jetpack-io/devbox/blob/987d51874e446fde6ef13dc751c6babafb356efc/internal/cloud/cloud.go#L480

PR #543, opened 2023-01-27
Improve Nix Sourcing

loreto:
## Summary
+ Don't rely on any binaries (like env) in order to source the nix environment.
+ Don't pollute the environment with all sorts of variables, only source the three we care about.
+ Use a well-tested library to parse the env variables

## How was it tested?
Removed nix from my path, used ubuntu container.


internal/nix/source.go
```diff
@@ -26,27 +25,38 @@ func SourceNixEnv() error {
...
-		fmt.Sprintf(". %s ; echo '<<<ENVIRONMENT>>>' ; env", srcFile),
+		script,
 	)
 
 	bs, err := cmd.CombinedOutput()
 	if err != nil {
-		return errors.WithStack(err)
+		// When there's an error, the output is usually an error message that
+		// was printed to stderr and that we want in the error for debugging.
+		return errors.WithMessage(err, string(bs))
```
mikeland86, line 49: You probably want `Wrap` instead of `WithMessage`. Wrap adds a stack but WithMessage does not.

(tbh, no idea why `WithMessage` exists)

internal/nix/source.go
```diff
@@ -26,27 +25,38 @@ func SourceNixEnv() error {
...
+	// Source the nix script that sets the environment, and print the environment
+	// variables that it sets, in a way we can parse.
+
+	// NOTE: Only use shell built-ins in this script so that we don't introduce
+	// any dependencies on an external binary.
+	script := heredoc.Docf(`
+		. %s;
+		echo PATH=$PATH;
+		echo NIX_PROFILEs=$NIX_PROFILES;
+		echo NIX_SSL_CERT_FILE=$NIX_SSL_CERT_FILE;
```
mikeland86, line 37: Did you confirm that single user install only has these as well?

mikeland86 (approved):
Nice! this is much better

internal/nix/source.go
loreto: Fixed

internal/nix/source.go
loreto: Yeah. Those the three it always sets.

Technically there's one more case for single user install: sometimes, but not always, it sets `MANPATH` (for man pages). Should I handle that one too?

loreto:
Added MANPATH case, just to be safe.

PR #548, opened 2023-01-27
[cloud] Improve cloud shell error messages

mikeland86:
## Summary

This shows better error messages when cloud shell fails. A possibly common case is when a nix package doesn't exist or doesn't build correctly. Ideally we don't end the shell and let the user fix it, but in the meantime this shows a more relevant error message.

To implement this efficiently and extensibly, this uses an io.Writer to collect pre-defined ssh session errors without explicitly storing the output.

## How was it tested?

* Added a bad package to my devbox.json and did `devbox cloud shell`
* To test generic failures did `devbox cloud shell`, turned off my wifi and waited for a timeout

image


savil:
@mikeland86 looking at the screenshot in the Test Plan. The red error message is much more helpful now. 

Should we omit the part about the full ssh command that comes after the custom error message about nix packages not existing or failing to build? That ssh command is an internal implementation detail of devbox.

savil (approved):
nice! 

See the previous comment for trimming the user-visible message, if possible.

mikeland86:
@savil I agree that omitting that would be bette. This is actually the current behavior of all user errors. It prints the "cause" below the human readable part. 

This was functionality I copied over from launchpad errors. At the time it was added because sometimes the underlying error (the cause) contained important information. In hindsight I think this was only needed because we were accidentally wrapping user errors over each other. That has since been fixed so I think we could get rid of that, but it makes me a bit nervous to fix it in this PR.

PR #550, opened 2023-01-27
[bugfix] Arbitrary commands should also run init and plugin hooks

ipince:
## Summary
Before this fix, if your had `export FOO=bar` in your init hook, and you had a script `"foo": "echo $FOO"`, then these two commands would have different results:
```
> devbox run foo
bar

> devbox run -- echo '$FOO'

```
The second invocation should _also_ return `bar`. This change makes it so.

Note that the implementation is a bit odd. I left a comment explaining why it's so.

## How was it tested?
Ran the commands in the description above and they both returned the same.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -40,8 +40,9 @@ const (
 	// shellHistoryFile keeps the history of commands invoked inside devbox shell
 	shellHistoryFile = ".devbox/shell_history"
 
-	scriptsDir    = ".devbox/gen/scripts"
-	hooksFilename = ".hooks"
+	scriptsDir           = ".devbox/gen/scripts"
+	hooksFilename        = ".hooks"
+	arbitraryCmdFilename = ".cmd"
 )
 
 func InitConfig(dir string, writer io.Writer) (created bool, err error) {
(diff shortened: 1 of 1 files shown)


internal/impl/devbox.go
```diff
@@ -280,10 +281,23 @@ func (d *Devbox) RunScript(cmdName string, cmdArgs []string) error {
...
+		// which we don't want. So, one solution is to write the entire command and its arguments into the
+		// file itself, but that may not be great if the variables contain sensitive information. Instead,
+		// we save the entire command (with args) into the DEVBOX_RUN_CMD var, and then the script evals it.
+		err := d.writeScriptFile(arbitraryCmdFilename, d.scriptBody("eval $DEVBOX_RUN_CMD\n"))
+		if err != nil {
+			return err
+		}
+		cmdWithArgs = []string{d.scriptPath(d.scriptFilename(arbitraryCmdFilename))}
+		// TODO: move this env var elsewhere. I will move all the env stuff into a single ComputeEnv() function.
+		pluginEnv = append(pluginEnv, fmt.Sprintf("DEVBOX_RUN_CMD=%s", strings.Join(append([]string{cmdName}, cmdArgs...), " ")))
```
savil, line 300: is pluginEnv the environment generated _after_ the init-hooks and plugin-hooks have run?

savil (approved):

internal/impl/devbox.go
ipince: no, the stuff in `pluginEnv` comes first. do you have a concern related to ordering?

PR #553, opened 2023-01-28
[perf] Make PkgInfo faster by using non-flakes lookup

mikeland86:
## Summary

Use legacy search for now to make `info` and `add` much faster. (first add is about 20 seconds faster)

## How was it tested?

```
devbox info php
devbox add curl
```


savil (approved):
yeah!!!!  🎉

PR #555, opened 2023-01-30
cloud: clean up mutagen sync ssh commands

gcurtis:
Use the `openssh` package where possible when executing the SSH commands for setting the mutagen sync status. This provides better logging when commands fail.

Also rename `openssh.Client` to `Cmd` to better reflect that we're calling an external process.


mikeland86 (approved):
Nice! This is cleaner

PR #556, opened 2023-01-30
[GHA] Bump versions, create daily golangci-lint cache

mikeland86:
## Summary

We're going way above our GHA cache limits because every branch creates new cache keys for golangci-lint and our go builds. This means we rarely get cache hits.

This bumps GHA versions and changes golangci-lint to use daily cache (instead of creating a new unique cache key on every branch)

This also fixes a an issue where we're doing tests twice (once on push and once on pull request)

## How was it tested?

CICD

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -7,6 +7,8 @@ concurrency:
 
 on:
   pull_request:
+    paths-ignore:
+      - 'docs/**'
   workflow_call:
   workflow_dispatch:
 
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -7,11 +7,6 @@ concurrency:
 
 on:
   pull_request:
-  push:
-    branches-ignore:
-      - main
-    paths-ignore:
-      - 'docs/**'
   workflow_call:
   workflow_dispatch:
 
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -49,9 +49,9 @@ jobs:
             ~/.cache/golangci-lint
             ~/.cache/go-build
             ~/go/pkg
-          key: golangci-lint-cache-${{ steps.current-time.outputs.day }}
+          key: golangci-lint-cache-${{ runner.os }}-${{ steps.current-time.outputs.day }}
           restore-keys: |
-            golangci-lint-cache-
+            golangci-lint-cache-${{ runner.os }}-
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3.4.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-tests.yaml
```diff
@@ -36,14 +31,27 @@ jobs:
...
+          path: |
+            ~/.cache/golangci-lint
+            ~/.cache/go-build
+            ~/go/pkg
+          key: golangci-lint-cache-${{ runner.os }}-${{ steps.current-time.outputs.day }}
+          restore-keys: |
+            golangci-lint-cache-${{ runner.os }}-
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.2.0
+        uses: golangci/golangci-lint-action@v3.4.0
```
gcurtis, line 51: Do you happen to know if the go action already handles these ones? Should the linter just be caching it's own directory?

gcurtis (approved):

.github/workflows/cli-tests.yaml
```diff
@@ -36,14 +31,27 @@ jobs:
...
+        uses: josStorer/get-current-time@v2.0.2
+        id: current-time
+      - name: Mount golangci-lint cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/golangci-lint
+            ~/.cache/go-build
+            ~/go/pkg
+          key: golangci-lint-cache-${{ runner.os }}-${{ steps.current-time.outputs.day }}
```
LucilleH, line 47: Should the key be the hash of `go.mod` file?

.github/workflows/cli-tests.yaml
```diff
@@ -7,11 +7,6 @@ concurrency:
 
 on:
   pull_request:
-  push:
-    branches-ignore:
-      - main
-    paths-ignore:
-      - 'docs/**'
```
LucilleH, line 14: We don't want to ignore `docs/`?

.github/workflows/cli-tests.yaml
mikeland86: That's a good point, I think the answer is possibly yes. But the linter will cache for all code instead of just a single go.mod (in the event we build more than one project in the repo)

I think the savings of this change plus a follow up for doing the same for golang will be more than enough.

.github/workflows/cli-tests.yaml
mikeland86: This ignore was only being applied to `push` and nog `pull_request`. Did you mean to ignore for both?

.github/workflows/cli-tests.yaml
mikeland86: Caching with go.mod key creates new keys very frequently (without much benefit). Currently (after weekend) we're using 12GB of 10GB max (so pretty much all new data causes eviction). On Friday we were using 30GB, which means we're cycling though keys every 8 hours or so. 

This is a tradeoff. It means the cache is a little less useful, (but in practice pretty good), and we never get cache misses and also use a lot less storage. If GHA had unlimited storage, I would run jobs on `main` that cache with both keys, but since storage is pretty limited this feels like the right approach.

PR #561, opened 2023-01-31
Add Fsharp to the docs

Lagoja:
## Summary

Add an F# Example

## How was it tested?

Localhost + Launchpad

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -7,8 +7,6 @@ concurrency:
 
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
   push:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


.github/workflows/cli-tests.yaml
```diff
@@ -7,8 +7,6 @@ concurrency:
 
 on:
   pull_request:
-    paths-ignore:
-      - 'docs/**'
```
LucilleH, line 11: why deleting this?

LucilleH (approved):

.github/workflows/cli-tests.yaml
Lagoja: This is part of Mike's quick fix for golangci-lint. Without it, the golangci-lint jobs will not run for docs changes, even though it's still required in order to merge

PR #562, opened 2023-01-31
[telemetry] update event name and change is_cloud to shell_access kind

savil:
## Summary

Updating:
1. Event name to be of the form: `[devbox] Shell event: interactive` which is consistent with the command event names.
2. Changing is_cloud to shell_access with values `local | ssh | browser`

## How was it tested?

- [x] Need to test

1. apply development keys for segment and sentry
2. `local`: run local `devbox shell` and verify event logged in segment dashboard.
3. `ssh`: deploy custom VM with custom devbox binary. verify event logged in segment dashboard.
4. `browser`: Follow devbox-cloud README for http-gateway, but how do I test the custom devbox binary?

### DIFF
--- a/internal/telemetry/segment.go
+++ b/internal/telemetry/segment.go
@@ -156,14 +156,13 @@ func UnixTimestampFromTime(t time.Time) string {
 }
 
 func shellAccess() shellAccessKind {
-	// START_WEB_TERMINAL is used to ensure a VM is started for a browser shell.
-	// In the future, we may change our model to reuse VMs for either browser or ssh shells
-	// and so should update this logic accordingly.
-	if os.Getenv("START_WEB_TERMINAL") == "1" {
+	// Check if running in devbox cloud
+	if os.Getenv("DEVBOX_REGION") != "" {
+		// Check if running via ssh tty (i.e. ssh shell)
+		if os.Getenv("SSH_TTY") != "" {
+			return ssh
+		}
 		return browser
-	} else if os.Getenv("DEVBOX_REGION") != "" {
-		return ssh
-	} else {
-		return local
 	}
+	return local
 }
--- a/internal/telemetry/segment.go
+++ b/internal/telemetry/segment.go
@@ -4,14 +4,15 @@ import (
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"log"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/samber/lo"
 	segment "github.com/segmentio/analytics-go"
 	"go.jetpack.io/devbox/internal/build"
 	"go.jetpack.io/devbox/internal/cloud/openssh"
@@ -31,12 +32,13 @@ type Event struct {
 	UserID      string
 }
 
-// ttiEvent contains fields used to log the time-to-interactive event. For now,
-// this is used for devbox shell (local and cloud).
-type ttiEvent struct {
-	Event
-	eventName string
-}
+type shellAccessKind string
+
+const (
+	local   shellAccessKind = "local"
+	ssh     shellAccessKind = "ssh"
+	browser shellAccessKind = "browser"
+)
 
 // NewSegmentClient returns a client object to use for segment logging.
 // Callers are responsible for calling client.Close().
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #562** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/562" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/562?utm_source=stack-comment).

internal/telemetry/segment.go
```diff
@@ -96,7 +95,8 @@ func LogShellDurationEvent(eventName string, startTime string) error {
 	// Ignore errors, telemetry is best effort
 	_ = segmentClient.Enqueue(segment.Track{
 		AnonymousId: evt.AnonymousID,
-		Event:       evt.eventName,
+		// Event name. We trim the prefix from shell-interactive/shell-ready to avoid redundancy.
+		Event: fmt.Sprintf("[%s] Shell Event: %s", evt.AppName, strings.TrimPrefix(eventName, "shell-")),
```
loreto, line 99: Are you sure this is what we want?

Wouldn't this add "Shell Event" to *all* events including the "Command" events that already exist?

I would have thought instead that you need to change the string wherever the `shell-interactive` event is triggered.

internal/telemetry/segment.go
```diff
@@ -154,3 +154,13 @@ func timeFromUnixTimestamp(timestamp string) (time.Time, error) {
 func UnixTimestampFromTime(t time.Time) string {
 	return strconv.FormatInt(t.Unix(), 10)
 }
+
+func shellAccess() shellAccessKind {
+	if os.Getenv("START_WEB_TERMINAL") == "1" {
```
loreto, line 159: OK, but ... Could you add a TODO/NOTE of some kind? At some point we'll have VMs with Gotty started that can be accessed via ssh *or* web, and this would treat them all as web. Really what we care about is the connection, that is running the command, not whether the VM has Gotty running or not.

internal/telemetry/segment.go
savil: no, those command events are still logged in `midcobra/telemetry.go`.
I have a cleanup I want to do to unify these further, but have held back since its a bit involved.

internal/telemetry/segment.go
savil: sure I can add a comment...

savil: If you look at how `START_WEB_TERMINAL` works, its purpose _is_ to ensure that we start the VM for browser shells. Whenever that logic changes, we'll need to update this as well.

@LucilleH suggested using this. Is there another more persistent way to detect that the shell is running as a browser shell?

loreto: I think this is the right behavior for now – I'm just saying we'll need to improve it and I like adding a TODO so it doesn't slip our minds.

savil: The other approach is to rely on an SSH env var like `SSH_USER_PUBLIC_KEY` so:

```
// Check if in cloud
if os.Getenv("DEVBOX_REGION") != "" {
   if os.Getenv("SSH_USER_PUBLIC_KEY") != "" {
       return ssh
   } else {
       return browser
    }
}
return local
```

(need to test this)

loreto (approved):
LGTM

internal/telemetry/segment.go
savil: I can use `SSH_TTY`  

I dumped `env` for both ssh and browser shells to compare:
```
❯ cat web_env.txt | grep SSH
SSH_HOST_KEY=
DEFAULT_SSH_USER=devbox
SSH_USER=savil
SSH_USER_PUBLIC_KEY=
SSH_LISTEN_IPV4=0.0.0.0
SSH_LISTEN_IPV6=::
SSH_PORT=2222

❯ cat ssh_env.txt | grep SSH
SSH_CONNECTION=fdaa:0:a780:a7b:b2e2:41a9:b82f:2 58826 fdaa:0:a780:a7b:a1:ef3c:55b5:2 2222
SSH_CLIENT=fdaa:0:a780:a7b:b2e2:41a9:b82f:2 58826 2222
SSH_TTY=/dev/pts/0
```

LucilleH: I think for now this is fine to use `START_WEB_TERMINAL` 👍

LucilleH (approved):

internal/telemetry/segment.go
```diff
@@ -154,3 +154,15 @@ func timeFromUnixTimestamp(timestamp string) (time.Time, error) {
 func UnixTimestampFromTime(t time.Time) string {
 	return strconv.FormatInt(t.Unix(), 10)
 }
+
+func shellAccess() shellAccessKind {
+	// Check if running in devbox cloud
+	if os.Getenv("DEVBOX_REGION") != "" {
+		// Check if running via ssh tty (i.e. ssh shell)
+		if os.Getenv("SSH_TTY") != "" {
```
savil, line 162: @loreto I think this approach is closer to what you wanted where we test for `ssh` versus `browser` based on access methodology.

loreto: Sounds good! Just make sure you test it  :)

savil: I verified that SSH_TTY isn't present in browser shell: 
https://github.com/jetpack-io/devbox/pull/562#discussion_r1094983619

PR #563, opened 2023-01-31
[CICD] Only run mac tests on main

mikeland86:
## Summary

mac tests are really slow (failing because they hit timeout). This increases the timeout to 20 minutes but only runs it on main.

## How was it tested?

CICD


loreto (approved):
LGTM

ipince (approved):

PR #572, opened 2023-02-02
[Test] Shellspec for end to end tests

mohsenari:
## Summary

- [x] Added shellspec basic test
- [x] Updated github actions to run the tests in bash, zsh, ksh

## How was it tested?
N/A


.github/workflows/cli-tests.yaml
```diff
@@ -77,14 +77,19 @@ jobs:
...
-      - name: Install additional shells (dash, zsh)
+      - name: Install additional tools and shells (dash, zsh)
         run: |
           if [ "$RUNNER_OS" == "Linux" ]; then
             sudo apt-get update
             sudo apt-get install dash zsh
+            sudo curl -sLo /home/shellspec.tar.gz https://github.com/shellspec/shellspec/releases/download/0.28.1/shellspec-dist.tar.gz
+            tar xvzf /home/shellspec.tar.gz
+            sudo mv shellspec/ /home/
+            sudo ln -s /home/shellspec/shellspec /usr/local/bin/
```
mohsenari, line 88: This is manual installation for shellspec on ubuntu since it's not in apt-get or snap package managers. I'm open to make this its own step

.github/workflows/cli-tests.yaml
```diff
@@ -95,8 +100,14 @@ jobs:
...
           echo "PATH=$PATH" >> $GITHUB_ENV
-      - name: Run tests
+      - name: Run go tests
         run: go test -race -cover ./...
+      - name: Run shellspec tests (bash)
+        run: shellspec testscripts/shellspec/ -s /bin/bash -f document --no-warning-as-failure
+      - name: Run shellspec tests (zsh)
+        run: shellspec testscripts/shellspec/ -s /bin/zsh -f document --no-warning-as-failure
+      - name: Run shellspec tests (dash)
+        run: shellspec testscripts/shellspec/ -s /bin/dash -f document --no-warning-as-failure
```
mohsenari, line 110: These tests run on every branch now. I'm open to change it so that it only runs on main.

.github/workflows/cli-tests.yaml
mikeland86: I think putting in own step has a lot of value. It let's us see how long the step takes more and spits up the different parts of the setup.

.github/workflows/cli-tests.yaml
mikeland86: I think on every branch is cool. It's much better to get feedback during PR than after merge.

testscripts/shellspec/spec/devbox_spec.sh
```diff
@@ -0,0 +1,32 @@
...
+    Before 'export DEVBOX_FEATURE_STRICT_RUN=1'
+    It 'Runs Hello'
+        When run devbox run hello
+        The stderr should include 'Ensuring packages are installed.'
+        The status should equal '0'
+        The stdout should equal 'Hello, world!'
+    End
+    It 'Removes package'
+        When run devbox rm hello
+        The stderr should include 'hello (hello-2.12.1) is now removed.'
```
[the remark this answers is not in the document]
mikeland86: You can fix this by creating a deterministic devbox.json (e.g. `echo "{...}" > devbox.json`)

.github/workflows/cli-tests.yaml
mikeland86, line 110: I'm not sure using `--no-warning-as-failure` is good (unless it's impossible to get around it). Right now some of the tests are failing because the output is to stderr is not checked. I think it's better to check it (ideally in a flexible way were it wont break if we change it). That way we can remove this flag and if stderr actually shows up we can catch it.

mikeland86: If performance of the test becomes an issue, we can split this test out and run it in parallel.

mikeland86 (commented):
I'm not super familiar with this tool, but looks cool! Is the main difference between this and what @loreto was trying out that this can test on multiple shells?

testscripts/shellspec/spec/devbox_spec.sh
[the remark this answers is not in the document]
mohsenari: I wanted to follow all the steps including devbox.json being generated by `devbox init`. 
I changed the matching method so that it won't rely on the version number.

mohsenari:
Closing this due to taking a different route for testing.

PR #574, opened 2023-02-02
[RFC] make Telemetry UserID be a uuid string

savil:
## Summary

**Motivation**:
From inspecting the segment logs, I am reminded that the userID is logged as a 64 char.

Segment says that ideally this userID is something we track in our main database. Longer term, I'm not sure we'd want our primary userIDs to be TEXT kind in Postgres data type terms, compared to smaller uuid or int variations.

**Approach**:
This PR amends the `UserIDFromGithubUsername` function to __deterministically__
return a UUID string (or empty string if not authenticated or error occurs).

It is important that the same UUID is returned for the same github username.

**RFC: Breaking Change**
Tradeoff: in Segment we'll not be able to track the same users' activity prior to this PR.

## How was it tested?

copied this function into a devbox-project for golang and executed it multiple times
with different usernames.

### DIFF
--- a/internal/telemetry/segment.go
+++ b/internal/telemetry/segment.go
@@ -1,9 +1,6 @@
 package telemetry
 
 import (
-	"crypto/hmac"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #574** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/574" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/574?utm_source=stack-comment).

internal/telemetry/segment.go
```diff
@@ -117,21 +118,34 @@ func LogShellDurationEvent(eventName string, startTime string) error {
...
 
-	return hex.EncodeToString(mac.Sum(nil))
+		const githubPrefix = "github:"
+		mac.Write([]byte(githubPrefix + username))
+
+		return hex.EncodeToString(mac.Sum(nil))
+	}
+
+	hash := hashFromUsername(username)
+	uid, err := uuid.FromBytes([]byte(hash)[:16])
```
gcurtis, line 144: I don't think this will create a UUID with a valid version/variant since all of the bytes are random. This might not matter for Sentry though (I'm not sure why they want a UUID specifically). `uuid.NewHash` might be a better way of hashing the username into a UUID.

internal/telemetry/segment.go
```diff
@@ -117,21 +118,34 @@ func LogShellDurationEvent(eventName string, startTime string) error {
...
-	const githubPrefix = "github:"
-	mac.Write([]byte(githubPrefix + username))
+		const salt = "d6134cd5-347d-4b7c-a2d0-295c0f677948"
+		mac := hmac.New(sha256.New, []byte(salt))
 
-	return hex.EncodeToString(mac.Sum(nil))
+		const githubPrefix = "github:"
+		mac.Write([]byte(githubPrefix + username))
+
+		return hex.EncodeToString(mac.Sum(nil))
```
gcurtis, line 140: You could not encode the hash to hex and use the byte slice directly.

gcurtis (approved):

internal/telemetry/segment.go
savil: Segment (not sentry) doesn't ask for a UUID. They do recommend choosing a `UserID` that will be stable for the same user, so we can track the same user over time.

Prior to this PR, I was using just the raw 64-char hash. But I felt that would be too large, and unstructured in case we choose to use that value later on in some internal database (hypothetically). So, I am exploring using a uuid type instead.

Thanks for pointing out the problem about the invalid version/variant. Will look into that.

PR #576, opened 2023-02-02
Add more metadata to error reports

ipince:
## Summary

Adds:
- The nixpkgs commit hash in `devbox.json`.
- Whether the user was inside a devbox shell or not.
- The Devbox-related env vars (those that start with `DEVBOX_`). These help us see which experimental features were enabled, if any.

## How was it tested?
Generated events, as such:
```
DO_NOT_TRACK=0 devbox run invalid-command
```
Checked [result](https://jetpack-io.sentry.io/issues/3900011083/events/ceddfdbf8b0649158d3df61cc73bb0fe/?project=4504249577242624).


internal/boxcli/midcobra/telemetry.go
```diff
@@ -144,11 +102,19 @@ func (m *telemetryMiddleware) newEventIfValid(cmd *cobra.Command, args []string,
...
+	pkgs, hash := getPackagesAndCommitHash(cmd)
 
 	// an empty userID means that we do not have a github username saved
 	userID := telemetry.UserIDFromGithubUsername()
 
+	devboxEnv := map[string]interface{}{}
+	for _, e := range os.Environ() {
+		if strings.HasPrefix(e, "DEVBOX") && strings.Contains(e, "=") {
+			key := strings.Split(e, "=")[0]
+			devboxEnv[key] = os.Getenv(key)
```
savil, line 114: instead of another call to Getenv, why not do:
```
parts := strings.Split(e, "=")
key = parts[0]
devboxEnv[key] = parts[1]
```
?

internal/boxcli/midcobra/telemetry.go
```diff
@@ -217,3 +206,32 @@ func (m *telemetryMiddleware) trackEvent(evt *event) {
...
+	}
+	return subcmd, subargs, err
+}
+
+func getPackagesAndCommitHash(c *cobra.Command) ([]string, string) {
+	configFlag := c.Flag("config")
+	// for shell, run, and add command, path can be set via --config
+	// if --config is not set, default to current directory which is ""
+	// the only exception is the init command, for the path can be set with args
+	// since after running init there will be no packages set in devbox.json
```
savil, line 224: unless we run `devbox init` on a project that already has a devbox.json (we do this for the gotty-vm, see start-gotty.sh in axiom) and that fails....

savil (approved):
nice!

savil:
the reordering is to keep in alphabetical order, or some other metric?

internal/boxcli/midcobra/telemetry.go
ipince: Because I actually don't know if `Split()` will return an array of size 2. it may return a larger array if the value of the variable happens to have a `=` in it...
It's not expensive to call `Getenv()` (I think), so it seems safer to just call it

ipince:
> the reordering is to keep in alphabetical order, or some other metric?

To make it more top-down readable... I like to be able to read a file from the top down, meaning the most important functions are up top, and small helper functions are on the bottom.

I feel like this is exactly the opposite of how you like to order functions though, lol. And we have no guideline about it, so it's a bit arbitrary I'll admit

internal/boxcli/midcobra/telemetry.go
ipince: hmm i see.. well I didn't change anything here, just moved it (sorry!)

PR #577, opened 2023-02-02
fix golang testdata

savil:
## Summary

These gave errors like:
```
expected version: go1.17, got: go1.17.13
```
when doing `go run main.go` in their respective `devbox shell`

## How was it tested?

opened a `devbox shell` and did `go run main.go` and got the version printed out without any panic

### DIFF
--- a/examples/testdata/go/go-1.17/main.go
+++ b/examples/testdata/go/go-1.17/main.go
@@ -3,13 +3,14 @@ package main
 import (
 	"fmt"
 	"runtime"
+	"strings"
 )
 
 func main() {
 	expected := "go1.17"
 	goVersion := runtime.Version()
 	fmt.Printf("Go version: %s\n", goVersion)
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #577** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/577" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/577?utm_source=stack-comment).

ipince (approved):
How come the tests were passing then? are these just manually run tests?

savil:
I think we don't execute them :(

the new testscripts tests should catch such issues, and we should hook these up

mikeland86:
Don't these examples all have `devbox.json` files? The version should be deterministic! We should be able to specify the exact version (and not just prefix)

PR #586, opened 2023-02-06
[cloud] Auto-port forward during cloud shell

mikeland86:
## Summary

This PR enables automatic port forwarding when services are started in devbox cloud. It does this by maintaining synced "status" files in `.devbox.cloud/[host-id]`

How to review this PR:

* `cloud.AutoPortForward` is a background thread that listens to changes in `.devbox.cloud/[host-id]` and starts port forwarding as needed. It only runs locally.
* `services.ListenToChanges` listens to changes in `.devbox.cloud/[host-id]` and calls a callback when an event occurs. The path of a service status file is `.devbox.cloud/[host-id]/[service-name].json`. It was modified to listen to statuses of individual services. I did this because having a single `services.json` could lead to file edit race conditions when starting/stoping multiple services.
* `services.listenToAutoPortForwardingChangesOnRemote` only runs on the cloud. It's the complement to `cloud.AutoPortForward`. It listens for when port forwarding has already been set up so it can print out the local ports on the cloud shell. It waits for 5 seconds and then gives up.

## How was it tested?

* Added nginx and apache.
* Started cloud shell. Copied over newest devbox binary.
* Ran `devbox services start/stop`  in several combinations and observed that port forwarding started and stopped as expected.

image

image


internal/services/services.go
```diff
@@ -77,5 +89,73 @@ func toggleServices(
...
+	}
+	hostname, err := hostname()
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	fmt.Fprintf(w, "Waiting for port forwarding to start/stop for service %q\n", serviceName)
+	waitGroup.Add(1)
+
+	ctx, cancel := context.WithCancel(ctx)
```
gcurtis, line 117: Can the cancel of this context be deferred? It ultimately needs to be called everywhere otherwise it'll leak resources.

internal/services/services.go
```diff
@@ -77,5 +89,73 @@ func toggleServices(
...
+		defer m.Unlock()
+		if ctx.Err() == nil {
+			cancel()
+			waitGroup.Done()
+		}
+	}
+
+	// After 5 seconds, if the context has not been canceled, go ahead and cancel it
+	// and also decrement the wait group so that the caller can continue.
+	time.AfterFunc(5*time.Second, func() {
```
gcurtis, line 133: Can this be done with `context.WithTimeout`?

internal/services/services.go
```diff
@@ -77,5 +89,73 @@ func toggleServices(
...
+	}
+
+	fmt.Fprintf(w, "Waiting for port forwarding to start/stop for service %q\n", serviceName)
+	waitGroup.Add(1)
+
+	ctx, cancel := context.WithCancel(ctx)
+
+	// Done function is thread safe and will cancel the context and decrement the wait group
+	// if context is already canceled, it will not do anything
+	var m sync.Mutex
```
gcurtis, line 121: Could this be a `sync.Once` instead of a mutex?

gcurtis (approved):

internal/boxcli/cloud.go
```diff
@@ -152,5 +125,5 @@ func runCloudShellCmd(cmd *cobra.Command, flags *cloudShellCmdFlags) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	return cloud.Shell(cmd.ErrOrStderr(), box.ProjectDir(), flags.githubUsername)
+	return cloud.Shell(cmd.Context(), cmd.ErrOrStderr(), box.ProjectDir(), flags.githubUsername)
```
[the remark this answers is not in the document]
mikeland86: It gets set when we execute cobra root command https://github.com/jetpack-io/devbox/blob/main/internal/boxcli/midcobra/midcobra.go#L64-L64

internal/services/services.go
mikeland86: Yes, it would need a timeout and also a cancel.

PR #584, opened 2023-02-04
[flakes | print-dev-env] enable flakes to work with print-dev-env

savil:
## Summary

This PR is a step towards enabling `devbox shell` to work with both Flakes and NixlessShell featureflags enabled.

Future PR TODOs: refer to notion doc on Post Devbox Cloud Clean up Tasks.

## How was it tested?

in `examples/testdata/go/go-1.19`:
```
> DEVBOX_FEATURE_FLAKES=1 DEVBOX_FEATURE_NIXLESS_SHELL=1 DEVBOX_DEBUG=1 devbox shell
(devbox) > go run main.go
Go version: go1.19.3

```

Also started shells with the following control cases:
1. `DEVBOX_FEATURE_FLAKES=0 DEVBOX_FEATURE_NIXLESS_SHELL=0 DEVBOX_DEBUG=1 devbox shell`
2. `DEVBOX_FEATURE_FLAKES=0 DEVBOX_FEATURE_NIXLESS_SHELL=1 DEVBOX_DEBUG=1 devbox shell`

For the control cases, did:
```
> devbox add vim
> which vim
> devbox rm vim
```
to verify that the package was added and removed

PR #591, opened 2023-02-07
Create unified function for nix env computation

ipince:
## Summary
This mainly refactors the env computation from `nix.Run()` into a place where it can be reused for both shell, run, and direnv integration. This should be a functional no-op.

## How was it tested?
`./devbox run -- echo '$PATH'` and other env vars


internal/impl/devbox.go
```diff
@@ -682,6 +681,47 @@ func (d *Devbox) printPackageUpdateMessage(
 	return nil
 }
 
+func (d *Devbox) computeNixEnv() ([]string, error) {
```
mikeland86, line 684: Should this be called in other places we need env? (`Shell` and `RunScriptInNewNixShell`)

savil: I believe that is the intention (in a future PR)

internal/impl/devbox.go
```diff
@@ -682,6 +681,47 @@ func (d *Devbox) printPackageUpdateMessage(
...
+	nixShellFilePath := filepath.Join(d.projectDir, ".devbox/gen/shell.nix")
+	vaf, err := nix.PrintDevEnv(nixShellFilePath)
```
savil, line 686: use: `d.nixShellFilePath()` but first please approve https://github.com/jetpack-io/devbox/pull/584 so it can be landed :)

savil (approved):

internal/impl/devbox.go
ipince: Correct!

internal/impl/devbox.go
ipince: oops, sorry for not seeing that earlier!  merging in main and updating...

PR #592, opened 2023-02-07
[Plugin] Add a plugin for `python3xxPackage.pip`

Lagoja:
## Summary

This plugin automatically sets up a virtual environment for python packages whenever a python3 version of `pip` is added.

- Sets a `VENV_DIR` environment variable that determines where the virtualenv is
- A `venvShellHook.sh` file is added in `devbox.d` which creates the venv directory at $VENV_DIR if it doesn't already exist
- The plugin's init hook runs the `venvShellHook.sh` script

I've left it so the user still needs to source the virtualenv manually, so they can choose whether or not to activate on starting devbox shell

## How was it tested?

Tested in devbox-examples, create a simple requirements.txt and ran it with the plugin

### DIFF
--- a/plugins/pip.json
+++ b/plugins/pip.json
@@ -1,0 +1,17 @@
+{
+    "name": "pip",
+    "version": "0.0.1",
+    "match": "^python3[0-9]{1,2}Packages.pip",
+    "readme": "This plugin adds a script for automatically creating a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `source $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
+    "env": {
+        "VENV_DIR": "{{ .Virtenv }}/pip/.venv"
+    },
+    "create_files": {
+        "{{ .DevboxDir }}/venvShellHook.sh": "pip/venvShellHook.sh"
+    },
+    "shell": {
+        "init_hook": [
+            "sh {{ .DevboxDir }}/venvShellHook.sh"
+        ]
+    }
+}
--- a/plugins/pip.json
+++ b/plugins/pip.json
@@ -4,14 +4,14 @@
     "match": "^python3[0-9]{1,2}Packages.pip",
     "readme": "This plugin adds a script for automatically creating a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `source $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
     "env": {
-        "VENV_DIR": "{{ .Virtenv }}/pip/.venv"
+        "VENV_DIR": "{{ .Virtenv }}/.venv"
     },
     "create_files": {
-        "{{ .DevboxDir }}/venvShellHook.sh": "pip/venvShellHook.sh"
+        "{{ .Virtenv }}/bin/venvShellHook.sh": "pip/venvShellHook.sh"
     },
     "shell": {
         "init_hook": [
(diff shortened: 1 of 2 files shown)


Lagoja:
Did this really quick because I got annoyed with setting up `venv` manually on Python examples

plugins/pip.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "pip",
+    "version": "0.0.1",
+    "match": "^python3[0-9]{1,2}Packages.pip",
+    "readme": "This plugin adds a script for automatically creating a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `source $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
+    "env": {
+        "VENV_DIR": "{{ .Virtenv }}/pip/.venv"
+    },
+    "create_files": {
+        "{{ .DevboxDir }}/venvShellHook.sh": "pip/venvShellHook.sh"
```
mikeland86, line 10: Do we want this in `devbox.d` (committed) vs `.devbox` (not-committed) ? Since the file is not meant to be modified, I would think `.devbox` (which is `{{ .Virtenv }}`).

plugins/pip.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "shell": {
+        "init_hook": [
+            "sh {{ .DevboxDir }}/venvShellHook.sh"
```
mikeland86, line 14: Is `sh` needed? I think you can just add `#!/bin/sh` to the file? If you put it into a `bin` directory it automatically gets run permissions https://github.com/jetpack-io/devbox/blob/landau/start-auto-forward-on-cloud/internal/plugin/pkgcfg.go#L89-L93

plugins/pip.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "pip",
+    "version": "0.0.1",
+    "match": "^python3[0-9]{1,2}Packages.pip",
+    "readme": "This plugin adds a script for automatically creating a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `source $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
+    "env": {
+        "VENV_DIR": "{{ .Virtenv }}/pip/.venv"
```
mikeland86, line 7: `pip` is not needed `Virtenv` is already specific to this package.  This can be `"VENV_DIR": "{{ .Virtenv }}/.venv"`

plugins/pip/venvShellHook.sh
```diff
@@ -0,0 +1,9 @@
+SOURCE_DATE_EPOCH=$(date +%s)
```
mikeland86, line 1: what's this?

mikeland86 (approved):

plugins/pip.json
Lagoja: I was thinking committed for visibility + if someone wants to modify the script, but could definitely go in `{{ .Virtenv }}`

plugins/pip/venvShellHook.sh
Lagoja: Based this on the Nix manual, but looking deeper i think this is only needed for python2. I'll remove it

plugins/pip.json
Lagoja: So I could do something like: 

```json
    "shell": {
        "init_hook": [
            "{{ .DevboxDir }}/bin/venvShellHook.sh"
        ]
    }
```
?

plugins/pip.json
Lagoja: I think i like this better in {{ .Virtenv }}. If we get a request we can move it to devbox.d or let someone define their own script.

PR #594, opened 2023-02-08
[easy][plugins] User init hooks should run after plugins

mikeland86:
## Summary

Change the order of init hooks to ensure user hooks are run last. Also ensure plugin hooks are run from the same directory as user hooks (the project dir)

## How was it tested?

```
devbox add ruby
# modified devbox.json to echo $PATH in init hook
devbox shell
# inspected PATH and saw GEM_HOME was in it.
```


mohsenari (approved):

mohsenari:
@mikeland86 the tests are failing because now the shellrc template doesn't match the one in test. We should update the test as well before merging.

PR #596, opened 2023-02-08
[devbox.json] Added env field in devbox.json

mohsenari:
## Summary
Added an `env` field in devbox.json inside shell object since it's related to shell environment.
Right now the env field only supports literal values (e.g., `"key"="value"`) and doesn't process shell commands (e.g., `"key"="$(ls -a)"`). This feature will be added in a future PR.

Also adding the empty `env` field in `devbox init` will be added in future PR.

## How was it tested?
add the `devbox.json` below in an empty directory.
run `devbox shell` from inside shell `echo $FOO` and confirm the env variable's value is set
or 
run `devbox run test`

```json
{
  "packages": [],
  "shell": {
    "env": {
      "FOO": "BAR"
    },
    "init_hook": [],
    "scripts": {
      "test": "echo $FOO"
    }
  },
  "nixpkgs": {
    "commit": "52e3e80afff4b16ccb7c52e9f0f5220552f03d04"
  }
}
```

### DIFF
--- a/internal/impl/config.go
+++ b/internal/impl/config.go
@@ -25,7 +25,10 @@ type Config struct {
 
 	// Shell configures the devbox shell environment.
 	Shell struct {
+		// Env contains environment variables that will be set in shell
 		// InitHook contains commands that will run at shell startup.
+		// Scripts contain commands that can be executed via devbox run
+		Env      map[string]string             `json:"env,omitempty"`
 		InitHook shellcmd.Commands             `json:"init_hook,omitempty"`
 		Scripts  map[string]*shellcmd.Commands `json:"scripts,omitempty"`
 	} `json:"shell,omitempty"`
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -238,12 +238,10 @@ func (d *Devbox) Shell() error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(env)
 	env, err = d.appendConfigEnv(env)
 	if err != nil {
 		return err
 	}
-	fmt.Println(env)
 
 	shellStartTime := os.Getenv("DEVBOX_SHELL_START_TIME")
 	if shellStartTime == "" {
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -238,10 +238,7 @@ func (d *Devbox) Shell() error {
 	if err != nil {
 		return err
 	}
-	env, err = d.appendConfigEnv(env)
-	if err != nil {
-		return err
-	}
+	env = d.appendConfigEnv(env)
 
 	shellStartTime := os.Getenv("DEVBOX_SHELL_START_TIME")
 	if shellStartTime == "" {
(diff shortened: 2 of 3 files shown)


internal/impl/config.go
```diff
@@ -25,7 +25,10 @@ type Config struct {
 
 	// Shell configures the devbox shell environment.
 	Shell struct {
+		// Env contains environment variables that will be set in shell
```
mikeland86, line 28: Add each comment above the field it describes

Lagoja:
Does this currently create a breaking change? Or is it just an optional change for now?

internal/impl/devbox.go
```diff
@@ -722,6 +725,15 @@ func (d *Devbox) computeNixEnv() ([]string, error) {
...
 
+// addConfigEnv takes env variables that are to be set (plugin Env vars)
+// and appends env variables defined in devbox.json to it.
+func (d *Devbox) appendConfigEnv(env []string) []string {
+	for key, value := range d.cfg.Shell.Env {
+		env = append(env, fmt.Sprintf("%s=%s", key, value))
+	}
+	return env
+}
+
```
mikeland86, line 736: Making *Devbox the receiver here is a bit weird. While it does use a field (`d.cfg.Shell.Env`) the behavior isn't completely obvious to me. I also think if we find ourselves repeating the same logic (get plugin env, get user env) then probably they should live in same function.

Could you instead put this logic into `computeNixEnv` ? 

Side note, I'm not sure if @ipince 's PR is ready, I only see `computeNixEnv` in one place but not in others.

mikeland86 (changes requested):
Is @ipince code ready? I've usually only stacked on my own PRs (since I have a lot of context about how ready they are and when they will get merged). I added some comments there because I think if `computeNixEnv` gets called everywhere then this logic should almost certainly live in that function.

internal/impl/devbox.go
mohsenari: `computNixEnv` is used only for devbox run. I spoke with @ipince and he's working consolidating all the env computation and implementations behind feature flags into just one function that is `computeNixEnv` I will then stack my PR on those changes so that the logic can live in `computeNixEnv`

mohsenari:
> Does this currently create a breaking change? Or is it just an optional change for now?

@Lagoja I'm adding a feature flag and will put this behind it. I can also make sure that it is backwards compatible so that if `env` field is not present in devbox.json, it won't cause an interruption/error.

internal/impl/config.go
mohsenari: I'll change it. The reason for this comment re-organizing was that go-lint was automatically format this and make it so that `json:"xyz,omitempty"` was misaligned see comparison below:
```go
Shell struct {
		// Env contains environment variables that will be set in shell
		// InitHook contains commands that will run at shell startup.
		// Scripts contain commands that can be executed via devbox run
		Env      map[string]string             `json:"env,omitempty"`
		InitHook shellcmd.Commands             `json:"init_hook,omitempty"`
		Scripts  map[string]*shellcmd.Commands `json:"scripts,omitempty"`
	} `json:"shell,omitempty"`
```
vs
```go
Shell struct {
		// Env contains environment variables that will be set in shell
		Env map[string]string `json:"env,omitempty"`
		// InitHook contains commands that will run at shell startup.
		InitHook shellcmd.Commands `json:"init_hook,omitempty"`
		// Scripts contain commands that can be executed via devbox run
		Scripts map[string]*shellcmd.Commands `json:"scripts,omitempty"`
	} `json:"shell,omitempty"`
```

mohsenari:
closing this in favor of #609

PR #599, opened 2023-02-08
[global] MVP of devbox global/home

mikeland86:
## Summary

This is MVP of devbox global/home. It uses the default nix profile. Since the default nix profile is already in path, we don't need anything different for that. This implements:

```bash
devbox global add <pkg>...
devbox global rm <pkg>...
devbox global list
```

A few caveats:
* This approach doesn't match devbox shell, but is simpler to implement and doesn't require making their user change their rc files. 
* This won't run setup-hooks.
* This causes anything in nix default profile to "leak" into devbox project shells. (we could fix this by using non-default profile, but I'm not sure this makes sense).
* The `global` command is hidden.

I think this approach certainly makes trade-offs, but gets most of the benefits fast and with a pretty nice UX. The setup-hooks issue is something we could fix later on.

Thoughts?

## How was it tested?

image


loreto (approved):
LGTM

I like the idea of implementing something quick and getting out in the next release so we can start marketing it and getting user feedback.

We can always change the internals later.

What would it take to make the command public?

gcurtis (commented):
I have a few questions/concerns about using the default Nix profile:

- We'll need to document the differences between this and a regular devbox shell. For example, `devbox global add python3Packages.*` won't work, but `devbox add python3Packages.*` will.
- Modifying the user's default profile seems a little risky. Nix users might already have their profile setup a certain way and we'd be overwriting that.
- There might be certain programs that shouldn't be added to the PATH. Homebrew has a concept of "keg-only" packages that don't get added because replacing the system-default could cause problems. For example, with `curl`:

> curl is keg-only, which means it was not symlinked into /opt/homebrew, because macOS already provides this software and installing another version in parallel can cause all kinds of trouble.

mikeland86:
@gcurtis, answers inline:

> We'll need to document the differences between this and a regular devbox shell
yeah, I think this is biggest drawback to this implementation. Could we show a warning/disclaimer?

FWIF, if you do 

```
devbox global add python3
python3 -m venv .venv
. ./.venv/bin/activate
pip install numpy
```

It works as expected. I'm more concerned about setup hooks on packages like compilers which could affect a larger and unknown set of packages.

> Modifying the user's default profile seems a little risky

I'm mixed on this. The big advantage is that it's already in path. It also doesn't affect non-nix users at all. What if we put some safeguards (e.g. `devbox global rm` wont remove packages that were not installed with devbox)?

> There might be certain programs that shouldn't be added to the PATH

I think eventually we can customize this with plugins, but right now the behavior will match nix right?

mikeland86:
@loreto 

> What would it take to make the command public?

* Fix the TODOs in code (handle failures when adding/removing multiple packages)
* Show a message telling the user to run `hash -r` in case the binary their installing was already in path so we need to reload the hashmap.
* Don't remove packages that were not installed by devbox

I'll do these in a stacked PR

mikeland86:
@gcurtis I reverted the change that excludes the default profile from PATH. This makes this PR essentially zero risk it doesn't modify existing (non-hidden) functionality at all.

My plan is too add a follow up PR that merged the global devbox.json when doing `devbox shell/run etc`. This will ensure global devbox installed packages are usable in shell (including setup hooks)

gcurtis (approved):

PR #600, opened 2023-02-08
remove unused pkgs.go from internal/impl

savil:
## Summary

In #597, I wanted to introduce a file called `packages.go` (currently, called install.go but that's a poor substitute IMO).
This file `pkgs.go` would be confusing to have at the same time.

Since it is unused, lets remove it. It can be brought back whenever needed.

## How was it tested?

compiles


savil:
* main
  * **PR #600** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/600" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/600?utm_source=stack-comment).

mikeland86 (approved):

PR #604, opened 2023-02-09
[home] Merge global packages when doing shell/run

mikeland86:
## Summary

This merges global packages into local devbox.json whenever `shell` and `run` are called.

This is behind a feature flag so no change is expected.

For cloud, we can leverage `.devbox.cloud` to sync global config and have the cloud vm use merge that. (TODO in future PR)

## How was it tested?

```bash
➜ devbox global add hello
➜ DEVBOX_FEATUR_HOME=1 devbox shell
➜ hello
Hello, world!
```


internal/impl/config.go
```diff
@@ -43,6 +48,29 @@ type Stage struct {
...
+	}
+	if c.Nixpkgs.Commit != global.Nixpkgs.Commit && !commitMismatchErrorShown {
+		commitMismatchErrorShown = true
+		color.New(color.FgHiYellow).Fprint(w, "Warning: ")
+		fmt.Fprintln(
+			w,
+			"local and global devbox.json have different nixpkgs commits. "+
+				"Will use the local version. This may lead to version mismatch and "+
+				"nix store bloat.")
+	}
```
gcurtis, line 70: The odds of hitting this warning sound pretty high. For example, if we update the default hash in devbox and then all new projects end up mismatching the global devbox.json. I don't think there's an obvious way to update the commit hash in the global debox.json either, especially if the user doesn't know where it is.

Instead, could we use the global nixpkgs commit hash for global packages and the project commit hash for project packages? That sounds pretty reasonable and then we wouldn't have to show a warning.

gcurtis (approved):

loreto (commented):
For a first pass ... do we need to do this merge? Our shells are not "pure" by default, so if `global add` adds things to your global profile, and that's already in your PATH (it would be if you want to use those tools). Then the tools will already be available on `shell/run`.

That's what I do today with tools I install via homebrew

loreto:
For cloud: I'm unsure whether we should merge the file locally and then file sync it.

A different approach:
+ Wait until we implement `devbox global push` (which saves your global environment in the cloud)
+ When we start your VM we know what your global environment is, since you've pushed it; and we use it.

internal/impl/config.go
```diff
@@ -43,6 +48,29 @@ type Stage struct {
 	Command string `cue:"string" json:"command"`
 }
 
+var commitMismatchErrorShown = false
+
+func (c *Config) Packages(w io.Writer) []string {
+	if featureflag.Home.Disabled() {
+		return c.RawPackages
+	}
+	path, _ := GlobalConfigPath()
```
savil, line 57: need to handle the error

internal/impl/config.go
```diff
@@ -43,6 +48,29 @@ type Stage struct {
 	Command string `cue:"string" json:"command"`
 }
 
+var commitMismatchErrorShown = false
```
savil, line 51: suggest: rename to `commitMismatchWarningShown`

savil (approved):
nice!

internal/impl/config.go
mikeland86: @gcurtis I want to do that, but need to wait for flakes stuff to be in. We can modify the flakes file to use two different nixpkgs. Though even in that case, I would want to show some sort of warning?

mikeland86:
@loreto not quite. We purposefully exclude the default nix profile from PATH. I had removed that in an earlier PR but then reverted.

My plan for cloud is not to merge the files, instead we sync a special hidden "global" file.

loreto:
I guess I still wonder if:
- Instead of using the default nix profile, we have a default devbox profile (in the other PR under discussion)
- We purposely include the devbox profile in the path. No merging of files needed.
- For cloud, I'm saying don't even sync a hidden `.cloud` file. We'll already have their global `devbox.json` from a `devbox global push`; on an S3 bucket or something like that, and we can use that from the server side when creating the VM. If you do it through the client, then it won't work for VMs started on the web browser.

mikeland86:
@loreto another downside of not merging is we don't get plugin behavior either.

mikeland86:
Chatted with @loreto offline. Will merge this as is but we may change merge implementation later on.

PR #607, opened 2023-02-09
Add Devbox Instructions to CONTRIBUTING.md

Lagoja:
## Summary

Update the Contributing Guide to describe the Devbox workflow

## How was it tested?

Localhost + Lint

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,35 +37,17 @@ devbox run build-linux
 devbox run lint
 ```
 
+* Run and Test the Devbox CLI:
+
+```bash
+./dist/devbox <test_command>
+```
+
 * For the best experience working on Devbox with VSCode, we recommend running:
 
 ```bash
 devbox shell -- code
 ```
-
-## Building and Testing
-
-Devbox is setup like a typical Go project. After installing the required tools and setting up your environment. You can make changes in the source code, build, and test your changes by following these steps:
-
-1. Install dependencies:
-
-    ```bash
-    go install
-    ```
-
-2. Build Devbox:
-
-    ```bash
-    go build -o ./devbox cmd/devbox/main.go
-    ```
-
-    This will build an executable file.
-
-3. Run and test Devbox:
-
-    ```bash
-    ./devbox <your_test_command>
-    ```
 
 ### Setting up the Environment Without Devbox
 
@@ -102,7 +84,7 @@ If you are unable to install or use Devbox, you can manually replicate the envir
 * Run and test Devbox:
 
     ```bash
-    ./devbox <your_test_command>
+    ./dist/devbox <your_test_command>
     ```
 
 ## Pull Request Process
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ Before making any changes to the source code (documentation excluded) make sure 
 
 ### With Devbox
 
-The easiest way to setup and build devbox is with Devbox!
+The easiest way to develop Devbox is with Devbox!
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -5,56 +5,107 @@ When contributing to this repository, please describe the change you wish to mak
...
 
 ## Setting Up Development Environment
-Before making any changes to the source code (documentation excluded) make sure you have installed all the required tools. 
-### Prerequisites
+
+Before making any changes to the source code (documentation excluded) make sure you have installed all the required tools.
+
+### With Devbox
+
+The easiest way to setup and build devbox is with Devbox!
```
mohsenari, line 13: > build devbox is with Devbox!
nit pick: devbox should be capitalized everywhere

CONTRIBUTING.md
```diff
@@ -5,56 +5,107 @@ When contributing to this repository, please describe the change you wish to mak
...
+devbox run build-linux
+
+# Run the CI Linter
+devbox run lint
+```
+
+* Run and Test the Devbox CLI:
+
+```bash
+./dist/devbox <test_command>
```
mohsenari, line 43: I think `devbox run build` prepends the binary (./dist/devbox) to $PATH so just running a devbox command would use the build binary

CONTRIBUTING.md
```diff
@@ -5,56 +5,107 @@ When contributing to this repository, please describe the change you wish to mak
...
+```
+
+* For the best experience working on Devbox with VSCode, we recommend running:
+
+```bash
+devbox shell -- code
```
mohsenari, line 49: We may need to change this to `devbox run code` after the feature flag for devbox run is removed.

CONTRIBUTING.md
```diff
@@ -5,56 +5,107 @@ When contributing to this repository, please describe the change you wish to mak
...
+* Install Devbox using the command below. If you don't have Nix installed, Devbox will automatically install it for you when you run a command.
+
+```bash
+curl -fsSL https://get.jetpack.io/devbox | bash
+```
+
+* Start a Devbox Shell for the project
+
+```bash
+devbox shell
```
mohsenari, line 24: we don't really need this here. It's confusing users thinking they have to be in a devbox shell to run the build commands below

mohsenari (approved):
approved for unblocking. See a couple of comments

CONTRIBUTING.md
Lagoja: Sounsd good, is the command right for the current build?

CONTRIBUTING.md
Lagoja: Would that change perisist if you're outside of a devbox shell?

mohsenari: @Lagoja right! it wouldn't because updating the path is in the init_hook. So ignore this comment. Running `dist/devbox` is a safer option.

CONTRIBUTING.md
mohsenari: Yes, it's just behind a feature flag

Lagoja: Actually, this is tricky because it won't work both inside or outside the shell. Maybe the easiest option is to add a script to our devbox.json

CONTRIBUTING.md
Lagoja: Edited a little bit, but wanted to have that command available somewhere

devbox.json
```diff
@@ -11,6 +11,7 @@
     "scripts": {
       "build": "go build -o dist/devbox cmd/devbox/main.go",
       "build-linux": "GOOS=linux go build -o dist/devbox-linux cmd/devbox/main.go",
+      "code": "code .",
```
mikeland86, line 14: did this sneak in by mistake?

PR #609, opened 2023-02-09
[devbox.json] Added env in config

mohsenari:
## Summary
Added ability to add env variables in devbox.json
Current implementation only supports literal string values as well as `$PATH` and `$PWD` variables.

## How was it tested?
1. Create a devbox.json similar to below:
```json
"shell": {
    "env": {
      "my_wd": "$PWD/test",
      "my_path": "$PATH:my_stuff/bin",
      "my_key": "string value"
    },
```
2. Run `DEVBOX_FEATURE_NIXLESS_SHELL=1 DEVBOX_FEATURE_ENV_CONFIG=1 ./devbox shell`

3. Inside devbox shell `echo $my_wd`, `echo $my_path` and `echo $my_key` and confirm values are correctly setup.

### DIFF
--- a/internal/impl/config.go
+++ b/internal/impl/config.go
@@ -29,6 +29,7 @@ type Config struct {
 
 	// Shell configures the devbox shell environment.
 	Shell struct {
+		Env map[string]string `json:"env,omitempty"`
 		// InitHook contains commands that will run at shell startup.
 		InitHook shellcmd.Commands             `json:"init_hook,omitempty"`
 		Scripts  map[string]*shellcmd.Commands `json:"scripts,omitempty"`
--- a/internal/impl/config.go
+++ b/internal/impl/config.go
@@ -27,9 +27,10 @@ type Config struct {
 	// It's differentiated from Packages() which also includes global packages.
 	RawPackages []string `cue:"[...string]" json:"packages"`
 
+	// Env allows specifying env variables
+	Env map[string]string `json:"env,omitempty"`
 	// Shell configures the devbox shell environment.
 	Shell struct {
-		Env map[string]string `json:"env,omitempty"`
 		// InitHook contains commands that will run at shell startup.
 		InitHook shellcmd.Commands             `json:"init_hook,omitempty"`
 		Scripts  map[string]*shellcmd.Commands `json:"scripts,omitempty"`
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -254,14 +254,6 @@ func (d *Devbox) Shell() error {
 		if err != nil {
 			return err
 		}
-	} else if featureflag.EnvConfig.Enabled() {
-		// TODO: this else-if can be removed when UnifiedEnv featureflag is
-		// removed. Since this logic already exists in computeNixEnv()
-
-		// Add env variables from config
-		for k, v := range d.configEnvs(env) {
-			env[k] = v
-		}
 	}
 
 	shellStartTime := os.Getenv("DEVBOX_SHELL_START_TIME")
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -249,6 +249,8 @@ func (d *Devbox) Shell() error {
 		if err != nil {
 			return err
 		}
+	} else {
+		env = append(env, d.getConfigEnvs(env)...)
 	}
 
 	shellStartTime := os.Getenv("DEVBOX_SHELL_START_TIME")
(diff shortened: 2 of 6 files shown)


internal/impl/devbox.go
```diff
@@ -232,16 +232,18 @@ func (d *Devbox) Shell() error {
...
 	}
 
-	env, err := plugin.Env(d.cfg.Packages, d.projectDir)
-	if err != nil {
-		return err
-	}
-
+	var env []string
 	if featureflag.NixlessShell.Enabled() {
+		// computeNixEnv() includes plugin env setup
```
ipince, line 237: Thanks. Sorry it wasn't as clear before!

internal/impl/devbox.go
```diff
@@ -727,6 +729,11 @@ func (d *Devbox) computeNixEnv() ([]string, error) {
 	}
 	envPairs = append(envPairs, pluginEnv...)
 
+	if featureflag.EnvConfig.Enabled() {
```
ipince, line 732: can we move this into `computeNixEnv`?  That way both devbox run and shell will get it, not just shell.  It also helps us guarantee that these two paths behave as identically as possible.

internal/impl/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 
 	// Shell configures the devbox shell environment.
 	Shell struct {
+		Env map[string]string `json:"env,omitempty"`
```
mikeland86, line 28: move out of shell struct

internal/impl/devbox.go
```diff
@@ -852,6 +859,35 @@ func (d *Devbox) nixFlakesFilePath() string {
 	return filepath.Join(d.projectDir, ".devbox/gen/flake/flake.nix")
 }
 
+func (d *Devbox) getConfigEnvs(computedEnv []string) []string {
+	// convert key=value strings to map of {key: value}
+	computedEnvMap := map[string]string{}
+	for _, kvpair := range computedEnv {
+		splitKV := strings.SplitN(kvpair, "=", 2)
```
ipince, line 866: if we move it into `computeNixEnv`, we can also avoid this silly splitting, but rather just iterate through the `map` before it gets, uhm, "pair-ified".

(We might have to refactor the function signature of the plugin envs though, which I just avoided because I was lazy).

internal/impl/devbox.go
```diff
@@ -232,16 +232,18 @@ func (d *Devbox) Shell() error {
...
-
+	var env []string
 	if featureflag.NixlessShell.Enabled() {
+		// computeNixEnv() includes plugin env setup
 		env, err = d.computeNixEnv()
 		if err != nil {
 			return err
 		}
+	} else {
+		env, err = plugin.Env(d.cfg.Packages, d.projectDir)
```
mikeland86, line 243: If you replace this with `env = append(envPairs, d.getConfigEnvs(envPairs)...)` then you are not blocked by nixsless shell flag right?

internal/impl/devbox.go
```diff
@@ -852,6 +859,35 @@ func (d *Devbox) nixFlakesFilePath() string {
 	return filepath.Join(d.projectDir, ".devbox/gen/flake/flake.nix")
 }
 
+func (d *Devbox) getConfigEnvs(computedEnv []string) []string {
```
mikeland86, line 862: This is  really well written! 

2 small nits:

* Just `configEnv` as the name (Golang style is to avoid `get` word when possible).
* Add a short comment on top explaining what it does since it's a bit non-trivial.

mikeland86 (approved):
This is great!

internal/impl/devbox.go
ipince: Btw, I did the refactor here: https://github.com/jetpack-io/devbox/pull/621

internal/impl/devbox.go
mohsenari: This is already in `computeNixEnv()` isn't it?

internal/impl/devbox.go
mohsenari: oh that's right! I'll make the change

internal/impl/devbox.go
mohsenari: Done

internal/impl/devbox.go
mohsenari: I ended up removing this else because it deviates from the logic we have in devbox run. I think it's ok to be blocked by unified_env feature flag

PR #612, opened 2023-02-10
[flakes] reset nix profile directory when Flakes feature is toggled

savil:
## Summary

If a nix-profile-directory was created for an old `nix-env` profile, then when
trying to use the new flake-y `nix profile` with the same directory, an error is 
reported.

This PR enables us to toggle `DEVBOX_FEATURE_FLAKES=` on and off for the same
devbox-projects by inspecting the contents of the profile:
1. a `manifest.nix` means that its an old style profile
2. a `manifest.json` means that its a new style profile (i.e. to be used with flakes feature)

The code will "reset" i.e. delete the directory and re-create it.

## How was it tested?

inserted a print statement prior to `os.Remove(dir)` in the reset function
```
> DEVBOX_FEATURE_FLAKES=0 devbox shell

> DEVBOX_FEATURE_FLAKES=1 devbox shell
reset = true

> DEVBOX_FEATURE_FLAKES=0 devbox shell
reset = true
```

Observe:
1. no error, we can toggle seamlessly
2. only one `reset = true` printed despite multiple calls to `devbox.profileDir()`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -590,25 +590,6 @@ func (d *Devbox) generateShellFiles() error {
 	return generateForShell(d.projectDir, plan, d.pluginManager)
 }
 
-// TODO savil. move to packages.go
-func (d *Devbox) profileDir() (string, error) {
-	absPath := filepath.Join(d.projectDir, nix.ProfilePath)
-	if err := os.MkdirAll(filepath.Dir(absPath), 0755); err != nil {
-		return "", errors.WithStack(err)
-	}
-
-	return absPath, nil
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #612** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/612" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/612?utm_source=stack-comment).

mikeland86 (approved):

PR #613, opened 2023-02-10
Add a default stale issue workflow

Lagoja:
## Summary

Adds a stale issue workflow that will close issues and PRs after 30 days of inactivity. 

This will exempt issues and prs with certain labels, so long drafts or issues that are marked for the future can be preserved

## How was it tested?

Untested, but marking this `debug-only` so we can try it as a dry run

### DIFF
--- a/.github/workflows/stale-issue-cleanup.yml
+++ b/.github/workflows/stale-issue-cleanup.yml
@@ -10,8 +10,9 @@ jobs:
     steps:
       - uses: actions/stale@v7
         with:
-          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
-          stale-pr-message: 'This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issue-cleanup.yml
```diff
@@ -0,0 +1,18 @@
...
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v7
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
```
savil, line 13: suggest: Remove stale label or comment, **else** this will be closed in 5 days.

savil: same in `stale-pr-message`

savil (approved):

.github/workflows/stale-issue-cleanup.yml
Lagoja: Updated

PR #616, opened 2023-02-10
Add virtenv path and nix profile path to computed nix path

ipince:
## Summary

This brings the `computeNixEnv` PATH handling to (almost) parity with `devbox shell`. I might remove the virtenv path after confirming we don't need it.

## How was it tested?
Compared `devbox shell` PATH with the PATH I get in `devbox shell` or `devbox run` when I have `NIXLESS_SHELL` or `STRICT_RUN` enabled, respectively.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -229,7 +229,7 @@ func (d *Devbox) Shell() error {
 	}
 	fmt.Fprintln(d.writer, "Starting a devbox shell...")
 
-	profileDir, err := d.profileDir()
+	profileDir, err := d.profilePath()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


internal/impl/devbox.go
```diff
@@ -683,6 +683,19 @@ func (d *Devbox) printPackageUpdateMessage(
...
+// calculated as follows:
+// 1. Start with the output of nix print-dev-env
+// 2. Allow a limited set of variables (leakedVars) in the host machine to "leak" in (e.g. HOME).
+// 3. Include any plugin env vars.
+//
+// The PATH variable has some special handling. In short:
+// 1. Start with the PATH as defined by nix (through nix print-dev-env).
+// 2. TODO: Clean the host PATH of any nix paths.
+// 3. Append the cleaned host PATH (tradeoff between reproducibility and ease of use).
+// 4. Prepend the devbox-managed nix profile path (which is needed to support devbox add inside shell--can we do without it?).
```
savil, line 697: we do need it for now

savil (approved):

internal/impl/devbox.go
ipince: But why exactly?

PR #621, opened 2023-02-10
Refactor env vars as map[string]string instead of []string

ipince:
## Summary
This makes it easier to manipulate vars in functions (like `computeNixEnv`). On the flip side, it does mean we need to convert them to `[]string` in order to pass them to a `Command`, which is a bit annoying, but I think it's reasonable. We can introduce a helper function `asPairs(map[string]string)` or something like that later on if we feel the need to clean up.

## How was it tested?
CICD

### DIFF
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -129,9 +129,11 @@ func WithHistoryFile(historyFile string) ShellOption {
 
 // TODO: Consider removing this once plugins add env vars directly to binaries
 // via wrapper scripts.
-func WithEnvVariables(envVariables []string) ShellOption {
+func WithEnvVariables(envVariables map[string]string) ShellOption {
 	return func(s *Shell) {
-		s.env = append(s.env, envVariables...)
+		for k, v := range envVariables {
+			s.env = append(s.env, fmt.Sprintf("%s=%s", k, v))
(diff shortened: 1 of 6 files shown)


internal/nix/shell.go
```diff
@@ -129,9 +129,11 @@ func WithHistoryFile(historyFile string) ShellOption {
 
 // TODO: Consider removing this once plugins add env vars directly to binaries
 // via wrapper scripts.
-func WithEnvVariables(envVariables []string) ShellOption {
+func WithEnvVariables(envVariables map[string]string) ShellOption {
 	return func(s *Shell) {
-		s.env = append(s.env, envVariables...)
+		for k, v := range envVariables {
+			s.env = append(s.env, fmt.Sprintf("%s=%s", k, v))
+		}
```
mohsenari, line 136: Can we have a utility function for this. Because we need both formats. I agree default form should be `map[string]string` but also have a converter from map[string]string to []string with `key=value` format can make things very easy.

mohsenari (approved):
Approved for unblocking. But see other comment.

internal/nix/shell.go
ipince: Agreed

PR #622, opened 2023-02-10
[global] Turn on feature, keep command hidden

mikeland86:
## Summary

Turns feature on. Adds warning if profile/bin path is not in $PATH

## How was it tested?

`devbox global add hello` 

image


Lagoja (approved):
LGTM

PR #623, opened 2023-02-10
[flakes] enable native nixpkgs download progress output

savil:
## Summary

`nix flakes prefetch` has a native progress indicator in its output. However,
it seems that using the `nixPackagesInstallWriter` would result in it not displaying.
Perhaps because of the tab indentation? Not entirely sure.

This PR uses the regular `io.writer` instead of the custom `nixPackagesInstallWriter`
for the nixpkgs prefetch

## How was it tested?

- set a new commit hash in devbox.json
- ran `devbox shell`
observed the progess indicator

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -96,6 +96,7 @@ func (d *Devbox) addPackagesToProfile(mode installMode) error {
 				"--extra-experimental-features", "nix-command flakes",
 				nix.FlakeNixpkgs(d.cfg.Nixpkgs.Commit),
 			)
+			cmd.Stdout = d.writer
 		} else {
 			cmd = exec.Command(
 				"nix", "profile", "install",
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #623** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/623" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/623?utm_source=stack-comment).

ipince (approved):

mikeland86:
👍 

We can probably remove `nixPackageInstallWriter` once we move completely to flakes.

savil:
@mikeland86 I need the tab-indentation that it does, but could remove the filtering for `devbox-development`.

PR #624, opened 2023-02-10
[flakes] explain better nixpkgs prefetch during devbox add

savil:
## Summary

Previously, we'd always have an explicit numbered step for `nixpkgs` when doing
`devbox add`. This would look odd like:
```
> devbox add emacs

[1/2] nixpkgs registry
...
[1/2] nixpkgs registry: Success
[2/2] emacs
   ...
[2/2] emacs: Success
```

I couldn't find a way to inspect local nix store to determine if we need to prefetch nixpkgs or not.
So, we continue to always do `nix flakes prefetch nixpkgs/<commit>`. 

But to make the UX more understandable, I pull it out into its own function with its own print statements.
IMO this is an improvement but still not awesome. We can continue iterating on it.


savil:
* main
  * **PR #624** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/624" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/624?utm_source=stack-comment).

mikeland86 (approved):
I like putting this in own function (instead of the empty string approach). I also wasn't able to figure out the mapping from commit hash to store (maybe it's saved int eh nix db?) but we can save it ourselves.

If you run

```bash
 nix flake prefetch --extra-experimental-features "nix-command flakes" nixpkgs/2a7ce2ac40c6b906c817b7511079d50cc2fe3643 --json
```

you get 

```json
{"hash":"sha256-ECdyiSgYXWDnCFnLcU+ZKb7Fxlv/csZffyeD8oVd8pY=","storePath":"/nix/store/1mmg9md4pj26hlsm65q0a0a85kpfzbw4-source"}
```

You can save this in `.devbox` somewhere and then use it to find the store path. If you can't find the file or the path, you can always fallback to the regular ensure.

gcurtis (approved):

PR #625, opened 2023-02-11
[bug fix] update Env caller to use map instead of key=val array

savil:
## Summary

I think this callsite was missed in #621

## How was it tested?

```
> cd devbox/examples/php
> devbox shell
```
previously this would fail with `runtime error: index out of range [1] with length 1`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -431,8 +431,8 @@ func (d *Devbox) PluginEnv() (string, error) {
 		return "", err
 	}
 	script := ""
-	for _, pluginEnv := range pluginEnvs {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #625** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/625" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/625?utm_source=stack-comment).

gcurtis (approved):

ipince:
Oops, apologies for the breakage! Thanks!

Issue #557, opened 2023-01-30
[Bug]: Direnv integration ignores init_hook changes


mohsenari:
**Current Behavior (bug)**
When using direnv integration. Adding `export FOO=bar` to init_hook won't be included in direnv load.

**Expected Behavior (fix)**
direnv loading should include `init_hook` changes as well. 

**Additional context**
generated `.envrc` file uses `use nix .devbox/gen/shell.nix` to compute the environment. But it should use `print-dev-env` feature of nix and make sure `init_hook` commands run as well (?)

Lagoja:
Released in 0.4.7 for `.envrc` and local project init_hooks

PR #626, opened 2023-02-11
[direnv] Added support for config envs in direnv

mohsenari:
## Summary
- Changed devbox shell --print-env to use computeNixEnv function (behind feature flag)
- There are certain exports that direnv couldn't process without escaping the quotes namely:
    - "shellHook" (had to change fix a comment line in the template for this as well)
    - "buildPhase"
    - "NIX_CFLAGS_COMPILE",
    - "NIX_HARDENING_ENABLE",
    - "NIX_LDFLAGS",
    - "__impureHostDeps",
    - "nativeBuildInputs",

## How was it tested?
0. compile
1. `./devbox generate direnv`
2. edit .envrc file to be like following:
```shell
use_devbox() {
    watch_file devbox.json
    if [ -f .devbox/gen/shell.nix ]; then
        DEVBOX_SHELL_ENABLED_BACKUP=$DEVBOX_SHELL_ENABLED
        use nix .devbox/gen/shell.nix
        eval $(DEVBOX_FEATURE_UNIFIED_ENV=1 DEVBOX_FEATURE_ENV_CONFIG=1 ./devbox shell --print-env)
        export DEVBOX_SHELL_ENABLED=$DEVBOX_SHELL_ENABLED_BACKUP
    fi
}
use devbox
```
3. cd out of directory and back in and confirm env variables are set correctly including the ones set in devbox.json .

### DIFF
--- a/internal/boxcli/generate/devcontainer_util.go
+++ b/internal/boxcli/generate/devcontainer_util.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 )
 
 type devcontainerObject struct {
@@ -78,11 +79,17 @@ func CreateEnvrc(tmplFS embed.FS, path string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	// setup values to pass to template
+	tmplData := struct {
+		EnvConfigFlagDisabled bool
+	}{
+		EnvConfigFlagDisabled: featureflag.EnvConfig.Disabled(),
+	}
 	// get .envrc content
 	tmplName := "envrc.tmpl"
 	t := template.Must(template.ParseFS(tmplFS, "tmpl/"+tmplName))
 	// write content into file
-	err = t.Execute(file, nil)
+	err = t.Execute(file, tmplData)
 	if err != nil {
 		return errors.WithStack(err)
 	}
--- a/internal/boxcli/generate/devcontainer_util.go
+++ b/internal/boxcli/generate/devcontainer_util.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 )
 
 type devcontainerObject struct {
@@ -79,17 +78,11 @@ func CreateEnvrc(tmplFS embed.FS, path string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	// setup values to pass to template
-	tmplData := struct {
-		EnvConfigFlagDisabled bool
-	}{
-		EnvConfigFlagDisabled: featureflag.EnvConfig.Disabled(),
-	}
 	// get .envrc content
 	tmplName := "envrc.tmpl"
 	t := template.Must(template.ParseFS(tmplFS, "tmpl/"+tmplName))
 	// write content into file
-	err = t.Execute(file, tmplData)
+	err = t.Execute(file, nil)
 	if err != nil {
 		return errors.WithStack(err)
 	}
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -430,15 +430,45 @@ func (d *Devbox) Exec(cmds ...string) error {
 	return nix.Exec(nixDir, cmds, env)
 }
 
-func (d *Devbox) PluginEnv() (string, error) {
-	pluginEnvs, err := plugin.Env(d.packages(), d.projectDir)
+func (d *Devbox) PrintEnv() (string, error) {
+	script := ""
+	if featureflag.UnifiedEnv.Disabled() {
+		envs, err := plugin.Env(d.packages(), d.projectDir)
+		if err != nil {
+			return "", err
+		}
+		for k, v := range envs {
+			script += fmt.Sprintf("export %s=%s\n", k, v)
+		}
+		return script, nil
+	}
+	envs, err := d.computeNixEnv()
+	// exclude variables that direnv can't process
+	varsToExclude := []string{
+		"buildPhase",
+		"shellHook",
+	}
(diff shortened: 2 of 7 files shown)


mohsenari:
closes #557

internal/impl/devbox.go
```diff
@@ -430,15 +430,45 @@ func (d *Devbox) Exec(cmds ...string) error {
...
+		}
+		return script, nil
+	}
+	envs, err := d.computeNixEnv()
+	// exclude variables that direnv can't process
+	varsToExclude := []string{
+		"buildPhase",
+		"shellHook",
+	}
+	// wrap variables in quotes
```
savil, line 451: Is there a downside to wrapping all env-vars in quotes? Should we do that inside `computeNixEnv`? cc @ipince

internal/impl/devbox.go
```diff
@@ -430,15 +430,45 @@ func (d *Devbox) Exec(cmds ...string) error {
...
+			script += fmt.Sprintf("export %s=%s\n", k, v)
+		}
+		return script, nil
+	}
+	envs, err := d.computeNixEnv()
+	// exclude variables that direnv can't process
+	varsToExclude := []string{
+		"buildPhase",
+		"shellHook",
+	}
```
savil, line 450: This feels a bit brittle in that we may modify computeNixEnv to add other variables which may not work.

In https://github.com/jetpack-io/devbox/pull/634, @ipince added a parameter for shellHook
Do we need to either add another parameter for buildPhase, or for the general concept of `isInteractive` ?

savil (commented):
will defer to @ipince , but had some questions about pushing the logic inside computeNixEnv to avoid special casing this. The special casing means this may not work robustly if we later change computeNixEnv.

internal/boxcli/generate/devcontainer_util.go
```diff
@@ -78,11 +79,17 @@ func CreateEnvrc(tmplFS embed.FS, path string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	// setup values to pass to template
+	tmplData := struct {
+		EnvConfigFlagDisabled bool
+	}{
+		EnvConfigFlagDisabled: featureflag.EnvConfig.Disabled(),
+	}
```
mikeland86, line 87: FYI, this could just be a map

internal/impl/devbox.go
mikeland86: Agree with this. The export line below could be:

`fmt.Sprintf("export %s=%q\n", k, v)` 

(I'm not 100% golang escaping and bash escaping are the same, but I think so)

internal/impl/devbox.go
ipince: Hmmm.... well, ideally we should actually _execute_ `shellHook` in order to have a properly configured environment. Luckily right now, the shellHook when using unified-env is a no-op... but the moment it's not, then this (i.e., the direnv integration) will break.

Maybe that moment will never happen though--it's hard to tell.

internal/impl/devbox.go
ipince: Yes, there is a downside. I tried that first and stuff broke. What broke in particular is that, the value of flags like `NIX_LDFLAGS` is something like ` -Lsomelib -Lsomeotherlib`, and it literally just gets replaced in a call to gcc/clang by nix. So nix will call something like `clang $NIX_LDFLAGS some-stuff`, and if `LD_FLAGS` has quotes, then the build will fail.

There must be a way to get around this?  FWIW, when I tried adding the quotes, I used `shellescape.Quote()`, which maybe won't have the same effect as if we use simple `""`? We can try it out and see...

ipince: To try it out, I suggest adding the `""`, and then attempting to run this script while using direnv integration:
https://github.com/jetpack-io/devbox-examples/blob/main/stacks/jekyll/devbox.json#L9

internal/boxcli/generate/devcontainer_util.go
mohsenari: I know there are 2 ways to pass values, a struct or a map. I like the struct better because for each filed you specify the type. For map it has to be `map[string]any`. It's just preference.

internal/impl/tmpl/flake.nix.tmpl
```diff
@@ -20,14 +20,14 @@
...
-                 unset SSL_CERT_FILE
-              fi
+                  # Undo the effects of "nix-shell --pure" on SSL certs.
+                  # See https://github.com/NixOS/nixpkgs/blob/dae204faa0243b4d0c0234a5f5f83a2549ecb5b7/pkgs/stdenv/generic/setup.sh#L677-L685
+                  if [ "$NIX_SSL_CERT_FILE" = "/no-cert-file.crt" ]; then
+                     unset NIX_SSL_CERT_FILE
+                  fi
+                  if [ "$SSL_CERT_FILE" = "/no-cert-file.crt" ]; then
+                     unset SSL_CERT_FILE
+                  fi
```
savil, line 30: why indent these?

savil (approved):
Approving, since code looks good. Lets also update the PR summary with the new intended changes?

internal/impl/tmpl/flake.nix.tmpl
mohsenari: This wasn't done by me. It was probably done automatic with linter/go-lint

internal/impl/tmpl/flake.nix.tmpl
```diff
@@ -20,7 +20,7 @@
               export IN_NIX_SHELL=0
               export DEVBOX_SHELL_ENABLED=1
 
-              # Undo the effects of `nix-shell --pure` on SSL certs.
+              # Undo the effects of "nix-shell --pure" on SSL certs.
```
ipince, line 23: Oh, so that was the default.nix error, lol. Must've been fun to catch that one!

ipince (approved):

PR #630, opened 2023-02-13
[global] Implement simple global pull

mikeland86:
## Summary

This "pulls" a global config either from a local path or URL. It merges the packages into the existing config.

## How was it tested?

```
devbox global pull devbox.json
devbox global pull https://raw.githubusercontent.com/jetpack-io/devbox/main/devbox.json
```


internal/boxcli/global.go
```diff
@@ -86,6 +87,16 @@ func globalListCmd() *cobra.Command {
 	}
 }
 
+func globalPullCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:     "pull <file> | <url>",
+		Short:   "Pulls a global config from a file or URL",
```
savil, line 93: it seems that url requires the scheme. That's non-obvious but very reasonable, so could we specify an example in the Long explanation?

internal/impl/global.go
```diff
@@ -70,21 +70,57 @@ func (d *Devbox) RemoveGlobal(pkgs ...string) error {
 	return d.saveCfg()
 }
 
+func (d *Devbox) PullGlobal(path string) error {
+	u, err := url.Parse(path)
+	if err == nil && u.Scheme != "" {
```
savil, line 75: scheme requirement is here (ref. previous comment above)

internal/impl/global.go
```diff
@@ -95,6 +131,16 @@ func GlobalConfigPath() (string, error) {
 	return filepath.Join(home, "/.config/devbox/"), nil
 }
 
+func globalProfilePath() (string, error) {
+	configPath, err := GlobalConfigPath()
+	if err != nil {
+		return "", err
+	}
+	nixDirPath := filepath.Join(configPath, "nix")
+	_ = os.MkdirAll(nixDirPath, 0755)
```
savil, line 140: lets check the err

savil (approved):

mikeland86:
@savil to avoid rebase conflicts, I addressed your comments in https://github.com/jetpack-io/devbox/pull/637

I'll merge both at same time.

PR #633, opened 2023-02-14
Add tests for PATH setting in run

ipince:
## Summary
TSIA

## How was it tested?
```
> go test ./testscripts/testscripts_test.go
```

### DIFF
--- a/testscripts/run/path.test.txt
+++ b/testscripts/run/path.test.txt
@@ -1,0 +1,31 @@
+exec devbox init
+exec devbox add which
+
+# Non-default nix profiles are removed from PATH
+# NOTE: I was unable to set NIX_PROFILES. It seems to be overriden somewhere somehow.
+env HOME=/home/test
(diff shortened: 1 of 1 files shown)


mohsenari (approved):

testscripts/run/path.test.txt
```diff
@@ -0,0 +1,32 @@
...
+exec devbox run echo '$PATH'
+! stdout 'relative/path'
+! stdout '/some//dirty/../clean/path'
+stdout '/some/clean/path'
+
+# Path contains path to installed nix packages
+stdout '-which-'
+
+# Path contains plugin virtual env path
+stdout '.devbox/virtenv/bin'
```
mohsenari, line 26: What plugin is this?

ipince: no plugin in particular.. but plugins that have executables will be put there. At least that was the design, though I think no plugin is using this today. I might just remove it altogether for simplicity.

PR #634, opened 2023-02-14
Set PATH prepend after user init file runs

ipince:
## Summary
This applies to `devbox shell` only. We defer the path prepending until _after_ any user init file is run. For `devbox run`, we set the path before creating the new shell.

## How was it tested?
```
DEVBOX_FEATURE_UNIFIED_ENV=1 devbox shell
echo $PATH
# inspect that .zshrc path prepends appear AFTER the nix paths
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -250,7 +250,7 @@ func (d *Devbox) Shell() error {
 	}
 
 	if featureflag.UnifiedEnv.Enabled() {
-		env, err = d.computeNixEnv()
+		env, err = d.computeNixEnv(false)
 		if err != nil {
 			return err
 		}
@@ -294,7 +294,7 @@ func (d *Devbox) RunScript(cmdName string, cmdArgs []string) error {
 		return err
 	}
 
-	env, err := d.computeNixEnv()
+	env, err := d.computeNixEnv(true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


savil (approved):
IIUC:

1. for `devbox shell`, the shellrc runs and we do `PATH=prepend:PATH`
2. for `devbox run`, the shellrc does NOT run but we still do `PATH=prepend:PATH`.

Questions:
1. did I understand correctly?
2. why do we not run shellrc for devbox run? is it because the shellrc is for interactive shells whereas devbox run is non-interactive?

internal/impl/devbox.go
```diff
@@ -754,7 +754,17 @@ func (d *Devbox) computeNixEnv() (map[string]string, error) {
 	nixPath := env["PATH"]
 	hostPath := nix.CleanEnvPath(os.Getenv("PATH"), os.Getenv("NIX_PROFILES"))
 
-	env["PATH"] = fmt.Sprintf("%s:%s:%s:%s", pluginVirtenvPath, nixProfilePath, nixPath, hostPath)
+	// NOTE: for devbox shell, we need to defer the PATH setting, because a user's init file may prepend
+	// stuff to PATH, which will then take precedence over the devbox-set PATH. Instead, we do the path
+	// prepending in shellrc.tmpl. I chose to use the `setFullPath` variable instead of something like
```
savil, line 759: > I chose to use...

btw, I like this design principle 👍🏾

ipince:
> did I understand correctly?
yes!

> why do we not run shellrc for devbox run? is it because the shellrc is for interactive shells whereas devbox run is non-interactive?

yes!  I guess it's a design decision to run bashrc (I'm using "bashrc" interchangeably here, even if the user isn't running bash), for usability. It does introduce a vector of non-reproducibility between shell and run, but I think that's ok.

PR #635, opened 2023-02-14
Include some more shell-specific vars, remove nix shell hook

ipince:
## Summary
Add some more variables that are required for shell: DEVBOX_ vars, a couple hard-coded nix vars, and a few more vars to leak.

Also removes the `shell.nix` hook, which we no longer need if we're using unified env.

## How was it tested?
go tests

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -742,19 +742,19 @@ func (d *Devbox) computeNixEnv(setFullPath bool) (map[string]string, error) {
 	// Copy over (and overwrite) vars that we explicitly "leak", as well as DEVBOX_ vars.
 	for _, kv := range os.Environ() {
 		parts := strings.SplitN(kv, "=", 2)
-		k := parts[0]
-		v := parts[1]
+		key := parts[0]
+		val := parts[1]
 
 		if strings.HasPrefix(parts[0], "DEVBOX_") {
-			env[k] = v
+			env[key] = val
 		}
 
-		if _, ok := leakedVars[k]; ok {
-			env[k] = v
+		if _, ok := leakedVars[key]; ok {
+			env[key] = val
 		}
 
-		if _, ok := leakedVarsForShell[k]; ok {
-			env[k] = v
+		if _, ok := leakedVarsForShell[key]; ok {
+			env[key] = val
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/impl/devbox.go
```diff
@@ -724,13 +725,31 @@ func (d *Devbox) computeNixEnv(setFullPath bool) (map[string]string, error) {
 		}
 	}
 
-	// Overwrite/leak whitelisted vars into nixEnv:
-	for name, leak := range leakedVars {
-		if leak {
-			env[name] = os.Getenv(name)
+	// Copy over (and overwrite) vars that we explicitly "leak", as well as DEVBOX_ vars.
+	for _, kv := range os.Environ() {
+		parts := strings.SplitN(kv, "=", 2)
```
savil, line 730:
```
key, val, found := strings.Cut(kv, "=")
if !found {
    return errors.Errorsf("expected = in kv: %s", kv)
}
```

internal/impl/devbox.go
```diff
@@ -724,13 +725,31 @@ func (d *Devbox) computeNixEnv(setFullPath bool) (map[string]string, error) {
...
+		}
+
+		if _, ok := leakedVarsForShell[key]; ok {
+			env[key] = val
 		}
 	}
 
+	// These variables are only needed for shell, but we include them here in the computed env
+	// for both shell and run in order to be as identical as possible.
+	env["NIX_PKGS_ALLOW_UNFREE"] = "1"     // Only for shell because we don't expect nix calls within run
```
savil, line 749: I think its good to include, because someone's script may do `devbox add` for them.

internal/impl/devbox.go
```diff
@@ -739,11 +758,13 @@ func (d *Devbox) computeNixEnv(setFullPath bool) (map[string]string, error) {
 		env[k] = v
 	}
 
-	// TODO: add shell-specific vars, including:
-	// - NIXPKGS_ALLOW_UNFREE=1 (not needed in run because we don't expect nix calls there)
-	// - __ETC_PROFILE_NIX_SOURCED=1 (not needed in run because we don't expect rc files to try to load nix profiles)
-	// - HISTFILE (not needed in run because it's non-interactive)
```
savil, line 745: would it be good to maintain somewhere a list of env vars like HISTFILE that are deliberately omitted from `devbox run` but included in `devbox shell`?

savil: maybe in LeakedVarsForShell with a value of `false` ?

savil (approved):

internal/impl/devbox.go
ipince: Well, I think looking at shellrc will be good enough for now (there's no other place in which the env can be changed)

PR #636, opened 2023-02-14
tmpl: run nixpkgs-fmt on nix files

gcurtis:
Run `nixpkgs-fmt *.nix.tmpl` to make everything follow common formatting/style conventions.


savil (approved):
thanks! good idea.

PR #637, opened 2023-02-15
[global] Move global to XDG_DATA_HOME

mikeland86:
## Summary

This is stacked on https://github.com/jetpack-io/devbox/pull/630

This moves global data to XDG_DATA_HOME or `~/.local/share`

* The devbox.json is stored in `~/.local/share/devbox/global/default/devbox.json`
* The nix profile is in `~/.local/share/devbox/global/default/nix-profile`
* the bin path is `~/.local/share/devbox/global/current/nix-profile/bin`

## How was it tested?

`devbox global add hello`

### DIFF
--- a/internal/impl/global.go
+++ b/internal/impl/global.go
@@ -128,7 +128,7 @@ func GlobalDataPath() string {
 }
 
 func GlobalNixProfilePath() string {
-	return filepath.Join(GlobalDataPath(), "nix-profile")
+	return filepath.Join(GlobalDataPath(), "profile")
 }
 
 // Checks if the global profile is in the path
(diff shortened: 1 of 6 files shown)


internal/impl/global.go
```diff
@@ -123,31 +121,25 @@ func (d *Devbox) addFromPull(pullCfg *Config) error {
...
-	if err != nil {
-		return err
+	currentPath := xdg.DataSubpath("devbox/global/current")
+	// For now default is always current. In the future we will support multiple
+	// and allow user to switch.
+	if err := os.Symlink(GlobalDataPath(), currentPath); err != nil && !os.IsExist(err) {
+		return errors.WithStack(err)
 	}
-	binPath := filepath.Join(profilePath, "bin")
+	binPath := filepath.Join(currentPath, "nix-profile", "bin")
```
loreto, line 142: Could we have the symlink point to the nix profile (`GlobalNixProfilePath`) to keep the path shorter when they add the path in their rc file?

(They can add `.../devbox/global/current/bin`)

Or are they not adding that to the path and instead doing something like `source $(devbox global activate)`?

loreto (approved):
LGTM

I think I would name the directory just `profile` instead of `nix-profile` but I don't care that much *as long as* that is hidden from the rc file (either by the symlink or by a `source $(devbox ...` as describe in the online comment.

PR #638, opened 2023-02-15
[dependabot] Fix vulnerability in example

mikeland86:
## Summary

Fixes https://github.com/jetpack-io/devbox/security/dependabot/16


savil (approved):
cool, this is the default hash used in the code.

> // The commit hash for nixpkgs-unstable on 2023-01-25 from status.nixos.org
const DefaultNixpkgsCommit = "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"

For some other time, it would be neat to have a script that updates all the examples, with a parameter for the default nixpkgsCommitHash

PR #631, opened 2023-02-14
[flake] php extensions example project

savil:
## Summary

Added an example devbox-project that has a php extension package. This has been useful for doing adhoc testing when developing the feature of "flakes with php extensions". In a future PR, I'll add a testscript (and possibly remove this one).

## How was it tested?

```
> devbox shell

# with package `php81Extensions.ds`
> php index.php
Original vector elements
idx: 0 and elem: hello
idx: 1 and elem: world
done

# without package `php81Extensions.ds` 
> php index.php
ds extension is not enabled
```

PR #639, opened 2023-02-15
[flakes] generate flake.nix file for php-related packages

savil:
## Summary

For `flake.nix`:
- we now also include the php "Definitions" and change the `buildInputs` list so that it uses the `php` packages from `Definitions` instead of from `nixpkgs` (if Definitions exist)

TODO for future PR:
- [ ] look into whether `buildInputs` should be `packages` in `flake.nix`

## How was it tested?

in `examples/testdata/php/php-extensions`, started `devbox shell` with `DEVBOX_FEATURE_FLAKES=0` and `DEVBOX_FEATURE_FLAKES=1`. Inspected the generated `flake.nix` files.

```
❯ cat .devbox/gen/flake/flake.nix
{
  description = "A devbox shell";

  inputs = {
    nixpkgs.url = "https://github.com/nixos/nixpkgs/archive/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62.tar.gz";

    flake-utils.url = "github:numtide/flake-utils";
  };

  outputs = { self, nixpkgs, flake-utils }:
    flake-utils.lib.eachDefaultSystem (system:
      let pkgs = nixpkgs.legacyPackages.${system};
          php81 = pkgs.php81.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds ]));
          php81Packages.composer = php81.packages.composer;

      in {
        devShell = pkgs.mkShell {
          shellHook =
            ''
              # We're technically no longer in a Nix shell after this hook because we
              # exec a devbox shell.
              export IN_NIX_SHELL=0
              export DEVBOX_SHELL_ENABLED=1

              # Undo the effects of "nix-shell --pure" on SSL certs.
              # See https://github.com/NixOS/nixpkgs/blob/dae204faa0243b4d0c0234a5f5f83a2549ecb5b7/pkgs/stdenv/generic/setup.sh#L677-L685
              if [ "$NIX_SSL_CERT_FILE" = "/no-cert-file.crt" ]; then
                 unset NIX_SSL_CERT_FILE
              fi
              if [ "$SSL_CERT_FILE" = "/no-cert-file.crt" ]; then
                 unset SSL_CERT_FILE
              fi

              # Append the parent shell's PATH so that we retain access to
              # non-Nix programs, while still preferring the Nix ones.
              export "PATH=$PATH:$PARENT_PATH"


            '';

          buildInputs = with pkgs; [
            php81

            php81Packages.composer

            php81Extensions.ds
            ];
        };
      }
    );
}
```


savil:
* main
  * **PR #631** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/631" target="_blank">Graphite</a> 
    * **PR #639** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/639" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/639?utm_source=stack-comment).

mikeland86:
@savil I think this PR might not be necessary if we solve a flakes bug detailed here: https://github.com/jetpack-io/devbox/pull/655

https://github.com/jetpack-io/devbox/pull/648 might similarly not be needed.

the tldr is that we should not add the profile path when starting shell (or if we do we need a way to control the binaries there to ensure they represent the flake.nix ones and not the individually installed ones).

gcurtis:
@savil I'm going to remove myself as a reviewer since I think you and Mike are going to figure out the best way to tackle this. Feel free to re-add me if you still need a review.

savil:
We are going to simplify this to just the flake.nix changes.

mikeland86 (approved):

PR #642, opened 2023-02-16
Deprecate shell -- <cmd> in favor of run <cmd>

ipince:
## Summary
Makes `devbox shell -- <cmd>` print a Warning and recommend using `devbox run <cmd>` instead. This behavior is protected behind the UNIFIED_ENV feature flag.

Note that `devbox services` internally uses `devbox shell -- <cmd>`, so when UNIFIED_ENV is enabled, it'll use `devbox run` instead. But it's a bit circular... when outside of a shell, running `devbox services start` is exactly equivalent to running `devbox run devbox services start`.

## How was it tested?
```
DEVBOX_FEATURE_UNIFIED_ENV=0 ./devbox shell --help
DEVBOX_FEATURE_UNIFIED_ENV=1 ./devbox shell --help

DEVBOX_FEATURE_UNIFIED_ENV=0 ./devbox shell -- echo "foo" # no warning
DEVBOX_FEATURE_UNIFIED_ENV=1 ./devbox shell -- echo "foo" # see warning

DEVBOX_FEATURE_UNIFIED_ENV=1 ./devbox shell -- echo '$PATH' 
DEVBOX_FEATURE_UNIFIED_ENV=1 ./devbox run echo '$PATH' # same paths
```

### DIFF
--- a/internal/boxcli/shell.go
+++ b/internal/boxcli/shell.go
@@ -6,6 +6,7 @@ package boxcli
 import (
 	"fmt"
 
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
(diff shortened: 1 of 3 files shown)


internal/boxcli/shell.go
```diff
@@ -68,6 +80,11 @@ func runShellCmd(cmd *cobra.Command, args []string, flags shellCmdFlags) error {
 	}
 
 	if len(cmds) > 0 {
+		if featureflag.UnifiedEnv.Enabled() {
+			fmt.Fprint(cmd.ErrOrStderr(),
+				color.HiYellowString("[Warning] \"devbox shell -- <cmd>\" is deprecated and will disappear "+
```
savil, line 85: I think Mike added a `ux.Warningf` or similar function....

savil (approved):

internal/boxcli/shell.go
ipince: Thanks! updated

PR #644, opened 2023-02-16
boxcli: add `devbox global shellenv` command

gcurtis:
## Summary

This command prints shell commands that add Devbox packages to the user's PATH. The output is intended to be evaluated in the user's shell rcfile:

	echo 'eval "$(devbox global shellenv)"' >> ~/.zshrc

We don't expand the environment variables in the generated commands so that it's clear to the user what the commands are doing.

## How was it tested?

Ran `eval $(devbox global shellenv)` with `PATH` and `XDG_DATA_HOME` set to various values.

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
+	"go.jetpack.io/devbox/internal/impl"
 	"go.jetpack.io/devbox/internal/nix"
 )
 
@@ -24,6 +25,7 @@ func globalCmd() *cobra.Command {
 	cmd.AddCommand(globalListCmd())
 	cmd.AddCommand(globalPullCmd())
 	cmd.AddCommand(globalRemoveCmd())
+	cmd.AddCommand(globalShellenvCmd())
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


loreto (approved):
LGTM

internal/impl/global.go
```diff
@@ -143,6 +143,24 @@ func GlobalNixProfilePath() (string, error) {
...
+// the user's shell rcfile. For example:
+//
+//	echo 'eval "$(devbox global shellenv)"' >> ~/.zshrc
+func GenerateShellEnv() string {
+	// If the user has "eval $(devbox global shellenv)" in their shell's
+	// rcfile, then running "devbox shell" will cause these commands to be
+	// evaluated twice (once by the "parent" shell and once by the devbox
+	// shell). Prevent this by making the eval a no-op when we're already
+	// inside a devbox shell.
+	if IsDevboxShellEnabled() {
```
mikeland86, line 158: This means that if a package is installed while in shell, there's not way for the user to update the current shell to make it work. They would have to quit the shell, add the eval statement to their rc files and start it again.

Maybe we add 

`devbox activate --force`  (or `--refresh`) or something like that? So we can let the user force the configuration while they are in a devbox shell.

internal/boxcli/global.go
```diff
@@ -98,6 +100,16 @@ func globalPullCmd() *cobra.Command {
...
 }
 
+func globalShellenvCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "shellenv",
+		Short: "Print shell commands that add Devbox packages to your PATH",
+		Run: func(*cobra.Command, []string) {
+			fmt.Print(impl.GenerateShellEnv())
+		},
+	}
```
mikeland86, line 110: @loreto had suggested using `devbox activate` for this. I like that a bit more because then it's not related to `global` but actually a more general command that let's us modify and improve rc configuration over time. Thoughts?

internal/impl/global.go
mikeland86: Another alternative is to have `GenerateShellEnv` set a new env variable `_DEBOX_SHELL_ENV_SET=1` which is independent of shell. This lets apply this configuration only in cases it hasn't been applied. 

We do still run into the issue @ipince brought up that we don't want global packages to supersede local ones. Maybe we splice in the new path instead?

mikeland86 (commented):
This is cool! My main comment is that maybe we should rename this to `devbox activate` so it can be more generic. Also we could add a flag that let's the user run this command manually to refresh their shell.

PR #646, opened 2023-02-16
Add testscript tests for PATH setting

ipince:
## Summary
Tests that the PATH is built in the desired order. Had to add a `print` function and `path.order` assertion function. Let me know if you have any other ideas!

## How was it tested?
```
go test ./testscripts/testscripts_test.go
```


loreto (approved):
LGTM

PR #651, opened 2023-02-17
Add revive linter and enable indent-error-flow rule

hezhizhen:
## Summary

1. Add `revive` linter to `.golangci.yml`
2. Enable `indent-error-flow` rule only for `revive`
3. Address related issues.

## How was it tested?

Linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,7 @@ linters:
     - misspell
     - nilerr
     - reassign
+    - revive
     - staticcheck
     - stylecheck
     - typecheck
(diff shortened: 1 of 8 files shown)


hezhizhen:
@ipince

hezhizhen:
@savil Please take a look.

savil:
@hezhizhen thanks! lgtm. Would you mind reverting the svg change?

hezhizhen:
> @hezhizhen thanks! lgtm. Would you mind reverting the svg change?

Okay. And skip checking svg files?

savil:
@hezhizhen I think that would make sense, unless there's some good reason you feel we should be checking svg files?

hezhizhen:
> @hezhizhen I think that would make sense, unless there's some good reason you feel we should be checking svg files?

I have no idea about how svg files are generated... I'll skip them.

hezhizhen:
@savil I rebased `main` and reverted changes in svg files. PTAL

savil (approved):

PR #653, opened 2023-02-17
Fix cli-release badge

ipince:
It's currently pointing to `release.yml`, which doesn't exist (but `cli-release.yml`) does.


loreto (approved):
LGTM

PR #663, opened 2023-02-22
[service-manager] Add more process compose yamls

mikeland86:
## Summary

This adds process compose for apache and php. It also fixes a hash to ensure we download a version of process compose that allows multiple config files to be merged.

## How was it tested?

```bash
devbox add php apacheHttpd
devbox services manager
```

### DIFF
--- a/plugins/apache/process-compose.yaml
+++ b/plugins/apache/process-compose.yaml
@@ -1,0 +1,15 @@
+version: "0.5"
+
+processes:
+  apache:
+    command: "echo \"Apache starting on port $HTTPD_PORT\ http://localhost:$HTTPD_PORT\" && apachectl start -f $HTTPD_CONFDIR/httpd.conf -D FOREGROUND"
+    availability:
+      restart: "always"
+  apache-error:
+    command: "tail -f $HTTPD_ERROR_LOG_FILE"
+    availability:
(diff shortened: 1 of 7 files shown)


plugins/apache/process-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+processes:
+  apache:
+    command: "echo \"Apache starting on port $HTTPD_PORT\ http://localhost:$HTTPD_PORT\" && apachectl start -f $HTTPD_CONFDIR/httpd.conf -D FOREGROUND"
+    availability:
+      restart: "always"
+  apache-error:
+    command: "tail -f $HTTPD_ERROR_LOG_FILE"
+    availability:
+      restart: "always"
```
savil, line 11: lets say the user wants to view these error logs, what command should they run to get the output from this `tail -f` ?

savil (approved):

plugins/apache/process-compose.yaml
mikeland86: They can `tail -f` directly or they can use `devbox service manager` and get a UI that looks like this:

image

savil: ah, that's neat

PR #667, opened 2023-02-22
[shellenv] Remove nix profile from path, add shellenv command

mikeland86:
## Summary

This removes nix profile from path when using unified environment. It adds a new shellenv command and prints instructions when package is added inside devbox shell.

@gcurtis I chose `eval $(devbox shellenv)` over `. <(devbox shellenv)` to be consistent with the global instructions and also because it's easy to miss the period when we print it. Not sure if there's a difference. Also is this fish compatible or do we need different instructions for fish?

Lastly, you'll notice `hash -r` is not there anymore. In my experiments running `eval $(devbox shellenv)` causes the hash table to clear. I believe this happens because `PATH` has changed, but I'm not 100% sure.

## How was it tested?

image

```bash
➜ devbox shell
➜ devbox add php82 php82Extensions.memcached
➜ which php
php not found
➜ eval $(devbox shellenv)
➜ which php
/nix/store/5yylkc71wmlzii2vl40021vgbfwqmajb-php-with-extensions-8.2.1/bin/php
➜ php -m | grep memcached
memcached
```


internal/impl/devbox.go
```diff
@@ -455,7 +455,7 @@ func (d *Devbox) PrintEnv() (string, error) {
 		}
 		return script, nil
 	}
-	envs, err := d.computeNixEnv(false)
```
ipince, line 458: huh.. this is weird and maybe a bug.. I think this should be `true`. This is what `direnv` integration uses, so we definitely want to set the full path here.

Or am I missing something?

ipince (approved):
LGTM, but see comment (unrelated to your change though).

Also, it'd be nice to add some tests

savil (approved):
awesome!

Does the test plan also work with `export DEVBOX_FEATURE_FLAKES=1`?

gcurtis:
> @gcurtis I chose `eval $(devbox shellenv)` over `. <(devbox shellenv)` to be consistent with the global instructions and also because it's easy to miss the period when we print it. Not sure if there's a difference. Also is this fish compatible or do we need different instructions for fish?

Using `eval` is more compatible. The process substitution `<(cmd)` syntax is specific to bash. I'm not very familiar with fish, so we should test to be sure.

gcurtis (approved):
LGTM. I second the suggestion to add a couple tests since shell stuff tends to be the source of bugs.

internal/impl/devbox.go
mikeland86: `false` assumes the path gets built in the rc files but that only happens for `devbox shell` right? for direnv, I assume it should be true, but not confident. I'll fix it in new PR and test properly

mikeland86:
@savil I pretty much only tested with `DEVBOX_FEATURE_FLAKES=1` but I'll make sure it works without flakes as well.

mikeland86:
@savil confirmed all working with without `DEVBOX_FEATURE_FLAKES`

PR #650, opened 2023-02-16
Enable unified env by default

ipince:
## Summary
I guess it _actually_ was ready right on time on 2/16 :). (but not for a 2/16 release)

## How was it tested?
testscript tests and using it while developing

PR #669, opened 2023-02-22
Add support for fish

ipince:
## Summary
Adds support for `fish` shell.

When running `devbox shell`:
We write the post-initialization steps into a "shellrc" file (see `shellrc_fish.tmpl`). I attempted to reuse the existing `shellrc.tmpl`, but the differences in syntax between fish and other shells were large enough that I had to separate them. Next, we invoke `fish -C <shellrc>`. This will make `fish` use the user's existing config first, and _then_ run whatever is in `<shellrc>`, which is similar to what we do for non-fish shells.

When running `devbox run`:
- If unified env is ON, then we invoke `sh -c <script>` just like we do for every other shell.
- If unified env is OFF, then we'll invoke `fish -C <shellrc> -c <script>`. However, note that #650 already enables unified env by default, so this branch is mainly here just in case something goes wrong with unified env and we need to revert.

Caveats:
- We don't check any of the syntax in devbox.json's init hook or scripts. If an init hook uses fish-specific syntax, then it will work for users using `fish`, but it won't for users using a different shell. Similarly for scripts. Conversely, if an init hook uses POSIX syntax that fish doesn't support (e.g. `FOO=bar`), then the init hook will fail for fish users.
- Any plugins that use init hooks should write them in such a way that they work for both fish and non-fish. So far, it seems only `pip` and `ruby` use init hooks, and they're written in a compatible way (AFAIK).

## How was it tested?
I tested that both `devbox shell` and `devbox run` worked with and without unified env. I verified that `PATH` was being set, user's config is being run (for shell only), init hooks are being run, history file is set (shell only), prompt is set (shell only), and scripts are written and run.
To test with fish, it's as easy as invoking devbox as such:
```
SHELL=fish devbox shell
```

### DIFF
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -19,6 +19,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/debug"
+	"go.jetpack.io/devbox/internal/xdg"
 )
 
 //go:embed shellrc.tmpl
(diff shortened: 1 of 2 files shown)


Lagoja:
Looks like this works as expected from my Fish testing, so I'll approve it. Someone else should do code review though

Lagoja (approved):

gcurtis (approved):
Nice! LGTM

PR #670, opened 2023-02-22
[flakes] enable unfree packages

savil:
## Summary

To enable unfree packages with flakes, we need to do both of:
1. set `NIXPKGS_ALLOW_UNFREE=1`
2. use `--impure` flag

There is also an option to do:
```
nixpkgs {
   config {
      allowUnfree = true;
   } 
}
```
However, I have not done this because:
1. we'd still need `--impure` for `nix profile install` that doesn't use `flake.nix`.
2. I couldn't actually figure out how to set `config.allowUnfree` in our flake.nix since we are using the num-tides flakes-utils to enable the flake to work for all `system` values.

## How was it tested?

in `examples/testdata/unfree-packages`, I can now do:
```
DEVBOX_FEATURE_FLAKES=1 DEVBOX_FEATURE_UNIFIED_ENV=1 DEVBOX_DEBUG=1 devbox shell
```

I was a little concerned that `--impure` would introduce a lot of other environment variables 
into the shell, but I think it is safe for the following reasons.

1. I tested with a specific env-var:
```
> export SAVIL=impure-test
> env | grep SAVIL
SAVIL=impure-test

> DEVBOX_FEATURE_FLAKES=1 DEVBOX_FEATURE_UNIFIED_ENV=1 DEVBOX_DEBUG=1 devbox shell
(devbox) > env | grep SAVIL
```
So, the `SAVIL` env-var was not included in the devbox shell.

2. I compared all the env-vars by doing `env > env_with_{property}.txt` for this branch and the `main` branch.

From visual inspection, the order of env-vars was different but the number of lines were same.

So, I sorted them alphabetically and then compared them:
```
❯ cat env_with_main.txt | sort > env_with_main_sorted.txt
(devbox)

❯ cat env_with_impure.txt | sort > env_with_impure_sorted.txt
(devbox)

❯ diff env_with_main_sorted.txt env_with_impure_sorted.txt
77c77
< STARSHIP_SESSION_KEY=7757126821727611
---
> STARSHIP_SESSION_KEY=3237459778369717
92c92
< ZDOTDIR=/var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/devbox3757196048
---
> ZDOTDIR=/var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/devbox1045668199
```
The differences here are session-specific, and I don't see a lot of other env-vars being introduced due to `--impure`.

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -154,8 +154,10 @@ func PrintDevEnv(nixShellFilePath, nixFlakesFilePath string) (*varsAndFuncs, err
 		"--extra-experimental-features", "nix-command",
 		"--extra-experimental-features", "ca-derivations",
 		"--option", "experimental-features", "nix-command flakes",
+		"--impure",
 		"--json")
 	debug.Log("Running print-dev-env cmd: %s\n", cmd)
+	cmd.Env = DefaultEnv()
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, errors.WithStack(err)
--- a/examples/testdata/go/go-1.18/devbox.json
+++ b/examples/testdata/go/go-1.18/devbox.json
@@ -1,11 +1,11 @@
 {
   "packages": [
-    "go"
+    "go_1_18"
   ],
   "shell": {
     "init_hook": null
   },
   "nixpkgs": {
-    "commit": "af9e00071d0971eb292fd5abef334e66eda3cb69"
+    "commit": "9e0eed654c705c7cafe192a8eba1610217f70544"
   }
 }
--- a/examples/testdata/go/go-1.18/shell_plan.json
+++ b/examples/testdata/go/go-1.18/shell_plan.json
@@ -1,5 +1,5 @@
 {
   "dev_packages": [
-    "go"
+    "go_1_18"
   ]
 }
--- a/examples/testdata/unfree-packages/devbox.json
+++ b/examples/testdata/unfree-packages/devbox.json
@@ -1,5 +1,6 @@
 {
   "packages": [
+    "slack",
     "vscode"
   ],
   "shell": {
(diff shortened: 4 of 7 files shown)


savil:
* main
  * **PR #670** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/670" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/670?utm_source=stack-comment).

ipince:
I think `--impure` doesn't affect things because we control which env vars make it through to the child shell, instead of just allowing it to inherit everything like `nix-shell` would do (if you passed `--impure`)

internal/nix/nix.go
```diff
@@ -154,8 +154,10 @@ func PrintDevEnv(nixShellFilePath, nixFlakesFilePath string) (*varsAndFuncs, err
 		"--extra-experimental-features", "nix-command",
 		"--extra-experimental-features", "ca-derivations",
 		"--option", "experimental-features", "nix-command flakes",
+		"--impure",
 		"--json")
 	debug.Log("Running print-dev-env cmd: %s\n", cmd)
+	cmd.Env = DefaultEnv()
```
ipince, line 160: hmmm is this enough?

what if we're inside a devbox shell and then `devbox add unfree-package`?

savil: yup, I tested that scenario as well.

There are two steps when running `devbox add unfree-package`:
1. `nix profile install`: in this PR, we add `--impure` and `cmd.Env = DefaultEnv` to this command.
2. `nix print-dev-env`: this code here (i.e. `internal/nix/nix.go:160`)  will run again with `cmd.Env = DefaultEnv`

gcurtis (approved):

internal/nix/nix.go
ipince: I'm just curious as to what the diff is for `nix print-dev-env`.. hmm i guess if you don't have the ALLOW_UNFREE flag, then maybe `nix print-dev-env` would fail, because the `flakes.nix` file doesn't have `allowUnfree = 1`

PR #672, opened 2023-02-22
[easy][global] Fix duplicate packages DEV-1281

mikeland86:
## Summary

TSIA

## How was it tested?

Added go_1_19 as global and local package.

`devbox run echo 5`

### DIFF
--- a/internal/impl/config.go
+++ b/internal/impl/config.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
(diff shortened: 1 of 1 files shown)


ipince (approved):
Is this why you had seen `go` and `go_1.20` be installed?  the packages are different though...

mikeland86:
Nope, this just a minor UX issue. When creating flake.nix we would duplicate packages if you had the same package in global and devbox.json. It would not break anything, but it just looked weird and it looked like we were installing things twice.

There's a separate potential issue (harder to fix) if global and devbox have competing binaries.

PR #674, opened 2023-02-22
[stderr] Fix a few prints that were going to out instead of err DEV-1282

mikeland86:
## Summary

This changes output for

* add
* global add
* global pull

(since these are not expected to print anything except diagnostic output)

It also fixes the debug middleware which prints all errors.

## How was it tested?

`devbox run test`


loreto (approved):
LGTM

ipince (approved):

PR #675, opened 2023-02-23
[error] Better error message for ExitErrors

savil:
## Summary

This PR makes three changes:

1. Handles ExitErrors specially (for non `usererr.ExitError`) by printing the `err.Stderr` when `DEVBOX_DEBUG=1`.

2. When `DEVBOX_DEBUG=0`, we now include the command and also suggest to the user to run with DEVBOX_DEBUG=1 for a more detailed error and suggest reporting the bug.

3. Renames `usererr.UserExecError` to `usererr.ExitError` which more accurately represents what that struct is doing: the error is only wrapped if it is ExitError.

## How was it tested?

in `testdata/unfree-packages`:

BEFORE:
```
❯ DEVBOX_FEATURE_FLAKES=1 DEVBOX_FEATURE_UNIFIED_ENV=1 DEVBOX_DEBUG=0 devbox shell
Ensuring nixpkgs registry is downloaded.
Downloaded 'github:NixOS/nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04' to '/nix/store/j5a7i3dvxslg2ychfy53wdhg1m3xfrwm-source' (hash 'sha256-FqZ7b2RpoHQ/jlG6JPcCNmG/DoUPCIvyaropUDFhF3Q=').
Ensuring nixpkgs registry is downloaded: Success
Starting a devbox shell...
Error: exit status 1
```

AFTER:

Screenshot 2023-02-23 at 1 21 30 PM

and AFTER with DEBUG=1:

Screenshot 2023-02-23 at 1 15 50 PM


savil:
* main
  * **PR #675** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/675" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/675?utm_source=stack-comment).

savil:
Should we only display this detailed error message when `DEVBOX_DEBUG=1` is true?

savil:
I chose to include more information so that if a user reports it, then it is much easier to see the problem.

savil:
But it could be overwhelming to a user who is new to nix, so we could show a simpler error like:
```
A nix command failed with error: exit status 1. Run with DEVBOX_DEBUG=1 for a detailed error message and please report to https://github.com/jetpack-io/devbox/issues/new/choose
```

internal/boxcli/midcobra/midcobra.go
```diff
@@ -64,7 +64,7 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	err := ex.cmd.ExecuteContext(ctx)
 
 	var postRunErr error
-	var userExecErr *usererr.UserExecError
+	var userExecErr *usererr.UserExitError
```
mikeland86, line 67: To avoid repetition, I wonder is just `usererr.ExitError`

mikeland86 (approved):
This is really good (and needed!) 

I do wonder if we can simplify a bit to avoid having custom error structs. Specifically, we could have

```golang
type UserExitError exec.ExitError
```

and not use any special error at all for generic exit errors. (this way, if we forget to wrap we still get good info!)

In the debug middleware we can have additional handling depending on the type:

```golang
var userExitError *UserExitError
var genericExitErr *exec.ExitError
if errors.As(err, userExitError) {
 //print very detailed stuff since it's their error
} else if errors.As(err, genericExitErr) {
 // print less but enough detail. If DEVBOX_DEBUG=1 is true, print everything
}
```

For printing the command, instead of using a custom struct you can wrap:

```golang
c := exec.Command("bad-echo", "hello")
return errors.Wrapf(c.Run(), "command: %s", c.String())
```

This should print something like `Error: command: bad-echo hello: exec: "bad-echo": executable file not found in $PATH` which is pretty helpful! I think this would be a more "golang" approach.

Anyway, don't want to block this PR so pre-approving. We can always follow up if you think the suggestions are valid.

savil:
@mikeland86 yes! I like your approach a lot. I was also feeling that this is a bit heavy for what I am trying to achieve. I'll update to that.

internal/boxcli/usererr/exiterr.go
```diff
@@ -1,12 +1,12 @@
...
 import (
-	"os/exec"
-
 	"errors"
+	"os/exec"
 )
 
-type UserExecError struct {
+// ExitError is an ExitError for a command run on behalf of a user
+type ExitError struct {
```
gcurtis, line 9: I'm curious why we have a new type for this. Why not just use `exec.ExitError` directly?

gcurtis (approved):

internal/boxcli/usererr/exiterr.go
mikeland86: @gcurtis the idea is that a user exit error and a generic exit error are treated differently in terms of what we show the user. For errors that were a result of an error caused by user code we want to always show as much info as possible. For non-user exit errors we only show all info if DEVBOX_DEBUG is true.

internal/boxcli/usererr/exiterr.go
```diff
@@ -19,22 +19,22 @@ func NewExecError(source error) error {
...
 		return source
 	}
-	return &UserExecError{
+	return &ExitError{
 		err: exitErr,
 	}
 }
 
-func (e *UserExecError) Error() string {
+func (e *ExitError) Error() string {
```
mikeland86, line 27: Not from this PR, but if `ExitError` is just an exec.ExitError you don't need to reimplement all this stuff.

internal/boxcli/usererr/exiterr.go
savil: This type is used for errors that are caused by the user's program, as opposed to a devbox CLI error. For example, when doing `devbox run -- <program>` the user's program may exit with an error. So, if we can detect that circumstance, we wrap the `exec.Command`'s error in this special type and then render the error message accordingly.

internal/boxcli/midcobra/debug.go
```diff
@@ -57,14 +59,17 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
...
-			color.Red("\nError: " + runErr.Error() + "\n\n")
+			color.Red(runErr.Error() + "\n\n")
 		}
-	} else {
-		fmt.Printf("Error: %v\n", runErr)
 	}
 
 	st := debug.EarliestStackTrace(runErr)
-	debug.Log("Error: %v\nExecutionID:%s\n%+v\n", runErr, d.executionID, st)
+	color.Red(fmt.Sprintf("Error: %v\n\n", runErr))
```
mikeland86, line 67: This should go to `cmd.ErrOrStderr()` instead.

internal/boxcli/midcobra/debug.go
```diff
@@ -57,14 +59,17 @@ func (d *DebugMiddleware) postRun(cmd *cobra.Command, args []string, runErr erro
 		if usererr.IsWarning(runErr) {
 			ux.Fwarning(cmd.ErrOrStderr(), runErr.Error())
 		} else {
-			color.Red("\nError: " + runErr.Error() + "\n\n")
+			color.Red(runErr.Error() + "\n\n")
```
mikeland86, line 62: This has been changed to go to cmd.ErrOrStderr() (you need to rebase)

mikeland86 (approved):

PR #678, opened 2023-02-23
[nix] create common ExperimentalFlags function

savil:
## Summary

I noticed that print-dev-env and nix profile commands were using a different set
of nix experimental features.

I think we should unify these so that they're writing to the store in a common fashion.

**Implementation Note**
It appears that: `"--extra-experimental-features", "nix-command flakes"`, is different from `"--option", "experimental-features", "nix-command flakes"` for our `nix print-dev-env` testscripts.

We must have the latter. The former is insufficient and not needed.

This confuses me because the docs suggest that `--extra-experimental-features` should append these experimental features, whereas `--option` replaces the experimental features:
>  --option name value
 Set the Nix configuration setting name to value (overriding nix.conf).

and

>  --extra-experimental-features value
 Append to the experimental-features setting.

Concretely, this works:
```
		"--extra-experimental-features", "ca-derivations",
		"--option", "experimental-features", "nix-command flakes",
```
but this does not work:
```
"--option", "experimental-features", "ca-derivations nix-command flakes",
```
and nor does this:
```
		"--extra-experimental-features", "nix-command flakes",
		"--extra-experimental-features", "ca-derivations",
```

## How was it tested?

`go test ./...`

in `examples/testdata/go/go-1.19` with flakes feature flag on for all commands:
```
> devbox shell
(devbox)> devbox add vim emacs
(devbox)> source activate shell...

(devbox)> devbox rm emacs
```

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -87,9 +87,8 @@ func flakesPkgInfo(nixpkgsCommit, pkg string) (*Info, bool) {
 		exactPackage = fmt.Sprintf("nixpkgs#%s", pkg)
 	}
 
-	cmd := exec.Command("nix", "search",
-		"--extra-experimental-features", "nix-command flakes",
-		"--json", exactPackage)
+	cmd := exec.Command("nix", "search", "--json", exactPackage)
+	cmd.Args = append(cmd.Args, ExperimentalFlags()...)
 	return pkgInfo(cmd, pkg)
 }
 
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #678** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/678" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/678?utm_source=stack-comment).

internal/nix/nix.go
```diff
@@ -87,9 +87,8 @@ func flakesPkgInfo(nixpkgsCommit, pkg string) (*Info, bool) {
 		exactPackage = fmt.Sprintf("nixpkgs#%s", pkg)
 	}
 
-	cmd := exec.Command("nix", "search",
-		"--extra-experimental-features", "nix-command flakes",
-		"--json", exactPackage)
+	cmd := exec.Command("nix", "search", "--json", exactPackage)
```
mikeland86, line 90: syntax candy suggestion:
```golang
cmd := exec.Command("nix", nix.Args("search", "--json", exactPackage))
```
or 

```golang
cmd := nix.Command("nix", "search", "--json", exactPackage)
```

```golang
func Args(args ...string) []string {
	return append(args,
		"--extra-experimental-features", "ca-derivations",
		"--option", "experimental-features", "nix-command flakes",
	)
}
```

mikeland86 (approved):

PR #679, opened 2023-02-23
add env to configuration

Lagoja:
## Summary

Add Details on the `env` portion of the devbox.json to the documentation

## How was it tested?

Localhost

### DIFF
--- a/docs/app/docs/configuration.md
+++ b/docs/app/docs/configuration.md
@@ -8,6 +8,7 @@ Your devbox configuration is stored in a `devbox.json` file, located in your pro
 ```json
 {
     "packages": [],
+    "env": {},
     "shell": {
         "init_hook": "...",
         "scripts": {}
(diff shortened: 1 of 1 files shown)


docs/app/docs/configuration.md
```diff
@@ -24,6 +25,23 @@ This is a list of Nix packages that should be installed in your Devbox shell and
...
+
+```json
+{
+    "env": {
+        "FOO": "bar"
+    }
+}
+```
+
+Currently, you can only set values using string literals, `$PWD`, and `$PATH`. Any other values will environment variables will not be expanded when starting your shell.
```
savil, line 42: edit needed: Any other values ~will~ with environment variables will not be expanded when starting your shell.

savil (approved):

LucilleH (approved):

PR #681, opened 2023-02-24
[flakes] save commit-hash to file, and inspect prior to prefetching nixpkgs

savil:
## Summary

This PR saves the nixpkgs commit hash to a file at `~/.cache/devbox/nixpkgs.txt`.
We then inspect this file for the nixpkgs commit hash of the devbox project, 
and if it is present then we skip the `nix flake prefetch nixpkgs/<commit>` call.

Risks:
This PR assumes that the user is not actively pruning the files downloaded by nix.
For example, if the user does `nix store gc`, then it is possible that the nixpkgs downloaded 
gets removed. Maybe.

Alternate approach:
I tried waiting 100 milliseconds to see if the `nix flake prefetch` command is still
working as a proxy for deciding if the nixpkgs is being downloaded or not. However,
this does NOT work because the nix progress indicator fails to render if we mess with
the output from `nix flake prefetch`. 

## How was it tested?

1. edited devbox.json to have a new nixpkgs commit hash, and started `devbox shell` . Observed the flake prefetch output as before.
2. re-ran `devbox shell` and this time there was no output about nixpkgs being downloaded. As expected.


savil:
* main
  * **PR #681** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/681" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/681?utm_source=stack-comment).

internal/impl/packages.go
```diff
@@ -244,7 +252,48 @@ func (d *Devbox) ensureNixpkgsPrefetched() error {
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (d *Devbox) saveToNixpkgsCommitFile() error {
+	path := d.nixpkgsCommitFilePath()
+
+	if err := os.Mkdir(filepath.Dir(path), 0755); err != nil && !errors.Is(err, fs.ErrExist) {
```
mikeland86, line 280: Should this be `MkdirAll`?

internal/impl/packages.go
```diff
@@ -244,7 +252,48 @@ func (d *Devbox) ensureNixpkgsPrefetched() error {
...
+		return errors.WithStack(err)
+	}
+
+	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer file.Close()
+
+	_, err = fmt.Fprintf(file, "%s", d.cfg.Nixpkgs.Commit)
```
mikeland86, line 290: The approach I had in mind was a bit different. Instead of saving only the nix commit (which doesn't guarantee that the nixpackges is actually downloaded) I was thinking of saving a map:

commit -> directory

by saving this we can check the directory actually exists. And we can make it a bit more general by calling file `nixpkgs.json`

and having:

```json
"commit_map" : {
  "hash-123": "/nix/store/..."
}
```

savil: oh that's a good idea. It'll get rid of the Risk I outline in the PR summary.

mikeland86 (approved):
Nice!

PR #691, opened 2023-02-25
devbox: unset TMPDIR in shell/run environment

gcurtis:
## Summary

Nix sets TMPDIR to its own path when building an environment and then prints it with `print-dev-env`. Devbox "leaks" TMPDIR, meaning it takes the TMPDIR from the parent shell and overwrites the Nix TMPDIR. However, this doesn't work when TMPDIR isn't set in the parent shell, which is the case in WSL and Docker.

This is a quick fix for just TMPDIR. We should double check the other variables that `nix develop` unsets when there's more time to test.

## How was it tested?

```
$ unset TMPDIR
$ devbox shell
$ echo $TMPDIR
<nothing set>
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -780,6 +780,15 @@ func (d *Devbox) computeNixEnv(setFullPath bool) (map[string]string, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


loreto (approved):
LGTM

PR #697, opened 2023-02-27
[flakes] ensure all golang tests pass when flakes flag is enabled

savil:
## Summary

Needed to make two fixes:

1. `RmTest`:
here, the test was failing because with flakes flag we now check that the user 
has the package installed previously and so we do `devbox run <command>` to trigger
package installation as a side-effect.

2. `ShellTest`:
Flakes no longer works with there being no default shell, so we call that out. But
in a subsequent PR, I'll revisit this since I think we can get rid of this block.
Wanted to keep the scope of this PR to just enabling these tests to pass.

## How was it tested?

enabled the Flakes feature flag in the code and ran `go test ./...`


savil:
* main
  * **PR #697** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/697" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/697?utm_source=stack-comment).

mikeland86 (approved):
Should we add a new job in cli-test.yaml that runs all tests with flakes enabled?

savil:
good idea

savil:
@mikeland86 one problem is that doing `DEVBOX_FEATURE_FLAKES=1 go test ./...` appears insufficient for having the tests run in flakes mode.

PR #700, opened 2023-02-28
Improve handling when no SHELL env-var is defined

savil:
## Summary

Previously, if `SHELL` env-var was not set, we'd return an error from `nix.DetectShell`
and initialize an empty struct `nix.Shell{}`. Then, in `Shell.Run` we'd see that `Shell.binPath` is empty
and call `nix-shell`.

This has the following problems:
1. Initializing an empty `nix.Shell{}` means that we drop all the other `ShellOptions`
in the constructor.
2. In `Shell.Run`, we'd use `nix-shell` but fail if the flakes feature was not enabled. This `nix-shell` also loses all the print-dev-env work that we have done for Unified-Env.

To solve and/or mitigate these issues, I propose the following changes:
- Rename `nix.Shell` to `nix.DevboxShell` to make it clear that its not a raw nix shell. TODO for future to move this out of `nix` package.
- Rename `nix.DetectShell` to `nix.NewDevboxShell` to make clear that it is a constructor function. Decompose its functionality into helper functions `shellPath` and `initShellBinaryFields`.
- If `SHELL` is undefined, ~we loop over a list of common shells to see if they are accessible in the `PATH`.~ we use the bash from nix.
- Failing to find ~any recognizable shell in `PATH`,~ bash from nix, we fail the program with an error.
- Remove the fallback to `nix-shell` in `Shell.Run`. This means we _always_ use the Unified-Env codepath.
- Update `shell_test` for Flakes feature, since it now finds a shell from PATH and uses that (previously, would error).

## How was it tested?

in `examples/testdata/go/go-1.19`:
```
❯ DEVBOX_DEBUG=0 SHELL= DEVBOX_FEATURE_FLAKES=0 devbox shell
Starting a devbox shell...
(devbox) Savil-Srivastavas-MacBook-Pro:go-1.19 savil$ echo $SHELL
/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16/bin/bash
(devbox) Savil-Srivastavas-MacBook-Pro:go-1.19 savil$ exit
exit

❯ DEVBOX_DEBUG=0 SHELL= DEVBOX_FEATURE_FLAKES=1 devbox shell

Installing package: go_1_19.

[1/1] go_1_19
[1/1] go_1_19: Success
Starting a devbox shell...
(devbox) Savil-Srivastavas-MacBook-Pro:go-1.19 savil$ echo $SHELL
/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16/bin/bash
(devbox) Savil-Srivastavas-MacBook-Pro:go-1.19 savil$
```

Also: `go test ./...` with flakes feature enabled and disabled.

### DIFF
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -6,6 +6,7 @@ package nix
 import (
 	"bytes"
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"io/fs"
 	"os"
@@ -19,6 +20,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/debug"
+	"go.jetpack.io/devbox/internal/planner/plansdk"
 	"go.jetpack.io/devbox/internal/xdg"
 )
 
@@ -73,8 +75,8 @@ type ShellOption func(*Shell)
 
 // NewShell initializes the Shell struct so it can be used to start a shell environment
 // for the devbox project.
-func NewShell(opts ...ShellOption) (*Shell, error) {
-	shPath, err := shellPath()
+func NewShell(nixpkgsCommitHash string, opts ...ShellOption) (*Shell, error) {
+	shPath, err := shellPath(nixpkgsCommitHash)
 	if err != nil {
 		return nil, err
 	}
@@ -90,7 +92,7 @@ func NewShell(opts ...ShellOption) (*Shell, error) {
 }
 
 // shellPath returns the path to a shell binary, or error if none found.
-func shellPath() (path string, err error) {
+func shellPath(nixpkgsCommitHash string) (path string, err error) {
 	defer func() {
 		if err != nil {
 			path = filepath.Clean(path)
@@ -104,19 +106,45 @@ func shellPath() (path string, err error) {
 		return path, nil
 	}
 
-	// Second, attempt to find the path to one of these common shells
-	shells := []string{"sh", "bash", "zsh", "ksh", "fish", "dash", "ash"}
-	for _, shell := range shells {
-		var err error
-		path, err = exec.LookPath(shell)
+	// Second, fallback to using the bash that nix uses by default.
+
+	var bashNixStorePath string // of the form /nix/store/{hash}-bash-{version}/
+	if featureflag.Flakes.Enabled() {
+		cmd := exec.Command(
+			"nix", "eval", "--raw",
+			fmt.Sprintf("%s#bash", FlakeNixpkgs(nixpkgsCommitHash)),
+		)
+		cmd.Args = append(cmd.Args, ExperimentalFlags()...)
+		out, err := cmd.Output()
 		if err != nil {
-			if errors.Is(err, exec.ErrNotFound) {
-				continue
-			}
 			return "", errors.WithStack(err)
 		}
-		debug.Log("SHELL env var is undefined. Falling back to a common shell accessible in PATH: %s\n", path)
-		return path, nil
+		bashNixStorePath = string(out)
+	} else {
+		nixpkgsInfo, err := plansdk.GetNixpkgsInfo(nixpkgsCommitHash)
+		if err != nil {
+			return "", err
+		}
+		expr := fmt.Sprintf(
+			"let pkgs = import (fetchTarball { url = \"%s\"; }) {}; in {inherit(pkgs.bash) outPath; }",
+			nixpkgsInfo.URL,
+		)
+		cmd := exec.Command("nix-instantiate", "--eval", "--strict",
+			"--json",
+			"--expr", expr,
+		)
+		out, err := cmd.Output()
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+		if err := json.Unmarshal(out, &bashNixStorePath); err != nil {
+			return "", errors.WithStack(err)
+		}
+	}
+
+	if bashNixStorePath != "" {
+		// the output is the raw path to the bash installation in the /nix/store
+		return fmt.Sprintf("%s/bin/bash", bashNixStorePath), nil
 	}
 
 	// Else, return an error
--- a/internal/nix/shell.go
+++ b/internal/nix/shell.go
@@ -159,7 +159,7 @@ func shellPath(nixpkgsCommitHash string) (path string, err error) {
 // for the devbox shell.
 func initShellBinaryFields(path string) *DevboxShell {
 
-	sh := &DevboxShell{binPath: path}
+	shell := &DevboxShell{binPath: path}
 	base := filepath.Base(path)
 	// Login shell
 	if base[0] == '-' {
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #700** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/700" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/700?utm_source=stack-comment).

internal/nix/shell.go
```diff
@@ -70,14 +71,63 @@ type Shell struct {
...
 type ShellOption func(*Shell)
 
-// DetectShell attempts to determine the user's default shell.
-func DetectShell(opts ...ShellOption) (*Shell, error) {
-	path := os.Getenv("SHELL")
-	if path == "" {
-		return nil, errors.New("unable to detect the current shell")
+// NewShell initializes the Shell struct so it can be used to start a shell environment
+// for the devbox project.
+func NewShell(opts ...ShellOption) (*Shell, error) {
```
gcurtis, line 76: What about `UserShell` to differentiate it from a Nix shell which is always `bash`? I wanted to avoid `New` because it doesn't actually create a shell yet.

internal/nix/shell.go
```diff
@@ -70,14 +71,63 @@ type Shell struct {
...
+
+	// First, check the SHELL environment variable.
+	path = os.Getenv("SHELL")
+	if path != "" {
+		debug.Log("Using SHELL env var for shell binary path: %s\n", path)
+		return path, nil
+	}
+
+	// Second, attempt to find the path to one of these common shells
+	shells := []string{"sh", "bash", "zsh", "ksh", "fish", "dash", "ash"}
```
gcurtis, line 108: Instead of searching in PATH what if we just use Nix's bash shell as the fallback? Otherwise we might find `/bin/sh` which could be something like dash, which isn't great for interactive shells.

internal/nix/shell.go
savil: yeah, the problem is that the word "shell" is overloaded here, because it could refer to both the struct `Shell` and the underlying system shell. When I write `NewShell`, I am using it in the sense of `New{StructName}` as a struct constructor.

To mitigate this confusion, would it make sense to rename the struct to `type DevboxShell struct`? This function would then become `NewDevboxShell`

internal/nix/shell.go
savil: how does one find the path to `nix`'s bash shell? 

I'd considered this approach initially. 

The problem I ran into is that the straightforward approach for this would be to do `nix develop`. But that needs other processing to make work properly. This would diverge logic from the regular implementation, which is what I was seeking to avoid.

Alternatively, we can get the path to bash in the `/nix/store` and use that here. Not fully sure how to do that, but I can look into it.

internal/nix/shell.go
savil: One reason I intentionally _do_ want to use the word `New` is to make it clear that we have this constructor function, so that callers do not do `nix.Shell{}` and feel they have a struct that is usable.

internal/nix/shell.go
savil: I think I found it: `nix eval --raw nixpkgs/<commit>#bash`

```
❯ /nix/var/nix/profiles/default/bin/nix eval --raw nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04#bash
/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16
```
and to confirm:
```
❯ /nix/var/nix/profiles/default/bin/nix develop .devbox/gen/flake/flake.nix
path '/Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/gen/flake/flake.nix' does not contain a 'flake.nix', searching up
warning: Git tree '/Users/savil/code/jetpack/devbox/examples/testdata/go/go-1.19/.devbox/gen/flake' is dirty

Savil-Srivastavas-MacBook-Pro:go-1.19 savil$ which bash
/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16/bin/bash

Savil-Srivastavas-MacBook-Pro:go-1.19 savil$ echo $SHELL
/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16/bin/bash
```

savil: oh need to find the non-flakes way to do this too...

savil: for non flakes:
```
❯ /nix/var/nix/profiles/default/bin/nix-instantiate --eval --strict --json --expr 'let
pkgs = import (fetchTarball { url = "https://github.com/nixos/nixpkgs/archive/52e3e80afff4b16ccb7c52e9f0f5220552f03d04.tar.gz"; })
{};
in {inherit(pkgs.bash) outPath; }'
"/nix/store/1bsjl5incfnszv7scdh4d02sh45vw2w1-bash-5.1-p16"
```

savil: useful command to get the attributes of a package:
```
/nix/var/nix/profiles/default/bin/nix-instantiate --eval --strict --json --expr 'let
pkgs = import (fetchTarball { url = "https://github.com/nixos/nixpkgs/archive/52e3e80afff4b16ccb7c52e9f0f5220552f03d04.tar.gz"; })
{};
in pkgs.lib.attrNames pkgs.bash'
```

savil (commented):
@gcurtis updated to use nix's bash.

I haven't yet changed any function names, since that discussion is still active.

internal/nix/shell.go
gcurtis: > yeah, the problem is that the word "shell" is overloaded here, because it could refer to both the struct Shell and the underlying system shell.

The original thought was to have the struct represent a shell, in the same way that `exec.Cmd` does. You call `DetectShell` and you get a shell that you can run or inspect. If you'd like to use `New`, I'd prefer if we could still keep it descriptive to what it's doing -- guessing the user's shell vs. using a default. `NewUserShell` or `NewDefaultShell`, etc.

> To mitigate this confusion, would it make sense to rename the struct to type DevboxShell struct

I agree that the name doesn't really make sense anymore. It was originally intended to be just a Nix shell, with the Devbox-specific logic going in the `devbox` package. IMO the best option would be to just move it out of the Nix package since it doesn't have anything to do with Nix anymore. I know what's a larger undertaking though, so just a rename to `DevboxShell` would be okay with me.

> One reason I intentionally do want to use the word New is to make it clear that we have this constructor function, so that callers do not do nix.Shell{} and feel they have a struct that is usable.

The zero value of `Shell` is usable - it starts a bash Nix shell. The original intention was for a `nix.Shell{}` to be usable and the caller could further modify the fields they needed to give it a different environment or init script. `DetectShell` was a convenience to go through the work of configuring those fields to point to the user's preferred shell and init files. I'd still prefer to keep the zero-value working as a default Nix/bash shell if possible. This is especially useful for tests.

savil: I've made the following updates:
1. Rename `nix.Shell` to `nix.DevboxShell`.
2. Added a TODO to move `DevboxShell` to `impl` package
3. Renamed `NewShell` to `NewDevboxShell`.

In a future PR, one of us can refactor out code to solve the TODO of (2) and also split it so that there's once again a rudimentary `nix.Shell` struct as before. 

I'm not convinced that its a good idea to have a usable raw struct, since it leads to the kinds of problems this PR is solving. Bugs like [this line](https://github.com/jetpack-io/devbox/blob/main/internal/impl/devbox.go#L277) being introduced which lose all the other shell options that should be applied to it. For tests, one can have a `NewTestShell` or similar constructor function if needed.

gcurtis (approved):

internal/nix/shell.go
gcurtis: That wasn't really a bug though. The intended behavior was to start a default Nix shell if we couldn't figure out the shell type. You can't set any of the shell options unless you know which flags and environment variables the specific shell uses. For example, how do you set the PATH if you don't know how to point the shell to the devbox rcfile?

savil: Maybe I was not quite understanding the desired behavior then. To me, it seemed better to fail if we can't resolve the shell such that we can apply all the plugins and devbox-project settings appropriately...which is why this PR errors out. I think the new logic to detect the nix-bash-shell does make that error scenario pretty unlikely though.

PR #706, opened 2023-02-28
nix: inherit all environment variables in shell

gcurtis:
This is a long description, but I wanted to document the exact changes being made to Devbox environments for reference.

## Context

Before release v0.4.0, devbox handled environment variables differently depending on the command:

- `devbox shell` would start with a clean environment. Nothing was inherited from the parent shell (but shellrc files would still be sourced)
- `devbox shell -- cmd` would inherit the environment.
- `devbox run` would inherit the environment.

With v0.4.0, all commands adopted the `devbox shell` behavior of starting with a clean environment. However, this broke some existing users and the workaround of manually specifying environment variables to inherit would be cumbersome. As discussed, this change flips the behavior such that all commands will now inherit their parent shell's environment. Eventually there will be a flag to enable an "isolated" mode to restore the previous behavior.

## Changes to `devbox shell` and `devbox run` environments

Devbox environments are now built with environment variables from the following steps:

1. Variables exported in the user's `~/.bashrc`, `~/.zshrc`, etc.
2. Variables from the current environment except for PWD, OLDPWD, SHLVL, and SHELL.
3. Variables from the `nix print-dev-env` environment except for BASHOPTS, HOME, NIX_BUILD_TOP, NIX_ENFORCE_PURITY, NIX_LOG_FD, NIX_REMOTE, PPID, SHELL, SHELLOPTS, TEMP, TEMPDIR, TERM, TMP, TMPDIR, TZ, and UID. The SSL_CERT_FILE variable is also ignored if it's set to `/no-cert-file.crt`.
4. Variables from Devbox plugins.
5. PATH is set to the concatenation of the PATHs from step 4, step 3, and step 2 (in that order). The PATH from the user's init files in step 1 is overwritten. The final PATH is cleaned and deduplicated.
5. HISTFILE is set to a Devbox project-specific file.
6. PS1 is set to "(devbox) $PS1"

Each step adds to the previous step, overwriting any keys that already exist.

## Bug fixes and improvements

- As suggested by @ipince, `shellrc.tmpl` simply sources the user's rcfile instead of copying it.
- The result of `PrintEnv` now escapes all environment variable values. This means that values such as `FOO=$(echo bar)` will become `FOO="\$(echo bar)"` instead of being evaluated by the shell. Allowing variables and command substitution would require implementing a full parser to determine which characters to escape.
- Some unquoted variables in `shellrc.tmpl` are now quoted so that devbox directories with spaces work.
- The `cd` commands in `shellrc.tmpl` exit on failure to prevent running further commands in the wrong directory.


internal/impl/devbox.go
```diff
@@ -1025,63 +1035,43 @@ func commandExists(command string) bool {
...
-	"TMPDIR":  true,
-	"TEMPDIR": true,
+// ignoreCurrentEnvVar contains environment variables that Devbox should remove
+// from the slice of [os.Environ] variables before sourcing them. These are
+// variables that are set automatically by a new shell.
+var ignoreCurrentEnvVar = map[string]bool{
+	// Devbox may change the working directory of the shell, so using the
+	// original PWD and OLDPWD would be wrong.
+	"PWD":    true,
+	"OLDPWD": true,
```
savil, line 1045: what scenario leads to working directory changing?

I thought doing `devbox --config=/path/to/config shell` would still start you in the current directory but using the `devbox.json` of that config path. But maybe there's some other scenario I'm not considering.

internal/nix/shell.go
```diff
@@ -582,62 +607,33 @@ func toKeepArgs(env []string, allowList map[string]bool) []string {
...
+				continue
 			}
-		}
-
-		if keep {
-			cleaned = append(cleaned, path)
+			if !seen[path] {
+				cleaned = append(cleaned, path)
+			}
+			seen[path] = true
```
savil, line 635: below: `filepath.Join(cleaned...)`

savil (approved):
awesome! This is clean and easy to understand.

> The result of PrintEnv now escapes all environment variable values.

I feel I'm missing something obvious. Could you explain the motivation for this again? perhaps an example may help.

gcurtis:
@savil the original motivation was to properly handle variables with special characters. We were previously doing `fmt.Sprintf("export %s=%q", k, v)` to escape quotes, but that can change the value of a variable if you don't escape the other special characters. For example, if you have:

```sh
FOO="$(ls "dir 01")" # lists a directory named `dir 01`
```

and then you only escape quotes, you get:

```sh
FOO="$(ls \"dir 01\")" # lists a directory named `"dir 01"`
```

Because quotes within a command substitution shouldn't be escaped, we'd need to implement a parser. Instead, if we just say all env var values need to be literals then it becomes easy. We just do:

```sh
FOO="\$(ls \"dir 01\")" # doesn't list anything, value is `$(ls "dir 01")`
```

internal/impl/devbox.go
gcurtis: I think it's primarily plugin + init hooks that always run from the devbox.json directory.

internal/nix/shell.go
gcurtis: Are you referring to line 638? That's joining on the list separator (`:`) instead of the path separator.

PR #707, opened 2023-02-28
ci: set NIX_PATH after installing Nix

gcurtis:
## Summary

For some reason Nix stopped being able to find nixpkgs by default after a fresh install. Explicitly setting `NIX_PATH` seems to fix it.

## How was it tested?

CI


mikeland86 (approved):
Does this affect end users in any way? I can't seem to repro locally

gcurtis:
@mikeland86 I'm not sure. I could only reproduce it by doing a multi-user install (as not root) on a fresh Ubuntu image without any XDG env vars set. Maybe it only affects Linux?

PR #712, opened 2023-03-01
[CICD] Fix mac tests

mikeland86:
## Summary

This fixes mac os tests by setting `NIX_PATH` on the test itself. This should not be needed so it makes me a bit nervous we are covering up another issue.

Added auto-install test for mac

## How was it tested?

CICD


loreto (approved):
LGTM

PR #717, opened 2023-03-03
[vscode extension] Fixed autorun shell on new terminal open bug

mohsenari:
## Summary
Fixed a bug where devbox extension would run `devbox shell` when opening a new terminal in vscode even if there was no devbox.json present in the workspace.
The change would make sure there is a devbox.json file present before running devbox shell in the newly opened terminal.

## How was it tested?
- Open vscode from devbox/vscode-extension
- from run and debug on the left bar click run extension 
- in the newly opened window create a devbox.json file and open a new terminal in vscode
- confirm `devbox shell` is run when devbox.json file is present and vice versa


mohsenari:
After this is merged the next step is to release a new version for the extension. 
cc @Lagoja

LucilleH (approved):

PR #721, opened 2023-03-06
Fix fish shell after #706

### LINKED ISSUE #719: [Bug]: Devbox 0.42 error on fish shell. (Error: execute shellrc template: shellrc_fish)
**Current Behavior (bug)**
`devbox shell` doesn't work when the default shell is set to `fish`, but it works fine when the default shell is set to `bash`. I tried to uninstall and reinstall, but it doesn't work.

The error output:

> `~ devbox shell`
> Ensuring packages are installed.
> Starting a devbox shell...
> Error: execute shellrc template: template: shellrc_fish:20:7: executing "shellrc_fish" at <.UnifiedEnv>: > can't evaluate field UnifiedEnv in type struct { ProjectDir string; OriginalInit string; OriginalInitPath string; UserHook string; PluginInitHook string; PathPrepend string; ScriptCommand string; ShellStartTime string; HistoryFile string; ExportEnv string }

**Expected Behavior (fix)**

**Additional context**
> `devbox version -v`

```text
Version:     0.4.2
Platform:    linux_amd64
Commit:      c249fa1da3a24eb686220da378bc9dcdc6d49680
Commit Time: 2023-03-02T21:16:52Z
Go Version:  go1.19.6
```

> *`devbox.json`*

```text
{
  "packages": [],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
  }
}
```

ipince:
## Summary
* Applies the same changes from #706 to the fish shellrc.
* Ignores the `_` variable. It's read-only and `fish` complains if we try to write it.
* Added a note in `shellrc` so we remember do edit `shellrc_fish` too. If we had tests for fish we could remove this silly note.
* Alphabetized vars so the generated shellrc is a bit cleaner (imo).

Fixes #719 

## How was it tested?
```
SHELL=fish ./devbox shell
```

I'll add some shellrc_fish tests in a future PR

### DIFF
--- a/internal/nix/shellrc.tmpl
+++ b/internal/nix/shellrc.tmpl
@@ -1,5 +1,7 @@
 {{- /*
 
+IF YOU EDIT THIS FILE, REMEMBER TO MAKE EQUIVALENT CHANGES TO shellrc_fish.tmpl
+
 This template defines the shellrc file that the devbox shell will run at
 startup.
 
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -1048,6 +1048,9 @@ var ignoreCurrentEnvVar = map[string]bool{
 
 	// The parent shell isn't guaranteed to be the same as the Devbox shell.
 	"SHELL": true,
+
+	// The "_" variable is read-only, so we ignore it to avoid attempting to write it later.
+	"_": true,
 }
 
 // ignoreDevEnvVar contains environment variables that Devbox should remove from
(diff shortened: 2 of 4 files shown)


internal/nix/shellrc.tmpl
```diff
@@ -1,5 +1,7 @@
 {{- /*
 
+IF YOU EDIT THIS FILE, REMEMBER TO MAKE EQUIVALENT CHANGES TO shellrc_fish.tmpl
```
gcurtis, line 3: Who would ever forget to do this? 😅

gcurtis (approved):

PR #725, opened 2023-03-07
Install Devbox Using Flakes

Lagoja:
## Context
We currently publish Devbox as a package in the `nixpkg` repository, but this version is often far behind our latest released version. For example: 22.11 currently has `devbox 0.1.0`, while unstable has `devbox 0.3.4` (we're on 0.4.2 as of this PR). 

By publishing a flake, we could make it easier for developers using NixOS, Nixpkg, or Home Manager to install Devbox on their system. Developers could add our repo + release tags as an input, and then have access to the latest versions as soon as we publish them.

With this installed, developers using Nix could install devbox with something like

```bash
nix build github:jetpack-io/devbox/latest
```

or adding 

```nix
{
  inputs = {
         devbox.url = "github:jetpack-io/devbox/latest";
   ...
  };
...
}
```

to their flake.

## Changes / Limitations with this PR:

1. **go.mod Hashes**  - We need to provide Nix with a sha256 of our vendored dependencies in order to build Devbox. We can do this by either adding a `vendorSHA256` to our Flake, or using `gomod2nix` to generate a module with the hashes of all the dependencies in our `go.mod` file. This PR goes with the `gomod2nix` approach. 

2. **Version Number** - The version number currently needs to be set and updated manually in the Flake. This is because Flake outputs don't have access to the repositories branches or tags. We could pull this version from a file in the repo, or by running a script that reads the tags. 

3. **Branch/Refs** - To make it easy to install the latest version, we'll need a ref in our git repo for `latest` or `stable`. This could be a branch or a tag, and should be updated whenever we make a new full release.

## How was it tested?

- [x]  aarch64-linux
- [x]  x86_64-linux
- [x]  aarch64-darwin
- [x]  x86_64-darwin
- [x]  WSL2

With `nix command` and `flakes` enabled: 

* `nix build`
* `nix flake`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,56 @@
+{
+  "nodes": {
+    "devbox-source": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1677791812,
+        "narHash": "sha256-9aGmX21TUINGDA3NB/OrGXi8W4i+b8fKS8ft8GtQxNM=",
+        "owner": "jetpack-io",
+        "repo": "devbox",
+        "rev": "c249fa1da3a24eb686220da378bc9dcdc6d49680",
+        "type": "github"
+      },
+      "original": {
+        "owner": "jetpack-io",
+        "ref": "0.4.2",
+        "repo": "devbox",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1676283394,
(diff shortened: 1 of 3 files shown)


adamdicarlo0:
I made https://github.com/adamdicarlo/devbox-nix-flake to use until this PR is completed - in case that helps anyone else. (Currently, devbox in nixpkgs is up to date, but this flake allows me to get the very latest release before nixpkgs gets updated.)

PR #729, opened 2023-03-07
Remove obsolete code after unified env

ipince:
## Summary
Should be a no-op.  Delete ALL the code.

## How was it tested?
CICD


savil (approved):
yay!

PR #734, opened 2023-03-08
plansdk: get cloud cache from environment

gcurtis:
Remove the hardcoded IP and instead get the base URL from the `DEVBOX_CACHE` environment variable.


loreto (approved):
LGTM

PR #735, opened 2023-03-08
Remove more dead code

ipince:
## Summary
TSIA

## How was it tested?
CICD

### DIFF
--- a/internal/impl/shell.go
+++ b/internal/impl/shell.go
@@ -18,7 +18,6 @@ import (
 	"github.com/alessio/shellescape"
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
-	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/planner/plansdk"
@@ -63,9 +62,6 @@ type DevboxShell struct {
 
 	// UserInitHook contains commands that will run at shell startup.
 	UserInitHook string
-
-	ScriptName    string
-	ScriptCommand string
 
 	// profileDir is the absolute path to the directory storing the nix-profile
 	profileDir  string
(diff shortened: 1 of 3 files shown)


internal/impl/shell.go
```diff
@@ -289,17 +285,10 @@ func (s *DevboxShell) Run() error {
 	// This could be from one of the generated shellrc commands, but more likely is from
 	// a user's command or script. So, we want to return nil for this.
 	if exitErr := (&exec.ExitError{}); errors.As(err, &exitErr) {
-
-		// The exception to the previous comment is if we are executing a shell script
-		// via `devbox run` or the deprecated `devbox shell -- <command>`. In this case,
-		// we do want to return the exit code of the script that was run.
```
savil, line 295: do we still have this behavior (elsewhere) of returning the exit code from a user's script?

internal/impl/shell.go
```diff
@@ -354,11 +343,6 @@ func (s *DevboxShell) writeDevboxShellrc() (path string, err error) {
 		}
 	}()
 
-	pathPrepend := s.profileDir + "/bin"
-	if s.pkgConfigDir != "" {
-		pathPrepend = s.pkgConfigDir + ":" + pathPrepend
-	}
```
savil, line 360: I understand that the profileDir should no longer be added to PATH
but why remove the plugins' pkgConfigDir?

savil (commented):
I may be missing something, but these look like logic changes rather than dead code?

internal/impl/shell.go
ipince: Executing scripts (i.e. `devbox run`) doesn't happen through this code path. So, basically, `s.ScriptCommand` is _always_ `""`

ipince: And yes, we do propagate the exit code from a script to the user (see `nix/run.go`)

internal/impl/shell.go
ipince: `pathPrepend` is only used when `ExportEnv` is empty. But `ExportEnv` is never empty ever since unified env is on.

savil (approved):
got it, thanks for the explanations!

PR #739, opened 2023-03-10
[config] early return if cannot find GlobalDataPath

savil:
## Summary

From inspecting the code, I think this should early return. The following statement
about `readConfig(filepath.Join(dataPath, "devbox.json"))` doesn't make sense
to execute if there's been an error in retrieving the `dataPath`.

## How was it tested?

buildkite testscripts


savil:
* main
  * **PR #739** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/739" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/739?utm_source=stack-comment).

gcurtis (approved):
Makes sense to me.

PR #744, opened 2023-03-10
[testscripts] setup all xdg folders to work inside testscripts

savil:
## Summary

TSIA

## How was it tested?

buildkite runs

### DIFF
--- a/testscripts/testrunner/setupenv.go
+++ b/testscripts/testrunner/setupenv.go
@@ -12,7 +12,11 @@ import (
 func setupTestEnv(env *testscript.Env) error {
 	setupPATH(env)
 
-	err := setupCacheHome(env)
+	if err := setupXDGHomes(env); err != nil {
+		return err
+	}
+
+	err := setupSharedCacheDirectories(env)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #744** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/744" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/744?utm_source=stack-comment).

gcurtis (commented):
I think these are great for test helpers, but should we be setting them by default? I think we should have tests that exercise cases where the XDG directories aren't set and/or don't exist.

savil:
@gcurtis That's a good question. 

For XDG folders in particular, we rely on the `xdg.resolveDir` function to ensure that the folders are set up correctly. It relies on `HOME` env-var to be set (or more specifically `os.UserHomeDir()` to work): https://github.com/jetpack-io/devbox/blob/7a14cb8b05e007395215e2a97ed0f1c10731e6df/internal/xdg/xdg.go#L29

For testscripts, it depends on whether we want to set `HOME` env-var to be a folder within the `env.WorkDir`. From the output, it seems that `HOME=/no-home`...@loreto was that a conscious choice or just the default from testscripts.

My take is:
- if we don't want to set `HOME` in testscripts, then this PR is the way to go.
- if we do want to set `HOME` to `workDir/home/` or something, then we can do that. But I need to look into the implications.

gcurtis:
If we had to pick a default, I'd say set `HOME` because that's the more common case. But really I think the environment should be explicitly laid out in each test. That is, a test should have to call a function like `setXDGEnvVars` or `setHome` so it's obvious what environment the test is running in.

loreto:
> If we had to pick a default, I'd say set `HOME` because that's the more common case. But really I think the environment should be explicitly laid out in each test. That is, a test should have to call a function like `setXDGEnvVars` or `setHome` so it's obvious what environment the test is running in.

Independent of where these get set ... I actually lean towards using XDG and never using HOME for files we create. Rationale: I think our code should always respect XDG and we should never write anything to a user's home directory when the XDG paths point elsewhere. I like that the home dir is read only for purposes of all the tests.

Let me know if it's better to have a live discussion.

gcurtis:
> Independent of where these get set ... I actually lean towards using XDG and never using HOME for files we create.

I think we should use explicit XDG variables too in some tests. But I think the far more common case is that the user doesn't have them set (for example, on macOS) so `HOME` would be needed -- even for just figuring out the default XDG dir.

gcurtis:
> For tests that explicitly want to amend the environment, they can opt to not call setCommonTestEnv and instead call other helper methods as needed.

Sure, that's fine too. But right now `setupTestEnv` is unconditionally applied to all tests. This makes it impossible to write a test that doesn't have XDG directories created + set.

savil:
@loreto I agree with Greg's argument that not all systems have XDG env vars defined but almost all would have HOME defined. So, using HOME as the lowest common denominator for env-vars makes sense, and we should be capable of deriving XDG env-vars from that HOME env-var.

savil:
Conclusions from IRL discussions: 
(1) Prefer setting HOME and have devbox code derive XDG paths, since many users have that instead of XDG env-vars being set 
(2) Instead of forcing all testscripts to get a "standard test environment", we should expose a setCommonEnv function so it is opt-in. This would enable testscripts that want to exercise alternate env-var situations.

Future PRs can do this.

PR #746, opened 2023-03-10
Include plugin subdirs

ipince:
## Summary
TSIA

## How was it tested?
Ran testscripts tests successfully


savil (approved):
nice! this was just what I was trying locally as well :D

PR #749, opened 2023-03-12
nix,impl: avoid modifying flake.nix; don't use --impure

gcurtis:
## Summary

The previous devbox shell/run startup time was taking about 3.5x longer than `nix develop` for "warm" runs where the flake had already been evaluated once:

	$ time nix develop .devbox/gen/flake -c echo hi
	hi
	0.12s user 0.04s system 45% cpu 0.356 total

	$ time devbox run -- echo hi
	hi
	0.66s user 0.17s system 62% cpu 1.330 total

This was due to two reasons:

1. Devbox was passing `--impure` so that `NIXPKGS_ALLOW_UNFREE` could be used. This prevents Nix from using its eval cache. Fix it by setting the `allowUnfree` attribute directly in the flake.
2. Under certain circumstances, Nix uses the flake file's modified time stamp to determine if it can use its eval cache. Devbox unconditionally writes generated files on every run, which was changing the modified timestamp of the flake. Update the `writeFromTemplate` function to not write to existing files if the contents haven't changed.

After these changes, the new devbox times are more in-line with Nix:

	$ time devbox run -- echo hi
	hi
	0.19s user 0.09s system 59% cpu 0.466 total

Also ran `nixpgs-fmt` on the flake file templates and added tests to ensure they stay correct.

## How was it tested?

Added tests.

### DIFF
--- a/internal/impl/generate.go
+++ b/internal/impl/generate.go
@@ -4,8 +4,10 @@
 package impl
 
 import (
+	"bytes"
 	"embed"
-	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -20,7 +22,7 @@ import (
 	"go.jetpack.io/devbox/internal/plugin"
 )
 
-//go:embed tmpl/* tmpl/.*
+//go:embed tmpl/*
 var tmplFS embed.FS
 
 var shellFiles = []string{"development.nix", "shell.nix"}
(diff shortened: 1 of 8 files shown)


Lagoja:
This looks really good. I can pull and run some tests with this, would be great to add to the release if we're confident in the changes.

Lagoja:
I'm seeing a pretty consistent improvement from 2.2 seconds to 0.6 seconds to start a devbox shell with no `init_hook`. Seems worth shipping,

Lagoja:
As an extra nice side effect -- evaluating installed packages (including those from global) is a lot faster as well

Lagoja (approved):
Looks good from my testing

ipince (approved):
Nice! and thanks for adding tests!

internal/impl/generate.go
```diff
@@ -62,26 +64,81 @@ func generateForShell(rootPath string, plan *plansdk.ShellPlan, pluginManager *p
...
+			return errors.WithStack(err)
+		}
+		tmplCache[tmplKey] = tmpl
+	}
+	tmplNewBuf.Reset()
+	if err := tmpl.Execute(tmplNewBuf, plan); err != nil {
+		return errors.WithStack(err)
+	}
+
+	// In some circumstances, Nix looks at the mod time of a file when
```
savil, line 95: is there any resource to read about which circumstances apply for this situation?

internal/nix/profile.go
```diff
@@ -230,16 +230,13 @@ func profileInstall(args *ProfileInstallArgs) error {
...
-		"--impure", // Needed to allow flags from environment to be used.
 		FlakeNixpkgs(args.NixpkgsCommit)+"#"+args.Package,
 	)
 	cmd.Args = append(cmd.Args, ExperimentalFlags()...)
 	if args.Priority != "" {
 		cmd.Args = append(cmd.Args, "--priority", args.Priority)
 	}
 	cmd.Args = append(cmd.Args, args.ExtraFlags...)
-
-	cmd.Env = DefaultEnv()
```
savil, line 242: I do think this is needed:

```
❯ devbox add slack
Warning: local and global devbox.json have different nixpkgs commits. Will use the local version. This may lead to version mismatch and nix store bloat.

Installing package: slack.

[1/1] slack
        error: Package ‘slack-4.29.149’ in /nix/store/3hzps14zspk0mkxdjjk3d85vhspibb9c-source/pkgs/applications/networking/instant-messengers/slack/default.nix:82 has an unfree license (‘unfree’), refusing to evaluate.
               a) To temporarily allow unfree packages, you can use an environment variable
                  for a single invocation of the nix tools.
                    $ export NIXPKGS_ALLOW_UNFREE=1
                Note: For `nix shell`, `nix build`, `nix develop` or any other Nix 2.4+
                (Flake) command, `--impure` must be passed in order to read this
                environment variable.
               b) For `nixos-rebuild` you can set
                 { nixpkgs.config.allowUnfree = true; }
               in configuration.nix to override this.
               Alternatively you can configure a predicate to allow specific packages:
                 { nixpkgs.config.allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [
                     "slack"
                   ];
                 }
               c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
                 { allowUnfree = true; }
               to ~/.config/nixpkgs/config.nix.
        (use '--show-trace' to show detailed location information)
[1/1] slack: Fail
There was an error installing nix packages: slack. Packages were not added to devbox.json
Error: Command: /nix/var/nix/profiles/default/bin/nix profile install --profile /Users/savil/code/jetpack/devbox/examples/testdata/unfree-packages/.devbox/nix/profile/default nixpkgs/9e0eed654c705c7cafe192a8eba1610217f70544#slack --extra-experimental-features ca-derivations --option experimental-features nix-command flakes --priority 5: exit status 1

```

But we can add this back, and I think it won't hurt the shell startup time, no?

savil:
btw, thanks so much for digging into this. This is awesome work!

PR #750, opened 2023-03-13
all: add tracing

gcurtis:
Add a hidden `--trace` flag that turns on tracing to measure execution time. When set, a trace file is written (by default to trace.out) which can be inspected with `go tool trace trace.out`.

This is an example of the trace output after running `devbox --trace run -- echo hello` and then `go tool trace trace.out`:

events
timeline

### DIFF
--- a/internal/boxcli/midcobra/midcobra.go
+++ b/internal/boxcli/midcobra/midcobra.go
@@ -55,7 +55,8 @@ func (ex *midcobraExecutable) AddMiddleware(mids ...Middleware) {
 
 func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	// Ensure cobra uses the same arguments
-	ex.cmd.SetArgs(args)
+	ex.cmd.SetContext(ctx)
+	ex.cmd.ParseFlags(args)
 
 	// Run the 'pre' hooks
 	for _, m := range ex.middlewares {
@@ -63,7 +64,7 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	}
 
 	// Execute the cobra command:
-	err := ex.cmd.ExecuteContext(ctx)
+	err := ex.cmd.Execute()
 
 	var postRunErr error
 	var userExecErr *usererr.ExitError
@@ -76,8 +77,8 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	// run even if the command resulted in an error. This is useful when we still want to clean up
 	// before the program exists or we want to log something. The error, if any, gets passed
 	// to the post hook.
-	for _, m := range ex.middlewares {
-		m.postRun(ex.cmd, args, postRunErr)
+	for i := len(ex.middlewares) - 1; i >= 0; i-- {
+		ex.middlewares[i].postRun(ex.cmd, args, postRunErr)
 	}
 
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/planner/planner.go
```diff
@@ -16,6 +19,8 @@ var PLANNERS = []plansdk.Planner{
 // Return a merged shell plan from shell planners if user defined packages
 // contain one or more dev packages from a shell planner.
 func GetShellPlan(srcDir string, userPkgs []string) *plansdk.ShellPlan {
+	defer trace.StartRegion(context.Background(), "getShellPlan").End()
```
ipince, line 22: does it matter that the context is different?

internal/boxcli/midcobra/trace.go
```diff
@@ -0,0 +1,67 @@
...
+	tracef      *os.File
+	flag        *pflag.Flag
+	task        *trace.Task
+}
+
+var _ Middleware = (*DebugMiddleware)(nil)
+
+func (t *TraceMiddleware) AttachToFlag(flags *pflag.FlagSet, flagName string) {
+	flags.String(flagName, "", "write a trace to a file")
+	t.flag = flags.Lookup(flagName)
```
ipince, line 26: does this behave correctly/reasonably if the flag doesn't exist?

internal/boxcli/midcobra/trace.go
```diff
@@ -0,0 +1,67 @@
...
+		return
+	}
+	path := t.flag.Value.String()
+	if path == "" {
+		return
+	}
+	var err error
+	t.tracef, err = os.Create(path)
+	if err != nil {
+		panic("error enabling tracing: " + err.Error())
```
ipince, line 42: should we panic here? why not log an error out but continue execution?

internal/boxcli/midcobra/midcobra.go
```diff
@@ -76,8 +77,8 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 	// run even if the command resulted in an error. This is useful when we still want to clean up
 	// before the program exists or we want to log something. The error, if any, gets passed
 	// to the post hook.
-	for _, m := range ex.middlewares {
-		m.postRun(ex.cmd, args, postRunErr)
+	for i := len(ex.middlewares) - 1; i >= 0; i-- {
```
ipince, line 80: why run them backwards? I mean, I think I understand, but is it obvious that preRuns are FIFO and postRuns are LIFO?

ipince (approved):

internal/planner/planner.go
gcurtis: The context groups regions into a single task, making it easier to see the overall timeline. This region will still be visible on its own, although it would be a bit nicer to use the same context. It was just a bit trickier to propagate the context into this function without changing the Devbox interface.

internal/boxcli/midcobra/trace.go
gcurtis: The flag should always exist because it's added on the line above. If it's not set on the command line, then this will still return a flag but it'll have an empty value.

internal/boxcli/midcobra/trace.go
gcurtis: I would've preferred to return an error, but I couldn't see an easy way of doing that with the middleware. I figured that if you're using the `--trace` flag you probably don't want things to continue if tracing can't be enabled.

I could alternatively log an error and then `os.Exit` if that seems cleaner?

savil (approved):
nice! thanks for adding this.

PR #753, opened 2023-03-13
[allow unfree] add --impure and env-var to nix profile commands

savil:
## Summary

Adding these back. They enable adding unfree packages and do not affect shell perf.

```
❯ time devbox run -- echo "hi"
Warning: local and global devbox.json have different nixpkgs commits. Will use the local version. This may lead to version mismatch and nix store bloat.
hi
devbox run -- echo "hi"  0.41s user 0.16s system 67% cpu 0.847 total
```

context: https://github.com/jetpack-io/devbox/pull/749#discussion_r1134404572

## How was it tested?

these now work
```
> devbox add slack
> devbox add vscode
```

~I will add a testscript unit test in a follow up PR~

Added testscript unit test


gcurtis (commented):
I enables unfree packages directly in the flake config. Is that not working?

savil:
@gcurtis the problem is that during `devbox add <package>`, we do `nix profile install nixpkgs/<commit>#package`. This doesn't use the flake.nix file, and so needs the `--impure` to be passed to it.

gcurtis (approved):
Ugh, I thought we were using the flake for everything. It sounds like the Nix CLI doesn't have a solution to this problem either (other than passing the env var). Using `--impure` will still slow things down because it completely disables the eval cache, but I guess that's no different than doing `nix profile install`.

savil:
> Using --impure will still slow things down because it completely disables the eval cache, but I guess that's no different than doing nix profile install.

That's a shame. When I ran `devbox run -- echo "hello"`, it seemed fast though?

We have two steps:
1. `nix profile install`
2. `nix print-dev-env`

The second step is still fast thanks to your changes...

PR #754, opened 2023-03-13
impl: don't error in resetProfileDir if profile doesn't exist

gcurtis:
## Summary

This was failing in macOS tests because the Nix profile didn't exist. Change it to only fail if the error is not an `os.ErrNotExist` error.

## How was it tested?

```
$ DEVBOX_FEATURE_FLAKES=0 devbox add hello
$ DEVBOX_FEATURE_FLAKES=1 devbox add hello
$ rm .devbox/nix/profile/default
$ DEVBOX_FEATURE_FLAKES=0 devbox add hello
```

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -15,7 +15,6 @@ import (
 	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/debug"
-	"go.jetpack.io/devbox/internal/fileutil"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/plugin"
 	"go.jetpack.io/devbox/internal/ux"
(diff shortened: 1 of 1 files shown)


internal/impl/packages.go
```diff
@@ -444,22 +443,31 @@ func resetProfileDirForFlakes(profileDir string) (err error) {
...
-
-	if !needsReset {
-		return nil
+		_, err := os.Stat(filepath.Join(dir, "manifest.json"))
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		if err != nil {
+			return errors.WithStack(err)
+		}
```
savil, line 470: I understand the changes in lines 446-448

Could we push this fix down inside `fileutil.Exists`? Using that api is easier to read (albeit masking a genuine error from `os.Stat`)

savil (approved):

internal/impl/packages.go
gcurtis: Do we want to mask the error? If there's something like a permissions error, then I figured returning that error would be helpful.

savil: that's a fair point....

PR #755, opened 2023-03-13
ci: use DeterminateSystems Nix installer

gcurtis:
Hopefully this can help fix the flaky installer bugs on macOS. Also remove a really slow integration test.


mikeland86 (approved):
Worth trying. FWIW, I tried using this installer on Ubuntu a few weeks ago and it failed. Maybe only for mac? 

Either way, we're still testing our own installer on every PR

gcurtis:
Yeah, I wanted to try it out in CI first to see how it goes.

.github/workflows/cli-tests.yaml
```diff
@@ -99,19 +99,13 @@ jobs:
...
-          nix-build '<nixpkgs>' -A stdenv -A bash -A hello
-          echo "__ETC_PROFILE_NIX_SOURCED=1" >> $GITHUB_ENV
-          echo "NIX_PROFILES=$NIX_PROFILES" >> $GITHUB_ENV
-          echo "NIX_SSL_CERT_FILE=$NIX_SSL_CERT_FILE" >> $GITHUB_ENV
-          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
-          echo "PATH=$PATH" >> $GITHUB_ENV
+        uses: DeterminateSystems/nix-installer-action@v1
+        with:
+          logger: pretty
+          nix-build-user-count: 4
```
mohsenari, line 105: is there a reason behind putting 4 here or it's just to have a multi-user mode

gcurtis: It defaults to 32, but creating 32 user accounts on macOS can be slow so I reduced it to 4. I don't think it should affect things too much since none of our jobs should be building from source.

PR #758, opened 2023-03-14
[Cloud] Added cloud init command for open in vscode

mohsenari:
## Summary
Added a `devbox cloud init` command similar to devbox cloud shell but it can be called in to create a cloud vm (if it's not created already) and sets up ssh configs without dropping user in an interactive cloud shell.
This will be used in the extension so that the extension can ensure:
1. the remote vm exists
2. get its ssh address (vm ID)
3. the client (vscode) can connect to it via ssh

Also, added `devbox generate ssh-config` command that only sets ssh-config without creating cloud VMs.

## How was it tested?
`devbox cloud init`
and 
`devbox generate ssh-config`

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -88,41 +88,13 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
 	color.New(color.FgMagenta, color.Bold).Fprint(w, "Devbox Cloud\n")
 	fmt.Fprint(w, "Remote development environments powered by Nix\n\n")
 	fmt.Fprint(w, "This is an open developer preview and may have some rough edges. Please report any issues to https://github.com/jetpack-io/devbox/issues\n\n")
-	if err := ensureProjectDirIsNotSensitive(projectDir); err != nil {
-		return err
-	}
-
-	username, vmHostname := parseVMEnvVar()
-	// The flag for githubUsername overrides any env-var, since flags are a more
-	// explicit action compared to an env-var which could be latently present.
-	if githubUsername != "" {
-		username = githubUsername
-	}
-	if username == "" {
-		var err error
-		username, err = getGithubUsername()
-		if err != nil {
-			return err
-		}
-	}
-	debug.Log("username: %s", username)
-
-	// setup ssh config
-	sshCmd, err := SSHSetup(username)
-	if err != nil {
-		return err
-	}
 	// Record the start time for telemetry, now that we are done with prompting
 	// for GitHub username.
 	telemetryShellStartTime := time.Now()
-
-	// creating vm for user if it doesn't exist
-	vmHostname, err = ensureVMForUser(vmHostname, w, username, sshCmd)
+	username, vmHostname, err := InitVM(ctx, w, projectDir, githubUsername)
 	if err != nil {
 		return err
 	}
-	debug.Log("vm_hostname: %s", vmHostname)
-
 	// file sync and shell
 	color.New(color.FgGreen).Fprintln(w, "Starting file syncing...")
 	err = syncFiles(username, vmHostname, projectDir)
@@ -146,9 +118,14 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
 }
 
 // Temporary function to create a vm and print vmHostname to be used by devbox extension
-func InitShell(ctx context.Context, w io.Writer, projectDir string, githubUsername string) error {
+func InitVM(
+	ctx context.Context,
+	w io.Writer,
+	projectDir string,
+	githubUsername string,
+) (string, string, error) {
 	if err := ensureProjectDirIsNotSensitive(projectDir); err != nil {
-		return err
+		return "", "", err
 	}
 
 	username, vmHostname := parseVMEnvVar()
@@ -161,7 +138,7 @@ func InitShell(ctx context.Context, w io.Writer, projectDir string, githubUserna
 		var err error
 		username, err = getGithubUsername()
 		if err != nil {
-			return err
+			return "", "", err
 		}
 	}
 	debug.Log("username: %s", username)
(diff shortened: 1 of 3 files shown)


savil (commented):
I fear `--interactive` will confuse users who will think they are getting an interactive shell, or a non-interactive shell.

If the motivation is to start a VM and have it ready, how about: `devbox cloud init`?

internal/cloud/cloud.go
```diff
@@ -29,7 +29,7 @@ import (
 	"go.jetpack.io/devbox/internal/ux/stepper"
 )
 
-func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername string) error {
+func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername string, interactive bool) error {
```
savil, line 32: instead of a boolean argument, could we split this function into:
- `cloud.EnsureVMForUser` 
- `cloud.Shell`

That would feel like a more reusable API

mohsenari:
> 🤔 taking a step back, I'm not convinced this is what we want.
> 
> For open-in-vscode, the `Gateway.VMInfo()` will ensure a VM that is in "Started" state, and return back the VM's address and public-private keys to access that VM. We then need the Extension to generate the ssh config via the `devbox` CLI binary.
> 
> I'd suggest a hidden command like `devbox generate ssh-config` which calls `openssh.SetupDevbox`.
> 
> Since the Gateway.VMInfo() API isn't yet ready, if you would like a hidden command that starts a cloud VM for testing purposes, then how about `devbox cloud init` that has a similar implementation as this PR, and prints the VM's address out?

that makes sense. I'll change the name for this command and make it hidden. And add a hidden `devbox generate ssh-config` as well and update this PR.

internal/cloud/cloud.go
mohsenari: you mean break this function into 2 different functions?
yeah I think it would make the logic simpler. Let me try it and see how the code turns out.

LucilleH:
> Since the Gateway.VMInfo() API isn't yet ready, if you would like a hidden command that starts a cloud VM for testing purposes, then how about `devbox cloud init` that has a similar implementation as this PR, and prints the VM's address out?

I'm not sure why we need `devbox cloud init`? We can just manually start a VM on fly.io to test right?

internal/cloud/cloud.go
```diff
@@ -127,6 +145,45 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
...
+
+	// creating vm for user if it doesn't exist
+	vmHostname, err = ensureVMForUser(vmHostname, w, username, sshCmd)
+	if err != nil {
+		return err
+	}
+	debug.Log("vm_hostname: %s", vmHostname)
+	// printing vmHostname so that the output of devbox cloud init can be read by
+	// devbox extension
+	fmt.Fprintln(w, vmHostname)
```
savil, line 183: if you call `InitShell` from `cloud.Shell`, then instead of printing this here, you could return this value and print it in the `devbox cloud init` command implementation function...

savil: you may also have to return `telemetryShellStartTime`, if we're sharing the implementation...

savil (commented):
@mohsenari nice! But could we share the implementation a bit more?

@LucilleH I think its: 
(a) a nice decomposition of the code
(b) a useful hidden command for testing. But I don't feel strongly: if you would prefer to not add it, and Mohsen has another way, then that's cool too.

internal/cloud/cloud.go
```diff
@@ -127,6 +145,45 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
 	return shell(username, vmHostname, projectDir, telemetryShellStartTime)
 }
 
+// Temporary function to create a vm and print vmHostname to be used by devbox extension
+func InitShell(ctx context.Context, w io.Writer, projectDir string, githubUsername string) error {
```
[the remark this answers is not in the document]
mohsenari: yeah sure I don't have a strong preference on the name.

internal/cloud/cloud.go
mohsenari: Can I just put `telemetryShellStartTime` in `Shell()` right before calling `InitVM`? 
That way I have 1 less item to return from InitVM.

savil: (Am on phone so can't quite see the code)

I think the problem is that getGithubUsername could fallback to a user prompt, and so we want the start time to begin after that...does that make sense?

internal/cloud/cloud.go
```diff
@@ -104,8 +81,49 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
...
+	debug.Log("username: %s", username)
+
+	// setup ssh config
+	sshCmd, err := SSHSetup(username)
+	if err != nil {
+		return err
+	}
+	// Record the start time for telemetry, now that we are done with prompting
+	// for GitHub username.
+	telemetryShellStartTime := time.Now()
```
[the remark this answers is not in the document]
savil: this doesn't quite work as per my comment [here](https://github.com/jetpack-io/devbox/pull/758#discussion_r1139705890). We'll get distorted tti-times due to waiting for users to enter their github username in the prompt.

Could we move it inside `InitVM` _after_ `getGithubUsername`?

To mitigate returning multiple values from the function being bad, you could make a helper struct that is returned with `username, vmHostname, telemetryShellStartTime`. WDYT?

internal/cloud/cloud.go
mohsenari: fair enough, moved the `telemetryShellStartTime` into `InitVM()`

internal/cloud/cloud.go
```diff
@@ -127,6 +115,51 @@ func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername s
 	return shell(username, vmHostname, projectDir, telemetryShellStartTime)
 }
 
+// Temporary function to create a vm and print vmHostname to be used by devbox extension
```
savil, line 118: I was assuming this was non-temporary!

> InitVM ensures a vm is started for the user

savil (approved):
nice!

internal/cloud/cloud.go
mohsenari: could be non-temporary, depending on how we utilize this for extension implementation. But I was thinking temporary till the gateway can send vm and username information so that I can continue with extension development work

PR #770, opened 2023-03-15
[nix] Simplify profile priority, make it match flake file

mikeland86:
## Summary

I was implementing a progress indicator for global install, and the profile install retry logic was making it a bit complicated. I then realized that the order we were assigning to priorities (latter package takes precedence) is actually inverse to what flake `buildInputs` does (it uses earlier precedence). In practice, this means the profile generated by devbox and the flake generated by devbox would have different behaviors. This is not terrible but also not ideal. So I removed the sorting and added some documentation to ensure we always keep the same (local, global) order.

Finally, I tried out a slightly different approach which doesn't require a profile install with retries. By reading the profile manifest file we're able to see what the highest number existing priority (lowest priority)  is and just increment it to achieve the next lowest priority. I tried this up to numbers in the thousands there doesn't seem to be any limits. 

Some benefits of this approach:

* It's easier to implement progress indicators that read stdout/err in a goroutine. 
* Better performance (no retries needed)
* more deterministic
* Code is a bit simpler

Drawbacks:
* Depends on the manifest
* Needlessly assigns more priorities than are needed

@savil curious your thoughts on this?

## How was it tested?

```bash
devbox global add php82
devbox add php
devbox run "php -v"
```

### DIFF
--- a/internal/nix/profile.go
+++ b/internal/nix/profile.go
@@ -3,18 +3,21 @@ package nix
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
-	"go.jetpack.io/devbox/internal/debug"
 )
+
+const DefaultPriority = 5
 
 // ProfileListItems returns a list of the installed packages
 func ProfileListItems(writer io.Writer, profileDir string) ([]*NixProfileListItem, error) {
@@ -180,7 +183,6 @@ func (item *NixProfileListItem) String() string {
 
 type ProfileInstallArgs struct {
 	CustomStepMessage string
-	Priority          string
 	ExtraFlags        []string
 	NixpkgsCommit     string
 	Package           string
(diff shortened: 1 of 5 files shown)


internal/nix/profile.go
```diff
@@ -201,90 +203,32 @@ func ProfileInstall(args *ProfileInstallArgs) error {
...
-					args.Priority,
-					args.Package,
-				)
-				return profileInstall(args)
-			}
-		}
-
 		return errors.Wrapf(err, "Command: %s", cmd)
 	}
+
```
savil, line 229: in line 227 above, lets add:
```
fmt.Fprintf(args.Writer, "%s: ", stepMsg)
color.New(color.FgRed).Fprintf(args.Writer, "Fail\n")
```

internal/nix/profile.go
```diff
@@ -311,3 +255,35 @@ func ProfileRemove(profilePath, nixpkgsCommit, pkg string) error {
...
+func readManifest(profilePath string) (manifest, error) {
+	data, err := os.ReadFile(filepath.Join(profilePath, "manifest.json"))
+	if os.IsNotExist(err) {
+		return manifest{}, nil
+	} else if err != nil {
+		return manifest{}, err
+	}
+
+	var m manifest
+	return m, json.Unmarshal(data, &m)
```
savil, line 274: I'm glad flakes changed the manifest from `.nix` to `.json` so we can do this :)

savil (approved):
Oh good catch on the `buildInputs` behavior. Just to confirm: did you try it out to verify?
Asking because I'd [heard](https://jetpack-io.slack.com/archives/C040L5DFHFT/p1678397002245089) the inverse last time and so implemented that order :)

I really like the clever implementation using the priority from `manifest.json`!

gcurtis (approved):

internal/nix/profile.go
mikeland86: Do you mind if I push this to the stacked diff? That changes this logic because it no longer uses custom stdout 

That way I can test how this will look with the new tty nix command.

mikeland86:
@savil yeah, I think I was wrong on precedence before:

image

image

PR #772, opened 2023-03-16
[nix] Show nix profile install progress

mikeland86:
## Summary

This shows inline progress when installing new packages (both global and local). `nix profile install` does this automatically but only in tty mode. @gcurtis curious if we can somehow make nix believe we're in tty mode without having to specify stdin and stdout?

Removed 2 blockers that are no longer needed:

* `PackageInstallWriter` is no longer needed I don't think. It could not see the lines it was meant to supress.
* Returning package not found is a bit redundant because all callsites of `ProfileInstall` already defensivelly check if package exists. If the user somehow does end up trying to install a package that doesn't exist, we still show a decent error  

Fixed 2 semi-related errors:

* global add  would print success message even if nothing was installed because of errors
* Removing a package from global devbox is impossible if the nix package was not installed

## How was it tested?

```bash
devbox add curl hello bazel
devbox add global curl hello bazel
```

### DIFF
--- a/internal/nix/profile.go
+++ b/internal/nix/profile.go
@@ -225,6 +225,8 @@ func ProfileInstall(args *ProfileInstallArgs) error {
 	cmd.Stderr = args.Writer
 
 	if err := cmd.Run(); err != nil {
+		fmt.Fprintf(args.Writer, "%s: ", stepMsg)
+		color.New(color.FgRed).Fprintf(args.Writer, "Fail\n")
 		return errors.Wrapf(err, "Command: %s", cmd)
 	}
 
--- a/internal/nix/profile.go
+++ b/internal/nix/profile.go
@@ -216,14 +216,15 @@ func ProfileInstall(args *ProfileInstallArgs) error {
 	cmd.Env = AllowUnfreeEnv()
 	cmd.Args = append(cmd.Args, ExperimentalFlags()...)
 	cmd.Args = append(cmd.Args, args.ExtraFlags...)
-	cmd.Stdout = &PackageInstallWriter{args.Writer}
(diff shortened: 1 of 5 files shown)


internal/nix/profile.go
```diff
@@ -216,14 +216,15 @@ func ProfileInstall(args *ProfileInstallArgs) error {
 	cmd.Env = AllowUnfreeEnv()
 	cmd.Args = append(cmd.Args, ExperimentalFlags()...)
 	cmd.Args = append(cmd.Args, args.ExtraFlags...)
-	cmd.Stdout = &PackageInstallWriter{args.Writer}
```
savil, line 219: Aside from ignoring some lines (which are no longer needed), I think we also needed this writer to indent the output:
> _, err = io.WriteString(fw.Writer, "\t"+line+"\n")

savil (commented):
Could you share a screenshot or video of what this looks like now? Not sure I am picturing it correctly. IIRC we needed the custom writer to indent the lines...


> Removing a package from global devbox is impossible if the nix package was not installed

what could add a package to devbox.json but have it be not installed?

mikeland86:
@savil it has to be a video or a gif because the output is dynamic inline. Which also means I don't think we need the indent to make it look good. Will share.

mikeland86:
I'm a bit too lazy to make a video, this is what it looks like:

image

The lower line is dynamic and shows progress.

this is what it looks like after installing multiple packages

image

I went ahead and removed the redundant "package is now installed" and now it looks like

image

Two weird behavior nits:

* If package is already installed, no output.
* When installing local packages we also install the global ones which is weird.

Will address this in follow up

Issue #29, opened 2022-08-28
devbox add ghc && devbox add haskellPackages.aeson does not add aeson to ghc package database


ParetoOptimalDev:
```
$ devbox add ghc
$ devbox add haskellPackages.aeson
$ devbox shell
$ ghc-pkg list | grep aeson
```

That should show a package listed. Compare to the nix equivalent:

```
$ ghc-pkg list | grep aeson
    aeson-1.5.6.0
```

This lets it be used by ghci for instance:

```
$ ghci 
GHCi, version 8.10.7: https://www.haskell.org/ghc/  :? for help
Prelude> import Data.Aeson
Prelude Data.Aeson> :t parseJSON
parseJSON
  :: FromJSON a =>
     Value -> aeson-1.5.6.0:Data.Aeson.Types.Internal.Parser a
```

I kind of expected this not to work, but figured I'd make an issue since I think it'd be nice if people could do this. Especially for things like the `lens` library with tons of dependencies which really make use of binary caching.

loreto:
Thanks for reporting, let me take a look this coming week and see if we can easily fix it.

ParetoOptimalDev:
> Thanks for reporting, let me take a look this coming week and see if we can easily fix it.

Sure, if it helps my first thought was "Can `devbox add haskellPackages.aeson` just always rewrite to implicitly be equivalent to `nix-shell -p "haskellPackages.ghcWithPackages (p: [p.aeson])"` rather than what I'm assuming it's now equivalent to `nix-shell -p haskellPackages.aeson`.

For haskell in particular this is very useful because the built packages are cached. For things like the cryptonite library that take a long time to compile, this can save you hours.

Where as with the binary cache of cryptonite I can get `ghci` with cryptonite available in 3.8s:

```haskell
$ time nix-shell -p "haskellPackages.ghcWithPackages (p: [p.cryptonite])" --run exit
... snip ... 
copying path '/nix/store/nkhj0j7vnvv0810g8zg15fpabvkwlzwy-memory-0.15.0' from 'https://cache.nixos.org'...
copying path '/nix/store/lyll680sx4cpbm2ivabq3abmhyc4z13q-cryptonite-0.29' from 'https://cache.nixos.org'...
building '/nix/store/dw00lxfnhhjnyfip8p6kd20ypcc4248s-ghc-8.10.7-with-packages.drv'...
```

Lagoja:
[...]

PR #773, opened 2023-03-16
First Pass at a Haskell Package Planner

Lagoja:
## Summary
This PR addresses #29, using a similar framework as our PHP planner. It makes it possible to use Haskell packages installed via `haskellPackages` alongside GHC.

This PR is incomplete - right now it only works for the default `ghc` in Nixpkg. We should make sure that we can detect when users request a different compiler version (like `haskell.compiler.ghc810`)

## How was it tested?

**Default GHC:** Use the following devbox.json:

```json
{
  "packages": [
    "haskellPackages.aeson",
    "ghc",
    "stack"
  ],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
  }
}
  ```
  
Run `devbox shell`, then run `ghc-pkg list | grep aeson` to make sure that it installed correctly

**Versioned GHC:** Use the following devbox.json

```json
{
  "packages": [
    "haskell.compiler.ghc92",
    "haskell.packages.ghc92.aeson",
    "tree",
    "direnv",
    "ghc",
  ],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
  }
}
```
Run `devbox shell`, then run `ghc-pkg list | grep aeson` to make sure that it installed correctly. Also confirm that other packages installed as expected

### DIFF
--- a/internal/planner/languages/haskell/haskell_v1_planner.go
+++ b/internal/planner/languages/haskell/haskell_v1_planner.go
@@ -24,11 +24,10 @@ type V2Planner struct {
 	userPackages []string
 }
 
-// PHPV2Planner implements interface PlannerForPackages (compile-time check)
 var _ plansdk.PlannerForPackages = (*V2Planner)(nil)
 
 func (p *V2Planner) Name() string {
-	return "php.v2.Planner"
+	return "haskell.v2.Planner"
 }
 
 func (p *V2Planner) IsRelevant(srcDir string) bool {
--- a/internal/planner/languages/haskell/haskell_v1_planner.go
+++ b/internal/planner/languages/haskell/haskell_v1_planner.go
@@ -1,0 +1,75 @@
+// Copyright 2022 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package haskell
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


internal/planner/languages/haskell/haskell_v1_planner.go
```diff
@@ -0,0 +1,74 @@
...
+func (p *V2Planner) Name() string {
+	return "haskell.v2.Planner"
+}
+
+func (p *V2Planner) IsRelevant(srcDir string) bool {
+	return false
+}
+
+func (p *V2Planner) IsRelevantForPackages(packages []string) bool {
+	p.userPackages = packages
```
savil, line 38: out of scope for this PR, but this is a strange API, where we save `packages` as a side-effect in a boolean IsRelevantForPackages function call....

savil (approved):
nice!

Lagoja:
This is now working for both the versioned and default GHC packages. My Go code may be a bit gnarly, so happy to make updates where needed

internal/planner/languages/haskell/haskell_v1_planner.go
Lagoja: It's also weird that we do 2 passes of the regex, one to detect if the planner is relevant, and another to actually create the shell plan -- should we be storing the results of the `IsRelevantForPackages` and then reusing it in `GetShellPlan`?

savil (commented):
@Lagoja mostly lgtm, but what do you think of the union idea below?

PR #774, opened 2023-03-16
feat: add scripts as commands for shell completion

hezhizhen:
## Summary

Add scripts defined in `devbox.json` as commands to `devbox run` for shell completion.


gcurtis:
Closing in favor of #775.

Issue #765, opened 2023-03-15
[Feature Request] Add completion for scripts


hezhizhen:
When I hit tab after `devbox run `, it can provide all the script names (just like Makefile).

savil:
this would be great to enable

mikeland86:
@hezhizhen that's a really good idea. Is this something you can take on? Happy to brainstorm approaches if needed

hezhizhen:
> @hezhizhen that's a really good idea. Is this something you can take on? Happy to brainstorm approaches if needed

Sure, I can try it. But I have no idea how completion works. Does it rely on shell completions?

hezhizhen:
I can add scripts to `devbox run` command as cobra commands and their names can be autocompleted. But it requires devbox to build itself every time users add or remove scripts.

hezhizhen:
> I can add scripts to `devbox run` command as cobra commands and their names can be autocompleted. But it requires devbox to build itself every time users add or remove scripts.

See https://github.com/jetpack-io/devbox/pull/774

hezhizhen:
> I've noticed that cobra command has field `ValidArgs` and `ValidArgsFunction`. Maybe I can use them.

See https://github.com/jetpack-io/devbox/pull/775 . I think this solution is better.

gcurtis:
Thanks again @hezhizhen for taking the time to implement this.

Closed by #775.

PR #775, opened 2023-03-16
feat: add scripts as valid args for shell completion

hezhizhen:
## Summary

Add scripts in `devbox.json` as `devbox run`'s valid arguments and it provides shell completion.

Here's the screenshot: 

Screenshot 2023-03-16 at 21 45 20

And here's a demo:

https://user-images.githubusercontent.com/7611700/225638325-e971c651-7383-4ddc-84db-861e6de4cbd2.mp4


gcurtis:
Just tried this out and it's super helpful! Thanks for adding it. I'll close out #774.

hezhizhen:
resolves https://github.com/jetpack-io/devbox/issues/765

PR #777, opened 2023-03-16
[flakes] Remove non-flakes code

mikeland86:
## Summary

Removes non-flakes code and tries to clean up some branching that is no longer needed.

@savil I removed `featureflag.disabled()` because it makes reasoning about these refactorings quite a bit more difficult and error prone. The words are really similar and some callsites use `!enabled()` which means now I have to think about 4 different combinations. Is this something you feel strongly about? I'm a bit more of the pythonic "only one correct way to do something" than the ruby-like "many ways to write this to make the code sound like english" thoughts?

## How was it tested?

```
devbox add hello
devbox run hello
```

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -12,7 +12,6 @@ import (
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
-	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/plugin"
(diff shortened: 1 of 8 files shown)


internal/impl/packages.go
```diff
@@ -390,24 +325,14 @@ func resetProfileDirForFlakes(profileDir string) (err error) {
...
-			return nil
-		}
-		if err != nil {
-			return errors.WithStack(err)
-		}
+	// older nix profiles have a manifest.nix file present
+	_, err = os.Stat(filepath.Join(dir, "manifest.nix"))
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
 	}
```
mikeland86, line 332: ugh, I'm not sure about this. I'm gonna read up on it a bit

mikeland86: actually, it's fine

savil:
I'd prefer the `Disabled` kept. I ended up making some mistakes during development by mis-typing/mis-reading `!featureflags.Flakes.Enabled()`.

savil (approved):
thanks for cleaning it up!

mikeland86:
> I'd prefer the Disabled kept. I ended up making some mistakes during development by mis-typing/mis-reading !featureflags.Flakes.Enabled().

I re-added it. 

Yeah, I understand that's confusing. I guess my counter is that it doesn't really keep people from using `!featureflags.Flakes.Enabled()` and now they can also use `!featureflags.Flakes.Disabled()` So you have 4 different statements that all ready similarly.

savil:
I agree having 4 ways is non-ideal. Let me think about it, and maybe we can come up with a better way which is both more explicit/readable and has just "one way".

PR #778, opened 2023-03-17
[ux] Add new refresh alias

mikeland86:
## Summary

This adds a new `refresh` alias that makes it a bit nicer when people install packages in shell. Ideally this goes away when we start wrapping binaries.

## How was it tested?

```bash
devbox shell
devbox add hello
refresh
hello
```

Also tested on fish:

```fish
SHELL=fish ./dist/devbox shell
devbox add hello
refresh
hello
```

image


gcurtis (approved):
Looks good, but do you mind updating the tests once #784 is merged?

PR #780, opened 2023-03-17
[CLI] Added install command

mohsenari:
## Summary
Added devbox install command. Re-used the same functions as devbox run since it's just an alias for calling `devbox run -- :` ( `:` is noop in shell).

## How was it tested?
`./devbox install`

### DIFF
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -48,7 +48,7 @@ func InstallCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "install",
 		Short: "Installs all packages mentioned in devbox.json",
-		Long: "Starts a new devbox shell and installs all packages mentioned in devbox.json in currect directory or" +
+		Long: "Starts a new devbox shell and installs all packages mentioned in devbox.json in current directory or" +
 			"a directory specified via --config. \n\n Then exits the shell when packages are done installing.\n\n ",
 		Args:    cobra.MaximumNArgs(0),
 		PreRunE: ensureNixInstalled,
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -4,6 +4,8 @@
 package boxcli
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
@@ -37,6 +39,30 @@ func RunCmd() *cobra.Command {
 	flags.config.register(command)
 
 	command.ValidArgs = listScripts(command, flags)
+
+	return command
+}
+
+func InstallCmd() *cobra.Command {
+	flags := runCmdFlags{}
+	command := &cobra.Command{
+		Use:   "install",
+		Short: "Installs all packages mentioned in devbox.json",
+		Long: "Starts a new shell and installs all packages mentioned in devbox.json or a json file" +
+			"specified via --config. \n\n Then exits the shell when packages are done installing.\n\n ",
+		Args:    cobra.MaximumNArgs(0),
+		PreRunE: ensureNixInstalled,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := runScriptCmd(cmd, []string{":"}, flags)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintln(cmd.ErrOrStderr(), "Finished installing packages.")
+			return nil
+		},
+	}
+
+	flags.config.register(command)
 
 	return command
 }
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -48,11 +48,13 @@ func InstallCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "install",
 		Short: "Installs all packages mentioned in devbox.json",
-		Long: "Starts a new shell and installs all packages mentioned in devbox.json or a json file" +
-			"specified via --config. \n\n Then exits the shell when packages are done installing.\n\n ",
+		Long: "Starts a new devbox shell and installs all packages mentioned in devbox.json in currect directory or" +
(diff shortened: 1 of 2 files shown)


hezhizhen:
I think it's not necessary because when users run `devbox shell`, it ensures that all packages specified in `devbox.json` are installed (see `internal/impl/devbox.go:L142`).

gcurtis:
I like the concept, but I find the command name to be confusing. It's not immediately obvious what the difference is between `devbox add` and `devbox install` by their names. I'd also expect `devbox install` to install a package.

Some random ideas for names:

1. Make `devbox add` (without a package) add missing packages that are in `devbox.json`.
2. Make `devbox run` (without any args) a no-op. It starts a shell and immediately exits, triggering the package downloads.
3. Have a `devbox shell --prefetch` or `devbox shell --download` flag.

gcurtis:
> I think it's not necessary because when users run `devbox shell`, it ensures that all packages specified in `devbox.json` are installed (see `internal/impl/devbox.go:L142`).

@hezhizhen this is true, but we've had a few users ask for a way to trigger the package install manually for things like CI. You may want to know that your entire environment is downloaded and installed (and maybe cache it) before proceeding with other CI steps. Right now the workaround is to run something like `devbox run -- exit`, but it's not easy to figure that out.

internal/boxcli/run.go
```diff
@@ -41,6 +43,30 @@ func RunCmd() *cobra.Command {
 	return command
 }
 
+func InstallCmd() *cobra.Command {
+	flags := runCmdFlags{}
+	command := &cobra.Command{
+		Use:   "install",
+		Short: "Installs all packages mentioned in devbox.json",
+		Long: "Starts a new shell and installs all packages mentioned in devbox.json or a json file" +
```
savil, line 51: I don't think this is accurate:
> or a json file specified via --config

I believe we always look for a `devbox.json` file at the directory specified by `--config`. See these lines where we first get the "projectDir" and then append the hardcoded `configFilename` which is `devbox.json`:
https://github.com/jetpack-io/devbox/blob/bf52b9a6324f3d4fb65dc92ce1f67240f151cc9d/internal/impl/devbox.go#L88-L92

savil, line 51: Should we clarify that the shell is a devbox shell?
i.e. suggestion: "Starts a new `devbox shell` and ...."

internal/boxcli/run.go
```diff
@@ -41,6 +43,30 @@ func RunCmd() *cobra.Command {
...
+			"specified via --config. \n\n Then exits the shell when packages are done installing.\n\n ",
+		Args:    cobra.MaximumNArgs(0),
+		PreRunE: ensureNixInstalled,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := runScriptCmd(cmd, []string{":"}, flags)
```
savil, line 56: for shell n00bs like myself, would you mind adding a small comment explaining that `:` is a noop in shell?

savil (commented):
My take:
1. I agree we need this functionality
2. I also agree with Greg that `devbox install` and `devbox add` are too similar.

I think `devbox init` should do an "install step" at the end. That way we avoid adding more top-level commands.

savil (approved):
Discussing IRL, the compelling argument is that `init`, `install` and `add` sub-commands are a common pattern for similar package managers like `yarn`, `bundler`, `poetry`.

Accepting modulo my suggestions below which I think are fairly straightforward to amend. Thanks!

hezhizhen:
> Discussing IRL, the compelling argument is that `init`, `install` and `add` sub-commands are a common pattern for similar package managers like `yarn`, `bundler`, `poetry`.
> 
> 
> 
> Accepting modulo my suggestions below which I think are fairly straightforward to amend. Thanks!

I think `add` and `install` are too similar and it's better to have one alone (add package to `devbox.json` and install it). Users can manually edit `devbox.json` file (add or remove packages), and devbox provides searching functionality (like `devbox search`) to get valid package names and then users can directly edit `devbox.json` or through `devbox install` command.

mohsenari:
> I think `add` and `install` are too similar and it's better to have one alone (add package to `devbox.json` and install it). 

@hezhizhen  We had people asking for a way to install all their packages separately before going into devbox shell. Mostly, in CICD stages or with Dockerfile where installed packages can be cached as a layer in building stage of a container.
Most of the time we end up suggesting a hacky solutions such as `devbox run -- echo "done"` and to make this command set as an alias in their shell e.g., `alias devbox-install="devbox run -- echo \"done\""`
So this is to cover those cases. The add command still adds a package and installs it. 

With all that said, do you have a nicer idea to implement instead of recommending `devbox run -- echo "done"` other than implementing `devbox install`?

> devbox provides searching functionality (like `devbox search`)

This is a separate issue. We have been experimenting with different ideas and implementations of `devbox search` but we want to commit to an implementation that is sustainable for future development and scale.

hezhizhen:
> We had people asking for a way to install all their packages separately before going into devbox shell. Mostly, in CICD stages or with Dockerfile where installed packages can be cached as a layer in building stage of a container.
Most of the time we end up suggesting a hacky solutions such as devbox run -- echo "done" and to make this command set as an alias in their shell e.g., alias devbox-install="devbox run -- echo \"done\""
So this is to cover those cases. The add command still adds a package and installs it.

You don't need to repeat what's been said before.

> With all that said, do you have a nicer idea to implement instead of recommending devbox run -- echo "done" other than implementing devbox install?

My point is, should we have both of `devbox add` and `devbox install`, or we can combine them into one (like `devbox install`)?

> This is a separate issue. We have been experimenting with different ideas and implementations of devbox search but we want to commit to an implementation that is sustainable for future development and scale.

Ignore it now.

hezhizhen:
Please ignore my comments, I'm not sure what's better.

PR #785, opened 2023-03-17
[services] Improve auto-port forward and proxy

mikeland86:
## Summary

This improves:

* port auto-forwarding only happens on CLI shells
* Print port on service proxy if available.
* Refactor common cloud env variables into a single package

## How was it tested?

builds

### DIFF
--- a/internal/services/services.go
+++ b/internal/services/services.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
+	"go.jetpack.io/devbox/internal/cloud/envir"
 	"go.jetpack.io/devbox/internal/plugin"
 )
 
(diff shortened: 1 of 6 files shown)


internal/services/services.go
```diff
@@ -163,10 +165,27 @@ func printProxyURL(w io.Writer) error {
...
+		}
+	}
+
+	if printGeneric {
+		color.New(color.FgHiGreen).Fprintf(
+			w,
+			"To access other services on this vm use: %s-<port>.svc.devbox.sh\n",
+			hostname,
+		)
+	}
```
savil, line 189: what if I have 3 services, but only one has a port, then this will print for just the one having-port-service. That seems problematic because users will worry about how to access the other-non-explicity-port-services.

Could we improve this to:
```
To access <having-port-service> on this vm use: ....
To access other services on this vm use: .....
```

savil (approved):
thanks for `envir` !

internal/services/services.go
mikeland86: I think the current behavior is what you are describing. If at least one service doesn't have a declared port, `printGeneric` will be `true` so we print the last message.

savil: oh you are right. booleans are hard!

PR #786, opened 2023-03-17
[config] Set PATH as ENV instead of hook

mikeland86:
## Summary

This change improves devbox.json and plugin jsons by allowing `PATH` to be set as an env variable instead of a hook.

For example, it allows `"PATH": "$PWD/dist:$PATH"`. 

In general any environment variable that is defined at the time of parsing is accessible. This means for plugins, the HOST environment is accessible, and for `devbox.json` the HOST environment and also the plugin environment is accessible. In practice this allows the user to override any plugin environment using an existing variable as base. On the other hand, the plugin environment variables will not take into account any changes that are made in the `devbox.json`

Some benefits of this change:

* We no longer need to export init hooks. One of the main use cases of init hooks is now moved to env instead.
* `refresh` no longer breaks changes to PATH. This is because init hooks don't get run by refresh, but the environment is reset correctly.

Related fixes:

* Removed code that was adding virtenv/<pkg>/bin to path. This is not used. 
* Removed code that was creating `env` files for each plugin. This is no longer used.
* Added `hash -r` to refresh functions
* No longer fetch plugin env vars when starting a service (this was only needed previously when adding a package while in shell would add it to profile but would not add env vars to current shell. Today `refresh` does this)

## How was it tested?

```bash
devbox shell
echo $PATH
devbox add ruby 
refresh
echo PATH
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -22,6 +22,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/generate"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
+	"go.jetpack.io/devbox/internal/conf"
 	"go.jetpack.io/devbox/internal/cuecfg"
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/fileutil"
@@ -170,7 +171,6 @@ func (d *Devbox) Shell() error {
 		WithHistoryFile(filepath.Join(d.projectDir, shellHistoryFile)),
 		WithProjectDir(d.projectDir),
 		WithEnvVariables(env),
-		WithPKGConfigDir(d.pluginVirtenvPath()),
 		WithShellStartTime(shellStartTime),
 	}
 
@@ -515,8 +515,6 @@ func (d *Devbox) computeNixEnv(ctx context.Context) (map[string]string, error) {
 
 		env[key] = val.Value.(string)
 	}
-	nixEnvPath := env["PATH"]
-	debug.Log("nix environment PATH is: %s", nixEnvPath)
 
 	// These variables are only needed for shell, but we include them here in the computed env
 	// for both shell and run in order to be as identical as possible.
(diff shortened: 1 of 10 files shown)


internal/impl/devbox.go
```diff
@@ -534,17 +532,15 @@ func (d *Devbox) computeNixEnv(ctx context.Context) (map[string]string, error) {
...
 		for k, v := range d.configEnvs(env) {
 			env[k] = v
 		}
 	}
 
-	// TODO: consider removing this; not being used?
-	pluginVirtenvPath := d.pluginVirtenvPath()
-	debug.Log("plugin virtual environment PATH is: %s", pluginVirtenvPath)
+	nixEnvPath := env["PATH"]
+	debug.Log("nix environment PATH is: %s", nixEnvPath)
```
gcurtis, line 541: Can this stay at the top and instead add more debug logs? It's helpful to be able to see what the `PATH` is at each stage (current env, after Nix, after plugins, after config, etc.).

internal/impl/devbox.go
```diff
@@ -534,17 +532,15 @@ func (d *Devbox) computeNixEnv(ctx context.Context) (map[string]string, error) {
...
 
-	env["PATH"] = JoinPathLists(pluginVirtenvPath, nixEnvPath, originalPath)
+	env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
 	debug.Log("computed unified environment PATH is: %s", env["PATH"])
```
gcurtis, line 544: Might as well while you're here:

```suggestion
	debug.Log("computed environment PATH is: %s", env["PATH"])
```

gcurtis (approved):
I vote +1 for this because it makes it easier to set env vars correctly. It's annoying needing to escape quotes in the init hooks and if you forget to do it things might break. Also, thanks for cleaning up the old plugin env stuff.

Related to this, should we also rerun init hooks? As an example, the `export GOROOT="$(go env GOROOT)"` command needs command substitution so it can't go in env. If you install a different version of Go, you'd want that hook to run again to pick up the new `GOROOT`.

mikeland86:
> elated to this, should we also rerun init hooks? As an example, the export GOROOT="$(go env GOROOT)" command needs command substitution so it can't go in env. If you install a different version of Go, you'd want that hook to run again to pick up the new GOROOT.

I'm mixed about re-running them. In this particular case, running `refresh` does not remove GOROOT, but you are correct that installing a new go version will not be captured correctly. The main reasons I lean towards not rerunning:

* Performance, some hooks could be slow.
* It forces hooks to be idempotent. I think generally it is good practice to make init hooks idempotent, but having them explicitly break if they are not idempotent could be bad UX.

Finally, once we wrap binaries with shellenv, it means we would need another solution for re-running init hooks, or just leave that responsibility to the user (which I don't think is that unreasonable)

Issue #781, opened 2023-03-17
[Bug]: another `devbox` binary is used


hezhizhen:
**Current Behavior (bug)**

<img width="1101" alt="Screenshot 2023-03-17 at 13 28 37" src="https://user-images.githubusercontent.com/7611700/225820630-f952c1f7-368a-48cd-8aac-04af55705a34.png">

I want to use the binary in `/usr/local/bin` whose version is `0.4.4`, but it seems that the binary in `dist` is used.

**Expected Behavior (fix)**

Use the former `devbox` instead of the latter one.

**Additional context**

gcurtis:
This has thrown me off too. I tend to build and run with `go install ./cmd/devbox && devbox` which puts it in my `$GOBIN` directory. I didn't realize I had been running an old `dist/devbox` binary instead.

The reason is because `devbox.json` sets `PATH=$(pwd)/dist:$PATH` in the init hooks. @mikeland86 is it okay to remove this? I think it's a bit surprising if you don't know about it.

Lagoja:
I think it makes sense to remove it. It's convenient, but easy to miss and pretty unexpected.

hezhizhen:
> This has thrown me off too. I tend to build and run with `go install ./cmd/devbox && devbox` which puts it in my `$GOBIN` directory. I didn't realize I had been running an old `dist/devbox` binary instead.
> 
> The reason is because `devbox.json` sets `PATH=$(pwd)/dist:$PATH` in the init hooks. @mikeland86 is it okay to remove this? I think it's a bit surprising if you don't know about it.

Oh, I forgot it...

PR #787, opened 2023-03-18
Adjust paths order

hezhizhen:
## Summary

Move `$(pwd)/dist` to the end of `PATH` as the last choice of `devbox` binary.

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -4,7 +4,7 @@
     "golangci-lint"
   ],
   "env": {
-    "PATH": "$PWD/dist:$PATH"
+    "PATH": "$PATH:$PWD/dist"
   },
   "shell": {
     "init_hook": [


hezhizhen:
Close https://github.com/jetpack-io/devbox/issues/781

hezhizhen:
PTAL @gcurtis @mikeland86

gcurtis:
@mikeland86 I think you added this. Are you okay with putting `dist/` at the end of the `PATH`?

mikeland86:
This is part of my workflow but if it's messing with other people's I'm OK removing it (I think removing is a bit better than appending it is very likely that everyone developing devbox has devbox installed). I thought this would be helpful to others, but If others don't use it this way, I can just modify my own PATH.

My workflow is usually using devbox shell to develop devbox. So I'll start a shell and then run `devbox run build` and test it out. 

@gcurtis can you share your GOBIN workflow? In some cases it can be important that the default devbox binary is the development one because in some cases devbox calls itself so if we don't change the default devbox we end up calling production one instead.

gcurtis:
@mikeland86 it's just `go install ./cmd/devbox; devbox`. I have `GOBIN` set to `~/bin` so that's where the built binary gets placed. But if I had run `devbox build` at some point earlier, then the build in `dist` always gets launched.

Sometimes I also test things out in a temp directory in a new terminal:

```
$ cd `mktemp -d`
$ devbox init
...
```

In that case I'm not in a devbox shell.

mikeland86:
@gcurtis should we also change build to be `go install ./cmd/devbox` ? And just let individual developers set GOBIN?

mikeland86 (approved):

Issue #760, opened 2023-03-14
Set LD_LIBRARY_PATH to point to the `.devbox/.../lib` folder


Lagoja:
A number of issues have been reported which are resolved by setting the `LD_LIBARY_PATH` variable to point to the devbox project's generated lib folder. For example:  #710  is resolved by setting `LD_LIBRARY_PATH=$PWD/.devbox/nix/profile/default/lib:$LD_LIBRARY_PATH`.

Since this is fairly obscure, we should set this value in Devbox Shell automatically. Developers can still append or overwrite it in their `devbox.json`, but this ensures that `ld` is aware of the libraries that Devbox installs for them.

Note: there are likely other flags that we should consider adding

Issue #710, opened 2023-03-01
[Bug]: ImportError: libstdc++.so.6: cannot open shared object file: No such file or directory


StephenStrickland:
**Current Behavior (bug)**
When using python packages like `gradio` the package fails when calling into `pandas`.

Steps:
1. `> devbox shell`
2. `> gradio`

**Error Message**
```sh
> gradio
Traceback (most recent call last):
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/bin/gradio", line 5, in <module>
    from gradio.reload import run_in_reload_mode
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/gradio/__init__.py", line 3, in <module>
    import gradio.components as components
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/gradio/components.py", line 21, in <module>
    import altair as alt
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/__init__.py", line 4, in <module>
    from .vegalite import *
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/vegalite/__init__.py", line 2, in <module>
    from .v4 import *
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/vegalite/v4/__init__.py", line 2, in <module>
    from .schema import *
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/vegalite/v4/schema/__init__.py", line 2, in <module>
    from .core import *
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/vegalite/v4/schema/core.py", line 4, in <module>
    from altair.utils.schemapi import SchemaBase, Undefined, _subclasses
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/utils/__init__.py", line 1, in <module>
    from .core import (
  File "/home/stephen/code/poc/.devbox/virtenv/python310Packages.pip/.venv/lib/python3.10/site-packages/altair/utils/core.py", line 14, in <module>
[...]

PR #792, opened 2023-03-21
[lib-path] Add a few common env variables

mikeland86:
## Summary

Adds `LD_LIBRARY_PATH` and `LIBRARY_PATH` env vars to shellenv (pointing to profile lib directory). 

This is a bit of a blunt approach, so not sure if it's best. This would fix https://github.com/jetpack-io/devbox/issues/760 and also lets us simplify the [rust plugin](https://github.com/jetpack-io/devbox/blob/main/plugins/rustup.json#L7-L7)

Two questions/considerations:

* I considered adding to the flake.nix file instead, but wasn't example sure how to point to the right directory (without hard coding). It might look something like `LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath...` 
* I also read the following:

> Not setting LD_LIBRARY_PATH is usually the right choice because usually the native binaries find their libraries using RUNPATH which is specific to the binary. Setting LD_LIBRARY_PATH has an obvious problem of potentially messing up other programs.

@gcurtis any idea how accurate this is?

## How was it tested?

`devbox shell` and confirmed env vars are set.

Also tested with this example: https://github.com/jetpack-io/devbox/issues/710#issuecomment-1451162069

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -546,6 +546,8 @@ func (d *Devbox) computeNixEnv(ctx context.Context) (map[string]string, error) {
 	env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
 	debug.Log("computed environment PATH is: %s", env["PATH"])
 
+	d.setCommonHelperEnvVars(env)
+
 	return env, nil
 }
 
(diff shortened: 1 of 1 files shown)


gcurtis (approved):

mikeland86:
@gcurtis awesome, thanks for the explanation! Gives me more confidence about this. I may follow up this PR to make 2 changes:

* Remove env var from rust plugin (since no longer needed)
* Move env variable to flake

PR #793, opened 2023-03-21
Move Devbox Examples to Devbox Repo

Lagoja:
## Summary
Moves the Devbox examples repo into the Devbox folder for testing purposes. Each example should have a `run_test` script that can be run for testing purposes

Tasks for Merging
- [x] Move examples into the Devbox Repo
- [x] Add `run_test` script to examples
    - [x] development
    - [x] stacks

## How was it tested?
`devbox run run_test`


examples/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
savil, line 1: do we need this new LICENSE compared to the LICENSE we already have in the root folder?

savil (approved):
Lets land this as-is (modulo the LICENSE file, see below) and polish it in future iterations?

examples/LICENSE
Lagoja: Removed

PR #794, opened 2023-03-21
Display devbox cloud timeout information

LucilleH:
## Summary
Screenshot 2023-03-21 at 11 18 56 AM

## How was it tested?
```
devbox run build
./dist/devbox cloud shell
```

### DIFF
--- a/internal/cloud/cloud.go
+++ b/internal/cloud/cloud.go
@@ -86,7 +86,8 @@ func ensureVMForUser(vmHostname string, w io.Writer, username string, sshCmd *op
 
 func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername string) error {
 	color.New(color.FgMagenta, color.Bold).Fprint(w, "Devbox Cloud\n")
-	fmt.Fprint(w, "Remote development environments powered by Nix\n\n")
+	fmt.Fprint(w, "Remote development environments powered by Nix\n")
+	fmt.Fprint(w, "This virtual environment will exit after 8 hours of inactivity\n\n")
 	fmt.Fprint(w, "This is an open developer preview and may have some rough edges. Please report any issues to https://github.com/jetpack-io/devbox/issues\n\n")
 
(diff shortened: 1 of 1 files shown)


internal/cloud/cloud.go
```diff
@@ -86,7 +86,8 @@ func ensureVMForUser(vmHostname string, w io.Writer, username string, sshCmd *op
 
 func Shell(ctx context.Context, w io.Writer, projectDir string, githubUsername string) error {
 	color.New(color.FgMagenta, color.Bold).Fprint(w, "Devbox Cloud\n")
-	fmt.Fprint(w, "Remote development environments powered by Nix\n\n")
+	fmt.Fprint(w, "Remote development environments powered by Nix\n")
+	fmt.Fprint(w, "This virtual environment will exit after 8 hours of inactivity\n\n")
```
gcurtis, line 90: I think we should be explicit that all data will be lost unless it's committed and pushed somewhere else. Exiting sounds like the environment might just shut down but won't be lost.

Lagoja: ++

LucilleH: All the data is synced to local file system. Do we still have the lost data concern for the CLI scenario? (I hear you on the browser side) @gcurtis @Lagoja

savil, line 90: could we make it `ux.Fwarning` ?

LucilleH: If the user disconnect before syncing completes, their local has all the changes, just the vm doesn't have the latest changes. I don't think users will lose their work tho

LucilleH:
Upon further reflection, user will not lose any data if the vm exits, since the CLI works differently from the web. I'm going to close this pull request.

internal/cloud/cloud.go
gcurtis: But the user could make some changes on the VM and disconnect before they finish syncing back locally.

PR #799, opened 2023-03-22
[Lagobot 2.0] part 1: generate testscript for a single devbox example project

savil:
## Summary

**Why**
Our goal is to make each Devbox release higher-quality by ensuring that all the devbox-projects in `examples/` are able to run.

**What**
We aim to do so by adding a `run_test` devbox script to each project in `examples/`, and ensuring that it can run. This has the additional benefit of making the `devbox.json` self-documenting for any user who is reading the code in the `examples/` directory: they can refer to the `run_test` to understand how to execute that example project.

**How**
To do so in our automated tests:
1. We will generate a simple testscript for each of these devbox-projects that calls
`devbox run run_test`.
2. We copy the devbox project's files and directories into the testscript's workdir.

As a proof of concept, this PR runs a testscript for the `examples/development/go/hello-world` project. A future PR will generalize this to all of the devbox projects in `examples/`

## How was it tested?

Ran with this command:
```
DEVBOX_DEBUG=0 go test -run TestExamples/development/go/hello-world -count 1 ./testscripts
```

### DIFF
--- a/testscripts/testrunner/examplesrunner.go
+++ b/testscripts/testrunner/examplesrunner.go
@@ -1,0 +1,97 @@
+package testrunner
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/rogpeppe/go-internal/testscript"
+	"go.jetpack.io/devbox/internal/debug"
+)
+
+// RunExamplesTestscripts generates testscripts for each example devbox-project.
+// For now, we prototype with the "go" example project. TODO savil: generalize.
+func RunExamplesTestscripts(t *testing.T, examplesDir string) {
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Error(err)
+	}
+
+	// TODO savil. Change this to handle _all_ of the example folder's devbox-projects
+	examplesDir = filepath.Join(wd, examplesDir)
+	projectDir := filepath.Join(examplesDir, "development", "go", "hello-world")
+	runSingleExampleTestscript(t, examplesDir, projectDir)
+}
+
+func runSingleExampleTestscript(t *testing.T, examplesDir, projectDir string) {
+	testscriptDir, err := generateTestscript()
+	if err != nil {
+		t.Error(err)
+	}
+	defer os.RemoveAll(testscriptDir)
+
+	testName, err := filepath.Rel(examplesDir, projectDir)
+	if err != nil {
+		t.Error(err)
+	}
+
+	t.Run(testName, func(t *testing.T) {
+		params := getTestscriptParams(testscriptDir)
+
+		// save a reference to the original params.Setup so that we can wrap it below
+		setup := params.Setup
+		params.Setup = func(env *testscript.Env) error {
+			// setup the devbox testscript environment
+			if err := setup(env); err != nil {
+				return errors.WithStack(err)
+			}
+
+			// copy all the files and folders of the devbox-project being tested to the workdir
+			debug.Log("copying projectDir: %s to env.WorkDir: %s\n", projectDir, env.WorkDir)
+			// implementation detail: the period at the end of the projectDir/.
+			// is important to ensure this works for both mac and linux.
+			// Ref.https://dev.to/ackshaey/macos-vs-linux-the-cp-command-will-trip-you-up-2p00
+			err = exec.Command("cp", "-r", projectDir+"/.", env.WorkDir).Run()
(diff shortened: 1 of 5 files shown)


savil:
* main
  * **PR #799** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/799" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/799?utm_source=stack-comment).

testscripts/testrunner/examplesrunner.go
```diff
@@ -0,0 +1,97 @@
...
+// and the caller is responsible for cleaning up the directory.
+func generateTestscript() (testscriptDir string, err error) {
+	defer func() {
+		// cleanup the temp-dir if there was any error
+		if err != nil {
+			os.RemoveAll(testscriptDir)
+		}
+	}()
+
+	// create a temp-dir to place the testscript file
```
gcurtis, line 80: Doesn't `testscript` run the test in a temp dir?

testscripts/testrunner/run_test.test.txt
```diff
@@ -0,0 +1 @@
+exec devbox run run_test
```
gcurtis, line 1: Will we be adding specific testscripts for other examples? If the testscript is just running `devbox run run_test` why not do that directly in a regular test?

testscripts/testrunner/examplesrunner.go
```diff
@@ -0,0 +1,97 @@
...
+			if err := setup(env); err != nil {
+				return errors.WithStack(err)
+			}
+
+			// copy all the files and folders of the devbox-project being tested to the workdir
+			debug.Log("copying projectDir: %s to env.WorkDir: %s\n", projectDir, env.WorkDir)
+			// implementation detail: the period at the end of the projectDir/.
+			// is important to ensure this works for both mac and linux.
+			// Ref.https://dev.to/ackshaey/macos-vs-linux-the-cp-command-will-trip-you-up-2p00
+			err = exec.Command("cp", "-r", projectDir+"/.", env.WorkDir).Run()
```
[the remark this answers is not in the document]
savil: hmm wouldn't the `cp` command still come from the underlying platform?

testscripts/testrunner/examplesrunner.go
savil: it does, which is the `workDir` we refer to elsewhere in this PR.

What this is doing is generating a directory that holds the "project that has to be tested", which would normally have the `.test.txt` file. I could simplify this by generating a single temp-dir, instead of re-generating this temp-dir for each example (as I do in part 2 PR). 

Will revisit if I can just do a single temp-dir for all examples projects.

testscripts/testrunner/run_test.test.txt
savil: The motivation is ensuring isolation is better...

testscripts/testrunner/setupenv.go
```diff
@@ -60,5 +60,12 @@ func setupCacheHome(env *testscript.Env) error {
 		return err
 	}
 
+	// Golang caches build outputs for reuse in future builds in GOCACHE
+	goBuildCache := filepath.Join(cacheHome, "go-build")
+	env.Setenv("GOCACHE", goBuildCache)
+	if err = os.MkdirAll(goBuildCache, 0755); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
savil: without this I get the following error:
```
failed to initialize build cache at /no-home/Library/Caches/go-build: mkdir /no-home: read-only file system
```

its because `GOCACHE` is likely set to `$HOME/Library/Caches/go-build` and testscripts sets `$HOME` to `/no-home`.

testscripts/testrunner/examplesrunner.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/rogpeppe/go-internal/testscript"
+	"go.jetpack.io/devbox/internal/debug"
+)
+
+// RunExamplesTestscripts generates testscripts for each example devbox-project.
+// For now, we prototype with the "go" example project. TODO savil: generalize.
+func RunExamplesTestscripts(t *testing.T, examplesDir string) {
+	wd, err := os.Getwd()
```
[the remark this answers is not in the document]
savil: ah, I see. I drop this in the part 2 PR then (to simplify merging).

PR #800, opened 2023-03-22
[global] Allow and use multiple nixpkgs commit hashes if needed

mikeland86:
## Summary

This allows using global packages with a different hash from the local devbox.json. It also gets rid of the warning.

Note, this introduces a minor regression. If a php extension is installed globally it will not be linked. Will fix this in a follow up, but it's a bit gnarly with current planner framework.

## How was it tested?

Changed commit hash on global devbox.json, added a few packages and ran `devbox run echo hello`. Inspected `flake.nix` with and without global packages.


gcurtis (approved):
Thank you!

PR #801, opened 2023-03-22
[Lagobot 2.0] part 2: run Examples whose run_test script passes

savil:
## Summary

This PR enables the generated testscript to run over more projects in the `examples/` directory.

- This PR enables those projects that happen to work without any further fixes.
  - It maintains a skipList of disabled projects.
- We remove the `t.Run` so that the testscripts can run in parallel. This improved execution
  time from 150 seconds to 60 seconds locally.
  - Internally, the `testscript` framework already does a `t.Run` with a `t.Parallel()`. 
    By having the outer `t.Run()`, these tests were not being run in parallel.
  - This required a bit of refactoring to rename the `run_test.test.txt` file to
    `<project_path>_run_test.test.txt`. We do this because the framework names
     the golang testcase after this `txt tar` file, and I wanted a nicer name.

## How was it tested?

```
DEVBOX_DEBUG=0 go test -v -run TestExamples -count 1 ./testscripts
```

### DIFF
--- a/testscripts/testrunner/examplesrunner.go
+++ b/testscripts/testrunner/examplesrunner.go
@@ -17,13 +17,7 @@ import (
 
 // RunExamplesTestscripts generates testscripts for each example devbox-project.
 func RunExamplesTestscripts(t *testing.T, examplesDir string) {
-	wd, err := os.Getwd()
-	if err != nil {
-		t.Error(err)
-	}
-
-	examplesDir = filepath.Join(wd, examplesDir)
-	err = filepath.WalkDir(examplesDir, func(path string, entry os.DirEntry, err error) error {
+	err := filepath.WalkDir(examplesDir, func(path string, entry os.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #799** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/799" target="_blank">Graphite</a> 
    * **PR #801** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/801" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/801?utm_source=stack-comment).

testscripts/testrunner/examplesrunner.go
```diff
@@ -1,82 +1,143 @@
...
+		if err != nil {
+			return err
+		}
+		if _, ok := config.Shell.Scripts["run_test"]; !ok {
+			t.Logf("skipping config due to missing run_test at: %s\n", path)
+			return nil
+		}
+
+		// TODO savil. Resolve these.
+		skipList := []string{
```
gcurtis, line 55: If you define this as a `map[string]bool` outside of the walk function then you can simplify down to:

```go
if shouldSkip[path] {
	t.Logf("skipping example %q because it's broken", path)
	return nil
}
```

testscripts/testrunner/examplesrunner.go
```diff
@@ -1,82 +1,143 @@
...
 
-	testName, err := filepath.Rel(examplesDir, projectDir)
+}
+
+func runSingleExampleTestscript(t *testing.T, examplesDir, projectDir string) {
+	testscriptDir, err := generateTestscript(examplesDir, projectDir)
 	if err != nil {
 		t.Error(err)
 	}
+	t.Cleanup(func() { _ = os.RemoveAll(testscriptDir) })
```
gcurtis, line 84: You can make `generateTestscript` take a `testing.T` and use `t.TempDir` so you don't need to create/clean it yourself.

gcurtis (approved):

PR #804, opened 2023-03-23
[vscode-extension] open vscode browser

mohsenari:
## Summary
Added capability to parse vscode:// urls which can be used for `open-in-vscode` button. 
The extension still needs handling when OS is not linux/macos and cover more edge cases. But Happy path works.

## How was it tested?
- go to devbox.sh and create a new VM (from examples or /new)
- inspect element from elements tab save vm_id and from application tab for cookies under `<vm_id>.vm.devbox.sh` save `id_token`'s value.
- create a link with this format: `vscode://jetpack-io.devbox?vm_id=<vm_id>&token=<id_token>` and save this.
- while on this branch open `vscode-extension` directory in vscode (`code devbox/vscode-extension`)
- Go to debug tab from the left side menu and click run and debug (this spawns a new vscode window with devbox extension loaded into it).
- open a random file in the newly opened vscode window then on the browser go to the link that you saved 3 steps before
- follow the prompts on browser and vscode and it should connect your new vscode window's environment to devbox cloud vm.

### DIFF
--- a/vscode-extension/package.json
+++ b/vscode-extension/package.json
@@ -100,16 +100,21 @@
     "test": "node ./out/test/runTest.js"
   },
   "devDependencies": {
-    "@types/vscode": "^1.72.0",
     "@types/glob": "^8.0.0",
     "@types/mocha": "^10.0.0",
     "@types/node": "16.x",
+    "@types/node-fetch": "2",
+    "@types/vscode": "^1.72.0",
     "@typescript-eslint/eslint-plugin": "^5.38.1",
     "@typescript-eslint/parser": "^5.38.1",
+    "@vscode/test-electron": "^2.1.5",
     "eslint": "^8.24.0",
     "glob": "^8.0.3",
     "mocha": "^10.0.0",
-    "typescript": "^4.8.4",
-    "@vscode/test-electron": "^2.1.5"
+    "typescript": "^4.8.4"
+  },
+  "dependencies": {
+    "form-data": "^4.0.0",
+    "node-fetch": "2"
   }
 }
--- a/vscode-extension/src/openinvscode.ts
+++ b/vscode-extension/src/openinvscode.ts
@@ -1,0 +1,73 @@
+import { Uri, commands, window } from 'vscode';
+import fetch from 'node-fetch';
+import { exec } from 'child_process';
+import FormData = require('form-data');
+import { open, writeFile } from 'fs/promises';
+
+export async function handleOpenInVSCode(uri: Uri) {
+    const queryParams = new URLSearchParams(uri.query);
+
+    if (queryParams.has('vm_id') && queryParams.has('token')) {
+        window.showInformationMessage('Setting up devbox');
+
+        // getting ssh keys
+        const response = await getVMInfo(queryParams.get('token'), queryParams.get('vm_id'));
+        const res = await response.json();
+        // TODO: remove debug
+        console.log("data:\n");
+        console.log(res);
+        // set ssh config
+        await setupSSHConfig(res['vm_id'], res['private_key']);
+        // connect to remote vm
+        connectToRemote(res['username'], res['vm_id']);
+    } else {
+        window.showErrorMessage('Error parsing information for remote environment.');
+    };
+}
+
+async function getVMInfo(token: string | null, vmId: string | null): Promise<any> {
+    // send post request to gateway to get vm info and ssh keys
+    const gatewayHost = 'https://api.devbox.sh/g/vm_info';
+    const data = new FormData();
+    data.append("vm_id", vmId);
+    const response = await fetch(gatewayHost, {
+        method: 'post',
+        body: data,
+        headers: {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            'Authorization': `Bearer ${token}`
+        }
+    });
+    return response;
+}
+
+async function setupSSHConfig(vmId: string, prKey: string) {
+    exec('./devbox generate ssh-config --config=./', (error, stdout, stderr) => {
+        if (error) {
+            console.error(`exec error: ${error}`);
+            return;
+        }
+        console.log(`stdout: ${stdout}`);
+        console.log(`stderr: ${stderr}`);
+    });
+    // save private key to file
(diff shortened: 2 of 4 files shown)


vscode-extension/src/openinvscode.ts
```diff
@@ -0,0 +1,77 @@
...
+        window.showInformationMessage('Setting up devbox');
+
+        // getting ssh keys
+        const response = await getVMInfo(queryParams.get('token'), queryParams.get('vm_id'));
+        const res = await response.json();
+        // TODO: remove debug
+        console.log("data:");
+        console.log(res);
+        // set ssh config
+        await setupSSHConfig(res['vm_id'], res['private_key']);
```
LucilleH, line 20: Define a type for `res` and you can do:
```
type VmInfo = {
  vm_id: string;
  private_key: string;
  username: string;
}

const res = await response.json() as VmInfo;
res.vm_id...
```

vscode-extension/src/openinvscode.ts
```diff
@@ -0,0 +1,77 @@
...
+    // send post request to gateway to get vm info and ssh keys
+    const gatewayHost = 'https://api.devbox.sh/g/vm_info';
+    const data = new FormData();
+    data.append("vm_id", vmId);
+    const response = await fetch(gatewayHost, {
+        method: 'post',
+        body: data,
+        headers: {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            'Authorization': `Bearer ${token}`
```
LucilleH, line 38: `authorization` or `Authorization` doesn't work? Does it have to be quoted?

LucilleH (approved):
🎉

vscode-extension/src/openinvscode.ts
```diff
@@ -0,0 +1,77 @@
...
+        console.log(`stdout: ${stdout}`);
+        console.log(`stderr: ${stderr}`);
+    });
+    // save private key to file
+    const prkeyPath = `${process.env['HOME']}/.config/devbox/ssh/keys/${vmId}.vm.devbox-vms.internal`;
+    try {
+        const prKeydata = new Uint8Array(Buffer.from(prKey));
+        const fileHandler = await open(prkeyPath, 'w');
+        await writeFile(fileHandler, prKeydata, { flag: 'w' });
+        await chmod(prkeyPath, 0o600);
```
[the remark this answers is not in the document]
mohsenari: 😄 no it is `0o600` I've never seen it like that either but this is from nodejs docs:

> mode [<string>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type) | [<integer>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type) Sets the file mode (permission and sticky bits) if the file is created. Default: 0o666 (readable and writable)
Returns: [<Promise>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) Fulfills with a [<FileHandle>](https://nodejs.org/api/fs.html#class-filehandle) object.

vscode-extension/src/openinvscode.ts
```diff
@@ -0,0 +1,77 @@
+import { Uri, commands, window } from 'vscode';
+import fetch from 'node-fetch';
+import { exec } from 'child_process';
+import FormData = require('form-data');
```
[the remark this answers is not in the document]
mohsenari: I know it's different than usual imports for typescript. But this was due to `form-data` not being exported properly. ESLint kept erroring the import and suggesting this way.
Note that this syntax is slightly different than `const FormData = require('form-data');`

vscode-extension/package.json
```diff
@@ -100,16 +100,21 @@
...
     "eslint": "^8.24.0",
     "glob": "^8.0.3",
     "mocha": "^10.0.0",
-    "typescript": "^4.8.4",
-    "@vscode/test-electron": "^2.1.5"
+    "typescript": "^4.8.4"
+  },
+  "dependencies": {
+    "form-data": "^4.0.0",
+    "node-fetch": "2"
```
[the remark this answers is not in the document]
mohsenari: good call I did `yarn add node-fetch@2` but I can change this to that it includes minor and patch updates.

vscode-extension/src/openinvscode.ts
mohsenari: I followed fetch's example from its own docs. I don't think it's necessary but I'll test

vscode-extension/src/openinvscode.ts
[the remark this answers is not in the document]
mohsenari: my guess is `0x` is hexadecimal and `0o` is decimal base

vscode-extension/src/openinvscode.ts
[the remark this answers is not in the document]
mohsenari: followup: just tried `import * as FormData from 'form-data';` and it worked.

vscode-extension/src/openinvscode.ts
mohsenari: followup: tried `authorization` without quotes and it worked. So will leave it as lowercase without quote `authorization`

vscode-extension/package.json
```diff
@@ -100,16 +103,21 @@
     "test": "node ./out/test/runTest.js"
   },
   "devDependencies": {
-    "@types/vscode": "^1.72.0",
     "@types/glob": "^8.0.0",
     "@types/mocha": "^10.0.0",
     "@types/node": "16.x",
+    "@types/node-fetch": "2",
```
LucilleH, line 109: Uh, this should probably also update to `^2`?

vscode-extension/src/openinvscode.ts
```diff
@@ -0,0 +1,83 @@
...
+import { chmod, open, writeFile } from 'fs/promises';
+
+export async function handleOpenInVSCode(uri: Uri) {
+    const queryParams = new URLSearchParams(uri.query);
+
+    if (queryParams.has('vm_id') && queryParams.has('token')) {
+        window.showInformationMessage('Setting up devbox');
+
+        // getting ssh keys
+        type VmInfo = {
```
LucilleH, line 14: Declare type outside of this function

LucilleH (commented):
Maybe do those in a follow up?

PR #805, opened 2023-03-23
Improve installation instructions

Lagoja:
## Summary

Clarifies Installation instructions for Linux, macOS, Windows/WSL, and Nix. This should also make clearer to developers how they can use us on Windows machines, and which platforms we support

## How was it tested?

Localhost + Launchpad


docs/app/devbox.json
```diff
@@ -1,12 +1,31 @@
...
+    "httpie",
+    "git",
+    "ripgrep",
+    "glow",
+    "gifsicle",
+    "epr",
+    "nixfmt",
+    "pulumi-bin",
+    "rnix-lsp",
+    "graphviz"
```
LucilleH, line 23: why all these packages?

Lagoja: That's weird, those are my global packages. They shouldn't be included

Lagoja: Fixed

LucilleH (approved):

PR #807, opened 2023-03-23
[vscode extension] Added getting working directory from `getvminfo()`

mohsenari:
## Summary
Title says it.

This is the follow up from https://github.com/jetpack-io/axiom/pull/3098

## How was it tested?
See #804


LucilleH:
@mohsenari can we address the follow ups here as well? https://github.com/jetpack-io/devbox/pull/804

LucilleH (approved):

PR #813, opened 2023-03-24
[hooks] Fix hook order

mikeland86:
## Summary

plugin hooks should run first, user hooks last

## How was it tested?

```bash
➜  devbox git:(main) ✗ cd examples/development/python/pip
➜  pip git:(main) ✗ devbox run run_test
```


savil (approved):
thanks!

PR #817, opened 2023-03-24
[lagobot] enable pip and django

savil:
## Summary

We use period `.` instead of `source`, and can now enable pip and django.

Not entirely sure why but I think the testscripts run in `sh` and not `bash`. `source` is a bash-ism.

## How was it tested?

buildkite run needs to be green

### DIFF
--- a/examples/development/python/pip/devbox.json
+++ b/examples/development/python/pip/devbox.json
@@ -5,7 +5,7 @@
     ],
     "shell": {
         "init_hook": [
-            "source $VENV_DIR/bin/activate",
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #817** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/817" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/817?utm_source=stack-comment).

mikeland86 (approved):

PR #820, opened 2023-03-24
[lagobot] enable pipenv

savil:
## Summary

The pipenv stuff is slow, because it seems to not respect the nix packages cache? Not fully sure. We should debug it.

But we can land it for now, since we are no longer running examples tests on PRs. They will be run on `main` soon.


savil:
* main
  * **PR #820** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/820" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/820?utm_source=stack-comment).

LucilleH (approved):

PR #822, opened 2023-03-24
[lagobot] add integration tag to separate the tests

savil:
## Summary

I'm in the middle of adding this but sending this pre-emptively so that Lagobot tests
can stop running in CICD.

The integration tests will only run via `go test -tags=integration`

References:
1. https://www.reddit.com/r/golang/comments/a1iuhg/comment/eat8uyn/?utm_source=reddit&utm_medium=web2x&context=3
2. https://mickey.dev/posts/go-build-tags-testing/

## How was it tested?

ran locally but rm_test testscript is failing (unrelated to this?)

### DIFF
--- a/testscripts/testrunner/examplesrunner.go
+++ b/testscripts/testrunner/examplesrunner.go
@@ -1,3 +1,6 @@
+//go:build integration
+// +build integration
+
 package testrunner
 
 import (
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #822** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/822" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/822?utm_source=stack-comment).

mikeland86 (approved):

gcurtis (commented):
The downside to build tags is that it's easy to break them because they're not built by default. It can also mess up autocomplete because the LSP server won't build them.

Could we add a `-run` flag in CI to not run the examples tests?

PR #824, opened 2023-03-24
[testscripts] respect DEVBOX_DEBUG flag instead of always setting it on

savil:
## Summary

It is very hard to parse the output from testscripts because we always turn on DEVBOX_DEBUG output as well.

Instead, we should respect the invoking user's DEVBOX_DEBUG value.

## How was it tested?

shows output as before:
```
 DEVBOX_DEBUG=1 go test -v ./testscripts/...
```

cleaner output:
```
go test -v ./testscripts/...
```

### DIFF
--- a/testscripts/testrunner/setupenv.go
+++ b/testscripts/testrunner/setupenv.go
@@ -17,10 +17,7 @@ func setupTestEnv(env *testscript.Env) error {
 		return err
 	}
 
-	// Enable new `devbox run` so we can use it in tests. This is temporary,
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #824** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/824" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/824?utm_source=stack-comment).

gcurtis (approved):

PR #834, opened 2023-03-28
[lagobot] ensure default nixpkgs commit by deleting the commits from devbox.json

savil:
## Summary

We want to ensure that the default Nixpkgs commit that we use works with all
Lagobot example projects. So, we delete the nixpkgs commit hash from the devbox.json.
When the testscript will execute `devbox run run_test`, the default nixpkgs commit
hash will be added to the devbox.json and used.

## How was it tested?

tests should pass in buildkite

### DIFF
--- a/examples/cloud_development/argo-workflows/devbox.json
+++ b/examples/cloud_development/argo-workflows/devbox.json
@@ -27,8 +27,5 @@
                 "minikube -p argo-test logs -f"
             ]
         }
-    },
-    "nixpkgs": {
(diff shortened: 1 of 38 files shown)


savil:
* main
  * **PR #834** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/834" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/834?utm_source=stack-comment).

Lagoja (approved):

gcurtis (approved):

PR #835, opened 2023-03-28
[vscode-extension] Handling for window os and devbox not installed

mohsenari:
## Summary
- Added handling for open in vscode on windows (show as unsupported will implement handling WSL in future).
- Added handling for devbox not installed for open in vscode by downloading and using launcher script.

 
## How was it tested?
See steps in #804

### DIFF
--- a/vscode-extension/src/openinvscode.ts
+++ b/vscode-extension/src/openinvscode.ts
@@ -1,5 +1,6 @@
 import { Uri, commands, window } from 'vscode';
 import fetch from 'node-fetch';
+import * as os from 'os';
 import { exec } from 'child_process';
 import * as FormData from 'form-data';
 import { chmod, open, writeFile } from 'fs/promises';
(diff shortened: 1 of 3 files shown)


vscode-extension/src/openinvscode.ts
```diff
@@ -49,13 +54,34 @@ async function getVMInfo(token: string | null, vmId: string | null): Promise<any
     return response;
 }
 
+async function setupDevboxLauncher(): Promise<any> {
+    // download devbox launcher script
+    const gatewayHost = 'https://releases.jetpack.io/devbox';
+    const response = await fetch(gatewayHost, {
+        method: 'get',
+    });
+    const launcherPath = `${process.env['HOME']}/.config/devbox/launcher.sh`;
```
LucilleH, line 63: Do we always download the launcher on every ssh setup? Should we check if devbox is already installed first by checking if this file exist already?

LucilleH (approved):

vscode-extension/src/openinvscode.ts
mohsenari: Fair point. But I don't think checking if the file exists is a good idea. Because we can't guarantee the content of the file and blindly running it doesn't seem safe to do. 
I can check if devbox is installed using this https://www.npmjs.com/package/which and if not I'll download launcher

PR #839, opened 2023-03-29
Update CLI reference up-to-date

hezhizhen:
## Summary

Update CLI reference based on the current help info.


hezhizhen:
I updated some commands' descriptions and then ran `devbox gen-docs docs/app/docs/cli_reference/`.

docs/app/docs/cli_reference/_category_.json
```diff
@@ -1,4 +0,0 @@
-{
```
hezhizhen, line 1: Should I keep this file? It seems cobra deletes everything and then generates new docs.

hezhizhen:
@savil You can see all the commands here.

Lagoja (changes requested):
Hey hezhiszhen, thanks for contributing these updates to the docs! The updates to the inline help are definitely needed, and there are a few commands like `completion` and `cloud forward` that were missing from our documentation

I think there's a few issues with this PR that make it hard to merge in it's current state: 

1. The `gen-docs` command creates the titles at H2 instead of H1, which means that Docusaurus uses the filename instead of a user friendly title in the sidebar + page heading. This leads to docs being named `devbox_shell` instead of `devbox shell`, which is less user friendly. At the moment there's no automated way to fix this. 

2. This PR ran `gen-docs` off of main, instead of our latest release tag (0.4.5). This means there are some unreleased or unfinished commands like `update` and `version update` included in the docs. We should always generate docs off the latest release tag. 

3. In a few pages, we've manually edited them to have some more information or examples than what's in our inline help. We should try to preserve these notes to help out users. 

I think what I'd like to do is merge just the changes you made in `internal/boxcli`. I can pull the docs changes which are relevant to 0.4.5 into a separate PR.

hezhizhen:
> Hey @hezhizhen , thanks for contributing these updates to the docs! The updates to the inline help are definitely needed, and there are a few commands like `completion` and `cloud forward` that were missing from our documentation
> 
> I think there's a few issues with this PR that make it hard to merge in it's current state:
> 
> 1. The `gen-docs` command creates the titles at H2 instead of H1, which means that Docusaurus uses the filename instead of a user friendly title in the sidebar + page heading. This leads to docs being named `devbox_shell` instead of `devbox shell`, which is less user friendly. At the moment there's no automated way to fix this.
> 2. This PR ran `gen-docs` off of main, instead of our latest release tag (0.4.5). This means there are some unreleased or unfinished commands like `update` and `version update` included in the docs. We should always generate docs off the latest release tag.
> 3. In a few pages, we've manually edited them to have some more information or examples than what's in our inline help. We should try to preserve these notes to help out users.
> 
> I think what I'd like to do is merge just the changes you made in `internal/boxcli`. I can pull the docs changes which are relevant to 0.4.5 into a separate PR.

I see, and that's reasonable. I'll create 2 PRs based on this one:

1. update descriptions of commands
2. update github action to run `gen-docs` for releases

PR #841, opened 2023-03-29
redact: package for redacting error messages

gcurtis:
Package redact implements functions to redact sensitive information from errors. A basic example is:

	name := "Alex"
	id := 5
	err := redact.Errorf("error getting user %s with ID %d",
		name, Safe(id))

	fmt.Println(err)
	// error getting user Alex with ID 5

	fmt.Println(redact.Error(err))
	// error getting user <redacted string> with ID 5

See the package docs and tests for more examples.

This will allow us to get more information in Sentry without sending any identifying information.

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -70,8 +70,10 @@ func Error(err error) error {
 	}
 
 	switch t := err.(type) {
-	case *safeError:
-		return t.redacted
+	case *redactedError:
+		// Don't redact an already redacted error, otherwise its redacted message
+		// will be replaced with a placeholder.
+		return err
 	case redactor:
 		return &redactedError{
 			msg:     t.Redact(),
@@ -176,6 +178,7 @@ type safeError struct {
 func (e *safeError) Error() string  { return e.err.Error() }
 func (e *safeError) Redact() string { return e.redacted.Error() }
 func (e *safeError) Unwrap() error  { return e.err }
+
 func (e *safeError) StackTrace() []runtime.Frame {
 	if len(e.callers) == 0 {
 		return nil
@@ -192,8 +195,23 @@ func (e *safeError) StackTrace() []runtime.Frame {
 	return frames
 }
 
+func (e *safeError) Format(f fmt.State, verb rune) {
+	switch verb {
+	case 'v', 's':
+		f.Write([]byte(e.Error()))
+		if f.Flag('+') {
+			for _, fr := range e.StackTrace() {
+				fmt.Fprintf(f, "\n%s\n\t%s:%d", fr.Function, fr.File, fr.Line)
+			}
+			return
+		}
+	default:
+		fmt.Fprintf(f, fmt.FormatString(f, verb), e.Error())
+	}
+}
+
 // redactedError is an error containing a redacted message. It is usually the
-// result of calling safeError.Redact.
+// result of calling Error(safeError).
 type redactedError struct {
 	msg     string
 	wrapped error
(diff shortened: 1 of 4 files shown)


internal/redact/redact_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+func TestSimple(t *testing.T) {
+	err := errors.New("simple")
+	checkUnredactedError(t, err, "simple")
+	checkRedactedError(t, err, "<redacted *errors.errorString>")
+}
+
+func TestSimpleWrapSimple(t *testing.T) {
+	wrapped := errors.New("error 2")
+	err := fmt.Errorf("error 1: %w", wrapped)
```
savil, line 69: does this also work with `errors.Wrap(err, "some wrapping msg")`  from `github.com/pkg/errors`?

savil (approved):
this is awesome!

internal/redact/redact.go
```diff
@@ -0,0 +1,218 @@
...
+//
+//	Errorf("cannot find user %s", Safe(username)).Error()
+//	// cannot find user bob
+func Errorf(format string, a ...any) error {
+	// Capture a stack trace.
+	safeErr := &safeError{
+		callers: make([]uintptr, 32),
+	}
+	n := runtime.Callers(2, safeErr.callers)
+	safeErr.callers = safeErr.callers[:n]
```
mikeland86, line 122: Is this going to print stack traces correctly with out existing error handling? (using `pkg/errors` ? )

mikeland86 (commented):
This is a really nice library and well documented!

I think this could be a good replacement for fmt.Errorf and errors.Errorf (assuming we handle stack traces similarly well). Two concerns:

* There's an additional mental overhead to determining if something should be safe or not. Redacted by default is good choice.
* We also add potential PII via `errors.Wrap` and  a few `errors.WithMessage`. These messages are kept independent of initial error and stack trace so in theory we can extract the errors without the messages. It won't have the benefit of the redact lib because we can't identify args from the messages (i.e. when using `errors.Wrapf`). I guess a potential follow up in this library would be to add `redact.Wrapf` as well.

internal/redact/redact.go
gcurtis: I updated `safeError` to format its stack trace the same way `pkg/errors` does. I double checked and the output is identical.

gcurtis:
@mikeland86 

> This is a really nice library and well documented!

Thanks!

> * There's an additional mental overhead to determining if something should be safe or not. Redacted by default is good choice.

My thinking is that you normally want to see the whole error message (whether you're a developer or a user), which is why errors don't redact their message by default. I felt like needing to calling an `Unredact` function everywhere to get a helpful error would be annoying. Instead, the error can be redacted right before it's sent somewhere with a single `redact.Error` call.

> * We also add potential PII via `errors.Wrap` and  a few `errors.WithMessage`. These messages are kept independent of initial error and stack trace so in theory we can extract the errors without the messages. It won't have the benefit of the redact lib because we can't identify args from the messages (i.e. when using `errors.Wrapf`). I guess a potential follow up in this library would be to add `redact.Wrapf` as well.

I don't think I follow. With the `errors.Wrap` and `errors.WithMessage` functions you still need to redact the error it's wrapping unless you know it's safe. For example, you can't just send an `os.Open` error because it will have a file path. We could print the error type (which is what this package does when there's no alternative message), but that's similar to what we do today.

The `redact.Errorf` is the same as `redact.Wrapf` in that it creates a new error that wraps another. For example, `redact.Errorf("bad id %d: %w", id, err) == errors.Wrapf(err, "bad id %d", id)`. Following the stdlib's `fmt.Errorf` lets `redact` be used as a drop-in replacement. Replacing all instances of `fmt.Errorf` with `redact.Errorf` will get you the same exact error messages until you call `redact.Error`, making it easy to add to an existing codebase.

PR #847, opened 2023-03-30
chore: update descriptions for commands

hezhizhen:
## Summary

1. Use imperative sentences for command descriptions
2. Use exported functions for commands added to the root

### DIFF
--- a/internal/boxcli/add.go
+++ b/internal/boxcli/add.go
@@ -19,7 +19,7 @@ type addCmdFlags struct {
 	config configFlags
 }
 
-func AddCmd() *cobra.Command {
+func addCmd() *cobra.Command {
 	flags := addCmdFlags{}
 
 	command := &cobra.Command{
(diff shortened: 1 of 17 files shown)


ipince (approved):
Thanks!

savil:
> Use exported functions for commands added to the root

why is this needed?

hezhizhen:
> > Use exported functions for commands added to the root
> 
> why is this needed?

Not needed, because all functions are in the same package. But most of functions for commands added to the root are exported and sub-commands for these commands are unexported. So I think it's better to follow the pattern.

savil:
Oh I see. I think that may be an oversight. In general, we should bias towards not visible _unless_ we intentionally want to export the function.

So, could we instead go the other way, and make the exported Command Functions be internal? I don't see a good reason for why they need to be exported.

hezhizhen:
> Oh I see. I think that may be an oversight. In general, we should bias towards not visible _unless_ we intentionally want to export the function.
> 
> So, could we instead go the other way, and make the exported Command Functions be internal? I don't see a good reason for why they need to be exported.

Okay, I've changed all functions to be internal.

savil (approved):

savil:
@hezhizhen thank you! for your PR and openness to modifying it.

PR #848, opened 2023-03-30
ci: update docs when deploying prod version

hezhizhen:
## Summary

When deploying the prod version, use `devbox gen-docs` to update cli reference up-to-date.

### DIFF
--- a/.github/workflows/docs-deploy-prod.yaml
+++ b/.github/workflows/docs-deploy-prod.yaml
@@ -20,6 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


savil:
@hezhizhen unfortunately, it seems that the generated docs are not good enough to always deploy. @Lagoja thoughts?

hezhizhen:
> @hezhizhen unfortunately, it seems that the generated docs are not good enough to always deploy. @Lagoja thoughts?

Maybe we can customize the output format?

savil:
true. @Lagoja any specific thoughts or guidance you could offer in this regard?

hezhizhen:
No good idea. Close it first.

PR #849, opened 2023-03-30
[linting] enable more rules for revive and address issues

hezhizhen:
## Summary

1. Enable more rules for revive.
2. Address existing issues.

## How was it tested?

devbox run lint

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,8 +35,25 @@ linters-settings:
   errorlint:
     errorf: false
   revive:
-    rules:
+    rules: # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md
+      - name: atomic
+      - name: bare-return
+      - name: bool-literal-in-expr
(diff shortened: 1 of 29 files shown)


.golangci.yml
```diff
@@ -35,7 +35,18 @@ linters-settings:
...
+      - name: atomic
+      - name: bare-return
+      - name: bool-literal-in-expr
+      - name: cognitive-complexity
+        arguments:
+          - 36 # TODO: gradually reduce it
+      - name: datarace
+      - name: duplicated-imports
+      - name: early-return
+      - name: empty-lines
```
hezhizhen, line 48: @savil If you don't like it, let's remove this.

savil: yeah, not a huge fan, lets remove please....

savil (approved):
thanks!

.golangci.yml
hezhizhen: Removed.

Issue #851, opened 2023-03-30
[Bug]: MariaDB not starting


gfriloux:
**Current Behavior (bug)**
Using examples for using `MariaDB` does not work for me.
The issue occurs when running `mysql_install_db` (from running `devbox shell`):
```
+ mysql_install_db --auth-root-authentication-method=normal --datadir=/home/kuri/test/conf/mysql/data --basedir=/home/kuri/test/.devbox/virtenv --pid-file=/home/kuri/test/conf/mysql/mysql.pid

FATAL ERROR: Could not find errmsg.sys

The following directories were searched:
```

**Expected Behavior (fix)**
`MariaDB` should start without issue

**Additional context**

I use:
*  `fish` 3.6.1
*  `devbox` 0.4.5
* Example `MariaDB` from : https://github.com/jetpack-io/devbox/tree/main/examples/databases/mariadb

If i change `conf/mysql/mysql.sh` to replace `--basedir=$MYSQL_BASEDIR` with `--basedir=/nix/store/faz61mkdds3q613qr6fpyj9i82qb20k0-mariadb-server-10.6.11` (which is where `nix` stores `mariadb` files), then everything works.

Of course, this path is not "normal" → not a correct fix.

I saw that `MYSQL_BASEDIR` is defined depending on the result of `which mariadb`.
In my case, this resolves to `/home/kuri/test/.devbox/virtenv/bin/mariadb`.
While this path is correct to find binaries, it won't have `MariaDB`'s `share` files

PR #852, opened 2023-03-30
[virtenv] Create symlinks for support dirs, move wrappers to a subdir.

mikeland86:
## Summary

This creates symlinks for all dirs in profile dir except `bin` (because that has wrappers). Also moved bin dir from

`.devbox/virtenv/bin` to `.devbox/virtenv/.wrappers/bin`

and all support symlinks are in `.devbox/virtenv/.wrappers`. This ensures we can't have collisions with plugin data which is stored in `.devbox/virtenv/[plugin-name]`

This fixes https://github.com/jetpack-io/devbox/issues/851

## How was it tested?

```
devbox shell 
ls -al .devbox/virtenv
ls -al .devbox/virtenv/.wrappers
which go
```

Also tested mariadb example

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -22,7 +22,7 @@ const (
 	VirtenvPath         = ".devbox/virtenv"
 )
 
-var WrapperPath = filepath.Join(VirtenvPath, "devbox-wrappers")
+var WrapperPath = filepath.Join(VirtenvPath, ".wrappers")
 var WrapperBinPath = filepath.Join(WrapperPath, "bin")
 
 type config struct {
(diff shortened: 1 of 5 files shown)


internal/plugin/plugin.go
```diff
@@ -19,10 +19,12 @@ import (
 const (
 	devboxDirName       = "devbox.d"
 	devboxHiddenDirName = ".devbox"
-	VirtenvBinPath      = ".devbox/virtenv/bin"
 	VirtenvPath         = ".devbox/virtenv"
 )
 
+var WrapperPath = filepath.Join(VirtenvPath, ".wrappers")
+var WrapperBinPath = filepath.Join(WrapperPath, "bin")
```
mikeland86, line 26: I plan on moving this out of here, but I just wanted to get this PR out quickly because we're not sure how many people this is affecting.

savil (approved):

PR #855, opened 2023-03-31
boxcli,nix: some better error messages

gcurtis:
Improve some misc. error messages so that they can be useful in Sentry. Also threw in some TODOs for a couple of bug related to error handling.

### DIFF
--- a/internal/nix/profile.go
+++ b/internal/nix/profile.go
@@ -13,7 +13,7 @@ import (
 	"strings"
 
 	"github.com/fatih/color"
-	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/redact"
 )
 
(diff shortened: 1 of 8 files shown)


internal/nix/profile.go
```diff
@@ -41,10 +41,10 @@ func ProfileListItems(writer io.Writer, profileDir string) ([]*NixProfileListIte
 
 	out, err := cmd.StdoutPipe()
 	if err != nil {
-		return nil, errors.WithStack(err)
+		return nil, redact.Errorf("create stdout pipe: %w", redact.Safe(err))
```
mohsenari, line 44: Should this be `"error creating stdout pipe: %w"` ?

mohsenari (approved):

PR #859, opened 2023-03-31
telemetry: report errors in separate process

gcurtis:
Instead of immediately sending errors to Sentry, write them to an XDG state directory as JSON and then upload them in a separate process. This allows devbox to exit faster (Segment still uploads synchronously).

Error reporting works by executing another devbox process with a single "bug" argument (i.e., `devbox bug`). This argument causes the child devbox process to immediately upload any Sentry events and then exit. It skips all other initialization. Because the parent devbox process exits without waiting for the child, the child process may become orphaned. There's a 3 second timeout for uploading events to help ensure that the process always exits.

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -13,6 +13,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/midcobra"
 	"go.jetpack.io/devbox/internal/cloud/openssh/sshshim"
 	"go.jetpack.io/devbox/internal/debug"
+	"go.jetpack.io/devbox/internal/telemetry"
 	"go.jetpack.io/devbox/internal/vercheck"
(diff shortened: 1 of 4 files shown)


internal/boxcli/root.go
```diff
@@ -83,6 +84,10 @@ func Main() {
 		code := sshshim.Execute(os.Args)
 		os.Exit(code)
 	}
+	if len(os.Args) > 1 && os.Args[1] == "bug" {
+		telemetry.ReportErrors()
+		return
+	}
```
mikeland86, line 90: Is this just for performance? Or to avoid having to deal with a cycle? 

Can you add comment.

savil (approved):

mikeland86 (approved):

PR #860, opened 2023-03-31
bump the timeouts for example tests

savil:
## Summary

The macos tests seem to timeout fairly regularly (though not always).

I'm bumping the non-macos tests timeout in anticipation of us hitting this limit
once we add more tests for redis, and other databases.

In a separate line of work, we should find a way to have these tests run faster.

## How was it tested?

buildkite run should be green

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -92,7 +92,7 @@ jobs:
           - is-main: true
             run-example-tests: false
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


mikeland86:
Let's do this to unblock, but this is not scalable. 

I think caching installed nix packages (maybe using artifacts?) might give us a huge benefit. My intuition is most of the slowdown is just downloading and installing nix packages.

mikeland86:
One simple solution is to use a docker container for these tests with preinstalled stuff.

PR #861, opened 2023-03-31
ci: create Sentry release

gcurtis:
Use the Sentry GitHub action to create releases for dev and prod. This should enable better tracking of issues + versions.


gcurtis:
@loreto it looks like we need a Sentry internal integration token for this. When you get a chance could you follow the steps [here](https://github.com/marketplace/actions/sentry-release#create-a-sentry-internal-integration) and then add the token as a `SENTRY_AUTH_TOKEN` secret in the release environment?

loreto (approved):
LGTM

PR #865, opened 2023-04-05
chore: simplify global commands

hezhizhen:
## Summary

Wrap common part in global commands in a function and use that function to simplify global commands.

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
 	"go.jetpack.io/devbox"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/impl"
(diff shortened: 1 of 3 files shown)


hezhizhen:
How about adding this function to `PreRunE` of `globalCmd`?

savil (approved):

savil:
Hmm I don't know if we necessarily want to add this to PreRunE because:
1. we may have some commands that works without any config (example: `init` and `version`).
2. we need the `path` in the RunE parts of the sub-commands here, and I don't know if we can pass that down from PreRunE

savil:
thanks!

hezhizhen:
> 1. we may have some commands that works without any config (example: `init` and `version`).

`PreRunE` only affects global command only, and `PersistentPostRunE` only affects children of global command, so `init` and `version` won't be affected.

> 2. we need the `path` in the RunE parts of the sub-commands here, and I don't know if we can pass that down from PreRunE

We can use persistent flags to pass the config path from global command to its children commands.

savil:
@hezhizhen I'm sorry I provided those examples with insufficient explanation. I was trying to illustrate how we may in the future have some other subcommands that don't require the Global Config.

> We can use persistent flags to pass the config path from global command to its children commands.

ah, yeah, this would work. 

While we could do this, i feel the benefits of moving to the PreRunE are not so big compared to the downside of making reading the logic have a bit more overhead: right now, its clear enough to read the code in each of these RunEs without trying to keep track of a flag and seeing who is writing to it, etc.

The main benefit is in centralizing the logic which you've done in this PR by refactoring to a helper function!

hezhizhen:
> @hezhizhen I'm sorry I provided those examples with insufficient explanation. I was trying to illustrate how we may in the future have some other subcommands that don't require the Global Config.
> 
> > We can use persistent flags to pass the config path from global command to its children commands.
> 
> ah, yeah, this would work.
> 
> While we could do this, i feel the benefits of moving to the PreRunE are not so big compared to the downside of making reading the logic have a bit more overhead: right now, its clear enough to read the code in each of these RunEs without trying to keep track of a flag and seeing who is writing to it, etc.
> 
> The main benefit is in centralizing the logic which you've done in this PR by refactoring to a helper function!

I see, and it makes sense.

PR #866, opened 2023-04-05
[flakes] Add ability to add custom local flakes to devbox project

mikeland86:
## Summary

This allows users to specify a local flake as a package in devbox.json. It can be added by using the following syntax:

`path:path/to/flake#output`

Where `output` can be a package that is part of `packages` or `legacyPackages` output. ([legacyPackages is just a hack that nixpkgs uses for performance](https://github.com/NixOS/nixpkgs/blob/master/flake.nix#L47))

The path to flake can be relative or absolute. If a relative path is used, it is relative to the project dir.

TODOs: (in this PR)

- [x] devbox add doesn't work because we still look for package in nixpkgs
- [x] devbox rm doesn't work because NixProfileListItem.AttributePath is broken for these. (see TODO in code)
- [x] NixProfileListItem.PackageName and callsites needs to modified to ensure we don't re-install the flake output to the profile every time (cc: @savil for visibility)

TODOs: (in follow up)

- [ ] Allow remote (github) flakes 
- [ ] Possible improvement: allow multiple outputs? (e.g. `path:path/to/flake#output1,output2`). This is not nix syntax so not sure we should support. cc: @Lagoja 

## How was it tested?

see flakes/php example.

`devbox run "php -m | grep memcached"`


examples/flakes/README.md
```diff
@@ -0,0 +1,28 @@
+# Flakes
+
+Examples that show how to add custom flakes to your devbox project.
+
+# Local flakes (usually committed to your project)
+
+In devbox.json use "path:/path/to/flake#output" as the package name.
```
gcurtis, line 7: Are these regular flake references? If so, could we call that out in case the user is already familiar with them?

internal/nix/input.go
gcurtis, line 1: Flake references are just URLs, so you should be able to parse them with `url.Parse`. For example: https://go.dev/play/p/KSyymaLk0CN

gcurtis (commented):
Pretty cool! I like using flakerefs for packages. It would be nice to eventually standardize on that for all packages in `devbox.json`.

Let me know when this is ready for a full review.

examples/flakes/README.md
mikeland86: Yep, they are exactly what you use when installing a package that is a flake output. That is why `nix profile install` works almost as is with this change.

mikeland86:
@gcurtis ready to review.

@savil can you take a look at https://github.com/jetpack-io/devbox/pull/866/commits/0e8e1eae68ba4ca3567d80e5eb071e7a71e3d3be it changes the logic to how we determine if a package is already installed in profile. It's a bit less efficient, but a lot simpler to follow. 

I'm gonna measure performance and possibly add some caching to avoid calling `nix profile list` more than once.

internal/nix/profile.go
```diff
@@ -142,13 +181,25 @@ func (item *NixProfileListItem) AttributePath() (string, error) {
...
+	if input.IsFlake() {
+		// TODO(landau): this needs to be normalized so that we can compare
+		// packages.aarch64-darwin.hello and hello and determine that they are
+		// the same package.
+		return item.unlockedReference, nil
+	}
+
+	// TODO(landau/savil): Should we use unlocked reference instead?
+	// I'm not sure we gain anything by using the locked reference. since
+	// the user specifies only the package name when installing
```
savil, line 202: hmm yeah, possibly. 

I can't remember why I chose to do the lockedReference, so its possible there was some scenario I ran into. But give it a try, and we can see if some edge-case doesn't work?

savil: would recommend a separate PR

examples/flakes/php/devbox.json
```diff
@@ -0,0 +1,12 @@
...
+    "path:my-php-flake#php",
+    "path:my-php-flake#hello"
+  ],
+  "shell": {
+    "init_hook": null
+  },
+  "nixpkgs": {
+    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
+  }
+}
```
savil, line 12: could you add a run_test script here so this gets exercised?

internal/impl/packages.go
```diff
@@ -214,34 +217,21 @@ func (d *Devbox) removePackagesFromProfile(ctx context.Context, pkgs []string) e
...
+		})
+		if err != nil {
+			ux.Ferror(d.writer, "Package %s not found in profile. Skipping.\n", pkg)
 		}
 
 		// TODO: unify this with nix.ProfileRemove
 		cmd := exec.Command("nix", "profile", "remove",
 			"--profile", profileDir,
-			attrPath,
+			fmt.Sprintf("%d", index),
```
savil, line 234: using index instead of (packageName, attributePath) sounds good

Could you double-check that the index is stable? Or at least stable enough for the purposes of these operations?

For example: what happens if I am installing 3 packages at the same time?

savil (commented):
Did a quick pass. Nice work!

There's a lot going on here. So, want to have another look in a bit.

Could we double-check the testscript unit tests have adequate coverage to ensure these changes work for various scenarios: with different forms of flake inputs, adding/removing multiple packages? 
Maybe citing them in the Test Plan would be good.

For some of the `nix.Input` stuff, we could add some unit tests as well.

internal/nix/input.go
```diff
@@ -0,0 +1,152 @@
...
+		return "default"
+	}
+	parts := strings.Split(i.Fragment, ".")
+	return parts[len(parts)-1]
+}
+
+func currentSystem() (string, error) {
+	cmd := exec.Command(
+		"nix", "eval",
+		"--impure", "--raw", "--expr",
```
gcurtis, line 127: Does this need to be impure?

gcurtis (approved):
Can be done in a followup, but we should make sure to have some testscript + golden file test coverage before we release.

internal/impl/packages.go
mikeland86: It's stable through adding but not removing. Which is why I added the ability to pass a pre-computed list. When we are removing we recompute the list every time. When adding we compute it once and then add all packages. 

Computing the list every time is not very slow, but we don't want it to happen on `ensure` since that gets called on every shell, run etc. 

In a follow up, we could even cache the list in a file which would allow a 20-30ms speed-up of ensure.

internal/nix/input.go
mikeland86: probably not. I copy pasted this from the internet. I'll fix.

internal/impl/packages.go
savil: I don't know if this is the biggest perf bottleneck. 20-30 ms doesn't seem that much.

PR #867, opened 2023-04-05
Skip interactive survey if devbox is not executed in a terminal

LucilleH:
## Summary
Skip interactive survey if devbox is not executed in a terminal. When `devbox init` or `devbox generate direnv` is called in a dockerfile for example, we do not want to prompt for user input.

I tried `echo 'y' | devbox generate direnv` or `yes | devbox generate direnv` or `devbox generate direnv <<< "y"`. None of them worked.

Bug in the survey package: https://github.com/go-survey/survey/issues/394

## How was it tested?
devbox run build
./dist/devbox init (interactive mode)

docker build -t devbox-cli .
docker run --entrypoint=/bin/sh -ti devbox-cli (non interactive mode)

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -363,8 +363,8 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
 	if commandExists("direnv") {
 		// prompt for direnv allow
 		var result string
-
-		if isatty.IsTerminal(os.Stdin.Fd()) {
+		isInteractiveMode := isatty.IsTerminal(os.Stdin.Fd())
+		if isInteractiveMode {
 			prompt := &survey.Input{
 				Message: "Do you want to enable direnv integration for this devbox project? [y/N]",
 			}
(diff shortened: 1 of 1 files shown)


mohsenari (approved):
Thanks for catching this

internal/impl/devbox.go
```diff
@@ -362,15 +363,18 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
 		// prompt for direnv allow
 		var result string
-		prompt := &survey.Input{
-			Message: "Do you want to enable direnv integration for this devbox project? [y/N]",
-		}
-		err := survey.AskOne(prompt, &result)
-		if err != nil {
-			return errors.WithStack(err)
+
+		if isatty.IsTerminal(os.Stdin.Fd()) {
```
gcurtis, line 367: Could we make the result a bool outside of the `if` so we only need to do `isatty` once? Something like:

```go
enableDirEnv := true
if isatty.IsTerminal(os.Stdin.Fd()) {
	// ...
	enableDirEnv := strings.ToLower(result) == "y"
}

if enableDirEnv {
	// ...
}
```

gcurtis (approved):

PR #869, opened 2023-04-05
[example tests] generate "set -e" to exit scripts upon failure

savil:
## Summary

RFC: we should have failures in the devbox-scripts reflect in a non-zero exit status
from `devbox run <script>`.

There is some risk that customers whose code was previously running will now fail. However, their scripts were indeed failing and its just that this failure will now be exposed. 

Update: Based on John's feedback in a comment below I am adding a featuregate that is activated in examples tests. We'll make a plan for rolling this out to users, in a way that is not a big surprise for them.

TODO:
- [x] fix examples which have failures that were previously masked. Done in previous PRs in the stack.

## How was it tested?

Removed `curl` from `lepp-stack` packages, and ran:
```
DEVBOX_EXAMPLE_TESTS=1 DEVBOX_DEBUG=0 go test -count 1 -v -run TestExamples/stacks_lepp ./testscripts/...
```
This failed, demonstrating that the `set -e` was working as intended.

I then added the `curl` back and the above command passed.

### DIFF
--- a/examples/development/rust/rust-stable-hello-world/devbox.json
+++ b/examples/development/rust/rust-stable-hello-world/devbox.json
@@ -13,7 +13,6 @@
             "build-docs": "cargo doc",
             "start": "cargo run",
             "run_test": [
-                "set -e",
                 "cargo test -- --show-output"
             ]
         }
--- a/examples/stacks/django/devbox.json
+++ b/examples/stacks/django/devbox.json
@@ -29,7 +29,6 @@
                 "python todo_project/manage.py runserver"
             ],
             "run_test": [
-                "set -e",
                 "initdb",
                 "devbox services start",
                 "devbox run create_db",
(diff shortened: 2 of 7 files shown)


savil:
* main
  * **PR #869** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/869" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/869?utm_source=stack-comment).

Lagoja:
Is this done only for testscripts? or for all devbox scripts?

savil:
@Lagoja this is for all users. Thoughts?

Lagoja:
Was thinking about this a bit -- since `-e` makes a script exit immediately upon an error, I worry that this could cause a big wave of unexpected errors if it's automatically pushed to users. 

I think we'd need to either make this something configurable, or do it in a roll out where it's explicitly described and developers have a chance to adapt.

savil:
Sounds good. Let me hold off on this then.

savil:
I'll revive this and gate it for example tests

savil:
oh not ready. Sorry. I'll convert to draft and debug the failing tests.

gcurtis (approved):

PR #870, opened 2023-04-05
List Services and Status via Process Compose

Lagoja:
## Summary

`devbox services ls` can now show the services available in your project, as well as the ones running in process-compose with their status: 

Screenshot 2023-04-05 at 3 35 52 PM

Still needs some tweaking to only show services, and not all processes

## How was it tested?
Localhost on examples. Should not affect backwards compatibility

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -12,14 +12,15 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime/trace"
-	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
+	"golang.org/x/exp/slices"
 
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/generate"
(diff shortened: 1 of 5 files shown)


internal/impl/devbox.go
```diff
@@ -480,6 +481,49 @@ func (d *Devbox) StopServices(ctx context.Context, serviceNames ...string) error
...
+			fmt.Fprintln(d.writer, "Error listing services: ", err)
+		} else {
+			fmt.Fprintln(d.writer, "Services running in process-compose:")
+			for _, s := range pcSvcs {
+				fmt.Fprintf(d.writer, "  %s (%s) \n", s.Name, s.Status)
+			}
+		}
+	}
+
+	fmt.Fprintln(d.writer, "\nServices available in your project:")
```
mikeland86, line 513: Do we always want to show this, even is process-compose is running? Don't we get double services?

internal/services/client.go
```diff
@@ -65,20 +74,34 @@ func RestartServices(ctx context.Context, serviceName string, projectDir string,
 	}
 }
 
-func ListServices(ctx context.Context, serviceName string, projectDir string, w io.Writer) (string, error) {
+func ListServices(ctx context.Context, projectDir string, w io.Writer) ([]ProcessSummary, error) {
 	path := "/processes"
 	method := "GET"
+	var results = []ProcessSummary{}
```
mikeland86, line 80: `results := []ProcessSummary{}`

internal/impl/devbox.go
```diff
@@ -480,6 +481,49 @@ func (d *Devbox) StopServices(ctx context.Context, serviceNames ...string) error
...
+	keys := make([]string, 0, len(svcSet))
+	for k := range svcSet {
+		keys = append(keys, k)
+	}
+	sort.StringSlice(keys).Sort()
+	for _, k := range keys {
+		fmt.Fprintf(d.writer, "  %s\n", svcSet[k].Name)
+	}
```
mikeland86, line 521: You could use `slices.SortFunc`

Also, might be a good idea to have `d.Services` always return sorted.

mikeland86 (approved):

internal/impl/devbox.go
```diff
@@ -480,6 +481,49 @@ func (d *Devbox) StopServices(ctx context.Context, serviceNames ...string) error
...
+		return err
+	}
+
+	if len(svcSet) == 0 {
+		fmt.Fprintln(d.writer, "No services found in your project")
+		return nil
+	}
+
+	if !services.ProcessManagerIsRunning() {
+		fmt.Fprintln(d.writer, "No services currently running. Run `devbox services up` to start them.")
```
savil, line 500: `return nil` after this?

savil (approved):
nice!

internal/impl/devbox.go
Lagoja: Yeah, I thought it might be helpful to see both available _and_ running services, but I think its a lot noisier. I'll only show one

Lagoja:
Updated UI

No Services Running: 
Screenshot 2023-04-05 at 5 23 43 PM


Services Running: 

Screenshot 2023-04-05 at 5 24 06 PM

PR #871, opened 2023-04-06
Add init hook flag to shellenv command

LucilleH:
## Summary
Add init hook flag to shellenv command. This prints out the env variables and the init hooks to stdout, and should replicate the behavior of a `devbox shell` within the context of the current shell.

## How was it tested?
```
devbox shellenv --init-hook
```
And changing the shell by
```
eval "$(devbox shellenv --init-hook)"
```

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -11,7 +11,8 @@ import (
 )
 
 type shellEnvCmdFlags struct {
-	config configFlags
+	config      configFlags
+	runInitHook bool
 }
 
 func shellEnvCmd() *cobra.Command {
@@ -27,9 +28,16 @@ func shellEnvCmd() *cobra.Command {
 				return err
 			}
 			fmt.Fprintln(cmd.OutOrStdout(), s)
+
+			if flags.runInitHook {
+				fmt.Fprintln(cmd.OutOrStdout(), shellEnvInitHookFunc(cmd, flags))
+			}
 			return nil
 		},
 	}
+
+	command.Flags().BoolVar(
+		&flags.runInitHook, "init-hook", false, "runs init hook after exporting shell environment")
 
 	flags.config.register(command)
 	return command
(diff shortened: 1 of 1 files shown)


internal/boxcli/shellenv.go
```diff
@@ -43,3 +51,12 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
...
 	return box.PrintEnv()
 }
+
+func shellEnvInitHookFunc(cmd *cobra.Command, flags shellEnvCmdFlags) string {
+	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
+	if err != nil {
+		return ""
+	}
+
+	return box.Config().Shell.InitHook.String()
```
mikeland86, line 61: Integrate with `shellEnvFunc`? and add error handling

mikeland86 (approved):

internal/boxcli/shellenv.go
mikeland86:
```golang
func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
	if err != nil {
		return "", err
	}

	result := box.PrintEnv()

	if flags.runInitHook {
		box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
		if err != nil {
			return ""
		}
		result := result + "\n" + box.Config().Shell.InitHook.String()
	}

	return result
}
```

ipince:
Couple questions/suggestions:

* Should `devbox shell --print-env` then print the init hook as well? or, remove that call altogether
* We could instead print `. .devbox/gen/scripts/.hooks.sh` instead of copying the entire hook here

LucilleH:
> Couple questions/suggestions:
> 
> * Should `devbox shell --print-env` then print the init hook as well? or, remove that call altogether

I think it should? I'm not sure. I know it is being used in direnv

> We could instead print . .devbox/gen/scripts/.hooks.sh instead of copying the entire hook here

Will there be a race condition of `.devbox/gen/scripts/.hooks.sh` not being generated in time?

PR #872, opened 2023-04-06
[wrappers] Don't re-source shellenv if already sourced

mikeland86:
## Summary

This fixes issue described by https://github.com/amithgeorge/devbox-nodejs-repro-20230406

It's not a perfect fix to all wrapper issues. Specifically:

`USER=foo node -e "console.log(process.env.USER);"`

will print the shellenv user instead of foo.

Will follow up with a fix for that.

cc: @amithgeorge

## How was it tested?

Followed steps in https://github.com/amithgeorge/devbox-nodejs-repro-20230406

Added example that breaks in current release, passes in this PR


loreto (approved):
LGTM

Although tempted to move that example to `testscripts/` instead ... it feels less like an example where we are teaching users how to use devbox in a particular use case, and more like a test-case designed to catch this edge case.

mikeland86:
> Although tempted to move that example to testscripts

@loreto yeah totally. I messed up

PR #873, opened 2023-04-06
Multi-instance Process Compose

Lagoja:
## Summary
This PR allows users to run process-compose for multiple devbox projects at the same time, without any collisions or undesirable behavior. This does _not_ provide any isolation between services in different projects, but it does let a user effectively manage them on a per project basis

## How it works

When starting process-compose in a devbox project, we create a UUID for the project, and then store the port + PID of the process-compose instance in a global `process-compose.json` file, with the UUID as the key. 

When making client requests to process-compose, a devbox project looks up the UUID for the project in the .devbox directory, gets the right port, and then makes the request. This ensures each project is communicating with the correct process-compose instance to start and stop services. 

Devbox also uses the same lookup to get the PID of a project's process-compose instance, so it can stop the right instance. 

## Questions: 
* Should the project UUID be permanent or set somewhere else? Is there a better way to identify a project that won't be affected by the user moving it?

## Future features: 
* Let users specify their own process-compose port in `devbox services up`
* ~Add `devbox services stop --all`, which stops all running process-compose files and resets the global config. This could help with situations where a developer accidentally leaves their projects running.~

## How was it tested?

Ran Example Tests to check backward compatibility, tested by running postgres + jekyll simultaneously, starting and stopping them

### DIFF
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -129,7 +129,7 @@ func stopServices(cmd *cobra.Command, services []string, flags serviceStopFlags)
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	if len(services) >= 0 && flags.allProjects {
+	if len(services) > 0 && flags.allProjects {
 		return errors.New("cannot both specify services and --all-projects")
 	}
 	return box.StopServices(cmd.Context(), flags.allProjects, services...)
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -40,7 +40,7 @@ func (flags *serviceUpFlags) register(cmd *cobra.Command) {
 func (flags *serviceStopFlags) register(cmd *cobra.Command) {
 	flags.configFlags.register(cmd)
 	cmd.Flags().BoolVar(
-		&flags.allProjects, "all-projects", false, "Stop all running services in all your projects")
+		&flags.allProjects, "all-projects", false, "Stop all running services across all your projects.\nThis flag cannot be used with the [service] argument")
 }
 
 func servicesCmd() *cobra.Command {
@@ -71,7 +71,8 @@ func servicesCmd() *cobra.Command {
 
 	stopCommand := &cobra.Command{
 		Use:   "stop [service]...",
-		Short: "Stop service. If no service is specified, stops all services",
+		Short: "Stop one or more services in the current project. If no service is specified, stops all services in the current project.",
+		Long:  `Stop one or more services in the current project. If no service is specified, stops all services in the current project. \nIf the --all-projects flag is specified, stops all running services across all your projects. This flag cannot be used with [service] arguments.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return stopServices(cmd, args, serviceStopFlags)
 		},
@@ -127,6 +128,9 @@ func stopServices(cmd *cobra.Command, services []string, flags serviceStopFlags)
 	box, err := devbox.Open(flags.configFlags.path, cmd.ErrOrStderr())
 	if err != nil {
 		return errors.WithStack(err)
+	}
+	if len(services) >= 0 && flags.allProjects {
+		return errors.New("cannot both specify services and --all-projects")
 	}
 	return box.StopServices(cmd.Context(), flags.allProjects, services...)
 }
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -19,6 +19,11 @@ type serviceUpFlags struct {
 	processComposeFile string
 }
 
+type serviceStopFlags struct {
+	configFlags
+	allProjects bool
+}
+
 func (flags *serviceUpFlags) register(cmd *cobra.Command) {
 	flags.configFlags.register(cmd)
 	cmd.Flags().StringVar(
@@ -32,9 +37,16 @@ func (flags *serviceUpFlags) register(cmd *cobra.Command) {
 		&flags.background, "background", "b", false, "Run service in background")
 }
 
+func (flags *serviceStopFlags) register(cmd *cobra.Command) {
+	flags.configFlags.register(cmd)
+	cmd.Flags().BoolVar(
+		&flags.allProjects, "all-projects", false, "Stop all running services in all your projects")
+}
+
 func servicesCmd() *cobra.Command {
 	flags := servicesCmdFlags{}
 	serviceUpFlags := serviceUpFlags{}
+	serviceStopFlags := serviceStopFlags{}
 	servicesCommand := &cobra.Command{
 		Use:   "services",
 		Short: "Interact with devbox services",
@@ -61,7 +73,7 @@ func servicesCmd() *cobra.Command {
 		Use:   "stop [service]...",
 		Short: "Stop service. If no service is specified, stops all services",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return stopServices(cmd, args, flags)
+			return stopServices(cmd, args, serviceStopFlags)
 		},
 	}
 
@@ -83,6 +95,7 @@ func servicesCmd() *cobra.Command {
 
 	flags.config.register(servicesCommand)
 	serviceUpFlags.register(upCommand)
+	serviceStopFlags.register(stopCommand)
 	servicesCommand.AddCommand(lsCommand)
 	servicesCommand.AddCommand(upCommand)
 	servicesCommand.AddCommand(restartCommand)
@@ -110,12 +123,12 @@ func startServices(cmd *cobra.Command, services []string, flags servicesCmdFlags
 	return box.StartServices(cmd.Context(), services...)
 }
 
-func stopServices(cmd *cobra.Command, services []string, flags servicesCmdFlags) error {
-	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
+func stopServices(cmd *cobra.Command, services []string, flags serviceStopFlags) error {
+	box, err := devbox.Open(flags.configFlags.path, cmd.ErrOrStderr())
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	return box.StopServices(cmd.Context(), services...)
+	return box.StopServices(cmd.Context(), flags.allProjects, services...)
 }
 
 func restartServices(
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -40,7 +40,7 @@ func (flags *serviceUpFlags) register(cmd *cobra.Command) {
 func (flags *serviceStopFlags) register(cmd *cobra.Command) {
 	flags.configFlags.register(cmd)
 	cmd.Flags().BoolVar(
-		&flags.allProjects, "all-projects", false, "Stop all running services across all your projects.\nThis flag cannot be used with the [service] argument")
+		&flags.allProjects, "all-projects", false, "Stop all running services across all your projects.\nThis flag cannot be used simultaneously with the [services] argument")
 }
 
 func servicesCmd() *cobra.Command {
@@ -130,7 +130,7 @@ func stopServices(cmd *cobra.Command, services []string, flags serviceStopFlags)
 		return errors.WithStack(err)
 	}
 	if len(services) > 0 && flags.allProjects {
-		return errors.New("cannot both specify services and --all-projects")
+		return errors.New("cannot use both services and --all-projects arguments simultaneously")
 	}
 	return box.StopServices(cmd.Context(), flags.allProjects, services...)
 }
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -459,9 +459,9 @@ func (d *Devbox) StartServices(ctx context.Context, serviceNames ...string) erro
 	for _, s := range serviceNames {
 		err := services.StartServices(ctx, d.writer, s, d.projectDir)
 		if err != nil {
-			fmt.Printf("Error starting service %s: %s", s, err)
+			fmt.Fprintf(d.writer, "Error starting service %s: %s", s, err)
 		} else {
-			fmt.Printf("Service %s started successfully", s)
+			fmt.Fprintf(d.writer, "Service %s started successfully", s)
 		}
 	}
 	return nil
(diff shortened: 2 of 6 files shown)


internal/impl/devbox.go
```diff
@@ -467,9 +467,18 @@ func (d *Devbox) StartServices(ctx context.Context, serviceNames ...string) erro
...
 
-func (d *Devbox) StopServices(ctx context.Context, serviceNames ...string) error {
+func (d *Devbox) StopServices(ctx context.Context, allProjects bool, serviceNames ...string) error {
 	if !IsDevboxShellEnabled() {
-		return d.RunScript("devbox", append([]string{"services", "stop"}, serviceNames...))
+		args := []string{"services", "stop"}
+		args = append(args, serviceNames...)
+		if allProjects {
+			args = append(args, "--all-projects")
+		}
```
savil, line 476: should both `--all-projects` _and_ specific serviceNames be allowed to be specified at the same time?

Should we make this:
```
if allProjects {
   args = append(args, "--all-projects")
   if len(serviceNames) > 0 {
      return errors.New("Cannot specify serviceNames alongwith --all-projects")
   }
} else if len(serviceNames) > 0 {
  args = append(args, serviceNames...)
} else {
   return errors.New("Expected either specific service names or --all-projects")
}
```
?

internal/services/manager.go
```diff
@@ -6,12 +6,109 @@ import (
...
+	// write json to file
+	path, err := globalProcessComposeConfigPath()
+	if err != nil {
+		return fmt.Errorf("failed to get config path: %w", err)
+	}
+
+	path = filepath.Join(path, "process-compose.json")
+
+	if err = os.WriteFile(path, json, 0666); err != nil {
+		return fmt.Errorf("failed to open process-compose log file: %w", err)
```
savil, line 94: the error message seems wrongly copy-pasted :)

internal/services/manager.go
```diff
@@ -49,46 +153,78 @@ func StartProcessManager(
...
+	globalProcessComposeConfig := readGlobalProcessComposeConfig()
+
+	projectConfig := projectConfig{
+		Pid:  cmd.Process.Pid,
+		Port: port,
+	}
+
+	globalProcessComposeConfig.Instances[runID] = projectConfig
+
+	if err := os.WriteFile(processComposeUIDFile, []byte(runID), 0666); err != nil {
```
savil, line 179: I'm not sure this is the best idea. Reasons:

1. The `.devbox` folder's semantics are currently that it can be blown away, and the devbox CLI will just re-create it, and one can continue.
2. This `.devbox/compose.uid` file changes those semantics. And I don't think its worth it for the goal of "move resiliency".

Suggestion:
Could we instead derive the ID as `hash(projectDir)`? This will not be resilient to moves, but I think we can document that services need to be stopped prior to relocating projects.

Not very opinionated about the `hash(projectDir)` implementation, but as a suggestion it could be like:
```
		h := fnv.New32a()
 		_, err := h.Write([]byte(projectDir))
 		if err != nil {
 		}
 		hashed := strconv.FormatUint(uint64(h.Sum32()), 10)
```
from: `plugin/files.go` in https://github.com/jetpack-io/devbox/pull/874/files

savil (commented):
Overall, I like the approach! 

but I'm afraid I have strong reservations about the UUID. Please see my comment below. What do you think?

internal/impl/devbox.go
mikeland86: Should we just get the full command and pass it in instead? 

I think we can do:

`d.RunScript("devbox", cmd.CommandPath())`

mikeland86: I think `--all-projects` should only work without passing specific service, but we should check that in the boxcli layer.

Lagoja: Agreed that if we're going to forbid using --all-projects with service names, we should error on boxcli. I'll implement it that way

internal/services/manager.go
Lagoja: Updated this: 

1. I switched to using projectDir as the key. In addition to being simpler, its more resilient to situations where process-compose crashes/fails and without cleaning up the config
2. I kept everything in one file. Splitting to multiple files meant I needed code to get all the files in a directory, parse them, etc. Seemed messy
3. Since everything is in one file, I added file locking/unlocking to avoid situations where 2 projects try to write/read the file at the same time. I’m using a `flock` syscall to lock the file between processes, but I’m keeping this best effort and not throwing an error if the lock fails

savil (commented):
looking great so far! I have to pause for morning-routine, and will resume review once back.

internal/services/manager.go
```diff
@@ -6,35 +6,181 @@ import (
...
+
+func removeInstance(projectDir string) error {
+	config, err := getGlobalConfig()
+	if err != nil {
+		return err
+	}
+
+	lockFile(config.File.Fd())
+	defer config.File.Close()
+	defer unlockFile(config.File.Fd())
```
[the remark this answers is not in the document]
savil: >  -u, --unlock
           Drop a lock. This is usually not required, since a lock is
           automatically dropped when the file is closed.
source: https://man7.org/linux/man-pages/man1/flock.1.html

So, this makes me think that you just want `defer config.File.Close()`.

internal/services/manager.go
```diff
@@ -6,35 +6,181 @@ import (
...
+}
+
+func addInstance(projectDir string, projectConfig instance) error {
+	config, err := getGlobalConfig()
+	if err != nil {
+		return err
+	}
+
+	lockFile(config.File.Fd())
+	defer unlockFile(config.File.Fd())
```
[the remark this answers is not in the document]
savil: but also see comment below: I think you just want `defer config.File.Close()`.

internal/services/manager.go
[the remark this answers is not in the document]
savil: I'd also suggest checking the error return values of `config.File.Close()` and doing:

```
defer func() {
     err := config.File.Close()
     if err  != nil {
         // prefer writer since we manage the --quiet flag using that, but if its hard to get a writer here
         // for now, you could do io.Stderr
          fmt.Fprintf(writer /* or io.Stderr */, "failed to close file (%s). Got error %v\n", someFilePath, err)
     } 

}
```

This way we may get alerted to any errors from conflicts, etc. with our file-locking-and-writing code.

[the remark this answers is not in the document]
savil: also, maybe move the `defer file.config.Close()` to be after the `getGlobalConfig's error check`?

internal/services/manager.go
```diff
@@ -122,23 +287,108 @@ func StopProcessManager(
...
 	}
 
+	config := readGlobalProcessComposeJSON(configPath)
+
+	project, ok := config.Instances[projectDir]
+	if !ok {
+		return false
+	}
+
+	process, _ := os.FindProcess(project.Pid)
```
savil, line 341: why skip the error check?

internal/services/manager.go
```diff
@@ -122,23 +287,108 @@ func StopProcessManager(
...
+	path, err := globalProcessComposeJSONPath()
+	if err != nil {
+		return 0, fmt.Errorf("process-compose is not running or it's config is missing. To start it, run `devbox services up`")
+	}
+
+	config := readGlobalProcessComposeJSON(path)
+
+	project, ok := config.Instances[projectDir]
+	if !ok {
+		return 0, fmt.Errorf("process-compose is not running or it's config is missing. To start it, run `devbox services up`")
```
savil, line 364: would it be better for debugging to have a more specific error message like:
`failed to find projectDir %s in config.Instances`?

If that is not a great user error message, then you can combine the two with:
```
usererr.WithUserMessage(
    fmt.Errorf("failed to find projectDir %s in config.Instances", projectDir),
    "process-compose is not running or it's config is missing. To start it, run `devbox services up`",
 )
```

internal/services/manager.go
```diff
@@ -122,23 +287,108 @@ func StopProcessManager(
...
+	lock := syscall.Flock_t{
+		Type:   syscall.F_WRLCK,
+		Whence: int16(io.SeekStart),
+		Start:  0,
+		Len:    0,
+	}
+
+	if err := syscall.FcntlFlock(fd, syscall.F_SETLK, &lock); err != nil {
+		fmt.Printf("Error acquiring lock: %s \n", err)
+	}
```
savil, line 381: why not return the error message and fail if we don't get the lock?

Continuing to write could lead to some weird errors because of clobbering writes.

savil (commented):
Nice! Left a few more comments.

One main concern is: should we fail if we get an error with locking or unlocking? I was leaning towards "yes" to avoid weird issues with multiple processing clobbering the shared file.

internal/services/manager.go
```diff
@@ -122,23 +287,108 @@ func StopProcessManager(
...
+	return project.Port, nil
+}
+
+func lockFile(fd uintptr) {
+
+	lock := syscall.Flock_t{
+		Type:   syscall.F_WRLCK,
+		Whence: int16(io.SeekStart),
+		Start:  0,
+		Len:    0,
```
savil, line 376: can we set a Timeout of N Seconds?

otherwise, flock seems to wait indefinitely.

internal/services/manager.go
```diff
@@ -122,23 +287,108 @@ func StopProcessManager(
...
+		fmt.Printf("Error acquiring lock: %s \n", err)
+	}
+}
+
+func unlockFile(fd uintptr) {
+	lock := syscall.Flock_t{
+		Type:   syscall.F_UNLCK,
+		Whence: int16(io.SeekStart),
+		Start:  0,
+		Len:    0,
```
savil, line 389: set a Timeout here too?

internal/boxcli/services.go
```diff
@@ -59,9 +71,10 @@ func servicesCmd() *cobra.Command {
 
 	stopCommand := &cobra.Command{
 		Use:   "stop [service]...",
-		Short: "Stop service. If no service is specified, stops all services",
+		Short: "Stop one or more services in the current project. If no service is specified, stops all services in the current project.",
+		Long:  `Stop one or more services in the current project. If no service is specified, stops all services in the current project. \nIf the --all-projects flag is specified, stops all running services across all your projects. This flag cannot be used with [service] arguments.`,
```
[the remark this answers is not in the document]
Lagoja: Pluralized 43

internal/services/manager.go
Lagoja: >  On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.

internal/services/manager.go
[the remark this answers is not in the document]
Lagoja: Ah that makes sense, if I can unlock the file by closing it, then I can just defer that and only call `unlock` specifically when I need it

internal/services/manager.go
Lagoja: We now return an error from the lock

internal/services/manager.go
Lagoja: I can't pass a timeout to Flock directly, but I added a timeout goroutine that will close the file if a lock is not obtained within 5 seconds

internal/services/manager.go
```diff
@@ -45,74 +191,93 @@ func StartProcessManager(
...
+	}
+
+	err := addInstance(projectDir, projectConfig)
+	if err != nil {
+		return fmt.Errorf("failed to add instance to global config: %w", err)
 	}
 
-	defer os.Remove(processComposePidfile)
-	return cmd.Wait()
+	err = cmd.Wait()
```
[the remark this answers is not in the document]
Lagoja: I don't think the command really outputs anything, since this starts the TUI.

internal/services/manager.go
```diff
@@ -288,24 +283,16 @@ func StopProcessManager(
...
 
-	// Open the config file, defer closing to the end of the scope
-	globalConfigFile, err := os.OpenFile(configPath, os.O_RDWR|os.O_CREATE, 0666)
-	if err != nil {
-		return fmt.Errorf("failed to open config file: %w", err)
-	}
-	defer globalConfigFile.Close()
+	defer configFile.Close()
 
+	config := readGlobalProcessComposeJSON(configFile)
```
savil, line 300: comment on line 294 can be moved up or removed

internal/services/manager.go
```diff
@@ -342,53 +331,48 @@ func ProcessManagerIsRunning(projectDir string) bool {
...
-		fmt.Printf("Error unlocking file: %s\n", err)
+	select {
+	case err := <-lockResult:
+		if err != nil {
+			file.Close()
+			return fmt.Errorf("failed to lock file: %w", err)
+		}
+		return nil
+
+	case <-time.After(fileLockTimeout):
```
savil, line 381: ah, this is a nice way to do a timeout

internal/services/manager.go
```diff
@@ -52,96 +53,62 @@ type globalProcessComposeConfig struct {
...
 
-	return nil
-}
-
-func removeInstance(projectDir string) error {
-	config, err := getGlobalConfig()
+	globalConfigFile, err := os.OpenFile(configPath, os.O_WRONLY|os.O_CREATE, 0664)
 	if err != nil {
-		return err
+		return globalConfigFile, fmt.Errorf("failed to open config file: %w", err)
```
savil, line 104: nit: I'd return `nil` in these error cases.

internal/services/manager.go
```diff
@@ -194,14 +170,14 @@ func StartProcessManager(
 	if processComposeBackground {
 		flags = append(flags, "-t=false")
 		cmd := exec.Command(processComposeBinPath, flags...)
-		return runProcessManagerInBackground(cmd, port, config, projectDir)
+		return runProcessManagerInBackground(cmd, config, port, projectDir)
 	}
 
 	cmd := exec.Command(processComposeBinPath, flags...)
-	return runProcessManagerInForeground(cmd, port, projectDir, w)
+	return runProcessManagerInForeground(cmd, config, port, projectDir, w)
```
savil, line 184: this feels a bit problematic:
1. above we are doing `defer configFile.Close()`
2. Inside `runProcessManagerInForeground`, we are closing the file, waiting for the command, and re-opening a file. 

I worry that since the step (1) defer runs after step (2) it will error because it is closing a file that's already closed.

Oh, but we are implicitly discarding the error from `file.Close()` in the `defer` block, so all good I think!

savil (approved):
great!

internal/services/manager.go
Lagoja: I guess another option is to defer the entire cleanup block after `cmd.Wait` call + error check in `runProcessManagerInForeground` to ensure that it runs -- but I agree that discarding the error from `defer configFile.Close()` avoids the issue as well.

PR #876, opened 2023-04-07
[bin-wrappers] Use shellenv command in binwrappers

mikeland86:
## Summary

This changes bin wrappers to use `eval $(devbox shellenv --use-cached-print-dev-env)`

### Motivation:
For performance reasons we were printing out the shellenv results in the bin wrappers. The problem with this approach is that the env vars in the wrappers overwrite whatever the user sets. For example:

`USER=foo node -e "console.log(process.env.USER);"`

will print `mike` because `export USER="mike"` is in the wrapper.

### Solution:

* Revert to using `eval $(devbox shellenv)` in the wrapper.
* For performance, add hash to environment and skip evaluation if the hash matches (this means shellenv has already been eval-ed).
* For performance add new `--use-cached-print-dev-env` flag to shellenv that uses a cached copy of `nix print-dev-env` if it exists. The cache is regenerated every time `computeNixEnv` is called so it happens on shell, run, adding/removing package, when wrappers are created, and also when shellenv is called without the flag.
* For performance added DO_NOT_TRACK to shellenv call in wrapper. This can be removed once segment moves to queuing instead of blocking.

Latency impact when hash matches: None, or maybe a bit < 1ms faster.
Latency impact when hash doesn't match: ~20ms slower

Potential additional optimization 1: skip launcher when calling `devbox shellenv`. This should reduce about ~10ms of latency.
Potential additional optimization 2: when hash matches, instead of saving a wrapper we could save a symlink. This would eliminate extra latency of wrapper.

## How was it tested?

* Tested by adding and removing packages while in shell.
* Timed calls to `hello` 
* Existing unit tests that test shellenv, bin wrappers, etc


mikeland86:
@ipince @gcurtis PTAL when you get a chance

internal/impl/devbox.go
```diff
@@ -981,3 +1028,14 @@ func (d *Devbox) NixBins(ctx context.Context) ([]string, error) {
...
+func addHashToEnv(env map[string]string) map[string]string {
+	data := []byte{}
+	for k, v := range env {
+		data = append(data, []byte(k)...)
+		data = append(data, []byte(v)...)
+	}
+	hash := md5.Sum(data)
+	env[devboxShellEnvHashVarName] = hex.EncodeToString(hash[:])
+	return env
+}
```
gcurtis, line 1041: `data` can be removed by hashing as you go:

```suggestion
func addHashToEnv(env map[string]string) map[string]string {
	h := md5.New()
	for k, v := range env {
		h.Write([]byte(k))
		h.Write([]byte(v))
	}
	env[devboxShellEnvHashVarName] = hex.EncodeToString(md5.Sum(data))
	return env
}
```

If you want, the return value can also be removed since the function is modifying the map directly.

internal/nix/nix.go
```diff
@@ -84,31 +84,72 @@ func parseInfo(pkg string, data []byte) *Info {
...
+	if args.UsePrintDevEnvCache {
+		if _, err := os.Stat(args.PrintDevEnvCachePath); err == nil {
+			data, err = os.ReadFile(args.PrintDevEnvCachePath)
+			if err != nil {
+				return nil, errors.WithStack(err)
+			}
+			if err := json.Unmarshal(data, &out); err != nil {
+				return nil, errors.WithStack(err)
+			}
+		}
```
gcurtis, line 120: You can remove the `os.Stat` call and just trying reading:

```suggestion
		cachedData, err := os.ReadFile(args.PrintDevEnvCachePath)
		if err == nil {
			if err := json.Unmarshal(data, &out); err != nil {
				return nil, errors.WithStack(err)
			}
			cachedData = data
		} else if !errors.Is(err, fs.ErrNotExist) {
			return nil, errors.WithStack(err)
		}
```

gcurtis (approved):

PR #883, opened 2023-04-10
[tests] Tests should not modify config

mikeland86:
## Summary

This allows us to run tests without modifying example devbox.json configs.

## How was it tested?

`go test ./...`


internal/impl/config.go
```diff
@@ -98,6 +99,11 @@ func readConfigFromURL(url *url.URL) (*Config, error) {
 }
 
 func upgradeConfig(cfg *Config, absFilePath string) error {
+	if dontUpgrade, _ := strconv.ParseBool(
+		os.Getenv("DEVBOX_DONT_UPGRADE_CONFIG"),
+	); dontUpgrade {
+		return nil
+	}
```
savil, line 106: I'm surprised this works. Does it actually use our desired defaultNixPkgsCommit? or just `nixpkgs` and "seem to work" because tests happen to pass?

 I would have thought you want to just gate the `writeConfig` below, but continue to write:
```
cfg.Nixpkgs.Commit = plansdk.DefaultNixpkgsCommit
```
so that the code works as we want by using the DefaultNixpkgsCommit

savil (approved):
approving but please see inline and lets try to verify its using the correct nixpkgs commit hash

internal/impl/config.go
mikeland86: The code correctly defaults to `DefaultNixpkgsCommit` if none is set.

Gating `writeConfig` felt wrong because what is specifically causing the writes is the upgrade. (This may change if we decide to add packages dynamically, but in that case gating writeConfig will break that).

I actually was thinking of adding something to the test that confirms that `devbox.json` has not changed (since it should not)

PR #884, opened 2023-04-10
nixstore: add package for querying and resolving Nix store paths

gcurtis:
Given the store name of a Nix package, a `nixstore.Root` can return a slice of transitive package dependencies in topological order without invoking a separate Nix process. This builds on top of the code from c009546.

Once we have a search index of attribute paths to store paths, this will allow Devbox to bypass nixpkgs evaluation by resolving and installing packages directly.

Follow-up improvements:

- Cache resolved package dependencies so that they don't need to be recomputed every time.
- Support for resolving dependencies using .narinfo files.


mikeland86 (approved):
Nice!

I'm not sure I'll be much help reviewing the actual implementation (without spending additional time to learn a lot of it myself), but the idea and structure look good.

I think if there are a few low hanging possible unit tests or integration tests that would be pretty nice (e.g. install some stuff with devbox and then use this to retrieve a package and dependencies)

PR #887, opened 2023-04-11
[vscode-extension] Added better error messages and debugs for openinvscode

mohsenari:
## Summary
Added better error messages and debugs for vscode extension.

## How was it tested?
run and debug when open vscode-extension directory in vscode


LucilleH (approved):
TY!

PR #889, opened 2023-04-11
chore: get config path from --config only

hezhizhen:
## Summary

Getting config path from arguments is deprecated, so I remove related code and let config path be fetched from --config flag only.


savil (approved):
thanks!

PR #892, opened 2023-04-11
nixstore: add support for remote stores

gcurtis:
Add a `Remote` function to access a remote store as opposed to a local one. For example:

	store, err := nixstore.Remote("https://cache.nixos.org")

creates a store that queries packages in the Nix binary cache. It accesses packages using the same `S3BucketFS` implemented in `nixed`. Resolving dependencies is done by looking at the `.narinfo` file for a package.


internal/nix/nixstore/fs.go
gcurtis, line 1: All of the additions to this file were taken directly from the `nixed` source.

mikeland86 (approved):
Accepting as experimental

PR #896, opened 2023-04-12
nixstore: implement package install

gcurtis:
Depends on #892.

---

This allows a `nixstore.Root` to install packages to the store without calling Nix directly:

```go
remoteStore, _ := Remote("https://cache.nixos.org")
remotePkg, _ := remoteStore.Package("b1kk0rp0yw1742rd88ql4379c2cmcqh2-zig-0.10.1")

localStore, _ := Local("/nix/store")
localStore.Install(remotePkg)
```

```sh
$ /nix/store/b1kk0rp0yw1742rd88ql4379c2cmcqh2-zig-0.10.1/bin/zig build-exe hello.zig
$ ./hello
Hello, world!
```

This has a caveat that the installed package isn't registered with the Nix database. This will cause Nix to think the path is invalid. This can be fixed by manually registering the path:

	$ printf "/nix/store/b1kk0rp0yw1742rd88ql4379c2cmcqh2-zig-0.10.1\n\n0\n" | sudo nix-store --register-validity

This implies that Devbox would need two modes to run with/without Nix:

1. With a Nix install - Devbox puts the packages in a temp directory and then runs `nix store add-path` instead of putting the package in the store directly.
2. Without a Nix install - just assume ownership of /nix/store

### DIFF
--- a/internal/nix/nixstore/nixstore.go
+++ b/internal/nix/nixstore/nixstore.go
@@ -4,17 +4,29 @@ package nixstore
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
+	"os"
 	"path"
+	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/nix-community/go-nix/pkg/nar"
 	"github.com/nix-community/go-nix/pkg/narinfo"
+	"github.com/ulikunitz/xz"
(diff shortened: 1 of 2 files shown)


mikeland86 (approved):
Accepting as experimental

PR #897, opened 2023-04-12
[lockfile] Add hidden lockfile for caching

mikeland86:
## Summary

This adds a hidden lock file for caching purposes. 

### Motivation:
This speeds up run/shell by only running `ensurePackagesAreInstalled` if config or nix profile manifest has changed.

### Why hidden?
I'm not ready for this to be an official checked-in lock file. So for now it lives in `.devbox` and does not get committed to repo.

## How was it tested?

In flakes example ran:

```bash
devbox run echo 5
```

and got a speed up of around 66%.

Also tested

```bash
cat .devbox/devbox.lock
devbox add hello
cat .devbox/devbox.lock  # does not match original
devbox run hello #success
devbox rm hello
cat .devbox/devbox.lock  # matches original
devbox run hello # error
```


gcurtis (approved):
I'm excited to have faster startup times!

mikeland86:
@gcurtis I'm going to rename this to possible `cache.json` since we can't have manifest hash in lock file.

mikeland86:
> Both names work for me since this is a hidden file. Why doesn't the manifest hash work? I think it should work for the local machine, so as long as it isn't committed it should be okay.

Yeah for non-committed files manifest is OK. I was just thinking we would eventually move this out of hidden.

PR #900, opened 2023-04-12
nixstore: search packages by attribute path

gcurtis:
Depends on #896.

---

This is a hack to get a rudimentary package search (by attribute path) working natively in Devbox. For example, to install Go 1.19:

```go
remoteStore, _ := Remote("https://cache.nixos.org")
remotePkg, _ := remoteStore.PackageAttrPath("go_1_19")

localStore, _ := Local("/nix/store")
localStore.Install(remotePkg)
```

The attribute path lookup relies on embedding a gzipped JSON index of all nixpkgs at commit 3fb8eedc450. This file adds 4.7 MiB to the size of the Devbox binary. Ideally this index would be downloaded as-needed instead.

### DIFF
--- a/internal/nix/nixstore/indexpkgs.nix
+++ b/internal/nix/nixstore/indexpkgs.nix
@@ -105,6 +105,14 @@ let
     */
     paths = [ ];
 
+    /* The full store path of the package.
+
+       Examples:
+         /nix/store/mzd0lvp4lqs5pkds7vihn68hspq50859-go-1.20.1
+         /nix/store/b1kk0rp0yw1742rd88ql4379c2cmcqh2-zig-0.10.1
+    */
+    out = unsafeDiscardStringContext drv.outPath;
+
     # The remaining attributes are all optional.
     ${if drv ? meta.mainProgram then "program" else null} = drv.meta.mainProgram;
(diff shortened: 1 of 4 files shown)


gcurtis:
If anyone things the embedded file is too large, I can hold off on this and get a better implementation working by downloading it at runtime.

mikeland86:
Not necessarily opposed to embedding, but generally prefer downloading. Is this part of a general strategy for search? or is it tangential?

loreto:
I think downloading (or API) is the way to go. But in the interest of Always Be Shipping, I think it's totally cool to embed as a first pass, and get the search feature out. 4.7 MiB feels reasonable to me, so I'd be inclined to ship this!

gcurtis:
> Not necessarily opposed to embedding, but generally prefer downloading. Is this part of a general strategy for search? or is it tangential?

It's more of a first step to get something working. I don't think this will (or should) be the final implementation.

mikeland86 (changes requested):
Should we shelve this in favor of server side search for now?

I think this is a good idea, but would prefer lazy download to avoid the increase in size but I don't think it's a priority

loreto:
Is this PR abandoned? (should we close it in that case?)

PR #904, opened 2023-04-13
[examples tests] insert one second sleep for services to initialize

savil:
## Summary

The lepp-stack test was sporadically failing with unable to find the working postgres socket.

I think but not 100% this could be because the `devbox services start` completes but postgres
may take a short while to fully initialize before it is ready. So, here I inject
a one second sleep to hopefully let postgres be ready.

## How was it tested?

`DEVBOX_EXAMPLE_TESTS=1 DEVBOX_DEBUG=0 go test -v -count 1 -run TestExamples/stacks_l ./testscripts/...`

### DIFF
--- a/examples/stacks/lapp-stack/devbox.json
+++ b/examples/stacks/lapp-stack/devbox.json
@@ -18,8 +18,9 @@
                 "psql devbox_lapp < setup_postgres_db.sql"
             ],
             "run_test": [
-                "initdb",
+                "devbox run init_db",
                 "devbox services start",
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #904** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/904" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/904?utm_source=stack-comment).

mikeland86 (approved):

PR #905, opened 2023-04-13
[cicd] bump timeouts for cli-tests

savil:
## Summary

The test-step `cli-test` timeout needs to be bigger than the `go test --timeout` value
since we run other steps prior to it.

Also bumped the go test timeout, just to be safe.

Obviously, this is not a long term solution. We need a way of caching the nix packages,
but in the meantime prefer to have the Github Actions be Green.

## How was it tested?

buildkite run should be green

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -92,7 +92,7 @@ jobs:
           - is-main: true
             run-example-tests: false
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #905** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/905" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/905?utm_source=stack-comment).

mikeland86 (approved):

PR #906, opened 2023-04-13
[CICD] stream go test output, and turn off DEVBOX_DEBUG for example tests

savil:
## Summary

The current test output when the macos tests timeout is not helpful. We don't know
if any of the specific example tests is taking longer:
https://github.com/jetpack-io/devbox/actions/runs/4685049031/jobs/8301872973

In this PR:
1. set `go test -v`. This should stream the output as the tests run. It also has 
the benefit of printing a summary of which testscripts passed versus failed (tho 
not useful for timeouts). I use this always locally when running these example testscripts.

2. turn off DEVBOX_DEBUG for example testscripts. The devbox-internal logging is
less useful for examples testscripts, and makes it hard to follow where in the examples's `run_test` the script
is failing or getting stuck on. 

DEVBOX_DEBUG _is_ useful for other unit tests so left it on for those.

## How was it tested?

buildkite run, will inspect.

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -23,6 +23,9 @@ on:
         type: boolean
       run-mac-tests:
         type: boolean
+      # run the example tests with DEVBOX_DEBUG=1
+      example-debug:
+        type: boolean
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #906** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/906" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/906?utm_source=stack-comment).

gcurtis (approved):

PR #907, opened 2023-04-13
[CICD] Don't run example tests on mac

mikeland86:
## Summary

Until we figure out a caching strategy we should disable example tests on macs.

## How was it tested?

CICD

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -94,6 +94,8 @@ jobs:
             os: "${{ inputs.run-mac-tests && 'dummy' || 'macos-latest' }}"
           - is-main: true
             run-example-tests: false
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-tests.yaml
```diff
@@ -94,6 +94,8 @@ jobs:
             os: "${{ inputs.run-mac-tests && 'dummy' || 'macos-latest' }}"
           - is-main: true
             run-example-tests: false
+          - run-example-tests: true
+            os: macos-latest
```
savil, line 98: can we make this line: `os: "${{ inputs.run-mac-tests && 'dummy' || 'macos-latest' }}"` ?

that way, we can kick off manual workflows that do run macos tests

savil (approved):

PR #910, opened 2023-04-14
[direnv] run init-hooks in envrc, and generate shell files prior to gen envrc

savil:
## Summary

Two changes:

1. Run `shellenv --init-hook` to ensure correctness. There will be a small perf hit if the init-hooks are very slow, but we are biasing towards correctness over perf.
  - To improve the perf, we can distinguish a generic init hook versus shell-specific init hook, later on.

2. Ensure packages are installed and shell files are generated. This is important because we check for existence of `.devbox/gen/flake/flake.nix` prior to running `devbox shellenv --init-hook`.
  - since this requires `nix`, we add `ensureNixInstalled` to `devbox init` and `devbox generate direnv` commands.
  - I refactored to call `ensureNixInstalled` from within `GenerateEnvrc` so that we don't block `devbox init` if envrc is not required.

## How was it tested?

```
# clear .devbox directory and nix profile contained therein
> rm -rf .devbox

# generate a new .envrc
> devbox generate direnv --force
# observe output ensuring packages are installed

# go outside the directory and check my global `go` path
> cd ..
> which go

# confirm `go` path is now from the /nix/store
> cd devbox
> which go
```

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -371,6 +371,14 @@ func (d *Devbox) GenerateDockerfile(force bool) error {
 
 // GenerateEnvrc generates a .envrc file that makes direnv integration convenient
 func (d *Devbox) GenerateEnvrc(force bool, source string) error {
+	ctx, task := trace.NewTask(context.Background(), "devboxGenerateEnvrc")
+	defer task.End()
+
+	// generate all shell files to ensure we can refer to them in the .envrc script
+	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
+		return err
+	}
+
 	envrcfilePath := filepath.Join(d.projectDir, ".envrc")
 	filesExist := fileutil.Exists(envrcfilePath)
 	if !force && filesExist {
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -374,11 +374,6 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
 	ctx, task := trace.NewTask(context.Background(), "devboxGenerateEnvrc")
 	defer task.End()
 
-	// generate all shell files to ensure we can refer to them in the .envrc script
-	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
-		return err
-	}
-
 	envrcfilePath := filepath.Join(d.projectDir, ".envrc")
 	filesExist := fileutil.Exists(envrcfilePath)
 	if !force && filesExist {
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #910** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/910" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/910?utm_source=stack-comment).

internal/nix/install.go
```diff
@@ -88,3 +89,43 @@ func DirExists() bool {
 func isRoot() bool {
 	return os.Geteuid() == 0
 }
+
+func EnsureNixInstalled(writer io.Writer, daemon *bool) error {
```
savil, line 93: a straight move, with the only change to line 120 `Install(writer, daemon)`. In the original code this was `cmd.ErrOrStderr, daemonFlag(..)` which is passed in from the caller.

internal/impl/devbox.go
```diff
@@ -394,20 +397,27 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
 
-		if strings.ToLower(result) == "y" || !isInteractiveMode {
+		if strings.ToLower(result) == "y" || !isInteractiveMode || source == "generate" {
+			nixDaemon := false
+			if err := nix.EnsureNixInstalled(d.writer, &nixDaemon); err != nil {
+				return err
+			}
+
+			// generate all shell files to ensure we can refer to them in the .envrc script
+			if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
```
ipince, line 407: this is basically a separate bugfix, right?  before this, we could've generated `.envrc` files that referred to non-existent `.nix` files. No we don't 👍

internal/impl/devbox.go
```diff
@@ -394,20 +397,27 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
+			}
+
 			// .envrc file creation
 			err := generate.CreateEnvrc(tmplFS, d.projectDir)
 			if err != nil {
 				return errors.WithStack(err)
 			}
+		}
+
+		if strings.ToLower(result) == "y" || !isInteractiveMode {
```
ipince, line 418: I'm confused--when does this happen and the previous if-statement does not? if source != "generate".. what's the other possible value of source? when do we want to run `direnv allow` but _not_ generate `.envrc`?

ipince (approved):

internal/impl/devbox.go
savil: yup

The envrc has `if [ -f .devbox/gen/flake/flake.nix]; then` so it wouldn't have errored but would've been a no-op. 

Now, we know it'll actually run the `devbox shellenv --init-hook`.

internal/impl/devbox.go
savil: if source is "generate", then we don't run `direnv allow` for the user. They need to run it themselves.

The other source is `devbox init`.

I don't know why we distinguish these two cases for `direnv allow`. Maybe @mohsenari knows?

I just refactored the existing logic so that the nix-and-ensurePackagesInstalled prequisite can run before generate.CreateEnvrc. I didn't intend to change this behavior.

PR #911, opened 2023-04-16
Reduce cognitive complexity

hezhizhen:
## Summary

1. Reduce cognitive complexity to 30
2. remove unnecessary nolint comments
3. Correct some comments
4. Remove unused `IsSymlink()`

## How was it tested?

Build and test.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -434,21 +434,21 @@ func (d *Devbox) saveCfg() error {
 }
 
 func (d *Devbox) Services() (services.Services, error) {
-	result := services.Services{}
 	pluginSvcs, err := plugin.GetServices(d.mergedPackages(), d.projectDir)
 	if err != nil {
-		return result, err
+		return nil, err
 	}
 
 	userSvcs := services.FromProcessComposeYaml(d.projectDir)
 
-	svcSet, err := lo.Assign(pluginSvcs, userSvcs), nil
-
+	svcSet := lo.Assign(pluginSvcs, userSvcs)
 	keys := make([]string, 0, len(svcSet))
 	for k := range svcSet {
 		keys = append(keys, k)
 	}
 	slices.Sort(keys)
+
+	result := services.Services{}
 	for _, k := range keys {
 		result[k] = svcSet[k]
 	}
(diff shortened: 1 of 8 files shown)


internal/impl/devbox.go
```diff
@@ -434,21 +434,21 @@ func (d *Devbox) saveCfg() error {
 }
 
 func (d *Devbox) Services() (services.Services, error) {
-	result := services.Services{}
 	pluginSvcs, err := plugin.GetServices(d.mergedPackages(), d.projectDir)
 	if err != nil {
-		return result, err
+		return nil, err
```
mohsenari, line 439: when `Services()` is referenced in `StartServices()` or `StopServices()`, etc. we check the output by doing 
```go
if len(svcSet) == 0 {
...
}
```
So, either the if statements need to change to `if svcSet == nil` or we need to keep the return an empty map. Because `len(nil)` throws an error saying: `invalid argument: nil for len`

mohsenari (changes requested):
@hezhizhen thanks a ton for doing this work. Just left a comment that needs a bit of change otherwise it'll result in unexpected error. Otherwise, it looks good 👍

internal/impl/devbox.go
hezhizhen: 1. The map returns `nil` when the error is not nil, so the usage should check the error first (and they do actually).
2. See the example https://go.dev/play/p/_HnoYvVdrfD . When we check the length of the a nil map, it returns 0 normally instead of an error.

So, I think there's no need to change `nil` to `Services{}`. What do you think?

hezhizhen: For more information, if we run `len(svcSet)` when `svcSet` is nil, we are not running `len(nil)`. We are running `len(svcSet)` when `svcSet`'s type is `map[string]Service` and its value is `nil`. It's different from a literal `nil`.

So, `len(nil)` throws an error `invalid argument: nil for len` but `len(svcSet)` returns 0 when `svcSet` is a nil `map[string]Services`.

mohsenari: I see. That makes sense. I tested `len(nil)` and saw the error. But you're right, even if the function returns `nil` the type is still `map[string]Services` so it works.

mohsenari (approved):

PR #922, opened 2023-04-19
chore: move ssh-config generation out of general generating process and wrap its error

hezhizhen:
## Summary

`ssh-config` is quite different from other sub-commands, so I move its part out of `runGenerateCmd`. I also wrap its error with `errors.WithStack`.

### DIFF
--- a/internal/boxcli/generate.go
+++ b/internal/boxcli/generate.go
@@ -71,8 +71,9 @@ func dockerfileCmd() *cobra.Command {
 	command := &cobra.Command{
 		Use:   "dockerfile",
 		Short: "Generate a Dockerfile that replicates devbox shell",
-		Long:  "Generate a Dockerfile that replicates devbox shell. Can be used to run devbox shell environment in an OCI container.",
-		Args:  cobra.MaximumNArgs(0),
+		Long: "Generate a Dockerfile that replicates devbox shell. " +
+			"Can be used to run devbox shell environment in an OCI container.",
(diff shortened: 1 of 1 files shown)


hezhizhen:
`Could not connect to azure.archive.ubuntu.com:80 (52.252.75.106), connection timed out`.

Could anyone re-run the tests?

internal/boxcli/generate.go
```diff
@@ -110,7 +111,9 @@ func sshConfigCmd() *cobra.Command {
 		Long:   "Check ssh config and if they don't exist, it generates the configs necessary to connect to devbox cloud VMs.",
 		Args:   cobra.MaximumNArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runGenerateCmd(cmd, flags)
+			// ssh-config command is exception and it should run without a config file present
+			_, err := cloud.SSHSetup(flags.githubUsername)
+			return errors.WithStack(err)
```
mohsenari, line 116: @hezhizhen is this gonna work when there is no error? I mean what would be the output of this function if it returns `errors.WithStack(nil)`?

hezhizhen: It will return nil.

```go
// WithStack annotates err with a stack trace at the point WithStack was called.
// If err is nil, WithStack returns nil.
func WithStack(err error) error {
	if err == nil {
		return nil
	}
	return &withStack{
		err,
		callers(),
	}
}
```

mohsenari (approved):

PR #924, opened 2023-04-19
Update add method of devbox interface

hezhizhen:
## Summary

1. update comments of `Add` based on its implementation
2. tweak functions used in the implementation


mohsenari (approved):
Thanks for cleaning this up. I think initially, `devbox add` didn't install a package. Later it was changed to install but the comment was left over.

PR #926, opened 2023-04-19
[source] Fix direnv

mikeland86:
## Summary

On older versions we are missing quotes in `.envrc`, this makes sure it is compatible.

## How was it tested?

direnv


gcurtis (approved):
I'm wondering if we should just put a semicolon at the end of every line in case there's ever an env var that'll break things again.

mikeland86:
@gcurtis it won't hurt

PR #928, opened 2023-04-19
[direnv] change default hooks for shell --print-env

mohsenari:
## Summary
title says it

## How was it tested?
- `devbox shell --print-env`
- confirm output does not contain `. /path/to/.hooks.sh`


mikeland86 (approved):
Could you add a comment saying init hooks are not compatible with old `.envrc`

PR #929, opened 2023-04-19
impl: semicolon at the end of env string commands

gcurtis:
## Summary

This hardens the `devbox shellenv --init-hook` command so that it works without quotes.

## How was it tested?

Did the following for `zsh`, `bash`, and `fish`:

- `eval "$(direnv hook ...)"`
- `cd devbox`
- edit .envrc to not quote the eval
- `cd ..; cd devbox`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -285,7 +285,7 @@ func (d *Devbox) PrintEnv(ctx context.Context, useCache bool, includeHooks bool)
 
 	if includeHooks {
 		hooksStr := ". " + d.scriptPath(hooksFilename)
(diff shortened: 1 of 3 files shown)


mikeland86 (approved):

mohsenari (approved):

ipince (approved):

mohsenari:
@gcurtis This is breaking the tests more specifically, `impl/testdata/basic/shellrc.golden` because that is a reference file for generated shellrc but it doesn't have semicolons there. Please add the semicolons in that file before merging.

gcurtis:
Yeah, I saw that. Will fix after lunch.

PR #930, opened 2023-04-20
[examples] drop binutils from java examples

savil:
## Summary

I think these were added when @mohsenari was developing these examples to 
work inside a alpine Dockerfile. Usually, binutils is not needed and can be sourced
from the regular OS.

## How was it tested?
```
DEVBOX_EXAMPLE_TESTS=1 go test -v -run TestExamples/development_java  ./testscripts/...
```

### DIFF
--- a/examples/development/java/gradle/hello-world/devbox.json
+++ b/examples/development/java/gradle/hello-world/devbox.json
@@ -1,8 +1,7 @@
 {
     "packages": [
         "gradle",
-        "jdk",
-        "binutils"
+        "jdk"
     ],
     "shell": {
         "init_hook": null,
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #930** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/930" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/930?utm_source=stack-comment).

savil:
@Lagoja what criteria do we want for the examples-projects? Should they have packages like `binutils` that work in completely isolated environments, even if they may be excessive for most regular OS use-cases? I'd imagine the latter scenario is more common, and so submitted this PR.

mohsenari:
At the time I was building the java planner I added binutils because it complained in the Docker build stage if I remember correctly. But if the example works without it, I think it should be removed.

savil:
eh, I tested gradle locally but not maven, and sure enough maven seems to require binutils

PR #935, opened 2023-04-21
[devbox global] Added devbox global install command

mohsenari:
## Summary
Pretty much title says it.

This is using the same concept as `devbox global shellenv` except it doesn't output the computed environment.
Since we check and install packages in shellenv, we can reuse the same logic for devbox install.

- Also updated `devbox install` implementation to be the same as `devbox global install`

## How was it tested?
- compile
- edit the global devbox.json (`~/.local/share/devbox/global/default/devbox.json`) and add packages to it
- ./devbox global install

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -152,6 +152,7 @@ func installGlobalCmdFunc(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	fmt.Fprintln(cmd.ErrOrStderr(), "Finished installing packages.")
 	return nil
 }
 
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -22,6 +22,7 @@ func globalCmd() *cobra.Command {
 	}
 
 	cmd.AddCommand(globalAddCmd())
+	cmd.AddCommand(globalInstallCmd())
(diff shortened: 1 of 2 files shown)


savil (approved):
can we update `devbox install` to use a similar (corresponding) implementation?

internal/boxcli/global.go
```diff
@@ -128,6 +138,23 @@ func addGlobalCmdFunc(cmd *cobra.Command, args []string) error {
...
+	path, err := ensureGlobalConfig(cmd)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	box, err := devbox.Open(path, cmd.ErrOrStderr())
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	_, err = box.PrintEnv(cmd.Context(), false /* run init hooks */)
```
savil, line 151: wait, should this be `true` ?

savil: I think `false` is okay, because we want `install` to download the nix packages to `/nix/store` and update the global-nix-profile.

init-hooks are necessary when we are _also_ setting up the environment (which we are not here)

mohsenari:
> can we update `devbox install` to use a similar (corresponding) implementation?

@savil I just updated the PR to include devbox install as well.

PR #936, opened 2023-04-21
[lockfile] Add lockfile and use search endpoint to resolve versions

mikeland86:
## Summary

This introduces a simple lock file and uses the search service to resolve versions. The lock file is created/updated whenever a package is added/removed. If a package is manually added to devbox.json the lockfile is updated on run/shell/install/shellenv.

Packages that don't have `@` are never added to lockfile.

This is behind a feature flag.

## How was it tested?

```bash
devbox add python@3.9
devbox run "python --version"
devbox rm python@3.9
```

* Inspected nix profile to ensure packages are added and removed as expected.
* Inspected lock file to ensure contents are as expected.


mikeland86:
@loreto @gcurtis this is ready to review.

loreto (approved):
LGTM

loreto:
It looks like there's some conflicts you need to resolve

gcurtis (approved):

internal/nix/input.go
```diff
@@ -57,19 +71,28 @@ func (i *Input) Name() string {
...
-	u := i.URL // get copy
-	u.Fragment = ""
-	// This will produce urls with extra slashes after the scheme, but that's ok
-	return u.String()
+func (i *Input) URLForInput() string {
+	if i.IsDevboxPackage() {
+		resolved, err := i.lockfile.Resolve(i.String())
+		if err != nil {
+			panic(err)
+			// TODO(landau): handle error
```
savil, line 79: @mikeland86 gentle reminder to handle this panic's TODO (in case it slips through the cracks)

savil (commented):
catching up on this PR so I can review #964...

PR #941, opened 2023-04-22
chore: move all env to one package

hezhizhen:
## Summary

Create a new package `env` and move all env defined and used in all packages to it.


hezhizhen:
I'll create a following PR to add a new file `utils.go` in package `env` to provide utility functions for these envs.

internal/env/env.go
```diff
@@ -0,0 +1,37 @@
+package env
+
+const (
+	DevboxCache            = "DEVBOX_CACHE"
+	DevboxCLICloudShell    = "DEVBOX_CLI_CLOUD_SHELL"
+	DevboxDebug            = "DEVBOX_DEBUG"
+	DevboxFeaturePrefix    = "DEVBOX_FEATURE_"
+	DevboxGateway          = "DEVBOX_GATEWAY"
+	DevboxNotUpgradeConfig = "DEVBOX_DONT_UPGRADE_CONFIG"
```
savil, line 9: nit: can we call this `DevboxDoNotUpgradeConfig` ?

savil (approved):
cool

internal/env/env.go
savil: maybe send a small followup PR?

savil:
has some conflicts. Could you rebase and update, please?

internal/env/env.go
hezhizhen: Done. PTAL

hezhizhen:
> has some conflicts. Could you rebase and update, please?

Done. PTAL

internal/env/env.go
```diff
@@ -0,0 +1,37 @@
+package env
```
mikeland86, line 1: Ideally package names should not use words that we would frequently use as variables or functions (because then it forces us to use more contrived names for those). In addition, using made-up or composite words for packages makes them easy to identify.

The reason we use `envir` was precisely this.

internal/env/env.go
```diff
@@ -0,0 +1,37 @@
...
+
+	SSHTTY = "SSH_TTY"
+
+	StartWebTerminal = "START_WEB_TERMINAL"
+
+	XDGDataHome   = "XDG_DATA_HOME"
+	XDGConfigHome = "XDG_CONFIG_HOME"
+	XDGCacheHome  = "XDG_CACHE_HOME"
+	XDGStateHome  = "XDG_STATE_HOME"
+)
```
mikeland86, line 28: Many of these don't need to be exported, we should use camelCase for those so we can use nicer names for the exported functions (e.g. `DoNotTrack()`)

internal/env/env.go
savil: Good point about avoiding words frequently used as variables.

Personally, I'd prefer `environ` over `envir`, since `envir` is not a common substitute for `environment`, and harder to understand what it may stand for if seen in isolation.

hezhizhen: Okay, I'll create a PR to rename the package. Which name is better? `envir` or `environ`?

internal/env/env.go
hezhizhen: Agree. I'll create a PR to unexport some constants and use functions outside.

hezhizhen: See https://github.com/jetpack-io/devbox/pull/954

internal/env/env.go
savil: eh, lets go with `envir` so a bit less to type?

hezhizhen: See https://github.com/jetpack-io/devbox/pull/962

PR #944, opened 2023-04-22
[cleanup] replace checks for file existing with fileutil.Exists

hezhizhen:
## Summary

Title says it.

### DIFF
--- a/internal/planner/plansdk/analyzer.go
+++ b/internal/planner/plansdk/analyzer.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/bmatcuk/doublestar/v4"
+
 	"go.jetpack.io/devbox/internal/cuecfg"
 )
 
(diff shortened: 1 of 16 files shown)


internal/planner/plansdk/analyzer.go
```diff
@@ -57,7 +58,7 @@ func (a *Analyzer) GlobFiles(patterns ...string) []string {
 	return results
 }
 
-func (a *Analyzer) FileExists(relPath string) bool {
+func (a *Analyzer) FileExists(relPath string) bool { // TODO: remove it?
```
savil, line 61: yeah, lets replace it in callers with `fileutil.Exists(a.AbsPath(relPath))` ?

savil (approved):
Great, thanks!

internal/planner/plansdk/analyzer.go
hezhizhen: I mean deleting this method because it's not used anywhere.

PR #946, opened 2023-04-25
chore: move all env utils to env package

hezhizhen:
## Summary

1. Move all env related util functions to package `env`
2. Delete `cloud/envir`


hezhizhen:
@savil

internal/impl/config.go
```diff
@@ -106,7 +105,7 @@ func readConfigFromURL(url *url.URL) (*Config, error) {
 }
 
 func upgradeConfig(cfg *Config, absFilePath string) error {
-	if notUpgrade, _ := strconv.ParseBool(os.Getenv(env.DevboxDoNotUpgradeConfig)); notUpgrade {
+	if env.NotUpgradeConfig() {
```
savil, line 108: nit: `env.DoNotUpgradeConfig` reads more naturally

internal/env/util.go
```diff
@@ -0,0 +1,46 @@
...
+func IsDevboxCloud() bool {
+	return os.Getenv(DevboxRegion) != ""
+}
+
+func IsDevboxShellEnabled() bool {
+	inDevboxShell, _ := strconv.ParseBool(os.Getenv(DevboxShellEnabled))
+	return inDevboxShell
+}
+
+func NotTrack() bool {
```
savil, line 22: nit: `DoNotTrack` reads more naturally

savil (approved):

internal/env/util.go
hezhizhen: But the name is occupied by the env constant...

internal/impl/config.go
hezhizhen: Okay, I'll update it in the next PR

internal/env/util.go
hezhizhen: I see. I'll create a PR to do so.

internal/impl/config.go
hezhizhen: See https://github.com/jetpack-io/devbox/pull/954

internal/env/util.go
hezhizhen: See https://github.com/jetpack-io/devbox/pull/954

PR #947, opened 2023-04-25
[cleanup] remove unused functions

hezhizhen:
## Summary

Title says it.

### DIFF
--- a/internal/services/services.go
+++ b/internal/services/services.go
@@ -2,7 +2,6 @@
 package services
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


internal/services/services.go
```diff
@@ -35,85 +28,6 @@ const (
 	stopService
 )
 
-func listenToAutoPortForwardingChangesOnRemote(
```
savil, line 38: @mikeland86 any plans for this?

mikeland86: let's kill it!

This was meant to auto-forward ports between devbox cloud and your local machine but in practice it's brittle compared to our existing service proxy.

internal/services/services.go
```diff
@@ -35,85 +28,6 @@ const (
...
-					color.New(color.FgYellow).Fprintf(w, "Port forwarding %s:%s -> localhost stopped\n", hostname, service.Port)
-					cancel()
-				}
-				return service, false
-			},
-		},
-	)
-}
-
-func printProxyURL(w io.Writer, services Services) error { // TODO: remove it?
```
mikeland86, line 82: Looks like the callsite for this got accidentally removed in https://github.com/jetpack-io/devbox/pull/836

cc: @Lagoja 

This function used to be called to print the proxy URL that users can use to access their services. Without it users won't know how to access their services on the cloud.

internal/services/services.go
hezhizhen: Okay.

savil (approved):

PR #948, opened 2023-04-25
[cleanup] replace pkgslice with lo

hezhizhen:
## Summary

Package `pkgslice` is rarely used and `lo` was introduced before, so I replace functions from `pkgslice` with `lo`'s and remove this package.


savil:
@mikeland86 why did we have pkgslice?

mikeland86 (approved):
@savil I think whoever wrote the code didn't know we had lo or didn't know these functions existed.

gcurtis:
`pkgslice` was from before `lo` and before generics. I'm not a huge fan of `lo` but I agree it's better to not have multiple ways of operating on package slices.

Issue #949, opened 2023-04-25
[Bug]: /bin/bash does not exist


ragingpastry:
**Current Behavior (bug)**
While using devbox on NixOS I receive an error message whenever I try to run any wrapped commands inside of a devbox shell.

```sh
devbox init
devbox add python310
devbox shell
(devbox) ╭─crepe@polis /tmp/testing 
╰─$ python --version
zsh: /tmp/testing/.devbox/virtenv/.wrappers/bin/python: bad interpreter: /bin/bash: no such file or directory
```

**Expected Behavior (fix)**
I expect the devbox wrappers to inherit the location of my current $SHELL instead of hardcoding `/bin/bash`

I believe all we have to do is modify https://github.com/jetpack-io/devbox/blob/main/internal/wrapnix/wrapper.sh.tmpl#L1 to use an environment variable

**Additional context**

devbox.json
```json
{
  "packages": [
    "python310"
  ],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
  }
}
```

```sh
(devbox) ╭─crepe@polis /tmp/testing 
╰─$ devbox version -v
Version:     0.4.6
Platform:    linux_amd64
Commit:      none
Commit Time: unknown
Go Version:  go1.20.3
```

savil:
@ragingpastry thanks for reporting. Will fix!

ragingpastry:
Pushed up a MR for this. Feel free to close if its not going in the right direction! 

https://github.com/jetpack-io/devbox/pull/950

PR #950, opened 2023-04-25
fix: inherit SHELL env variables in wrappers

ragingpastry:
Possible fix for https://github.com/jetpack-io/devbox/issues/949

Edit by @mikeland86 

Fixes in this PR:
* Use correct bash path for bin wrappers (thanks @ragingpastry!)
* Using nix-shell instead of nix to determine if nix is installed.
* Don't print multi-user message if we're not installing nix.
* Only source `.bashrc` file if it exists

### Test Plan

* In docker started nixos, did `devbox init`, `devbox add hello` and then `devbox run hello`


mikeland86:
Thanks @ragingpastry! I tried this out and it works perfectly is `$SHELL` is exported. Unfortunately it seems like in nixos (at least the docker image I'm using for testing) does not export SHELL automatically.

I switched it to use `exec.LookPath` which seems to work reliably.

I also found a few other issues that can affect nixOS well:

* Using nix-shell instead of nix to determine if nix is installed.
* We print a message about root/multi-user every time even if nix is already installed

Included fixes for these.

internal/boxcli/setup.go
```diff
@@ -44,28 +44,30 @@ func runInstallNixCmd(cmd *cobra.Command) error {
...
 
-func nixDaemonFlagVal(cmd *cobra.Command) *bool {
-	if !cmd.Flags().Changed(nixDaemonFlag) {
-		if os.Geteuid() == 0 {
-			ux.Fwarning(
-				cmd.ErrOrStderr(),
-				"Running as root. Installing Nix in multi-user mode.\n",
-			)
-			return lo.ToPtr(true)
+func nixDaemonFlagVal(cmd *cobra.Command) func() *bool {
```
savil, line 54: its a bit unusual to return a `func`, so could you write a comment explaining the motivation: 
we don't want to print the multi-user warning unless this function is actually being used by nix.Install...

its won't be obvious to the casual reader, especially without the context of this PR

savil (approved):
awesome! thanks for digging into it @mikeland86 and thanks for taking a stab at it @ragingpastry !

PR #952, opened 2023-04-25
[global] Don't add global packages to local

mikeland86:
## Summary

This removes the concept of "merged packages". Now all packages are local and global are not installed at all. They are ignored for plugin, init hook purposes and everything else.

global will only work if it's been previously shellenv'd.

Minor fix involved renaming the `DEVBOX_OG_PATH` env var to ensure it doesn't conflict (between global and different projects).

## How was it tested?

```bash
devbox global add hello
devbox run hello
```


internal/impl/devbox.go
```diff
@@ -712,9 +700,12 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
...
 	// Use the original path, if available. If not available, set it for future calls.
 	// See https://github.com/jetpack-io/devbox/issues/687
-	originalPath, ok := env["DEVBOX_OG_PATH"]
+	// We add the project dir hash to ensure that we don't have conflicts
+	// between different projects (including global)
+	// (moving a project would change the hash and that's fine)
+	originalPath, ok := env["DEVBOX_OG_PATH_"+d.projectDirHash()]
 	if !ok {
-		env["DEVBOX_OG_PATH"] = currentEnvPath
+		env["DEVBOX_OG_PATH_"+d.projectDirHash()] = currentEnvPath
```
savil, line 708: this feels like an important but easy-to-overlook-or-break scenario. Could we augment `testscripts/shell/shellenv.test.txt` to test for the `devbox rm` scenario mentioned in the issue? cc @ipince

mikeland86: For some reason `devbox` gets removed from path in `testscripts/shell/shellenv.test.txt` (not sure if sourcing a new path is doing it and somehow the path that gets used by devbox is different than the path in testscripts)

So instead, wrote a few unit tests directly.

mikeland86:
New unit tests https://github.com/jetpack-io/devbox/pull/952/commits/4cda405cbe02320d6698a0f481cb71c02959d269

PR #954, opened 2023-04-26
chore: update env constants and functions

hezhizhen:
## Summary

1. Unexport some constants in `env`
2. Rename some functions in `env`


hezhizhen:
@mikeland86 PTAL

mikeland86 (approved):
Thanks!

PR #955, opened 2023-04-26
[perf] move ensureNixpkgsPrefetched outside nix.ProfileInstall for-loops

savil:
## Summary

I tried refactoring the nix.ProfileInstall callers to make it work for multiple
packages, but it is a bit challenging because:
1. local installs via `devbox add`: these need to fail-fast as soon as the first package fails
2. global installs: these need to continue trying to install for all

I'll send a PR that does that but sending this first.

### DIFF
--- a/internal/impl/global.go
+++ b/internal/impl/global.go
@@ -55,6 +55,10 @@ func (d *Devbox) AddGlobal(pkgs ...string) error {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


savil:
* main
  * **PR #955** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/955" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/955?utm_source=stack-comment).

PR #958, opened 2023-04-26
[tests] global-add should not fail if global-bins-not-in-path and add testscript unit tests for add and global-add

savil:
## Summary

Makes a couple of fixes:
1. `devbox global add` should avoid returning an error status code if devbox-global-bin-path is not in PATH. We should instead print a warning, and return a 0 code.
2. `devbox global shellenv` should call `ensureNixInstalled` to find `nix` if not in `PATH`.

Adds basic test cases for multiple packages for `devbox add` and `devbox global add`

## How was it tested?

`go test -v -run TestScripts/add\.test ./testscripts/...`
`go test -v -run TestScripts/global_add\.test ./testscripts/...`

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -103,8 +103,9 @@ func globalPullCmd() *cobra.Command {
 func globalShellenvCmd() *cobra.Command {
 	flags := shellEnvCmdFlags{}
 	command := &cobra.Command{
-		Use:   "shellenv",
-		Short: "Print shell commands that add global Devbox packages to your PATH",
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #958** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/958" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/958?utm_source=stack-comment).

gcurtis (approved):

PR #960, opened 2023-04-26
[easy][wrappers] Ensure wrappers use correct project dir

mikeland86:
## Summary

We're relying on binaries getting called within the same directory hierarchy where the original devbox.json lives but this can break down in the following cases:

* global
* Multiple devbox.json within the hirarchy
* if you navigate outisde the directory with the devbox.json

## How was it tested?

```bash
devbox shell
devbox add hello # to change the env hash
cd ~/
go version
```

Also tested global:

```
devbox global add hello  # to change the env hash
go version
```


savil (approved):
ooh good find!

PR #961, opened 2023-04-26
[easy][global] Don't show eval message if already shellenved

mikeland86:
## Summary

Two fixes:

* We were computing the wrong hash for projectDir. (using config contents instead). We want it to be stable even when adding packages so dir is the correct one.
* In devbox global we were showing the shellenv warning even when user had already shellenved.

## How was it tested?

`devbox global add hello`

### DIFF
--- a/internal/impl/global.go
+++ b/internal/impl/global.go
@@ -73,7 +73,7 @@ func (d *Devbox) AddGlobal(pkgs ...string) error {
 	if err := d.saveCfg(); err != nil {
 		return err
 	}
-	d.ensureGlobalProfileInPath()
+	d.ensureDevboxGlobalShellenvEnabled()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/impl/global.go
```diff
@@ -174,29 +173,9 @@ func GlobalNixProfilePath() (string, error) {
...
-
-// Checks if the global profile is in the path
-func (d *Devbox) ensureGlobalProfileInPath() error {
-	binPath, err := globalBinPath()
-	if err != nil {
-		return err
-	}
-	if !strings.Contains(os.Getenv(env.Path), binPath) {
+// Checks if the global has been shellenv'd and warns the user if not
+func (d *Devbox) ensureGlobalProfileInPath() {
```
savil, line 177: should this be renamed to `ensureDevboxGlobalShellenvEnabled()` ?

savil (approved):

PR #964, opened 2023-04-27
[lockfile] Use lockfile for legacy packages

mikeland86:
## Summary

This changes saves legacy packages to lockfile. It does not save absolute flake references to lockfile (e.g. `github:/my/flake#foo`) but we could in the future.

Merging legacy and versioned packages let's us simplify the code quite a bit. Specifically now all inputs are treated like flakes (either explicit or resolved). 

One complication this exposes is the php and haskell planners. Specifically, each planner creates definitions must be included as packages. As a quick hack I just include the definitions as is at the top of the list. This ensures these take precedence over any other package. It also means that packages can show up twice if these planners are activated. That doesn't break the functionality, it just looks weird in the flake.nix

@Lagoja are there any example that exercises the Haskell planner?

A few possible follow ups:

- [ ] We only create lockfile when packages need to be resolved. This may lead to some perf issues is people use an existing project on a read only platform (e.g. CICD) because not having a lockfile makes things a bit slower.
- [ ] Out flake.nix is a bit ugly because now it has multiple duplicate inputs. We can clean this up.

## How was it tested?

Need to do a lot more testing but what I did:

```
devbox rm go_1_20 && devbox add go_1_20
rm -rf .devbox
devbox install 
devbox run echo 5
devbox add python@3.11
devbox run "python --version"
devbox shell
```

Need to test:

- [x] LOCKFILE flag off on legacy project
- [x] LOCKFILE flag on on legacy project
- [x] LOCKFILE flag off on project with lockfile
- [x] LOCKFILE flag on on project with lockfile


mikeland86:
Adding @gcurtis for all the flake.nix stuff.

mikeland86:
There's 3 failing tests that may be failing do to a perf regression in this PR. (they pass locally on my machine). I'm gonna test adding a longer delay to see if fixes them. If it does, I'll investigate the perf regression.

savil:
> @Lagoja are there any example that exercises the Haskell planner?

@mikeland86  see `examples/development/haskell`

mikeland86:
@savil examples/development/haskell runs, not sure if there's anything else I should do to test out the planner.

internal/nix/profile.go
```diff
@@ -188,42 +183,6 @@ func (item *NixProfileListItem) AttributePath() (string, error) {
...
-		return "", redact.Errorf(
-			"expected >= 2 parts for AttributePath in \"nix profile list\" output, but got %d parts: %s",
-			redact.Safe(len(parts)),
-			redact.Safe(attrPath),
-		)
-	}
-
-	return strings.Join(parts[2:], "."), nil
-}
-
```
savil, line 226: good riddance :)

internal/impl/flakes.go
```diff
@@ -14,22 +14,20 @@ func (d *Devbox) flakeInputs() []*plansdk.FlakeInput {
...
-				inputs[pkg.URLForInput()] = &plansdk.FlakeInput{
-					Name:     pkg.Name(),
-					URL:      pkg.URLForInput(),
-					Packages: []string{AttributePath},
-				}
-			} else {
-				input.Packages = lo.Uniq(
-					append(inputs[pkg.URLForInput()].Packages, AttributePath),
-				)
+		AttributePath, err := pkg.PackageAttributePath()
```
savil, line 17: To clarify: everything will be a flake input now, right? 

Perhaps we could add a function comment here, or maybe in the `plansdk.FlakeInput` file explaining that we can treat user-specified packages and explicit flake inputs uniformly as flakeInputs. Or something along those lines.

For some later cleanup: should this function be renamed `Inputs()`, and and `plansdk.FlakeInput` be renamed `plansdk.Input`, by dropping the explicit "flake" keyword?
By having "flake" keyword be in the name, the question arises: what are the other kinds of inputs?

savil (approved):
Nice!

internal/boxcli/featureflag/lockfile.go
```diff
@@ -0,0 +1,3 @@
+package featureflag
+
+var LockFile = disabled("LOCKFILE")
```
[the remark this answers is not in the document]
mikeland86: I merged the flags (because versioned doesn't make a ton of sense without a lockfile) so only this flag exists

internal/impl/flakes.go
mikeland86: On second thought, I think `flakeInputs()` makes sense (it's not inputs that are flakes, but inputs to a flake).

I think `plansdk. FlakeInput ` could be renamed to `flake.Input` maybe.

PR #965, opened 2023-04-28
[version] part 1: update command for devbox and launcher updates

savil:
## Summary

**Motivation:**
We want to disable auto-update for devbox CLIs. This will minimize unexpected interruptions for our users.

To do so, we are going to do the following:

1. Update `launch.sh` file in `axiom` repo to write new versions to `{XDG_CACHE_HOME}/devbox/available-version`. https://github.com/jetpack-io/axiom/pull/3150
2. Change `devbox version update` to update the launcher (if needed) and else, update the CLI binary. (this PR).
3. Change the `vercheck.CheckVersion` that runs in `boxcli/root.go` on every command to check if the launcher or CLI binary versions are out-of-date, and to print a notice if so. (reviewed in #966, but merged into this PR)

We will cherry-pick ~these two PRs~ this PR for the next release so auto-update stops for users.

**Implementation:**
This PR enables `jetpack version update` to:

1. Update the launcher: We detect if the launcher version is outdated, and update it, if needed. This also updates the devbox CLI binary as a side-effect. This requires `sudo` so we only do it, if necessary.

2. Update just the devbox CLI binary: If the launcher is up-to-date, then we update just the devbox CLI binary.

3. Remove dependency on new code like `env` package to minimize cherry-picking merge conflicts.

TODO:
- [x] confirm that we can switch the `devbox version` operation after an update to `StdOut`. I feel this is better so user is informed about the new version.
cc @mikeland86 for the above question.

Will only land this after the launch.sh file is updated in https://github.com/jetpack-io/axiom/pull/3150 

## How was it tested?

Did some basic testing:

hardcoded `true` to ensure `selfUpdateLauncher` runs, and did `devbox version update`:

Screenshot 2023-05-01 at 8 57 46 PM

regular `devbox version update`
Screenshot 2023-05-01 at 8 48 20 PM

### DIFF
--- a/internal/vercheck/vercheck.go
+++ b/internal/vercheck/vercheck.go
@@ -4,18 +4,21 @@
 package vercheck
 
 import (
+	"bytes"
 	"fmt"
 	"io"
-	"net/http"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/build"
 	"golang.org/x/mod/semver"
 
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
-	"go.jetpack.io/devbox/internal/envir"
 	"go.jetpack.io/devbox/internal/ux"
 	"go.jetpack.io/devbox/internal/xdg"
 )
@@ -24,12 +27,9 @@ import (
 // Than the version in launch.sh, we'll print a warning.
 const expectedLauncherVersion = "v0.1.0"
 
-// currentChannel represents the CLI update channel. For now, we set it to a
-// constant, but we'll expand this to be a variable in the future.
-const currentChannel = "stable"
-
 func CheckLauncherVersion(w io.Writer) {
-	if envir.IsDevboxCloud() {
+	// Replace with envir.IsDevboxCloud(). Not doing so to minimize merge conflicts.
+	if os.Getenv("DEVBOX_REGION") != "" {
 		return
 	}
 
@@ -44,14 +44,18 @@ func CheckLauncherVersion(w io.Writer) {
 	}
 }
 
-// SelfUpdate updates the devbox launcher and binary. It ignores and deletes the
-// version cache
+// SelfUpdate updates the devbox launcher and devbox CLI binary.
+// It ignores and deletes the version cache.
+//
+// The launcher is a wrapper bash script introduced to manage the auto-update process
+// for devbox. The production devbox application is actually this launcher script
+// that acts as "devbox" and delegates commands to the devbox CLI binary.
 func SelfUpdate(stdOut, stdErr io.Writer) error {
 	if isNewLauncherAvailable() {
 		return selfUpdateLauncher(stdOut, stdErr)
 	}
 
-	return selfUpdateBinary(stdOut, stdErr)
+	return selfUpdateDevbox(stdErr)
 }
 
 func selfUpdateLauncher(stdOut, stdErr io.Writer) error {
@@ -64,9 +68,13 @@ func selfUpdateLauncher(stdOut, stdErr io.Writer) error {
 		return usererr.New("curl or wget is required to update devbox. Please install either and try again.")
 	}
 
-	// Delete version cache.
-	_ = os.Remove(versionCacheFilePath())
+	// Delete current version file. This will trigger an update when invoking any devbox command;
+	// in this case, inside triggerUpdate function.
+	if err := removeCurrentVersionFile(); err != nil {
+		return err
+	}
 
+	// Fetch the new launcher.
 	cmd := exec.Command("sh", "-c", installScript)
 	cmd.Stdout = stdOut
 	cmd.Stderr = stdErr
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #965** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/965" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/965?utm_source=stack-comment).

internal/vercheck/vercheck.go
```diff
@@ -50,10 +61,8 @@ func SelfUpdate(stdOut, stdErr io.Writer) error {
 		return usererr.New("curl or wget is required to update devbox. Please install either and try again.")
 	}
 
-	// Delete version cache. Keep this in-sync with whatever logic is in launch.sh
-	cacheDir := xdg.CacheSubpath("devbox")
-	versionCacheFile := filepath.Join(cacheDir, "latest-version")
-	_ = os.Remove(versionCacheFile)
+	// Delete version cache.
+	_ = os.Remove(versionCacheFilePath())
```
mikeland86, line 65: I think we should handle this error. If this file can't be deleted we may end up in a bad state:
```golang
if err := os.Remove(versionCacheFilePath()); err !=nil && !os.IsNotExist(err) {
  return ...
}
```

internal/vercheck/vercheck.go
```diff
@@ -62,14 +71,84 @@ func SelfUpdate(stdOut, stdErr io.Writer) error {
...
+func triggerUpdateAndPrintNewVersion(stdOut, stdErr io.Writer) error {
+	fmt.Fprintln(stdOut, "triggering update...")
+
 	exe, err := os.Executable()
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	cmd = exec.Command(exe, "version")
-	// The output of version is incidental, so just send it all to stdErr
-	cmd.Stdout = stdErr
```
mikeland86, line 72: I'm a bit confused about 

> I feel this is better so user is informed about the new version.

Both stdout and stderr get printed to terminal by default so anything we want to warn the user about can go in stderr. The main heuristic I use is:

* If the command explicitly asking for output e.g. `devbox global ls` or `devbox shellenv`. The output is the _reason_ for these commands. From a user perspective, they are not meant to perform actions as much as simply return/print text.
* If the command meant to do some action "x" and then print the status or information associated to the performance of this command, in that case I send to stderr. 

Anyway, we don't strictly follow this and since update requires a tti anyway (it asks a question and does sudo) it's not a big deal so not a blocker. Just sharing my process.

internal/vercheck/vercheck.go
```diff
@@ -20,27 +21,37 @@ import (
 // Than the version in launch.sh, we'll print a warning.
 const expectedLauncherVersion = "v0.1.0"
 
+// currentChannel represents the CLI update channel. For now, we set it to a
+// constant, but we'll expand this to be a variable in the future.
+const currentChannel = "stable"
```
mikeland86, line 26: The CLI doesn't need to know about channels 

We could, but I would prefer to encapsulate the logic in the launcher only. Having logic in both places introduces coordination problems (as we're currently experiencing)

internal/vercheck/vercheck.go
```diff
@@ -62,14 +71,84 @@ func SelfUpdate(stdOut, stdErr io.Writer) error {
...
+	// time is reset.
+	_ = os.Remove(versionCacheFilePath())
+
+	// Write the new version to the version cache file.
+	if err := os.WriteFile(versionCacheFilePath(), newVersion, 0644); err != nil {
+		return errors.WithStack(err)
+	}
+
+	return triggerUpdateAndPrintNewVersion(stdOut, stdErr)
+}
```
mikeland86, line 106: This can be simplified to:

```golang
func selfUpdateBinary(stdOut, stdErr io.Writer) error {
  os.Remove(filepath.Join(xdg.CacheSubpath("devbox"), "current-version"))
  return triggerUpdateAndPrintNewVersion(stdOut, stdErr)
}
```

By removing the file and calling the launcher again you trigger a new fetch (we always fetch new version if current-version is missing even if auto update is turned off). 

A small change I can add to the launcher is to pass this file path as an ENV var (e.g. `DEVBOX_CURRENT_VERSION_FILE_PATH`) to reduce the coupling. 

A minor UX improvement could be to have `triggerUpdateAndPrintNewVersion` return the new version.

```golang
func selfUpdateBinary(stdOut, stdErr io.Writer) error {
  os.Remove(filepath.Join(xdg.CacheSubpath("devbox"), "current-version"))
  version, err := triggerUpdateAndPrintNewVersion(stdOut, stdErr)
  if err != nil {
    return err
  } 
  if semver.Compare(version, build.Version) == 0 {
    print you are already at latest version, blah...
  }
}
```

mikeland86 (changes requested):

savil (commented):
update coming...

savil:
Updated code, but need to test properly.

internal/vercheck/vercheck.go
mikeland86: Just to clarify, stderr is visible to the user. I do want users to see diagnostic info especially errors that occur during normal operation. 

So `devbox version update` will show the latest version and `devbox add <pkg>` will show success (but to stderr).

internal/vercheck/vercheck.go
```diff
@@ -53,26 +64,142 @@ func SelfUpdate(stdOut, stdErr io.Writer) error {
...
+// triggerUpdate runs `devbox version -v` and triggers an update since a new
+// version is available. It parses the output to get the new launcher and
+// devbox versions.
+func triggerUpdate(stdErr io.Writer) (*updatedVersions, error) {
+
+	exe, err := os.Executable()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	cmd := exec.Command(exe, "version", "-v")
```
mikeland86, line 122: Possibly future hardening is adding `--json`

mikeland86 (approved):

internal/vercheck/vercheck.go
savil: that would be a good idea

savil:
I'm going to merge #966 into this PR, prior to merging to `main` so there is one commit to cherrypick

PR #973, opened 2023-05-01
Update Dockerfile to match Devbox Cloud

Lagoja:
## Summary

TSIA -- moves our generated dockerfile to a `debain:stable-slim` base

## How was it tested?

`docker build .` on an example project


internal/impl/tmpl/Dockerfile.dockerignore.tmpl
mohsenari, line 1: @Lagoja This is the wrong file to update.
This is dockerignore. You should update `devcontainerDockerfile.tmpl`

mohsenari (changes requested):
Wrong file is updated.

internal/impl/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -1,30 +1,30 @@
...
 
-# Setting up devbox user
-ENV DEVBOX_USER=devbox
-RUN adduser -h /home/$DEVBOX_USER -D -s /bin/bash $DEVBOX_USER
-RUN addgroup sudo
-RUN addgroup $DEVBOX_USER sudo
-RUN echo " $DEVBOX_USER      ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers
+# Step 1: Installing dependencies
+RUN apt-get update
+RUN apt-get -y install bash binutils git xz-utils wget sudo
```
hezhizhen, line 5: How about choose a base image that contains these tools (or some of them)?

internal/impl/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -1,30 +1,30 @@
...
 
-# installing devbox
-RUN wget --quiet --output-document=/dev/stdout https://get.jetpack.io/devbox | bash -s -- -f
+# Step 3: Installing devbox
+RUN wget --quiet --output-document=/dev/stdout https://get.jetpack.io/devbox   | bash -s -- -f
 RUN chown -R "${DEVBOX_USER}:${DEVBOX_USER}" /usr/local/bin/devbox
 
-# nix installer script
-RUN wget --quiet --output-document=/dev/stdout https://nixos.org/nix/install | sh -s -- --no-daemon
+# Step 4: Install Nix
```
hezhizhen, line 19: Use `Installing` to follow the pattern

internal/impl/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -1,30 +1,30 @@
...
+RUN wget --output-document=/dev/stdout https://nixos.org/nix/install | sh -s -- --no-daemon
 RUN . ~/.nix-profile/etc/profile.d/nix.sh
 # updating PATH
 ENV PATH="/home/${DEVBOX_USER}/.nix-profile/bin:/home/${DEVBOX_USER}/.devbox/nix/profile/default/bin:${PATH}"
 
+# Step 5: Install your devbox project
```
hezhizhen, line 25: `Installing`

internal/impl/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -1,30 +1,30 @@
...
+RUN apt-get update
+RUN apt-get -y install bash binutils git xz-utils wget sudo
 
-# installing dependencies
-RUN apk add --no-cache bash binutils git libstdc++ xz sudo
+# Step 2: Setting up devbox user
+ENV DEVBOX_USER=devbox
+RUN adduser $DEVBOX_USER
+RUN usermod -aG sudo $DEVBOX_USER
+RUN echo " devbox      ALL=(ALL:ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/$DEVBOX_USER
```
hezhizhen, line 11: Why do we need some many spaces between `devbox` and `ALL`?

internal/impl/tmpl/devcontainerDockerfile.tmpl
mohsenari: @hezhizhen The motivation for using this base image is that, we use the same base image in devbox cloud instances. To keep the image size small, it's better to use a minimal image and install the tools you need.

mohsenari (approved):
There are some small linting updates that I can take care. Otherwise, looks good.

internal/impl/tmpl/devcontainerDockerfile.tmpl
hezhizhen: I see.

PR #974, opened 2023-05-02
[direnv] `.envrc` simplified

mohsenari:
## Summary

Updated .envrc template to make it simpler and also future proof. So that if the contents of `use devbox` function changes in the future, we won't have mismatch between devbox binary and user's `.envrc` file.

Added `--envrc` flag to `devbox generate direnv` so that it outputs the content of `use devbox` function. 
Without the flag, the command generates the following:
```
# Automatically sets up your devbox environment whenever you cd into this
# directory via our direnv integration:

eval $(devbox generate direnv --envrc)

# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
# for more details
```

## How was it tested?
- compile
- in a new directory run `devbox generate direnv` (answer No to the prompt)
- update the .envrc file and change `devbox generate ...` to `./devbox generate` (so that it uses the compiled binary).
- cd out of directory and back in to check if direnv works as expected.


internal/boxcli/generate.go
```diff
@@ -98,6 +99,8 @@ func direnvCmd() *cobra.Command {
 	}
 	command.Flags().BoolVarP(
 		&flags.force, "force", "f", false, "force overwrite existing files")
+	command.Flags().BoolVarP(
+		&flags.envrcContent, "envrc", "e", false, "output contents of a configuration to use in .envrc")
```
hezhizhen, line 103: A clearer name would be better. Add `output (o)` or `print (p)` to the name.

internal/impl/devbox.go
```diff
@@ -384,48 +385,32 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
-				return errors.WithStack(err)
-			}
+		// .envrc file creation
+		err := generate.CreateEnvrc(tmplFS, d.projectDir)
+		if err != nil {
+			return errors.WithStack(err)
 		}
 
-		if strings.ToLower(result) == "y" || !isInteractiveMode {
+		if cmdutil.Exists("direnv") {
```
savil, line 406: based on the conversation with @Lagoja, I thought we were gonna print some information messages, like:
```
Success: generated .envrc file
Success: ran `direnv allow`
```

Do you still wanna do that?

savil: You could add a `ux.Fsuccess(msg)` method that prints a green "Success:" and the rest of the `msg`

similar to `ux.Fwarning` with the yellow "Warn"

savil (approved):

internal/impl/devbox.go
mohsenari: yeah sure I'll add that.

internal/impl/tmpl/direnvrc.tmpl
```diff
@@ -0,0 +1,5 @@
+use_devbox() {
```
mikeland86, line 1: is direnvrc  a standard name? It looks like direnv uses it for custom extensions. Is this an extension?

internal/impl/devbox.go
```diff
@@ -384,48 +386,33 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
-		isInteractiveMode := isatty.IsTerminal(os.Stdin.Fd())
-		if isInteractiveMode {
-			prompt := &survey.Input{
-				Message: "Do you want to enable direnv integration for this devbox project? [y/N]",
-			}
-			err := survey.AskOne(prompt, &result)
-			if err != nil {
-				return errors.WithStack(err)
-			}
+	if source == "generate" {
```
mikeland86, line 389: Not from this PR, but what is the purpose of this brancing logic? Don't we always want to do this check?

I'm a afraid if we don't do this check during `init` and then we add packages manually to devbox.json then these errors might not show up because we are calling it from inside a file and discarding the stderr.

mikeland86 (changes requested):

internal/impl/tmpl/direnvrc.tmpl
mikeland86, line 1: is direnvrc a standard name? It looks like direnv uses it for custom extensions. Is this an extension?

devbox.go
```diff
@@ -27,11 +27,12 @@ type Devbox interface {
 	Generate() error
 	GenerateDevcontainer(force bool) error
 	GenerateDockerfile(force bool) error
-	GenerateEnvrc(force bool, source string) error
+	CreateEnvrcFile(force bool) error
```
hezhizhen, line 30: Should `GenerateDevcontainer` and `GenerateDockerfile` and `CreateEnvrcFile` use the same verb?

devbox.go
```diff
@@ -27,11 +27,12 @@ type Devbox interface {
...
 	Info(pkg string, markdown bool) error
 	ListScripts() []string
 	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
 	PrintGlobalList() error
+	PrintEnvrcContent(w io.Writer) error
```
hezhizhen, line 35: Should these print methods have similar parameters (`io.Writer`)?

internal/boxcli/generate.go
```diff
@@ -16,6 +16,7 @@ import (
 type generateCmdFlags struct {
 	config         configFlags
 	force          bool
+	envrcContent   bool
```
hezhizhen, line 19: `printEnvrcContent` would be better

internal/boxcli/generate.go
```diff
@@ -98,6 +99,11 @@ func direnvCmd() *cobra.Command {
 	}
 	command.Flags().BoolVarP(
 		&flags.force, "force", "f", false, "force overwrite existing files")
+	command.Flags().BoolVarP(
+		&flags.envrcContent, "print-envrc", "p", false, "output contents of devbox configuration to use in .envrc")
+	// this command marks a flag as hidden. Error handling for it is not necessary.
+	// _ = command.Flags().MarkHidden("print-envrc")
```
hezhizhen, line 105: remove these two lines?

internal/boxcli/generate.go
```diff
@@ -137,7 +143,11 @@ func runGenerateCmd(cmd *cobra.Command, flags *generateCmdFlags) error {
 	case "dockerfile":
 		return box.GenerateDockerfile(flags.force)
 	case "direnv":
-		return box.GenerateEnvrc(flags.force, "generate")
+		if flags.envrcContent {
+			fmt.Println("alkdsfhlaskghslkdfj")
```
hezhizhen, line 147: unnecessary print

internal/impl/tmpl/envrc.tmpl
```diff
@@ -1,15 +1,7 @@
 # Automatically sets up your devbox environment whenever you cd into this
```
hezhizhen, line 1: What's the difference between `envrc.tmpl` and `envrcContent.tmpl`?

internal/boxcli/generate/devcontainer_util.go
```diff
@@ -70,6 +70,14 @@ func CreateDevcontainer(path string, pkgs []string) error {
 	return err
 }
 
+func OutputEnvrcContent(tmplFS embed.FS) error {
```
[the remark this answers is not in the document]
mohsenari: Moved this function to impl/devbox.go and named it `PrintEnvrcContent` broke the `createEnvrc` function into 2 separate functions.

internal/impl/tmpl/envrc.tmpl
mohsenari: `envrc.tmpl` is the file itself. But inside the file we use a devbox command and that command prints what is inside `envrcContent.tmpl`

devbox.go
mohsenari: This function in particular should print to stdout. The others get their writer from devbox struct when we do `box.open()`.

internal/impl/tmpl/direnvrc.tmpl
mohsenari: @mikeland86 can I get another review so that I can potentially merge this?

devbox.go
mikeland86: @hezhizhen I think that's a good point.

@mohsenari you are right that the other functions get their writer from box struct, but they only really make sense as stdout (and in that's how they are used).

In fact, `devbox.Open` is almost always called with stderr (except for list, generate, and info cases).

I think having `print` functions implicitly print to whatever is in devbox struct can be a bit error prone. Another way to think about it is that the `Writer` field in devbox is actually a logger (or just were we send diagnostics).

I would not suggest any action in this PR (I think the current signature is good). But maybe for the future:

* Default devbox.Writer to Stderr and remove from open() method.
* Refactor all print functions to take writer, or alternatively just have them all return the string.

internal/boxcli/init.go
```diff
@@ -32,10 +32,6 @@ func runInitCmd(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	box, err := devbox.Open(path, cmd.ErrOrStderr())
-	if err != nil {
-		return errors.WithStack(err)
-	}
 
-	return errors.WithStack(box.GenerateEnvrc(false, "init"))
+	return nil
```
mikeland86, line 36: nit, above if statement can be removed and this can just be `return errors.WithStack(err)`

internal/impl/devbox.go
```diff
@@ -384,48 +391,31 @@ func (d *Devbox) GenerateEnvrc(force bool, source string) error {
...
-			err := cmd.Run()
-			if err != nil {
-				return errors.WithStack(err)
-			}
+	// .envrc file creation
+	err := generate.CreateEnvrc(tmplFS, d.projectDir)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	ux.Fsuccess(d.writer, "generated .envrc file\n")
```
mikeland86, line 410: This should go in if statement right? We should not create `.envrc` file if user does not have direnv installed?

mikeland86 (approved):
Only blocking comment is that `generate.CreateEnvrc(tmplFS, d.projectDir)` should go in if statement so we don't create the file if the user doesn't have direnv installed.

mikeland86 (approved):
Only blocking comment is that `generate.CreateEnvrc(tmplFS, d.projectDir)` should go in if statement so we don't create the file if the user doesn't have direnv installed.

mikeland86 (approved):
Only blocking comment is that `generate.CreateEnvrc(tmplFS, d.projectDir)` should go in if statement so we don't create the file if the user doesn't have direnv installed.

internal/impl/devbox.go
mohsenari: @mikeland86 No we only check for direnv being installed to run `direnv allow`. I don't think it's necessary to check direnv being installed and prevent user from generating the file. They might have the direnv binary compiled and not in path or they might be generating the file for cicd or other other environments. With that logic we should also check if user has Docker installed before for `devbox generate dockerfile`.

devbox.go
savil: > Default devbox.Writer to Stderr and remove from open() method.

reminder: the `--quiet` flag is needs to be respected

PR #975, opened 2023-05-02
[version update] use LAUNCHER_PATH in triggerUpdate function

savil:
## Summary

For trigger update to work, we need to invoke the "launcher" and not the "devbox CLI binary".

To do so, we can use the `LAUNCHER_PATH` env-var. And fallback to "devbox" if in PATH,
or finally as a fail-safe to the prior `os.Executable`.

## How was it tested?

:eyes:

carefully pair-programmed with @mikeland86. We'll test after building the pre-release
in `0.4.9-dev`.

### DIFF
--- a/internal/vercheck/vercheck.go
+++ b/internal/vercheck/vercheck.go
@@ -16,6 +16,7 @@ import (
 	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/build"
+	"go.jetpack.io/devbox/internal/ux"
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #975** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/975" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/975?utm_source=stack-comment).

mikeland86 (approved):

PR #976, opened 2023-05-03
chore: use constants defined in envir in version check

hezhizhen:
## Summary

1. replace constants with existing envir constants
2. add new envir constants

### DIFF
--- a/internal/boxcli/version.go
+++ b/internal/boxcli/version.go
@@ -9,7 +9,9 @@ import (
 	"runtime"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


hezhizhen:
@savil Please check if I've left anything.

savil (approved):
thanks! this looks good to me cc @mikeland86 

We're hoping to deploy a release this morning to stop auto-updates. If there's some problems with that, we may need to update `vercheck` again. So, I'm going to wait on merging this PR until we have confidence that the release went well.

PR #978, opened 2023-05-03
[update] Implement devbox update

mikeland86:
## Summary

Implements `devbox update [pkg]...`

It basically just re-resolves version. If there's a new version that satisfies constraints, it installs it and removes the previous one.

Other improvements:

* implemented `findPackageByName` which allows us to use `devbox update go` instead of `devbox update go@1.20`. In follow up we can also use this for `devbox rm` 
* flake.nix optimization: If there are no dev packages (this is a legacy planner concept that is almost never used) and we add at least one nixpkgs flake input, we use that input for mkShell. This avoids having to download an additional nixpkgs.

## How was it tested?

* Manually modified lockfile to point to older versions and ran `devbox update` and `devbox update go`
* Inspected flake.nix 
* Inspected nix profile

### DIFF
--- a/internal/boxcli/update.go
+++ b/internal/boxcli/update.go
@@ -18,15 +18,13 @@ func updateCmd() *cobra.Command {
 
 	command := &cobra.Command{
 		Use:   "update [pkg]...",
-		Short: "Updates one, many, or all packages in your devbox",
-		Long: "Updates one, many, or all packages in your devbox. If no " +
-			"packages are specified, all packages will be updated. Only updates " +
-			"versioned packages (e.g. `python@3.11`), not packages that are " +
-			"pinned to a nix channel (e.g. `python3`)",
+		Short: "Updates packages in your devbox",
+		Long: "Updates one, many, or all packages in your devbox. " +
+			"If no packages are specified, all packages will be updated. " +
+			"Only updates versioned packages (e.g. `python@3.11`), not packages that are pinned to a nix channel (e.g. `python3`)",
 		PreRunE: ensureNixInstalled,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return updateCmdFunc(cmd, args, flags)
-
 		},
 	}
 
--- a/internal/boxcli/update.go
+++ b/internal/boxcli/update.go
@@ -1,0 +1,48 @@
+// Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package boxcli
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"go.jetpack.io/devbox"
+)
+
+type updateCmdFlags struct {
+	config configFlags
+}
+
+func updateCmd() *cobra.Command {
+	flags := &updateCmdFlags{}
+
+	command := &cobra.Command{
+		Use:   "update [pkg]...",
+		Short: "Updates one, many, or all packages in your devbox",
+		Long: "Updates one, many, or all packages in your devbox. If no " +
+			"packages are specified, all packages will be updated. Only updates " +
+			"versioned packages (e.g. `python@3.11`), not packages that are " +
+			"pinned to a nix channel (e.g. `python3`)",
+		PreRunE: ensureNixInstalled,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return updateCmdFunc(cmd, args, flags)
+
+		},
+	}
+
+	flags.config.register(command)
+	return command
+}
+
+func updateCmdFunc(
+	cmd *cobra.Command,
+	args []string,
+	flags *updateCmdFlags,
+) error {
+	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return box.Update(cmd.Context(), args...)
+}
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -150,6 +150,19 @@ func (d *Devbox) ShellPlan() (*plansdk.ShellPlan, error) {
 	shellPlan.FlakeInputs = d.flakeInputs()
 
 	nixpkgsInfo := plansdk.GetNixpkgsInfo(d.cfg.Nixpkgs.Commit)
+
+	// This is an optimization. If there are no dev packages (which we only use
+	// for php/haskell planners) we can use nixpkgs from one of the flakes.
+	// That saves us from downloading nixpkgs an additional time for mkShell.
+	if len(shellPlan.DevPackages) == 0 {
+		for _, input := range shellPlan.FlakeInputs {
+			if input.IsNixpkgs() {
+				nixpkgsInfo = plansdk.GetNixpkgsInfo(input.HashFromNixPkgsURL())
+				break
+			}
+		}
+	}
+
 	shellPlan.NixpkgsInfo = nixpkgsInfo
 
 	return shellPlan, nil
@@ -940,6 +953,23 @@ func (d *Devbox) nixFlakesFilePath() string {
 // packages returns the list of packages to be installed in the nix shell.
 func (d *Devbox) packages() []string {
 	return d.cfg.Packages
+}
+
+func (d *Devbox) findPackageByName(name string) (string, error) {
+	results := []string{}
+	for _, pkg := range d.cfg.Packages {
+		i := nix.InputFromString(pkg, d.lockfile)
+		if i.String() == name || i.CanonicalName() == name {
+			results = append(results, pkg)
+		}
+	}
+	if len(results) > 1 {
+		return "", usererr.New(
+			"found multiple packages with name %s: %s", name, results)
+	} else if len(results) == 0 {
+		return "", usererr.New("no package found with name %s", name)
+	}
+	return results[0], nil
 }
 
 // configEnvs takes the computed env variables (nix + plugin) and adds env
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -965,8 +965,12 @@ func (d *Devbox) findPackageByName(name string) (string, error) {
 	}
 	if len(results) > 1 {
 		return "", usererr.New(
-			"found multiple packages with name %s: %s", name, results)
-	} else if len(results) == 0 {
+			"found multiple packages with name %s: %s. Please specify version",
+			name,
+			results,
+		)
+	}
+	if len(results) == 0 {
 		return "", usererr.New("no package found with name %s", name)
 	}
 	return results[0], nil
--- a/internal/impl/update.go
+++ b/internal/impl/update.go
@@ -1,0 +1,46 @@
+package impl
+
+import (
+	"context"
+	"fmt"
+)
+
+func (d *Devbox) Update(ctx context.Context, pkgs ...string) error {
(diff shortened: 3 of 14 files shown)


internal/impl/devbox.go
```diff
@@ -942,6 +955,23 @@ func (d *Devbox) packages() []string {
...
+	for _, pkg := range d.cfg.Packages {
+		i := nix.InputFromString(pkg, d.lockfile)
+		if i.String() == name || i.CanonicalName() == name {
+			results = append(results, pkg)
+		}
+	}
+	if len(results) > 1 {
+		return "", usererr.New(
+			"found multiple packages with name %s: %s", name, results)
+	} else if len(results) == 0 {
```
hezhizhen, line 969: unnecessary `else`

internal/impl/update.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+func (d *Devbox) Update(ctx context.Context, pkgs ...string) error {
+	pkgsToUpdate := []string{}
+	if len(pkgs) == 0 {
+		pkgsToUpdate = append([]string(nil), d.packages()...)
```
hezhizhen, line 11: maybe we need a deep copy utilization function to replace this way?

internal/impl/update.go
```diff
@@ -0,0 +1,46 @@
+package impl
+
+import (
+	"context"
+	"fmt"
+)
+
+func (d *Devbox) Update(ctx context.Context, pkgs ...string) error {
+	pkgsToUpdate := []string{}
```
hezhizhen, line 9: it's okay to let `pkgToUpdate` be `nil` (like `var pkgToUpdate []string`)

internal/impl/update.go
```diff
@@ -0,0 +1,46 @@
...
+		pkgsToUpdate = append([]string(nil), d.packages()...)
+	} else {
+		for _, pkg := range pkgs {
+			found, err := d.findPackageByName(pkg)
+			if err != nil {
+				return err
+			}
+			pkgsToUpdate = append(pkgsToUpdate, found)
+		}
+	}
```
hezhizhen, line 20: How about this?

```go
var pkgToUpdate []string // declare pkgToUpdate
for _, pkg := range pkgs { ... } // add pkgs to pkgToUpdate
// no pkgs to update or no pkgs found
if len(pkgToUpdate) == 0 { pkgToUpdate = d.packages() } // TODO: deep copy it
```

internal/searcher/client.go
```diff
@@ -55,13 +55,14 @@ func (c *client) SearchVersion(query, version string) (*SearchResult, error) {
 	)
 }
 
-func (c *client) Resolve(pkg, version string) (*lock.Package, error) {
-	result, err := c.SearchVersion(pkg, version)
+func (c *client) Resolve(pkg string) (*lock.Package, error) {
+	name, version, _ := strings.Cut(pkg, "@")
```
hezhizhen, line 59: Should we resolve versioned package only? What if name or version is missing?

internal/searcher/client.go
```diff
@@ -55,13 +55,14 @@ func (c *client) SearchVersion(query, version string) (*SearchResult, error) {
...
+	result, err := c.SearchVersion(name, version)
 	if err != nil {
 		return nil, err
 	}
 	if len(result.Results) == 0 {
-		return nil, usererr.New("No results found for %q.", pkg)
+		return nil, usererr.New("No results found for %q.", name)
```
hezhizhen, line 65: I think `pkg` would be clearer because it's supposed to have both name and version.

internal/searcher/client.go
```diff
@@ -55,13 +55,14 @@ func (c *client) SearchVersion(query, version string) (*SearchResult, error) {
 	)
 }
 
-func (c *client) Resolve(pkg, version string) (*lock.Package, error) {
-	result, err := c.SearchVersion(pkg, version)
+func (c *client) Resolve(pkg string) (*lock.Package, error) {
+	name, version, _ := strings.Cut(pkg, "@")
+	result, err := c.SearchVersion(name, version)
```
hezhizhen, line 60: Maybe we should rename it as `SearchPackageWithVersion` or something else?

internal/boxcli/update.go
```diff
@@ -0,0 +1,48 @@
...
+type updateCmdFlags struct {
+	config configFlags
+}
+
+func updateCmd() *cobra.Command {
+	flags := &updateCmdFlags{}
+
+	command := &cobra.Command{
+		Use:   "update [pkg]...",
+		Short: "Updates one, many, or all packages in your devbox",
```
hezhizhen, line 21: I think there's no need to emphasize the number of packages. `Updates packages in your devbox` would be just fine.

internal/boxcli/update.go
```diff
@@ -0,0 +1,48 @@
...
+		Long: "Updates one, many, or all packages in your devbox. If no " +
+			"packages are specified, all packages will be updated. Only updates " +
+			"versioned packages (e.g. `python@3.11`), not packages that are " +
+			"pinned to a nix channel (e.g. `python3`)",
+		PreRunE: ensureNixInstalled,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return updateCmdFunc(cmd, args, flags)
+
```
hezhizhen, line 29: unnecessary trailing new line

internal/boxcli/update.go
```diff
@@ -0,0 +1,48 @@
...
+	config configFlags
+}
+
+func updateCmd() *cobra.Command {
+	flags := &updateCmdFlags{}
+
+	command := &cobra.Command{
+		Use:   "update [pkg]...",
+		Short: "Updates one, many, or all packages in your devbox",
+		Long: "Updates one, many, or all packages in your devbox. If no " +
```
hezhizhen, line 22: How about one sentence in one line (one pair of quotes)?

internal/nix/input_test.go
```diff
@@ -117,17 +118,24 @@ func (l *lockfile) ProjectDir() string {
 	return l.projectDir
 }
 
-func (lockfile) Resolve(pkg string) (string, error) {
+func (l *lockfile) IsVersionedPackage(pkg string) bool {
```
hezhizhen, line 121: Is it necessary for this given package to be in the lockfile?

internal/nix/input_test.go
```diff
@@ -117,17 +118,24 @@ func (l *lockfile) ProjectDir() string {
...
+	name, version, found := strings.Cut(pkg, "@")
+	return found && name != "" && version != ""
+}
+
+func (lockfile) Resolve(pkg string) (*lock.Package, error) {
 	if strings.Contains(pkg, "path:") {
-		return pkg, nil
+		return &lock.Package{Resolved: pkg}, nil
 	} else if strings.Contains(pkg, "github:") {
```
hezhizhen, line 129: unnecessary `else`

hezhizhen: Or `switch` would be clearer

```go
switch {
  case strings.Contains(pkg, "path:"): return xxx
  case strings.Contains(pkg, "github:"): return xxx
  default: return xxx
}
```

internal/searcher/client.go
gcurtis: I kinda like the succinctness of `Search` and `SearchVersion`.

gcurtis (approved):
LGTM. I think any feedback I had was already mentioned (thanks @hezhizhen!).

internal/impl/devbox.go
```diff
@@ -942,6 +955,23 @@ func (d *Devbox) packages() []string {
...
+	results := []string{}
+	for _, pkg := range d.cfg.Packages {
+		i := nix.InputFromString(pkg, d.lockfile)
+		if i.String() == name || i.CanonicalName() == name {
+			results = append(results, pkg)
+		}
+	}
+	if len(results) > 1 {
+		return "", usererr.New(
+			"found multiple packages with name %s: %s", name, results)
```
savil, line 968: how would the user proceed in this scenario? Can we print additional information about the packages and suggest an alternative command argument?

🤔 this function is kinda generic, so hard to give corrective suggestions here. 

So, maybe the above advice could happen in one of the callers of this function where the user-scenario is more clear.

internal/impl/update.go
```diff
@@ -0,0 +1,46 @@
...
+
+	for _, pkg := range pkgsToUpdate {
+		if !d.lockfile.IsVersionedPackage(pkg) {
+			fmt.Fprintf(d.writer, "Skipping %s because it is not a versioned package\n", pkg)
```
savil, line 24: Not sure. Do we want the full advice to be something like:

"Skipping %s because it is not a versioned package, and so we derive its version from the nixpkgs commit hash in devbox.json. Consider making it a versioned package, or changing the nixpkgs commit hash to get a newer version.\n"

savil: If we do want to give this longer advice, I'd suggest buffering the skipped packageNames, and printing this advice once for all of them at the end of the entire `update` execution.

internal/impl/update.go
```diff
@@ -0,0 +1,46 @@
...
+			continue
+		}
+		existing := d.lockfile.Entry(pkg)
+		newEntry, err := d.lockfile.ForceResolve(pkg)
+		if err != nil {
+			return err
+		}
+		if existing != nil && existing.Version != newEntry.Version {
+			fmt.Fprintf(d.writer, "Updating %s %s -> %s\n", pkg, existing.Version, newEntry.Version)
+			if err := d.removePackagesFromProfile(ctx, []string{pkg}); err != nil {
```
savil, line 34: I'm struggling to follow the logic:
- why are we removing the package from the profile here?
- where are we adding the new version of the package?

internal/impl/update.go
mikeland86: The idea was to print one line per package to give you an idea of status. Don't disagree that we should explain to user how to transition but not sure this is the best place.

Generally I lean to less explaining and just more telling/fixing for example:

```
Warning: your devbox.json contains legacy unversioned packages. 
Please run `devbox config upgrade` to fix.
```

mikeland86: Also:

* `nixpkgs commit hash` is not really something non-nixers will be familiar with.
* I'm not sure I want to encourage people to mess with the hash and instead they should migrate to versioned packages.

internal/impl/update.go
mikeland86: * `ForceResolve` updates the lockfile.
* We are removing the old package from the nix profile.
* `ensurePackagesAreInstalled` installs the new (newly resolved) version

internal/searcher/client.go
mikeland86: good call

internal/nix/input_test.go
mikeland86: Good call. I can't put it in searcher because of a dependency cycle, but I can remove the receiver so it no longer needs to be mocked.

internal/searcher/client.go
mikeland86: agree!

internal/impl/devbox.go
mikeland86: You need to provide exact match. So if you have `go@1.20` and `go@1.19` and you try to update `go` you will get this error.

I don't disagree that the error is a bit terse, but this is a minor edge case, generally users will not install multiple versions of same package. I don't think we need special handling?

I'll change to `"found multiple packages with name %s: %s. Please specify version",`

PR #981, opened 2023-05-04
feat: add retry function

hezhizhen:
## Summary

1. Add `WithRetry` in `cmdutil` to provide general retry function
2. Replace retry logic in `ExecRemoteWithRetry` and `checkActiveVMWithRetries` with the new function


hezhizhen:
@savil @gcurtis @ipince PTAL. Thanks.

hezhizhen:
I think it's not well considered. Close it first.

PR #982, opened 2023-05-04
[update] Silence update notice for devbox-global-shellenv and development-devbox

savil:
## Summary

For `devbox global shellenv` which may be used in the user's shellrc file, its
not ideal to show the update notice everytime the user opens their shell (likely
for a non-devbox purpose).

For devbox engineers, we need not see the notice everytime we execute a devbox command
during development.

## How was it tested?

Added `fmt.Fprintf(os.Stderr, "version was checked")` in `boxcli/root.go` just prior
to calling `vercheck.BuildVersion`. Then did `devbox global shellenv 1> /dev/null`,
and saw that "version was checked" was not printed. As a sanity test, verified
that it was printed for `devbox shellenv 1> /dev/null`.

Also ran `devbox run build` and saw that the notice was not printed since I'd built
the `devbox` originally with the code of this PR.

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"go.jetpack.io/devbox/internal/build"
 
 	"go.jetpack.io/devbox/internal/boxcli/midcobra"
 	"go.jetpack.io/devbox/internal/cloud/openssh/sshshim"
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #982** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/982" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/982?utm_source=stack-comment).

internal/boxcli/root.go
```diff
@@ -34,7 +35,15 @@ func RootCmd() *cobra.Command {
...
+			// Skip CheckVersion for `devbox global shellenv` because users will include that
+			// command in their shellrc files, and we don't want to bother them everytime they
+			// open their terminals.
+			//
+			// Skip CheckVersion for engineers building devbox during development.
+			if !strings.HasPrefix(cmd.CommandPath(), "devbox global shellenv") &&
+				!build.IsDev {
+				vercheck.CheckVersion(cmd.ErrOrStderr())
+			}
 			if flags.quiet {
```
hezhizhen, line 47: Should this block be at the top most?

mohsenari (approved):

internal/boxcli/root.go
savil: good call, will fix

PR #983, opened 2023-05-04
[lockfile] Enable feature flag

mikeland86:
## Summary

TSIA


hezhizhen:
Maybe tests should be updated? @mikeland86

savil (approved):

PR #984, opened 2023-05-05
chore: add GetValueOrDefault for environment variables

hezhizhen:
## Summary

1. Add `GetValueOrDefault` in package `envir`
2. Use the function in `searcher` and `impl`
3. simplify some functions in `initrec` and `plansdk`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -197,10 +197,10 @@ func (d *Devbox) Shell(ctx context.Context) error {
 		return err
 	}
 
-	shellStartTime := os.Getenv(envir.DevboxShellStartTime)
-	if shellStartTime == "" {
-		shellStartTime = telemetry.UnixTimestampFromTime(telemetry.CommandStartTime())
-	}
+	shellStartTime := envir.GetValueOrDefault(
+		envir.DevboxShellStartTime,
+		telemetry.UnixTimestampFromTime(telemetry.CommandStartTime()),
+	)
 
 	opts := []ShellOption{
 		WithHooksFilePath(d.scriptPath(hooksFilename)),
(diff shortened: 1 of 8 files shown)


internal/initrec/recommenders/ruby/ruby.go
```diff
@@ -64,8 +64,6 @@ func parseRubyVersion(gemfile string) string {
 			return matches[2]
 		}
 	}
-	if err := s.Err(); err != nil {
-		return ""
-	}
-	return "" // not found
```
LucilleH, line 70: I wonder why it was written this way before 😂

LucilleH (approved):
Thank you!

internal/impl/devbox.go
```diff
@@ -197,10 +197,10 @@ func (d *Devbox) Shell(ctx context.Context) error {
 		return err
 	}
 
-	shellStartTime := os.Getenv(envir.DevboxShellStartTime)
-	if shellStartTime == "" {
-		shellStartTime = telemetry.UnixTimestampFromTime(telemetry.CommandStartTime())
-	}
+	shellStartTime := envir.GetValueOrDefault(
+		envir.DevboxShellStartTime,
+		telemetry.UnixTimestampFromTime(telemetry.CommandStartTime()),
```
mikeland86, line 202: It's not a big deal in this case because this is cheap to compute, but keep in mind that this logic is a bit different. Specifically `telemetry.UnixTimestampFromTime` will get called every time (even if `DevboxShellStartTime` exists). So when using `envir.GetValueOrDefault` it's important to check that the default is not slow and doesn't have side effects.

internal/initrec/recommenders/ruby/ruby.go
mikeland86: do we want to do anything with the error? (this is a strict improvement so not really a blocker).

mikeland86 (approved):

mikeland86 (approved):

internal/impl/devbox.go
hezhizhen: Oh, I missed it. It's really something that should be taken into consideration.

PR #979, opened 2023-05-03
[Release] remove prerelease step that is no longer being used

savil:
## Summary

We don't seem to be using this step for our "prereleases". The current workflow
is to make a tag with `-dev`, like `0.4.9-dev`.

## How was it tested?

didn't test

PR #985, opened 2023-05-05
[CICD] Fix pre-release job and create new edge release process.

mikeland86:
## Summary

This needed some changes for it to work again. Changes:

* Renamed to alpha
* Create a new tag of the form `ALPHA_TAG=0.0.0-alpha.$(date +%Y-%m-%d)` This tag is semver.
* Run goreleaser --snapshot with this tag as version.
* Create/update github release with same name as tag.

This almost works with new CLI launcher. It has a regex that is not completely accurate so it complains that a version ending in `.2023-01-1` is not valid.

## How was it tested?

Github actions

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -7,13 +7,13 @@ on:
   # Build/Release on demand
   workflow_dispatch:
     inputs:
-      update_nightly:
-        description: "Update nightly?"
+      update_weekly:
+        description: "Update weekly?"
         required: false
         default: false
         type: boolean
   schedule:
-    - cron: "0 0 * * *" # Nightly
+    - cron: "45 8 * * 4" # Weekly pre-release on Thursdays.
   push:
     tags:
       - "*" # Tags that trigger a new release version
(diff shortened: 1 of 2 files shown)


savil:
Could you run it and confirm it generates .tar.gz assets?

John had been unable to use this, so I had sent a PR for removing this entirely: https://github.com/jetpack-io/devbox/pull/979

Also see: https://jetpack-io.slack.com/archives/C040L5DFHFT/p1683134673760769

mikeland86:
@savil PTAL

.github/workflows/cli-release.yml
```diff
@@ -23,23 +23,32 @@ permissions:
...
-    if: ${{ inputs.is_prerelease || github.event.schedule }}
+    if: ${{ inputs.update_weekly || github.event.schedule }}
     steps:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
           fetch-depth: 0 # Needed by goreleaser to browse history.
+      - name: Determine weekly tag
+        # This tag is semver and works with semver.Compare
+        run: echo "WEEKLY_TAG=0.0.0-weekly.$(date +%Y.%U)" >> $GITHUB_ENV
```
loreto, line 44: Could we do `0.0.0-dev.2023-04-03`, which should be:
`$(date +%Y-%m-%d)`

Open to changing `dev` to `rc`, `alpha`, `edge` or something similar ... but trying to avoid promising a particular cadence in the tag.

Rationale:
1. Even though we are choosing to run this weekly, I prefer to not promise weekly in the name. That way if we later want to change to nightly, or something else, we don't need to change how the tags get generated.

2. For the same reason, switching to Month + Day over Week number gives us the flexibility to do multiple releases in a week if we ever need to.

3. `YYYY-MM-DD` over `YYYY.MM.DD` because the first one matches the ISO 8601 standard, and I think we should err on choosing the standard.

savil: nit:
- `edge` would be more suitable for this IMO
- `trunk` is usually specifically referred to as the version of `main` branch
- `alpha` has the downside that it builds the expectation of a `beta` version, etc.

savil (approved):
suggestion: call it `edge` :)

savil:
to be clear: just stating my opinion, totally fine to land this and carry on

PR #986, opened 2023-05-05
[devbox.json] Allowing including plugins explicitly via "include" field.

mikeland86:
## Summary

This adds the ability to include plugins explicitly via:

```
 "include": [
    "plugin:php"
  ]
```

in devbox.json. The idea is to allow including any devbox.json into an existing one. The proposed syntax is:

* path: for local files
* https:// for remote files
* plugin: for built-in plugins

only plugin is implemented in this PR.

Improved existing plugin mechanism by using `lockfile` to decide is `devbox.d` directory needs to be created by keeping track of the plugin version. In the future we could allow upgrading of configs if there is version mismatch.

A few design questions: (cc: @Lagoja @loreto )

* Should include be merged into `packges` instead? We could introspect the include to see if it's a flake or a plugin. This would simplify the devbox.json but it might be a weird way to include other configs. 
* One kinda ugly thing of current design is that the php plugin doesn't actually install PHP. That feels a bit wrong, but the way we've currently designed plugins is that they trigger on packages so not sure how to improve this. One option is to rename `builtin` to something that makes it more explicit that it extends functionality on existing nix packages. Maybe `extend:php`. Another option: In the future I expect includes to contain packages which would be installed as well. So could parametrize the php plugin so it can take the top level php package as an input (but otherwise it could install its own)

Edit:

After chatting with @loreto and @Lagoja we agreed to keep this design with following:
 
* replace "builtin" with "plugin". 
* Including plugins works the same way current plugins. Specifically init hooks, and environment get added to existing environment. No new packages are installed.

## How was it tested?

In php flake example, did `devbox services up`


loreto:
I like the functionality, but I'm unsure of the interface/schema in devbox.json (mainly due to similar concerns to the product questions you are raising). Worth chatting in person to flesh the interface out?

Lagoja:
> Should include be merged into packges instead? We could introspect the include to see if it's a flake or a plugin. This would simplify the devbox.json but it might be a weird way to include other configs.

I lean towards not including this in packages for now. While it might simplify the config, plugins and other `devbox.json` configs feel like a separate thing from packages (which ultimately tie back to a Nix package, either from a Flake or the nixpkgs repo). Having it in `includes` is a bit clearer.

> One kinda ugly thing of current design is that the php plugin doesn't actually install PHP. That feels a bit wrong, but the way we've currently designed plugins is that they trigger on packages so not sure how to improve this. One option is to rename builtin to something that makes it more explicit that it extends functionality on existing nix packages. Maybe extend:php. Another option: In the future I expect includes to contain packages which would be installed as well. So could parametrize the php plugin so it can take the top level php package as an input (but otherwise it could install its own)

I think we might want to rename the plugin from `php` to something else? Maybe being explicit and calling it `builtin:php-plugin` or `builtin:php-config` might make this clearer. Based on the answer above, we should make sure these are distinct from packages

Happy to chat about this in person to flesh out.

savil:
suggestion: avoid calling it `builtin`, because that usually implies a concept that ships with `devbox`. For example, in programming languages builtin functions ship with the language runtime. In our case, we do envision a world where plugins could be written by users and included, and those would not be "builtin".

Alternatives: "plugin: php", or "extension: php-plugin".

mikeland86:
> suggestion: avoid calling it builtin, because that usually implies a concept that ships with `devbox`

Yep, same page. This is precisely why I called them `builtin`. `builtin` can only be used with plugins in our plugin directory which are compiled into devbox. For any other plugins, they would be referred to as `path` or `https` URLs that would point to a relative local path or a remote address.

internal/impl/config.go
```diff
@@ -40,6 +40,13 @@ type Config struct {
 
 	// Nixpkgs specifies the repository to pull packages from
 	Nixpkgs NixpkgsConfig `json:"nixpkgs,omitempty"`
+
+	// Reserved to allow including other config files. Proposed format is:
+	// path: for local files
+	// https:// for remote files
+	// plugin: for built-in plugins
+	// This is similar to nix inputs
```
savil, line 48: Mind elaborating on what you mean by "similar to nix inputs": in what way?

internal/impl/packages.go
```diff
@@ -59,7 +59,6 @@ func (d *Devbox) Add(ctx context.Context, pkgs ...string) error {
 		return err
 	}
 
-	d.pluginManager.ApplyOptions(plugin.WithAddMode())
```
savil, line 62: when does this get applied now, in the flow for adding a package?

internal/plugin/plugin.go
```diff
@@ -207,9 +241,10 @@ func createSymlink(root, filePath string) error {
 	return errors.WithStack(os.Symlink(filePath, newname))
 }
 
-func (m *Manager) shouldCreateFile(filePath string) bool {
-	// Only create devboxDir files in add mode.
-	if strings.Contains(filePath, devboxDirName) && !m.addMode {
+func (m *Manager) shouldCreateFile(pkg *lock.Package, filePath string) bool {
+	// Only create files in devboxDir is they are not in the lockfile
```
savil, line 245: typo: if

savil: 🤔 for older plugins (i.e. not in the lockfile), this code is re-creating their `devbox.d` files.

- I thought users were encouraged to commit them to source-control and possibly change them.
- so, shouldn't we avoid re-generating these files for existing devbox projects

internal/impl/generate.go
```diff
@@ -61,7 +61,16 @@ func (d *Devbox) generateShellFiles() error {
...
+		if err := d.pluginManager.Create(d.writer, pkg, d.projectDir); err != nil {
+			return err
+		}
+	}
+
+	for _, includes := range d.cfg.Include {
+		if err := d.lockfile.Add(includes); err != nil {
+			return err
+		}
+		if err := d.pluginManager.Include(d.writer, includes, d.projectDir); err != nil {
```
savil, line 73: It'll be good for some future PR  to make the plugin manager life-cycle more clear. 

Suggestion:
- `pluginManager.New(lockfile)` function that does `applyOptions`
- `pluginManager.Generate(packages, includes)`
   - this can do `lockfile.Add(includes)` internally

internal/impl/generate.go
```diff
@@ -61,7 +61,16 @@ func (d *Devbox) generateShellFiles() error {
 	}
 
 	for _, pkg := range d.packages() {
-		if err := d.pluginManager.CreateFilesAndShowReadme(d.writer, pkg, d.projectDir); err != nil {
+		if err := d.pluginManager.Create(d.writer, pkg, d.projectDir); err != nil {
+			return err
+		}
+	}
+
+	for _, includes := range d.cfg.Include {
```
savil, line 69: nit: shouldn't this be singular `include`?

savil (commented):
I have one concern with the implementation on `plugin/plugin.go:245`. I think it may be backwards incompatible: it may lead to undesirable overwriting of files for existing devbox-projects. WDYT?

-----
How does a user discover which plugins are available? 

- (short term) Could we have a docs page that we reference in the help notes?
- (long term) implement `devbox plugins ls`?

> Including plugins works the same way current plugins. Specifically init hooks, and environment get added to existing environment. No new packages are installed.

what is the behavior a user would experience if they have `includes: [ plugins: php ]` but no package `php`?

internal/impl/packages.go
savil: Oh I understand this one. The AddMode is replaced with Lockfile.

internal/impl/config.go
mikeland86: nix flake inputs are urls of this format (path, https, github, etc) so we are using the same format.

internal/plugin/plugin.go
mikeland86: I'll make sure we don't overwrite them, but yeah this will cause a one time re-creation if they previously deleted/moved them. I think that's on ok tradeoff in order to migrate to a better system.

internal/impl/generate.go
mikeland86: `NewManager` already applies options. The reason we call it separately is because of a circular dependency (see `impl.Open` function)

Agree with unifying `Create` and `Includes`

Not sure about calling `lockfile.Add(includes)` internally. That feels like logic that is unrelated to plugins. Insterad, plugins should be saved to lock file and then we can pass in resolved packages directly to plugin manager, which will mean we no longer need the lockfile in there.

internal/impl/config.go
savil: I see, got it. Perhaps "This **format** is similar to nix inputs"

internal/impl/generate.go
savil: actually, in addition, I'm being thrown off by `d.cfg.Include` being singular....should that be plural? 

Relatedly, in the `devbox.json`, should the field be `includes`?

internal/impl/devbox.go
```diff
@@ -125,6 +125,7 @@ func Open(path string, writer io.Writer) (*Devbox, error) {
 	if err != nil {
 		return nil, err
 	}
+	box.pluginManager.ApplyOptions(plugin.WithLockfile(lock))
```
savil, line 128: I'd suggest doing:

1. remove `pluginmanager: plugin.NewManager(),` above
2. make this `box.pluginManager = plugin.NewManager(plugin.WithLockfile(lock))` 
3. make `applyOptions` an internal function, 

(3) would make NewManager a more genuine constructor. That would make it easier to reason about that all the pluginManager's fields are ready and methods can be applied.

internal/impl/generate.go
savil: > NewManager already applies options. The reason we call it separately is because of a circular dependency (see impl.Open function)

Oh I see. I left a comment above about improving that so its easier to reason about.

> Not sure about calling lockfile.Add(includes) internally. That feels like logic that is unrelated to plugins. Insterad, plugins should be saved to lock file and then we can pass in resolved packages directly to plugin manager, which will mean we no longer need the lockfile in there.

Yup, you're right. Lets not call lockfile.Add from within pluginManager. 👍🏾

internal/plugin/plugin.go
savil: Got it

Yeah, I was concerned about the overwriting scenario where users may have made changes.

internal/impl/generate.go
mikeland86: "include" is a verb, not a noun so I think it should be singular (e.g. tsconfig.json also uses "include" and not "includes"). @Lagoja curious if you have opinion.

For the range variable, I'll change it, maybe to `included`?

mikeland86:
> How does a user discover which plugins are available?

> (short term) Could we have a docs page that we reference in the help notes?

Built in plugin functionality is already documented. Plugins are not new, they were just only included by package whereas this allows them to be included explicitly.

> (long term) implement devbox plugins ls?

Maybe. There's also a world where plugins are as ubiquitous as packages, and most of them are third party. So a search might be more appropriate than an ls.

> what is the behavior a user would experience if they have includes: [ plugins: php ] but no package php?

The service will fail. For other plugins the init hooks will fail. I think this is OK. The `match` mechanism in built in plugins is a bit too magical, and I'm not sure we should expand it to check if you have correct packages installed. Not sure.

internal/impl/generate.go
savil: oh its a verb! I see i see. 
Okay, we can leave as is for devbox.json, and yeah I like `included` for the range var 👍🏾

savil (approved):

PR #990, opened 2023-05-09
[fileutil] move paths to fileutil

hezhizhen:
## Summary

1. Move paths from different packages to `fileutil`
2. Add default values as comments for every path
3. Distinguish file paths and dir paths with different suffixes
4. Add functions to ensure dirs exist

### DIFF
--- a/internal/cloud/mutagen/install.go
+++ b/internal/cloud/mutagen/install.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 
 	"github.com/cavaliergopher/grab/v3"
+
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/fileutil"
 )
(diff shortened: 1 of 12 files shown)


mohsenari:
@hezhizhen thanks for this. I'll review it thoroughly once the failed test and linter errors gets addressed.

mikeland86:
@hezhizhen 

> make sure all the directories are created if they are not

I think that's the sort of thing that could go in a util (e.g. EnsureFile, EnsureDir) 

>  centralize all these paths (to let anyone know which paths are used and what they are, etc)

For stuff that is already shared (e.g. the devbox config, state, cache paths) that could also go in a single place. (either in the xdg package or a wrapper on it that is more devbox specific).

> And I'm available on the devbox discord as littlepoint. Feel free to DM or mention me.

Will reach out there!

hezhizhen:
close it first

PR #991, opened 2023-05-09
[easy][bugs] Fix 2 package input bugs

mikeland86:
## Summary

Fixes 2 bugs:

* Avoid triple slashes that can make comparing urls break when adding flakes (this was affecting correct removal of profile packages and also detecting that they were already installed)
* Fix package validation for non-versioned packages (They should use nix search instead of new search service)

## How was it tested?

* `devbox add sqlite3`  showed error and was not added to devbox.json
* In examples/flake/php added and removed packages and confirmed profile remained in sync.

### DIFF
--- a/internal/nix/input.go
+++ b/internal/nix/input.go
@@ -27,8 +27,9 @@ type Input struct {
 func InputFromString(s string, l lock.Locker) *Input {
 	u, _ := url.Parse(s)
 	if u.Path == "" && u.Opaque != "" && u.Scheme == "path" {
-		u.Path = filepath.Join(l.ProjectDir(), u.Opaque)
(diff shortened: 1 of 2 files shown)


internal/nix/input.go
```diff
@@ -142,13 +143,12 @@ func (i *Input) PackageAttributePath() (string, error) {
 		)
 	}
 
-	return "", usererr.New("Flake \"%s\" was not found", i.String())
+	return "", usererr.New("Package \"%s\" was not found", i.String())
```
mohsenari, line 146: Can you update the message at line 140 as well?

mohsenari (approved):
approving to unblock, see small comment.

PR #993, opened 2023-05-10
chore: add more envs and functions

hezhizhen:
## Summary

1. Add more envs: `DevboxRunCmd`, `DevboxOGPathPrefix`
2. Add function to set env `devboxShellEnabled` and get `PATH` values
3. use existing envs and functions

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -196,7 +196,7 @@ func (d *Devbox) Shell(ctx context.Context) error {
 		return err
 	}
 	// Used to determine whether we're inside a shell (e.g. to prevent shell inception)
-	envs[envir.DevboxShellEnabled] = "1"
+	envir.EnableDevboxShell(envs)
 
 	if err := wrapnix.CreateWrappers(ctx, d); err != nil {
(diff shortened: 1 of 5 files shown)


internal/impl/devbox.go
```diff
@@ -714,7 +714,7 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 		}
 		env[key] = val
 	}
-	currentEnvPath := env["PATH"]
+	currentEnvPath := envir.GetPath(env)
```
savil, line 717: I don't know if `GetPath` is adding much value here, over `env[envir.Path]`. 

`env[envir.Path]` feels easier to read since one can readily understand what it is doing.

What's the motivation?

hezhizhen: Yeah, it seems a little over-wrapping.

I want to limit the adding/getting/removing/updating one specific environment variable operations in functions so that callers use functions instead of envs.

hezhizhen:
I think it's not well considered. Close it first.

PR #994, opened 2023-05-11
[services] Fix services when using versioned packages

mikeland86:
## Summary

This addresses related 3 bugs:

* "versioned" packages where not triggering plugins/services. This was solved by updating plugin packages to use `nix.Input` and then using `CanonicalName()` as the triggering name (In short, this means we use `python` instead of `python@3`)
* Updated plugins to recognize canonical names in `match` field. (e.g. `apache` plugin should trigger for `apache` and `apacheHttpd`). This was implemented by augmenting the regular expression. Another approach would have been to save the attribute path returned by search and use that instead. Since it only affects a very limited number of plugins and we don't really want to expand usage of `match` too much, I think this is an OK tradeoff.
* minor bug: `PackageAttributePath()` in `input.go` was causing legacy packages that don't exist to get resolved (e.g. when doing `devbox add foobar`). This wasn't breaking anything because no packages get added to `devbox.json` but it would leave bad data in lockfile.

## How was it tested?

`devbox add apach@2`
`devbox add php`
`devbox services ls`
`devbox add foobar` shows error and doesn't add to lockfile.


internal/plugin/plugin.go
```diff
@@ -136,7 +142,7 @@ func (m *Manager) CreateFilesAndShowReadme(w io.Writer, pkg, projectDir string)
 // TODO: We should associate the env variables with the individual plugin
 // binaries via wrappers instead of adding to the environment everywhere.
 func Env(
-	pkgs []string,
+	pkgs []*nix.Input,
```
savil, line 145: should line 165 also have `pkg nix.Input`?

savil (approved):

savil:
> "versioned" packages where not triggering plugins/services.

could we write a test for this?

savil:
actually, seems like we could add tests for all 3 bugs here....maybe in a follow up

mikeland86:
Will add test in follow up

PR #996, opened 2023-05-11
[versioned] Deal with packages that don't build on system

mikeland86:
## Summary

Fixes a few bugs:

* Remove panic
* If package can't be found for current system, search on all systems so we can return better error message.
* Don't automatically write to disk when resolving lockfile packages. That way if there's an error we don't leave the lockfile in a bad state.

### package that exists but not for my system

```bash
➜  devbox git:(landau/fix-panic) ✗ devbox add ripgrep@12
There was an error installing nix packages: ripgrep@12. Packages were not added to devbox.json

Error: Package "ripgrep@12" was found, but we're unable to build it for your system. You may need to choose another version or write a custom flake.
```

### package that doesn't exist

```bash
➜  devbox git:(landau/fix-panic) ✗ devbox add foobar1@8

Error: foobar1@8: package not found

```

### regular versioned install:

```bash
devbox add php@8
```

### Regular install:

```bash
devbox add php
```

### Manual add

* Edited devbox.json and added `php@8` and confirmed lockfile was updated after `devbox run`

### DIFF
--- a/internal/nix/search.go
+++ b/internal/nix/search.go
@@ -67,16 +67,23 @@ func parseSearchResults(data []byte) map[string]*Info {
 	return infos
 }
 
-// pkgExistsForAnySystem is a bit slow. Only use it if there's already been
-// an error and we want to provide a better error message.
+// pkgExistsForAnySystem is a bit slow (~600ms). Only use it if there's already
+// been an error and we want to provide a better error message.
 func pkgExistsForAnySystem(pkg string) bool {
 	systems := []string{
+		// Check most common systems first.
+		"x86_64-linux",
+		"x86_64-darwin",
 		"aarch64-linux",
+		"aarch64-darwin",
+
+		"armv5tel-linux",
 		"armv6l-linux",
 		"armv7l-linux",
 		"i686-linux",
-		"x86_64-darwin",
-		"x86_64-linux",
+		"mipsel-linux",
+		"powerpc64le-linux",
+		"riscv64-linux",
 	}
 	for _, system := range systems {
 		if len(searchSystem(pkg, system)) > 0 {
--- a/internal/nix/search.go
+++ b/internal/nix/search.go
@@ -1,0 +1,103 @@
+package nix
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/pkg/errors"
+	"github.com/samber/lo"
+	"go.jetpack.io/devbox/internal/debug"
+	"go.jetpack.io/devbox/internal/lock"
+)
+
+var ErrPackageNotFound = errors.New("package not found")
+var ErrPackageNotInstalled = errors.New("package not installed")
+
+func PkgExists(pkg string, lock *lock.File) (bool, error) {
+	return InputFromString(pkg, lock).validateExists()
+}
+
+type Info struct {
+	// attribute key is different in flakes vs legacy so we should only use it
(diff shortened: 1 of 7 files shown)


internal/nix/search.go
```diff
@@ -0,0 +1,103 @@
...
+			Version:      result["version"].(string),
+		}
+
+	}
+	return infos
+}
+
+// pkgExistsForAnySystem is a bit slow. Only use it if there's already been
+// an error and we want to provide a better error message.
+func pkgExistsForAnySystem(pkg string) bool {
```
mikeland86, line 72: By a bit slow I mean ~1 second if nixpkgs is already downloaded. So not super slow. Keep in mind this is only used if we find ourselves in a bad state so it's not called in any happy paths. It will usually get called only when doing

`devbox add somePacakgeThatExistsButNotOnMySystem@<version>`

internal/nix/search.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (i *Info) String() string {
+	return fmt.Sprintf("%s-%s", i.PName, i.Version)
+}
+
+func PkgInfo(nixpkgsCommit, pkg string) *Info {
+	exactPackage := fmt.Sprintf("%s#%s", FlakeNixpkgs(nixpkgsCommit), pkg)
+	if nixpkgsCommit == "" {
+		exactPackage = fmt.Sprintf("nixpkgs#%s", pkg)
```
savil, line 37: should we modify `FlakeNixpkgs` to:
```
func FlakeNixpkgs(commit string) string {
  if commit == "" {
     return "nixpkgs"
  }
  return "nixpkgs/" + commit
}
```
?

internal/nix/search.go
```diff
@@ -0,0 +1,103 @@
...
+
+func search(url string) map[string]*Info {
+	return searchSystem(url, "")
+}
+
+func parseSearchResults(data []byte) map[string]*Info {
+	var results map[string]map[string]any
+	err := json.Unmarshal(data, &results)
+	if err != nil {
+		panic(err)
```
savil, line 56: Let's change this `panic` to `error` ?

internal/nix/search.go
```diff
@@ -0,0 +1,103 @@
...
+// pkgExistsForAnySystem is a bit slow. Only use it if there's already been
+// an error and we want to provide a better error message.
+func pkgExistsForAnySystem(pkg string) bool {
+	systems := []string{
+		"aarch64-linux",
+		"armv6l-linux",
+		"armv7l-linux",
+		"i686-linux",
+		"x86_64-darwin",
+		"x86_64-linux",
```
savil, line 79: I'm surprised `aarch64-darwin` is not included!

savil (approved):

internal/nix/search.go
mikeland86: that was a mistake, though in practice if x86 darwin is there, so is aarch64

internal/nix/search.go
mikeland86: @savil for sure. This is not new code though, I just moved it to other file. (and will likely get removed/changed very soon)

internal/nix/search.go
mikeland86: similar to above comment. I agree, but not introduced in this PR. Propagating an error is gonna cause a lot more changes, so I would prefer to push that off to a follow up.

PR #997, opened 2023-05-11
[info] Ensure devbox info works with versioned packages

mikeland86:
## Summary

`devbox info php@8` now works correctly

## How was it tested?

```bash
devbox info php
devbox info php@8
devbox info php@latest
```


savil (approved):
could we add scenarios to `testscripts/info.txt`?

PR #998, opened 2023-05-11
[versioned] Add @latest to unversioned packages

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/997

This adds `@latest` whenever an unversioned non-flake package is added.

This PR will be blocked until the php and haskell planners are updated to work with versioned packages.

## How was it tested?

```bash
devbox add apache php curl
```

inspected devbox.json and lockfile.


internal/impl/packages.go
```diff
@@ -33,9 +33,17 @@ func (d *Devbox) Add(ctx context.Context, pkgs ...string) error {
...
+	// Add to Packages of the config only if it's not already there. We do this
+	// before addin @latest to ensure we don't accidentally add a package that
+	// is already in the config.
+	for _, pkg := range nix.InputsFromStrings(lo.Uniq(pkgs), d.lockfile) {
+		versionedPackages = append(versionedPackages, pkg.Versioned())
+		if !slices.Contains(d.cfg.Packages, pkg.Raw) {
+			d.cfg.Packages = append(d.cfg.Packages, pkg.Versioned())
+		}
+	}
+	pkgs = versionedPackages
```
savil, line 46: Going back-and-forth between `pkg string` and `input nix.Input` in this function is pretty confusing.

1. can the next for-loop for PkgExists also operate on `nix.Input`?
2. can the for-loop on line 78 also operate on `nix.Input` via `versionedPackages` ?

internal/impl/packages.go
```diff
@@ -33,9 +33,17 @@ func (d *Devbox) Add(ctx context.Context, pkgs ...string) error {
...
 	defer task.End()
 
-	pkgs = lo.Uniq(pkgs)
-
-	original := d.cfg.Packages
+	versionedPackages := []string{}
+	// Add to Packages of the config only if it's not already there. We do this
+	// before addin @latest to ensure we don't accidentally add a package that
+	// is already in the config.
+	for _, pkg := range nix.InputsFromStrings(lo.Uniq(pkgs), d.lockfile) {
```
savil, line 40: can we call `pkg` as `input` here? In the context of this function since the `pkgs ...string` param is there, then I'll always assume that `pkg` is of type `string`.

savil (approved):

PR #1000, opened 2023-05-12
[planner->plugin] Turn php planner into a plugin

mikeland86:
## Summary

This PR turns our PHP planner into a plugin. Our plugins a are actually almost able to do this with some minimal changes.

Re-used existing functionality:

* Create `flake.nix` file in the php plugin
* packages (inputs) can be local flakes

Changes:

* Pass `Packages`, `System`, and `URLForInput` to plugin file templates. This allows plugins to build and customize some interesting flakes.
* Add `packages` field to plugins. We wanted to do this for a while. In this case it simply points to the flake created by the plugin. These packages are added to the devbox environment.

Semi-related changes:

* Use `--recreate-lock-file` every time we `print-dev-env`. The reason this is needed is because input flakes might change se we need to frequently update. It doesn't hurt performance much because we cache print-dev-env. This means any unlocked package will update frequently. I think that's fine because this lockfile is not shared so there's no expectation of stability.

cc: @Lagoja @savil 

## How was it tested?

```bash
devbox add php phpExtensions.ds
devbox run 'php -m | grep ds'
```


Lagoja:
Is the Flake in this example a Go template? If so, to move the Haskell planner we would need to create a similar flake, and use the builtins to match the Haskell packages in the `.Packages` list?

mikeland86:
@Lagoja yeah it's a template. It would look pretty similar, but probably with more complicated logic in the `let` clause.

It could output a default with what today is called `haskell-pkg` in the flake and can be included in the same exact way:

```json
"packages": [
    "path:{{ .Virtenv }}"
  ],
```

In the flake:


```
packages.{{ .System }} = {
  default = nixpkgs.legacyPackages.{{ .System }}.${haskellVersion}.ghcWithPackages (ps: with ps; [ 
    map (pkg: ps.${pkg}) haskellPackages
  ]);
};
```

`haskellVersion` would be set to the versioned or default (as in planner)
`haskellPackages` would be built in a similar way to php flake (but following the haskell package regex)

internal/cuecfg/hash.go
```diff
@@ -13,7 +13,9 @@ import (
 
 func FileHash(path string) (string, error) {
 	data, err := os.ReadFile(path)
-	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+	if errors.Is(err, fs.ErrNotExist) {
+		return "", nil
+	} else if err != nil {
```
gcurtis, line 18:
```suggestion
	}
	if err != nil {
```

internal/impl/devbox.go
```diff
@@ -125,7 +125,10 @@ func Open(path string, writer io.Writer) (*Devbox, error) {
 	if err != nil {
 		return nil, err
 	}
-	box.pluginManager.ApplyOptions(plugin.WithLockfile(lock))
+	box.pluginManager.ApplyOptions(
+		plugin.WithDevbox(box),
+		plugin.WithLockfile(lock),
```
gcurtis, line 130: Is providing a Devbox and/or lockfile optional?

internal/impl/devbox_test.go
```diff
@@ -78,22 +78,25 @@ func (n *testNix) PrintDevEnv(ctx context.Context, args *nix.PrintDevEnvArgs) (*
 }
 
 func TestComputeNixEnv(t *testing.T) {
-	d := &Devbox{
-		cfg: &Config{},
-		nix: &testNix{},
-	}
+	path, _ := os.MkdirTemp("", "TestComputeNixEnv")
```
gcurtis, line 81: `t.TempDir()`

internal/impl/devbox_test.go
```diff
@@ -78,22 +78,25 @@ func (n *testNix) PrintDevEnv(ctx context.Context, args *nix.PrintDevEnvArgs) (*
...
 	assert.NotNil(t, env, "computeNixEnv should return a valid env")
 }
 
 func TestComputeNixPathIsIdempotent(t *testing.T) {
-	devbox := &Devbox{
-		cfg:        &Config{},
-		nix:        &testNix{"/tmp/my/path"},
-		projectDir: "/tmp/TestComputeNixPathIsIdempotent",
-	}
+	dir, _ := os.MkdirTemp("", "TestComputeNixPathIsIdempotent")
```
gcurtis, line 94: `t.TempDir()`

internal/impl/devbox_test.go
```diff
@@ -114,11 +117,12 @@ func TestComputeNixPathIsIdempotent(t *testing.T) {
 }
 
 func TestComputeNixPathWhenRemoving(t *testing.T) {
-	devbox := &Devbox{
-		cfg:        &Config{},
-		nix:        &testNix{"/tmp/my/path"},
-		projectDir: "/tmp/TestComputeNixPathWhenRemoving",
-	}
+	dir, _ := os.MkdirTemp("", "TestComputeNixPathWhenRemoving")
```
gcurtis, line 120: `t.TempDir()`

internal/impl/devbox_test.go
```diff
@@ -78,22 +78,25 @@ func (n *testNix) PrintDevEnv(ctx context.Context, args *nix.PrintDevEnvArgs) (*
 }
 
 func TestComputeNixEnv(t *testing.T) {
-	d := &Devbox{
-		cfg: &Config{},
-		nix: &testNix{},
-	}
+	path, _ := os.MkdirTemp("", "TestComputeNixEnv")
+	_, err := InitConfig(path, os.Stdout)
+	assert.NoError(t, err, "InitConfig should not fail")
```
gcurtis, line 83: `assert` doesn't abort the test. I think you need `require`.

internal/plugin/plugin.go
```diff
@@ -246,15 +266,20 @@ func createSymlink(root, filePath string) error {
...
 	// Only create files in devboxDir if they are not in the lockfile
 	pluginInstalled := pkg != nil && pkg.PluginVersion != ""
 	if strings.Contains(filePath, devboxDirName) && pluginInstalled {
 		return false
 	}
 
 	// Hidden .devbox files are always replaceable, so ok to recreate
-	if strings.Contains(filePath, devboxHiddenDirName) {
+	if strings.Contains(filePath, devboxHiddenDirName) ||
+		strings.HasPrefix(filePath, virtenvPath) {
```
gcurtis, line 282: Will these always be relative paths for `HasPrefix`?

gcurtis (approved):

internal/impl/devbox.go
mikeland86: @gcurtis not really, but due to circular dependency needs to be added after construction

internal/plugin/plugin.go
mikeland86: Both `filePath` and `virtenvPath` should be absolute.

savil (commented):
This is cool!

PR #1007, opened 2023-05-16
[version check] Avoid re-printing update notice in child devbox commands

savil:
## Summary

Some devbox commands like `devbox shell` may invoke other devbox commands. For example,
`devbox shell` generates a shellrc file that calls `devbox log`. This results in 
multiple "update version" notices being printed for the same devbox command invocation.

To avoid this, we set an env-var after printing the notice, and unset that env-var
when this devbox command will exit. Note, I keep this env-var inside vercheck
to maintain encapsulation, instead of putting it in envir package.

I also pushed some of the skipping checks from `root.go` inside `vercheck` for
(1) better encapsulation (2) adding test cases for them.

## How was it tested?

`go test ./internal/vercheck/...`

For manual testing:
1. `set -gx DEVBOX_LATEST_VERSION 0.5.0`
2. modify binary to `vercheck` to have `currentDevboxVersion = 0.4.8` and `isDevBuild = false`.
3. run `devbox shell`
BEFORE: multiple update notices would print
AFTER: one notice was printed

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -41,9 +41,6 @@ func RootCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return cmd.Help()
 		},
-		PersistentPostRun: func(cmd *cobra.Command, args []string) {
-			vercheck.ClearCheckEnvVar()
-		},
 		SilenceErrors: true,
 		SilenceUsage:  true,
 	}
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -11,8 +11,6 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
-	"go.jetpack.io/devbox/internal/build"
-
 	"go.jetpack.io/devbox/internal/boxcli/midcobra"
 	"go.jetpack.io/devbox/internal/cloud/openssh/sshshim"
 	"go.jetpack.io/devbox/internal/debug"
@@ -38,19 +36,13 @@ func RootCmd() *cobra.Command {
 			if flags.quiet {
 				cmd.SetErr(io.Discard)
 			}
-
-			// Skip CheckVersion for `devbox global shellenv` because users will include that
-			// command in their shellrc files, and we don't want to bother them everytime they
-			// open their terminals.
-			//
-			// Skip CheckVersion for engineers building devbox during development.
-			if !strings.HasPrefix(cmd.CommandPath(), "devbox global shellenv") &&
-				!build.IsDev {
-				vercheck.CheckVersion(cmd.ErrOrStderr())
-			}
+			vercheck.CheckVersion(cmd.ErrOrStderr(), cmd.CommandPath())
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return cmd.Help()
+		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+			vercheck.ClearCheckEnvVar()
 		},
 		SilenceErrors: true,
 		SilenceUsage:  true,
--- a/internal/vercheck/vercheck.go
+++ b/internal/vercheck/vercheck.go
@@ -17,7 +17,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
-	"go.jetpack.io/devbox/internal/build"
 	"go.jetpack.io/devbox/internal/cmdutil"
 	"go.jetpack.io/devbox/internal/envir"
 	"go.jetpack.io/devbox/internal/ux"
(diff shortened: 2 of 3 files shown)


savil:
* main
  * **PR #1007** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1007" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1007?utm_source=stack-comment).

internal/vercheck/vercheck.go
```diff
@@ -15,32 +15,60 @@ import (
...
 )
 
 // Keep this in-sync with latest version in launch.sh.
 // If this version is newer than the version in launch.sh, we'll print a notice.
 const expectedLauncherVersion = "v0.2.0"
 
+// versionCheckEnvName determines whether the version check has already occurred.
+// We set this env-var so that this devbox command invoking other devbox commands
+// do not re-run the version check and print the notice again.
+const versionCheckEnvName = "DEVBOX_VERSION_CHECK"
```
savil, line 35: do we want __DEVBOX_VERSION_CHECK?

savil, line 35: I'll also rename this variable to `envName`.

internal/boxcli/root.go
```diff
@@ -38,20 +36,14 @@ func RootCmd() *cobra.Command {
...
-				vercheck.CheckVersion(cmd.ErrOrStderr())
-			}
+			vercheck.CheckVersion(cmd.ErrOrStderr(), cmd.CommandPath())
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return cmd.Help()
 		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+			vercheck.ClearCheckEnvVar()
+		},
```
mikeland86, line 46: This is not needed. Environment variables set during the execution do not leak to the environment. (in fact that is no way to update the current shell environment which is why the user needs to source shellenv)

internal/vercheck/vercheck.go
mikeland86: I've been inconsistent, but I think starting with `__` is better.

I try to follow:

`DEVBOX_[name]` when it is a variable that a user might want to set/change.

`__DEVBOX_[name]` for automated values that the user should never touch. (i.e. private)

internal/vercheck/vercheck.go
```diff
@@ -56,6 +84,13 @@ func CheckVersion(w io.Writer) {
 		// TODO: use ux.FNotice
 		color.New(color.FgYellow).Fprintf(w, devboxNotice)
 	}
+
+	os.Setenv(versionCheckEnvName, "1")
+}
+
+// ClearCheckEnvVar will unset the env-var that indicates a version check was done.
+func ClearCheckEnvVar() {
+	os.Unsetenv(versionCheckEnvName)
```
mikeland86, line 93: See comment above. Not needed.

internal/vercheck/vercheck.go
```diff
@@ -15,32 +15,60 @@ import (
...
 var currentDevboxVersion = build.Version
 
+// isDevBuild determines whether this CLI binary was built during development, or published
+// as a release.
+// We use this variable so we can mock it in tests.
+var isDevBuild = build.IsDev
+
+var commandSkipList = []string{
+	"devbox global shellenv",
+}
```
mikeland86, line 48: Do we also want to skip `devbox shellenv` ? 

Or should we flip this around and only show warning for `add` `rm` `init` `run` shell` ? 

I don't feel strongly so no action needed. Just throwing it out there.

internal/vercheck/vercheck.go
```diff
@@ -15,32 +15,60 @@ import (
...
+const versionCheckEnvName = "DEVBOX_VERSION_CHECK"
+
 // currentDevboxVersion is the version of the devbox CLI binary that is currently running.
 // We use this variable so that we can mock it in tests.
 var currentDevboxVersion = build.Version
 
+// isDevBuild determines whether this CLI binary was built during development, or published
+// as a release.
+// We use this variable so we can mock it in tests.
+var isDevBuild = build.IsDev
```
mikeland86, line 44: One thing we could do is make all the build values unexported (I think ld flags work with unexported values) and expose everything via a public struct. That way we can just mock the struct.

mikeland86 (approved):

mikeland86 (approved):

mikeland86 (approved):

internal/vercheck/vercheck.go
savil: yeah, this is the central question I was struggling with. I first implemented an `allowList` and then switched to this env-var approach. 

🤔 perhaps we should have both allowList _and_ this env-var approach to minimize the places we show this update notice?

internal/boxcli/root.go
savil: oh good point

internal/vercheck/vercheck.go
mikeland86: Yeah, I would not remove the env var even if we switch to allowList, it's nice to have it since we can't always prefix what will be called in a command.

PR #1009, opened 2023-05-16
[haskell] Move planner to plugin

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1000

This moves Haskell planner to plugin framework. Note that this plugin is less strict than planner. Specifically, it adds packages even if versions don't match. So it's not really able to do multi-version Haskell setups. @Lagoja not sure what you think about that. Making it version specific may be hard, especially once we move towards all versioned packages.

For example:

```
ghc@latest
haskellPackages.stack@latest
```

should work even if there's a mismatch. In fact, the PHP plugin ignores the version of the extensions and just uses the version of php (This is needed because the versions are inconsistent, some extensions mimic php and others have their own version). I think it's OK for the Haskell planner to do something similar?
 
TODO: Will remove planner framework in follow up.

## How was it tested?

`devbox run run_test -c examples/development/haskell/ `

Also tried example using `haskell.compiler.ghc92`


Lagoja:
I think ignoring the package versions is the right way to go. Totally possible my original planner was too strict 🙃

plugins/haskell.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "haskell",
+  "version": "0.0.1",
+  "match": "^(ghc|haskell\\.compiler\\.(.*))$",
+  "readme": "Haskell plugin",
+  "packages": [
+    "path:{{ .Virtenv }}"
```
savil, line 7: I'm struggling to interpret this line. How should I read it in my head?

plugins/haskell.json
```diff
@@ -0,0 +1,12 @@
...
+  ],
+  "create_files": {
+    "{{ .Virtenv }}/flake.nix": "haskell/flake.nix"
```
savil, line 10: this map is also confusing me a bit: are there some docs on how to interpret the plugin format?

savil (approved):

plugins/haskell.json
mikeland86: Packages are allowed to be flakes. For example

`path:relative/path/to/my/flake`

where `relative/path/to/my/flake` has `flake.nix` in it.

This plugin creates a `{{ .Virtenv }}/flake.nix`

plugins/haskell.json
mikeland86: I'll write some docs in follow up

PR #1012, opened 2023-05-17
[planners] Remove unused planner code

mikeland86:
## Summary

Clean up old code. TSIA.

@savil what is `shell.nix.tmpl` for? I remember you removed it/reverted it.

## How was it tested?

Unit tests, it builds.

Edit: confirmed that `nix-shell .devbox/gen/shell.nix` works.


savil:
The `shell.nix` is needed for users who have the old `.envrc` that relied on it. 

 They may not yet have run `devbox generate direnv` again to get the newer slimmed-down `.envrc`.

PR #1013, opened 2023-05-17
[devbox.json] Don't create nixpkgs commit hash

mikeland86:
## Summary

This stops adding nixpkgs commit hash to devbox.json. User can still add their own. In the future I can imagine us deprecating and eventually removing this field.

e.g. something like this:

```bash
Warning: nixpkgs commit hash in devbox.json is deprecated. Please run `devbox config upgrade` to remove and use lockfile to resolve package versions.
```

Turned `Nixpkgs` into a pointer so that it doesn't show up in devbox.json

## How was it tested?

```
devbox init
devbox add curl php@latest
devbox run which curl
devbox run which php
```


internal/impl/devbox.go
```diff
@@ -181,7 +173,7 @@ func (d *Devbox) ShellPlan() (*plansdk.ShellPlan, error) {
 		return nil, err
 	}
 
-	nixpkgsInfo := plansdk.GetNixpkgsInfo(d.cfg.Nixpkgs.Commit)
+	nixpkgsInfo := plansdk.GetNixpkgsInfo(d.cfg.NixPkgsCommitHash())
```
savil, line 176: This doesn't look safe if an empty-string is passed into GetNixpkgsInfo:

> mirrorURL := fmt.Sprintf("%s/nixos/nixpkgs/archive/%s.tar.gz", baseURL, commitHash)

internal/impl/devbox.go
```diff
@@ -243,7 +235,7 @@ func (d *Devbox) Shell(ctx context.Context) error {
 		WithShellStartTime(shellStartTime),
 	}
 
-	shell, err := NewDevboxShell(d.cfg.Nixpkgs.Commit, opts...)
+	shell, err := NewDevboxShell(d.cfg.NixPkgsCommitHash(), opts...)
```
savil, line 238: similarly, here, we need to change `nix.FlakeNixpkgs` to work with empty commit hashes

internal/impl/devbox.go
```diff
@@ -146,7 +138,7 @@ func (d *Devbox) ConfigHash() (string, error) {
 }
 
 func (d *Devbox) NixPkgsCommitHash() string {
-	if hash := d.cfg.Nixpkgs.Commit; hash != "" {
+	if hash := d.cfg.NixPkgsCommitHash(); hash != "" {
```
savil, line 141: do we need line 146 below, even with this change?

internal/impl/global.go
```diff
@@ -142,7 +142,7 @@ func (d *Devbox) pullGlobalFromPath(path string) error {
 }
 
 func (d *Devbox) addFromPull(pullCfg *Config) error {
-	if pullCfg.Nixpkgs.Commit != plansdk.DefaultNixpkgsCommit {
+	if pullCfg.NixPkgsCommitHash() != plansdk.DefaultNixpkgsCommit {
```
savil, line 145: 1. Why do we still need this change?
2. Do we still require `plansdk.DefaultNixpkgsCommit` or could we get rid of it too?

savil (commented):
some questions inline:

mikeland86:
@savil I think I addressed all the issues by defaulting `NixPkgsCommitHash()` if there's nothing in the config. I also removed the code comparing hashes in global pull which is no longer needed.

savil (approved):
Cool, I think the code works.

We could even say we require `plansdk.DefaultNixpkgsCommit` until we update our `examples/` to use versioned packages.

However, assuming we do update the `examples/` to use versioned packages, then longer term what role does `plansdk.DefaultNixpkgsCommit` play? Do we envision needing to keep it around?

mikeland86:
`DefaultNixpkgsCommit` is currently used for mkShell in flake.nix.tmpl (in some cases). We also want to continue to be backwards compatible with unversioned packages so we may need it for that as well. 

I'm unsure we will get rid of it for a while, but it will be only used in limited situations.

PR #1014, opened 2023-05-17
[devbox.json] Turn shell into pointer

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1013

This makes devbox.json config a bit nicer by omitting the "shell" field if it is empty.

Added default values to give the user an idea of what goes in those fields.

## How was it tested?

Created test project added hooks and scripts.


gcurtis:
No strong opinions here, but in the past we didn't make it a pointer because wanted the user to see the fields they could set. I think the idea was to set the prompt in a default init hook.

mikeland86:
@gcurtis my understanding is we only wanted `packages" to show up which is why it says `Deliberately do not omitempty.` 

Everything else had `omitempty` which doesn't work for non-pointer structs. 

@Lagoja thoughts on this? I agree it can help discoverability, but it makes config uglier and doesn't seem like something other configs (e.g. package.json) do.

Lagoja:
I think I prefer the visibility, package.json doesn't do this, but `npm init` walks you through filling out the critical fields, and still leaves placeholders if you don't provide an answer: 

```
{
  "name": "npmtest",
  "version": "1.0.0",
  "description": "test",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC"
}
```

mikeland86:
@Lagoja what if we add a default script and then if the user removes it it goes away? (so same as npm init)

savil (approved):

Issue #1010, opened 2023-05-17
[Bug]: nix bad illegal url error after auto upgrading to 0.5.0


tclain:
**Current Behavior (bug)**
Since auto upgrade to 0.5.0, I get the following message when activating devbox in my production repo:

```
direnv: loading ~/z/next/.envrc
direnv: using devbox
direnv: using nix .devbox/gen/shell.nix
error: unable to download 'github:NixOS/nixpkgs/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62': URL using bad/illegal format or missing URL (3)

       … while realising the context of a path

       at /Users/tclain/z/next/.devbox/gen/shell.nix:2:10:
```

**Expected Behavior (fix)**
No errors should happened

**Additional context**
- fish shell

Please include the output of `devbox version -v` and
`0.5.0`
a copy of your `devbox.json` file.

```
{
  "packages": [
    "nodejs",
    "nodePackages.pnpm",
    "postgresql",
    "nodePackages.typescript-language-server",
    "nodePackages.prettier",
    "colima",
    "kubectl"
  ],
  "nixpkgs": {
    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
  },
  "shell": {
    "scripts": {
      "pnpm-store-path": "pnpm store path",
      "install": "pnpm i",
      "test": "pnpm turbo test:unit"
    }
  }
}
```

Lagoja:
Hey @tclain , thanks for reporting. Looks like a pretty serious issue, so we'll take a look and try to resolve ASAP

If you need to unblock yourself, you can export `DEVBOX_USE_VERSION=0.4.9` to revert to the 0.4.9 version of Devbox. Once we resolve the issue, you can unset the variable to revert to the latest version.

Lagoja:
One quick request, can you run `nix --version` and share the output?

tclain:
```
(devbox) ◆  ❯❯❯ nix --version
nix (Nix) 2.11.1
````
sorry for the wait, twas nitght time :-)

mohsenari:
#1015

mohsenari:
@tclain we're working on a fix for this. This was caused because we made changes to our envrc file generation and how we handle direnv. The fix will be in the next release.
To prevent an issue like this happening for you, I recommend doing the following:
1. make a backup of your `.envrc` file `cp .envrc .oldenvrc`
2. regenerate `.envrc` by running `devbox generate direnv --force`

[...]

PR #1015, opened 2023-05-17
[bugs] Fix shell.nix bug and ensure current symlink

mikeland86:
## Summary

Fixes 2 unrelated bugs:

* Create current symlink if needed. This has no user impact, but is causing devbox cloud to break. It is also needed for future multi-profile support.
* Ensure shell.nix works by including correct URL in fetchTarball. This affects direnv users that have a legacy .envrc file. New/updated users are unaffected. Fixes https://github.com/jetpack-io/devbox/issues/1010

cc: @Lagoja 

## How was it tested?

* Deleted current symlink, and saw it being recreated.
* `nix-shell .devbox/gen/shell.nix`

### DIFF
--- a/internal/impl/global.go
+++ b/internal/impl/global.go
@@ -5,6 +5,7 @@ package impl
 
 import (
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


mohsenari:
#1010

mohsenari (approved):

PR #1017, opened 2023-05-17
Go Mod Example + Doc Links

Lagoja:
## Summary

* Add a Go Mod Flake Example
* Add links back to our docs for some key examples

## How was it tested?

Markdown Only

### DIFF
--- a/examples/flakes/go-mod/devbox.json
+++ b/examples/flakes/go-mod/devbox.json
@@ -1,0 +1,11 @@
+{
+  "packages": [
+    "path:ory-cli"
+  ],
+  "shell": {
+    "init_hook": null
+  },
+  "nixpkgs": {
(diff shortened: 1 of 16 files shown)


examples/flakes/go-mod/devbox.json
```diff
@@ -0,0 +1,11 @@
+{
+  "packages": [
+    "path:ory-cli"
+  ],
+  "shell": {
+    "init_hook": null
+  },
+  "nixpkgs": {
+    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
```
LucilleH, line 9: Should we get rid of `nixpkgs` field?

LucilleH (approved):

examples/flakes/go-mod/devbox.json
Lagoja: Once we have the auto-latest stuff merged, I'll go through and do that.

PR #1018, opened 2023-05-17
[direnv] Added warning for old envrc content

mohsenari:
## Summary
Checks for old .envrc file if all following is true:
- usePrintDevEnvCache is false
- .envrc file exists
- DEVBOX_NO_ENVRC_UPDATE env variable is not set to 1

Then checks if the .envrc file contains the 1 line that we have in new .envrc file:
```eval "$(devbox generate direnv --print-envrc)"```

if the file doesn't contain this line it prints out a warning.

## How was it tested?
- compile
- prepend path to compiled binary to PATH
- create an empty dir
- `devbox init`
- `touch .envrc` (don't run `devbox generate direnv`)
- copy the following to .envrc
```
eval "$(../devbox/dist/devbox generate direnv --print-envrc)"
```
replace `../devbox/dist/devbox` with a relative path to the compiled devbox binary

- then run `direnv allow` or run `devbox shell`
- it should show the warning about the content of your envrc being old.
NOTE: subsequent cd ins and out or devbox shells won't show the warning because of `usePrintDevEnvCache`

### DIFF
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -48,7 +48,7 @@ func Exists(path string) bool {
 }
 
 // Contains checks if a given file at 'path' contains the 'substring'
-func Contains(path string, substring string) (bool, error) {
+func FileContains(path string, substring string) (bool, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return false, err
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -5,6 +5,7 @@ package fileutil
 
 import (
 	"os"
+	"strings"
 )
 
 // TODO: publish as it's own shared package that other binaries can use.
@@ -44,4 +45,13 @@ func IsFile(path string) bool {
 func Exists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
+}
+
+// Contains checks if a given file at 'path' contains the 'substring'
+func Contains(path string, substring string) (bool, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return false, err
+	}
+	return strings.Contains(string(data), substring), nil
 }
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -47,7 +47,7 @@ func Exists(path string) bool {
 	return err == nil
 }
 
-// Contains checks if a given file at 'path' contains the 'substring'
+// FileContains checks if a given file at 'path' contains the 'substring'
 func FileContains(path string, substring string) (bool, error) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by mikeland86 — CHANGES_REQUESTED** [review]

File: internal/impl/devbox.go [hunk 1]
```diff
@@ -1005,6 +1013,27 @@ func (d *Devbox) findPackageByName(name string) (string, error) {
 	return lo.Keys(results)[0], nil
 }
 
+func (d *Devbox) checkOldEnvrc() error {
+	// check if user has an old version of envrc
+	if fileutil.Exists(".envrc") && os.Getenv("DEVBOX_NO_ENVRC_UPDATE") != "1" {
```
mikeland86, line 1018: nit `strconv.ParseBool`

internal/impl/devbox.go
```diff
@@ -1005,6 +1013,27 @@ func (d *Devbox) findPackageByName(name string) (string, error) {
...
+		// fmt.Println("inside general if")
+		isNewEnvrc, err := fileutil.FileContains(".envrc", "eval \"$(devbox generate direnv --print-envrc)\"")
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
```
mikeland86, line 1021: I missed that you had `fileutil.Exists(".envrc")` in the outside if statement. If you have that, then you don't need to check here again.

internal/impl/devbox.go
mikeland86, line 1018: Just realized you are using `".envrc"` directly. Instead you should use `filepath.Join(d.ProjecDir(), ".envrc")` in case devbox is called from another dir.

mikeland86 (approved):

mikeland86 (approved):

internal/fileutil/fileutil.go
```diff
@@ -45,3 +46,12 @@ func Exists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
 }
+
+// Contains checks if a given file at 'path' contains the 'substring'
+func FileContains(path string, substring string) (bool, error) {
```
hezhizhen, line 51: No need to have `File` prefix?

hezhizhen: LGTM.
Update the comment btw.

mohsenari: Good call! I updated 👍

internal/impl/devbox.go
mikeland86: @mohsenari this comment applies to all instances on .envrc

This currently breaks if we call it from a directory that contains `.envrc` but the target  project does not (e.g doing anything `global` from a directory with .envrc)

PR #1019, opened 2023-05-17
[cicd] Simplify go CICD cache

mikeland86:
## Summary

We were double caching in some places. 

I remove `cache: true` because this is default in v4 go action.

@loreto any reason cache was disabled for release?

## How was it tested?

CICD


loreto:
It was mainly to 100% guarantee that release is never accidentally "polluted" by invalid cache. That said, I think go does a good job with reproducibility, so it might be fine to turn on.

gcurtis (approved):

PR #1036, opened 2023-05-20
[devcontainer] Added shellenv to dockerfile template

### LINKED ISSUE #1028: [Bug]: Devbox environment in dev container not integrating with VS Code processes
**Current Behavior (bug)**

Running `devbox generate devcontainer` produces a dev container that includes the Devbox environment, but it does not integrate with all VS Code processes (as noted in #1001). For example, if I have a Devbox environment with Terraform installed and I configure this task in `.vscode/tasks.json`:

```json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "terraform version",
            "type": "shell",
            "command": "terraform version"
        }
    ]
}
```

Running it with `Tasks: Run Task` gives this:

```
 *  Executing task: terraform version 

/bin/bash: line 1: terraform: command not found

 *  The terminal process "/bin/bash '-c', 'terraform version'" failed to launch (exit code: 127). 
```

Accessing the Devbox environment requires `devbox shell` to be run first. Although the [Devbox VS Code extension](https://marketplace.visualstudio.com/items?itemName=jetpack-io.devbox) does this for you automatically when opening a terminal, it does not help with other VS Code features such as tasks and extensions.

**Expected Behavior (fix)**

When connecting to the dev container, the Devbox environment should be immediately available. A proposed solution is to simply replace the `CMD ["devbox", "shell"]` line in the `Dockerfile` with this:

```dockerfile
RUN devbox shellenv --init-hook >> ~/.profile
```

This persists the Devbox environment to `~/.profile`, including the all-important `PATH` environment variable. Assuming `userEnvProbe` is left at its default value, dev container tooling will "probe" this file and pass the environment into the container. I've tried this and it allows tasks, debugging and extensions to work correctly.

The `devbox.autoShellOnTerminal` behaviour provided by the Devbox extension then becomes unnecessary and should be disabled in `devcontainer.json`:

```json
  "customizations": {
    "vscode": {
      "settings": {
        "devbox.autoShellOnTerminal": false
      },
      "extensions": [
        "jetpack-io.devbox"
      ]
    }
  },
```

**Additional context**
`devbox version -v` gives:

```
Version:     0.5.2
Platform:    linux_amd64
Commit:      8f579b2c45c730742cbf9b7c400f247abf902775
Commit Time: 2023-05-18T16:18:52Z
Go Version:  go1.20.4
Launcher:    0.2.0
```

mohsenari:
## Summary
title says it.
fixes #1028 

## How was it tested?
- devbox init
- devbox add terraform
- devbox generate devcontainer
- add a file in `.vscode/tasks.json` and put the following json in it:
```
 {
    "version": "2.0.0",
    "tasks": [
        {
            "label": "terraform version",
            "type": "shell",
            "command": "terraform version"
        }
    ]
}
```

- `cmd + shift + p` and select "Rebuild and Reopen in Container"
- `cmd + shift + p` and select "Run Task" 
- Choose terraform task to run
- Assert that the output shows terraform version is printed in output

### DIFF
--- a/internal/impl/tmpl/devcontainerDockerfile.tmpl
+++ b/internal/impl/tmpl/devcontainerDockerfile.tmpl
@@ -27,5 +27,8 @@ WORKDIR /code
 RUN sudo chown $DEVBOX_USER:root /code
 COPY devbox.json devbox.json
 RUN devbox install
+{{if .IsDevcontainer}}
 RUN devbox shellenv --init-hook >> ~/.profile
+{{- else}}
 CMD ["devbox", "shell"]
+{{- end}}
--- a/internal/impl/tmpl/devcontainerDockerfile.tmpl
+++ b/internal/impl/tmpl/devcontainerDockerfile.tmpl
@@ -27,4 +27,5 @@ WORKDIR /code
 RUN sudo chown $DEVBOX_USER:root /code
 COPY devbox.json devbox.json
 RUN devbox install
+devbox shellenv --init-hook >> ~/.profile
 CMD ["devbox", "shell"]
--- a/internal/impl/tmpl/devcontainerDockerfile.tmpl
+++ b/internal/impl/tmpl/devcontainerDockerfile.tmpl
@@ -27,5 +27,5 @@ WORKDIR /code
 RUN sudo chown $DEVBOX_USER:root /code
 COPY devbox.json devbox.json
 RUN devbox install
-devbox shellenv --init-hook >> ~/.profile
+RUN devbox shellenv --init-hook >> ~/.profile
 CMD ["devbox", "shell"]
--- a/internal/boxcli/generate/devcontainer_util.go
+++ b/internal/boxcli/generate/devcontainer_util.go
@@ -36,8 +36,12 @@ type vscode struct {
 	Extensions []string `json:"extensions"`
 }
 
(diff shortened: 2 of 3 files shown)


Lagoja (approved):
LGTM

internal/impl/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -27,4 +27,5 @@ WORKDIR /code
 RUN sudo chown $DEVBOX_USER:root /code
 COPY devbox.json devbox.json
 RUN devbox install
+RUN devbox shellenv --init-hook >> ~/.profile
 CMD ["devbox", "shell"]
```
apgrucza, line 31: This line seems redundant

mohsenari: @apgrucza true. But the problem is we use this template file for both devcontainer setup and for general docker setup. 
I think the best way is to remove the CMD and keep shellenv for devcontainer and vice versa for `devbox generate dockerfile`.

apgrucza: Ok, but maybe the same change would be beneficial for the general Dockerfile too?

Issue #1035, opened 2023-05-20
[Plugin Update] PHP Plugin should include a basic php.ini in the user's project


Lagoja:
First reported by @jrseliga in the Devbox Discord

## Current Experience
 
 - The PHP Nix package installs the `php.ini` file in the Nix Store, where it cannot be edited through normal means. 
 
## Desired Experience

- Developers should have a way to modify the settings in `php.ini` for their project. 

## Current Workarounds

- Developers can currently install PHP extensions by including them in their package list. This should work for installing any PHP extensions that do not require additional php.ini config. See our [example](https://github.com/jetpack-io/devbox/tree/main/examples/development/php/php8.1) for more details
```json
"packages": [
     "php@latest",
     "phpExtensions.xdebug"
]
```

- If developers want to add a custom extension or modify the config via Nix, they can do this with a Flake as shown in [this example](https://github.com/jetpack-io/devbox/blob/main/examples/flakes/php-extension/my-php-flake/flake.nix). 

## Some Ideas for Improvement: 

1. Users can configure a `.user.ini` file in their local project to override some settings for their local project. This is a bit limited in which settings can be overridden, but may work for most cases. We can update our plugin to create this file by default + point PHP to it's location. 
2. Our plugin can modify PHP so that it looks for the php.ini file in the local project directory. This would make it editable, but would also result in a copy of PHP being installed for each project. Might be best to set as an option
3. Add a way to pass a php.ini file or additional config that will be added to the php.ini file in the Nix store when building PHP for the project.

Lagoja:
PR #1039 went with option 2 -- the php.ini file is now stored in the local project directory under `devbox.d`, where it can be edited by the user.

PR #1039, opened 2023-05-21
[plugin] Add `php.ini` in the PHP plugin

Lagoja:
## Summary

Adds an editable `php.ini` to the PHP plugin. This is useful when developers need to configure their PHP instance, and by creating it in `devbox.d` we let developers customize it for each of their PHP projects. This addresses #1035 and the issue reported by @jrseliga

## How was it tested?

* On a clean PHP project, run `devbox add php@latest` and verify that the php.ini file is created in devbox.d
* Also check the updated example in this PR


jrseliga:
Thanks @Lagoja! Looks good to me, the `php.ini` file is published and it seems the settings defined within take precedent.

mikeland86:
@Lagoja does php pick up the php.ini automatically? I thought it had to be in a special dir.

Lagoja:
You can use the `$PHPRC` environment variable (which I added to the plugin) to tell PHP where to look for the php.ini file. It's the second location that PHP will look, after the SAPI module specific location: 

https://www.php.net/manual/en/configuration.file.php

It seems to work on testing -- I can use it to set the xdebug mode, or to bump the memory limit for Django

mikeland86:
missed that, awesome!

plugins/php.json
```diff
@@ -10,10 +10,12 @@
   "env": {
     "PHPFPM_ERROR_LOG_FILE": "{{ .Virtenv }}/php-fpm.log",
     "PHPFPM_PID_FILE": "{{ .Virtenv }}/php-fpm.log",
-    "PHPFPM_PORT": "8082"
+    "PHPFPM_PORT": "8082",
+    "PHPRC": "{{ .DevboxDir }}"
```
mikeland86, line 14: Does this break if file/directory are missing? (user might delete `devbox.d/php` and we don't recreate.

mikeland86 (approved):

mikeland86 (approved):

plugins/php.json
Lagoja: Yes, if no `php.ini` is detected, php will use the default values for all directives.

Lagoja:
Going to tweak the examples to be backwards compatible, and then I'll merge

PR #1040, opened 2023-05-21
[UX] Improve ux when nixpkgs has not been downloaded

mikeland86:
## Summary

Ensure that nixkgs has been downloaded in a few places we were expecting it and causing poor UX.

## How was it tested?

* Installed a few packages
* unit tests

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -219,6 +219,15 @@ func (d *Devbox) addPackagesToProfile(ctx context.Context, mode installMode) err
 		return nil
 	}
 
+	// If packages are in profile but nixpkgs has been purged, the experience
+	// will be poor when we try to run print-dev-env. So we ensure nixpkgs is
+	// prefetched for all our packages.
+	for _, input := range d.packagesAsInputs() {
+		if err := input.EnsureNixpkgsPrefetched(d.writer); err != nil {
+			return err
+		}
+	}
+
 	pkgs, err := d.pendingPackagesForInstallation(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


internal/impl/packages.go
```diff
@@ -219,6 +219,15 @@ func (d *Devbox) addPackagesToProfile(ctx context.Context, mode installMode) err
...
 	}
 
+	// If packages are in profile but nixpkgs has been purged, the experience
+	// will be poor when we try to run print-dev-env. So we ensure nixpkgs is
+	// prefetched for all our packages.
+	for _, input := range d.packagesAsInputs() {
+		if err := input.EnsureNixpkgsPrefetched(d.writer); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
savil: Greg clarified offline that nixpkgs is not normally considered a dependency

internal/nix/input_test.go
```diff
@@ -142,3 +142,47 @@ func (l *lockfile) Resolve(pkg string) (*lock.Package, error) {
...
+		},
+		{
+			url:      "github:NixOS/other-repo/12345",
+			expected: "",
+		},
+		{
+			url:      "",
+			expected: "",
+		},
+	}
```
savil, line 175: thanks for adding these test cases

internal/nix/search.go
```diff
@@ -94,6 +94,14 @@ func pkgExistsForAnySystem(pkg string) bool {
 }
 
 func searchSystem(url string, system string) map[string]*Info {
+	// Search will download nixpkgs if it's not already downloaded. Adding this
+	// check here provides a slightly better UX.
+	if IsGithubNixpkgsURL(url) {
+		hash := HashFromNixPkgsURL(url)
+		// TODO, pass legit writer, handle error
```
savil, line 101: could we fix this prior to landing, otherwise we'll likely forget

savil (approved):
cool

Does this impact `devbox shell` start time?

internal/nix/search.go
mikeland86: I added some comments on why stderr is OK and why we want to ignore the error. Passing in a writer is a bit complicated. I think longer term we can restructure the nix package to have an struct. Another option is to hard code everything to stderr.

PR #1042, opened 2023-05-23
[docs] Correct direnv.md reference to `devbox generate direnv`

chancecarey:
## Summary

Docs update. Docs refer to `devbox generate envrc` when it should be `devbox generate direnv`

## How was it tested?

Visually? 🤷


mohsenari:
ah good catch! Thanks @chancecarey

mohsenari (approved):

PR #1043, opened 2023-05-24
[pull] Implement pull tar

mikeland86:
## Summary

This implements:

```bash
devbox global pull https://example.com/file.tar.gz
```

Since this can generate conflicts (because of existing files), there's a new `--force` flag that automatically overwrites.

If flag is not specified and there's a conflict, we ask the user if they wan to overwrite.

TODO:

- [x] We should do an install after a successful pull

TODO in follow up:

- [ ] Currently we always ask the survey question because `devbox.json` is created by default. So ideally if the devbox.json is empty, we always overwrite. This requires some refactoring that I would prefer not to include in this PR.

cc: @bketelsen

## How was it tested?

```bash
devbox global pull https://fleekgen.fly.dev/high
devbox global pull https://fleekgen.fly.dev/high --force
```


loreto:
Let me know if worth discussing live, but I think we should always override and remove the flag. If we want, we can prompt for a confirmation before overriding and have a `--force` flag that doesn't prompt.

bketelsen:
Plus one on only overriding Sent from my iPhoneOn May 23, 2023, at 9:04 PM, Daniel Loreto ***@***.***> wrote:﻿
Let me know if worth discussing live, but I think we should always override and remove the flag. If we want, we can prompt for a confirmation before overriding and have a --force flag that doesn't prompt.

mikeland86:
@loreto @bketelsen sounds good! Will add flag and prompt.

Always overriding is simplest by fair, but the first time I tested this I blew away my global, so I prefer to at least add some friction.

internal/impl/config.go
```diff
@@ -101,6 +102,20 @@ func (c *Config) InitHook() *shellcmd.Commands {
...
+		return
+	}
+	c.Packages = lo.Uniq(append(c.Packages, other.Packages...))
+	c.Env = lo.Assign(c.Env, other.Env)
+	c.InitHook().Merge(other.InitHook())
+	if c.Shell == nil {
+		c.Shell = &shellConfig{}
+	}
+	c.Shell.Scripts = lo.Assign(c.Scripts(), other.Scripts())
+	c.Include = append(c.Include, other.Include...)
```
savil, line 116: Can you add a comment near `Config` struct fields that we need to update this function when adding a new field?

Bonus points for adding a testcase with two Config structs to be merged, which asserts:
1. both structs have all non-zero-value fields. This ensures that new fields get added to the test case
2. the merged struct has fields that are different from either source struct's corresponding field. This ensures this Merge function was updated

internal/impl/config.go
```diff
@@ -101,6 +102,20 @@ func (c *Config) InitHook() *shellcmd.Commands {
 	return c.Shell.InitHook
 }
 
+func (c *Config) Merge(other *Config) {
```
savil, line 105: could we add a comment that `other`'s fields get precedence over `c`'s fields?

internal/pullbox/download.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to download file: %s", response.Status)
+	}
+
+	data, err := ioutil.ReadAll(response.Body)
```
savil, line 24: need to update to non-deprecated

internal/pullbox/tar.go
```diff
@@ -0,0 +1,88 @@
...
+	"os/exec"
+	"path/filepath"
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+var ErrFileExists = errors.New("file already exists")
+
+// extract decompresses a tar file and saves it to the specified directory
```
savil, line 18: is this comment correct? It doesn't save to a specified dir....its a temp dir, no?

internal/pullbox/tar.go
```diff
@@ -0,0 +1,88 @@
...
+
+		_, err := os.Stat(dstPath)
+		// Special case for devbox.json and merge if needed.
+		if err == nil && action == MergeAction && srcFile.Name() == "devbox.json" {
+			fmt.Fprintf(os.Stderr, "Merging devbox.json\n")
+			if err := p.merger(srcPath, dstPath); err != nil {
+				return err
+			}
+			// Copy if overwrite or file doesn't exist.
+		} else if action == OverwriteAction || err != nil {
```
savil, line 79: the `err != nil` should be `!errors.Is(err, fs.ErrNotExist)`

and then we need an `else if err != nil { return err }`

savil (commented):
looks good so far. Please re-request review when prompt is in. 

I'm also somewhat uneasy about always overriding, since the original global-devbox-json is not in source control and cannot be recovered.

I'd suggest also saving a backup of the devbox-global (could be a temp-dir, or in `XDG_CACHE_HOME`) which we print for the user. That way they could recover if unhappy with the result of the `pull`. 

Actually, maybe saving this backup copy is sufficient to avoid the prompt?

bketelsen:
testing this on ubuntu, determinate installer, bash shell. There's a repeating error 
```
path '/home/ubuntu/nixpkgs' does not contain a 'flake.nix', searching up
error: getting status of '/home/ubuntu/nixpkgs': No such file or directory
Ensuring nixpkgs registry is downloaded: Fail
Ensuring nixpkgs registry is downloaded.```
it looks like the command it's trying to run is 
```
/nix/var/nix/profiles/default/bin/nix flake prefetch nixpkgs/ --extra-experimental-features ca-derivations --option experimental-features nix-command flakes: exit status 1
```
which would correspond to the prefetch command in nixpkgs.go maybe being called with an empty `commit`?

**Comment by mikeland86** [comment]
@bketelsen yep, I think you caught an unrelated bug related to prefetching nixpkgs. Fix [here](https://github.com/jetpack-io/devbox/pull/1047)

btw, you can use new versioned package notation, e.g. 

```
python@3.11
go@latest
```

Which automatically resolves the version to the correct nixpkgs.

**Comment by bketelsen** [comment]
Yes @mikeland86 that change fixed it.

File: internal/boxcli/global.go [hunk 6]
```diff
@@ -85,7 +105,23 @@ func pullGlobalCmdFunc(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	return box.PullGlobal(cmd.Context(), args[0])
+	err = box.PullGlobal(cmd.Context(), overwrite, args[0])
+	if errors.Is(err, pullbox.ErrFileExists) {
```
savil, line 109: This breaks the Devbox interface abstraction by having callers check for an internal package (`pullbox`) error kind. This works for internal callers but some external user of the devbox package won't be able to use it.

Could we re-use `fs.ErrExist` instead of defining a new Err kind?

savil (approved):

PR #1044, opened 2023-05-24
Add a new-project example

Lagoja:
## Summary

Adds a blank project that can be used with Devbox Cloud

## How was it tested?

Trivial

### DIFF
--- a/examples/new-project/.envrc
+++ b/examples/new-project/.envrc
@@ -1,0 +1,7 @@
+# Automatically sets up your devbox environment whenever you cd into this
+# directory via our direnv integration:
+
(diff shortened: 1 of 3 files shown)


savil (approved):
thanks!

LucilleH:
Wait we don't need template - we just need `devbox init` to power cloud

hezhizhen:
> Wait we don't need this particular template - we just need `devbox init` to power cloud

Will `devbox init` create an `.envrc` file now?

Lagoja:
Closing since we can use devbox directly to create a blank project

PR #1045, opened 2023-05-24
[refactor] Move config to own package. Don't warn if overwriting defa…

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1043

This is mostly a straight refactor. Just moved everything to new package, renamed a few methods/constants if it made sense.

Only added functionality is `isModifiedConfig` in pullbox that doesn't return error if config that is being overwritten is a default config. (we need this because configs are created automatically so we always get the prompt even if it's empty)

## How was it tested?

Ran `devbox global pull https://fleekgen.fly.dev/high` with global directory:

* Empty
* Only with unmodified config
* Only with modified config
* With other files.

### DIFF
--- a/internal/pullbox/tar.go
+++ b/internal/pullbox/tar.go
@@ -59,9 +59,9 @@ func (p *pullbox) copy(overwrite bool, src, dst string) error {
 
 	if !overwrite {
 		for _, srcFile := range srcFiles {
-			// Only show error if file exists and is not a non-modified config
-			_, err := os.Stat(filepath.Join(dst, srcFile.Name()))
-			if err == nil && !isModifiedConfig(srcFile.Name()) {
+			dstPath := filepath.Join(dst, srcFile.Name())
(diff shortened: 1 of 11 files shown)


internal/pullbox/tar.go
```diff
@@ -59,7 +60,9 @@ func (p *pullbox) copy(overwrite bool, src, dst string) error {
 
 	if !overwrite {
 		for _, srcFile := range srcFiles {
-			if _, err := os.Stat(filepath.Join(dst, srcFile.Name())); err == nil {
+			// Only show error if file exists and is not a non-modified config
```
savil, line 63: should this be:
- is a non-modified config
- OR: is not a modified config
?

savil (approved):

internal/pullbox/tar.go
mikeland86: ugh, this is wrong (on several levels). I'll fix.

PR #1047, opened 2023-05-25
[easy][bug] Fix legacy resolving when nixpkgs hash is not set

mikeland86:
## Summary

TSIA

cc: @bketelsen


gcurtis (approved):

bketelsen:
> I'm curious how this scenario arises, since usually we either have `config.Nixpkgs` omitted or the entire `config.Nixpkgs.Commit` specified....

I think this is my fault. The config fleekgen creates has an empty `Nixpkgs` key. Sounds like I should have just omitted it?

PR #1048, opened 2023-05-25
MariaDB Plugin

Lagoja:
## Summary

Draft PR for a Maria DB Plugin.

- [x] add wrappers for binaries
- [x] add init_hook to install + configure the DB
- [x] add a service
- [x] check if other wrappers are needed 

## How was it tested?

- `devbox add mariadb` on a new project
- `devbox services up` should start the server
- `devbox run mysql` should connect the console


examples/databases/mariadb/devbox.json
```diff
@@ -1,18 +1,22 @@
...
+  "packages": [
+    "mariadb@latest"
+  ],
+  "shell": {
+    "init_hook": [],
+    "scripts": {
+      "connect_db": [
+        "mysql -u devbox_user -p -D devbox_lamp"
+      ],
+      "setup_db": [
```
mikeland86, line 11: Should we rename this to avoid using same name as plugin hook?

mikeland86 (approved):

PR #1054, opened 2023-05-26
feat: add list to devbox run if no args are given

bketelsen:
## Summary
If merged this change will print available scripts if no arguments are given on the `devbox run` command.

## How was it tested?
Trivial change, tested with CLI.

### DIFF
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -4,6 +4,8 @@
 package boxcli
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 
 	"go.jetpack.io/devbox"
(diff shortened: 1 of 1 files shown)


internal/boxcli/run.go
```diff
@@ -53,6 +54,18 @@ func listScripts(cmd *cobra.Command, flags runCmdFlags) []string {
 }
 
 func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
+	if len(args) == 0 {
+		scripts := listScripts(cmd, flags)
+		if len(scripts) == 0 {
+			return redact.Errorf("no scripts defined in devbox.json")
```
savil, line 60: this will return a non-zero exit code, which I feel isn't what we want: its _okay_ for there to not be any scripts to list :)

Could we change this to a regular `fmt.Fprintln(cmd.OutOrStdout(), "no scripts defined in devbox.json")`?

internal/boxcli/run.go
```diff
@@ -53,6 +54,18 @@ func listScripts(cmd *cobra.Command, flags runCmdFlags) []string {
...
+		}
+		fmt.Println("Available scripts:")
+		for _, p := range scripts {
+			fmt.Printf("* %s\n", p)
```
savil: Could we change these to explicitly take in the stdout writer, like `fmt.Fprintln(cmd.OutOrStdout(), "<msg>")`? 

That way, it'll make it easier to change the writer for tests or the --quiet flag, or similar future use-cases.

savil (commented):
Thanks!

a couple small requests inline

internal/boxcli/run.go
savil: ah, you beat me to it :)

savil (approved):

bketelsen:
PTAL

savil:
@bketelsen lgtm, thanks! I'll merge it once the required tests pass...

PR #1058, opened 2023-05-28
[WIP] Export the devbox global prefix of the packages that have been installed.

rbelem:

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -818,6 +818,8 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 		os.Getenv("XDG_DATA_DIRS"),
 	)
 
+	env["DEVBOX_GLOBAL_PREFIX"] = filepath.Join(d.projectDir, nix.ProfilePath)
+
 	return env, addHashToEnv(env)
 }
 


bketelsen:
This is useful! While you're here, can you add `DEVBOX_GLOBAL_ROOT` pointing to `.local/share/devbox/global/current`. That's useful to find files that are needed in the global profile, since the global profile is usually not executed in the same directory where the global devbox.json file lives.

rbelem:
@mikeland86, I think this is needed for projects and global. I saw myself making an env var for both.
One of my first iterations in the devbox channel was about creating an env var DEVBOX_PREFIX to build the path to a script that was in the share dir. And now DEVBOX_GLOBAL_PREFIX for my global :sweat_smile:

mikeland86:
@rbelem thoughts on a new command:

```bash
devbox path
devbox path -c path/to/my/config
devbox global path
devbox path --nix-profile
devbox path --bins
```

So you could do

`DEVBOX_GLOBAL_NIX_PROFILE="$(devbox global path --nix-profile)"`

cc: @Lagoja

rbelem:
@mikeland86 Looks flawless to me! And how would the ROOT path that @bketelsen mentioned work? `devbox path --projectDir` o `devbox path --root`?

mikeland86:
@rbelem I was thinking making `devbox path` (and `devbox global path`) point to the root.

bketelsen:
This is a winnerOn May 31, 2023, at 5:45 PM, Mike Landau ***@***.***> wrote:﻿
@rbelem I was thinking making devbox path (and devbox global path) point to the root.

rbelem:
@mikeland86 and `devbox global path --nix-profile` would point to `$HOME/.local/share/devbox/global/default/.devbox/nix/profile/default`? `devbox global path` to `$HOME/.local/share/devbox/global/default` ;

mikeland86:
@rbelem yep I think so. Open to suggestions if you have a strong opinion.

Main point of contention is deciding if this merits a new top level command. But we're on board with adding the functionality.

rbelem:
@mikeland86 Look good to me :D ty \o/

PR #1059, opened 2023-05-28
[packages] Recommend go from devbox index

mauricioabreu:
That would let Devbox handle which specific Nix package to use instead of making the user figure it out

## Summary

When using go 1.20 (go mod), it recommends the default version go_1_19. We could add 1_20 to the list but I feel like always running behind.

### DIFF
--- a/internal/initrec/recommenders/golang/golang.go
+++ b/internal/initrec/recommenders/golang/golang.go
@@ -4,6 +4,7 @@
 package golang
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


internal/initrec/recommenders/golang/golang.go
```diff
@@ -13,14 +14,7 @@ import (
 	"go.jetpack.io/devbox/internal/initrec/recommenders"
 )
 
-var versionMap = map[string]string{
-	// Map go versions to the corresponding nixpkgs:
-	"1.19": "go_1_19",
-	"1.18": "go",
-	"1.17": "go_1_17",
-}
```
mohsenari, line 21: The reason we had these versions was that in nixpkgs, these versions were available. Right now we also index older hashes of nixpkgs as far back as `go@1.13` you can see that by doing `devbox search go`.
If we don't have this versionMap, there will be an (unlikely) edge case where a user can have go version 1.12 which gets parsed but will see an error package not found. 
So, imo, we should have a map of supported versions (1.20 to 1.13) and default to 1.20 in the case of the user's goVersion not being in the map.

mohsenari: Alternatively, we can also handle this by showing a specific error message (e.g., "go version not supported please use `devbox search go` to see a list of supported versions)"

mohsenari (changes requested):

PR #1063, opened 2023-05-30
[Flakes] Changes to local flakes should reflect in shell after devbox install

savil:
## Summary

**Problem:** Changes to local flakes would not be reflected in the current shell. 

**Product workflow after this fix:**
1. User should edit the flake.nix file.
2. User should run `devbox install`.
3. The newly updated binaries from the flake should be available in their environment.

This PR makes two fixes:
1. Changes the `devbox.ConfigHash` to incorporate the contents of local flake files i.e. local inputs.
2. Calls `wrapnix.CreateWrappers` in `devbox.PrintEnv` to ensure that we re-create the binary wrappers
   to point to the latest binaries installed by the local flake. 

## How was it tested?

the below was run within `devbox shell` of each example.

in `examples/flakes/php/my-php-flake/flake.nix`, removed `ds` extension. Ran `devbox install`. And confirmed `ds` was removed via getting no output from `php -m | grep ds`.
I then re-added `ds` extension back, re-ran `devbox install` and this time `php -m | grep ds` returned a result.

in `examples/flakes/go-mod/ory-cli`, I changed the post-install binary's name from `ory` to `ory-renamed`. Ran `devbox install` and `which ory-renamed` returned a path to the binary.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -110,7 +110,12 @@ func (d *Devbox) Config() *devconfig.Config {
 }
 
 func (d *Devbox) ConfigHash() (string, error) {
-	return d.cfg.Hash()
+	hashes := lo.Map(d.packagesAsInputs(), func(i *nix.Input, _ int) string { return i.Hash() })
+	h, err := d.cfg.Hash()
+	if err != nil {
+		return "", err
+	}
+	return cuecfg.Hash(h + strings.Join(hashes, ""))
 }
 
 func (d *Devbox) NixPkgsCommitHash() string {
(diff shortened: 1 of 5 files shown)


savil:
* main
  * **PR #1063** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1063" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1063?utm_source=stack-comment).

mikeland86 (approved):

mikeland86 (approved):

PR #1061, opened 2023-05-29
[pull] Implement pull <git>

mikeland86:
## Summary

This implements `devbox global pull <git>` and offers familiar `--force` flag to overwrite conflicts.

### Existing functionality change:
It also changes the behavior of existing `pull <url>` and `pull <local-path>` to also warn and overwrite on `--force` instead of merging. We do this to standardize behavior.

## How was it tested?

```
devbox global pull git@github.com:mikeland86/global.git
devbox global pull https://fleekgen.fly.dev/high --force
devbox global pull devbox.json --force
devbox global pull https://raw.githubusercontent.com/mikeland86/global/main/devbox.json\?token=<token>
```

PR #1064, opened 2023-05-30
[push] Add devbox global push

mikeland86:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1061

This allows users to push global config (and any other files) to a repo. Pushing applies the current global profile on top of the latest commit in the repo (clone, commit, and push). It does not require force because it always applies on the latest commit.

## How was it tested?

```bash
devbox global pull git@github.com:mikeland86/global.git
devbox global push git@github.com:mikeland86/global.git
devbox global add curl
devbox global push git@github.com:mikeland86/global.git
```


bketelsen:
game on!

internal/boxcli/push.go
```diff
@@ -0,0 +1,76 @@
...
+		"Force push even if the remote has diverged",
+	)
+
+	flags.config.register(cmd)
+
+	return cmd
+}
+
+func pushCmdFunc(cmd *cobra.Command, flags pushCmdFlags) error {
+	box, err := devbox.Open(flags.config.path, cmd.ErrOrStderr())
```
savil, line 44: how is `flags.config.path` set for global configs?

savil (approved):
cool - I didn't quite follow how the global config's path is set, but otherwise looks good

internal/boxcli/push.go
mikeland86: the global parent command does some hackiness and basically sets the path for all subcommands in `setGlobalConfigForDelegatedCommands`

mikeland86:
@savil this changed a bunch. PTAL

internal/pullbox/ioutil/ioutil.go
```diff
@@ -0,0 +1,32 @@
...
+		}
+	}
+	return nil
+}
+
+func ClearDir(dir string) error {
+	if err := os.RemoveAll(dir); err != nil {
+		return errors.WithStack(err)
+	}
+	return os.MkdirAll(dir, 0755)
```
hezhizhen, line 31: Since the error returned above is wrapped, I think it's better to wrap it as well.

devbox.go
```diff
@@ -35,6 +35,7 @@ type Devbox interface {
 	PrintGlobalList() error
 	PrintEnvrcContent(w io.Writer) error
 	Pull(ctx context.Context, overwrite bool, path string) error
+	Push(url string) error
```
hezhizhen, line 38: I think it's better to have a `ctx` parameter like `Pull`

internal/pullbox/ioutil/exec.go
```diff
@@ -0,0 +1,33 @@
+package ioutil
```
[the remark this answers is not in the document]
mikeland86: I was mostly reusing cause `iotuil` is deprecated, but you are right it's confusing. Will change.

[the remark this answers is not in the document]
mikeland86: I was mostly reusing cause `iotuil` is deprecated, but you are right it's confusing. Will change.

internal/impl/pushpull.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.jetpack.io/devbox/internal/pullbox"
+)
+
+func (d *Devbox) Pull(
+	ctx context.Context,
```
savil, line 11: `ctx` parameter is unused

internal/pullbox/git/git.go
```diff
@@ -20,21 +19,17 @@ func CloneToTmp(repo string) (string, error) {
 	if err := clone(repo, tmpDir); err != nil {
 		return "", errors.WithStack(err)
 	}
-	if err := os.RemoveAll(filepath.Join(tmpDir, ".git")); err != nil {
-		return "", errors.WithStack(err)
-	}
```
savil, line 25: why remove this?

internal/pullbox/git/git.go
```diff
@@ -20,21 +19,17 @@ func CloneToTmp(repo string) (string, error) {
...
 	return tmpDir, nil
 }
 
 func IsRepoURL(url string) bool {
 	// For now only support ssh
-	return strings.HasPrefix(url, "git@")
+	return strings.HasPrefix(url, "git@") ||
```
savil, line 27: lets delete the comment about ssh

internal/pullbox/git/git.go
```diff
@@ -20,21 +19,17 @@ func CloneToTmp(repo string) (string, error) {
...
+		(strings.HasPrefix(url, "https://") && strings.HasSuffix(url, ".git"))
 }
 
 func clone(repo, dir string) error {
-	cmd := exec.Command("git", "clone", repo, dir)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
+	cmd := ioutil.CommandTTY("git", "clone", repo, dir)
```
savil: One motivation to pass in `context.Context` to `devbox.Push` and `devbox.Pull` is so that we set it in this `cmd.Context = ctx`.

Then if the `git` command is taking too long and getting stuck, the user could Ctrl+C and our top-level code would `cancel` the context to also cancel this....

savil: That said, I'm not sure what happens (even without explicitly setting a context in this `cmd`):
If the user Ctrl+C terminates the `devbox push/pull` command, then I think that would also terminate this "child process". But I could be wrong.

savil: Doing "research", it seems the answer is a bit subtle:
1. if the user interrupts the parent program with SIGINT, then the child processes also terminate (since they share the process group): https://stackoverflow.com/questions/19243431/why-process-created-by-exec-start-quits-if-its-parent-is-killed-by-sigint
2. But this suggests that one needs some manual setup to ensure that all spawened processes are truly killed. Way overkill for our situation: https://stackoverflow.com/questions/71714228/go-exec-commandcontext-is-not-being-terminated-after-context-timeout
3. This way suggests `exec.CommandContext` which is what I had in mind: https://stackoverflow.com/questions/74215347/kill-exec-if-the-parent-process-terminates

Anyway, I'm inclined to say this isn't a problem to be solved right now

internal/pullbox/ioutil/exec.go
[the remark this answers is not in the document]
savil: yeah, +1, I'd get confused thinking that `ioutil.CommandTTY` refers to the stdlib ioutil. 

Also the stdlib ioutil deals with `io.Reader` and `io.Writer` whereas this also deals with executing commands.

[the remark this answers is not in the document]
savil: suggestions:
- `osutil`
- `iops` (for `io operations`)
- `exe` for commands and `osutil` for the directory operations in the next file

don't feel strongly about any of these suggestions

savil (approved):

PR #1069, opened 2023-05-31
[flakes] Fix flakes with fragments

mikeland86:
## Summary

URLs were getting normalized without fragments which breaks flakes for for non default exports or flakes that export more than 1 package. This fixes it.

## How was it tested?

PHP flake example

### DIFF
--- a/examples/flakes/php/my-php-flake/flake.nix
+++ b/examples/flakes/php/my-php-flake/flake.nix
@@ -20,7 +20,7 @@
 
           # If you only want to export a single package, you can name it default which allows
           # installation without using url fragment (.e.g. "path:my-flake")
-          default = pkgs.php.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds memcached ]));
+          # default = pkgs.php.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds memcached ]));
         };
       });
 }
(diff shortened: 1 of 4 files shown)


gcurtis (approved):

examples/flakes/php/my-php-flake/flake.nix
```diff
@@ -20,7 +20,7 @@
 
           # If you only want to export a single package, you can name it default which allows
           # installation without using url fragment (.e.g. "path:my-flake")
-          default = pkgs.php.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds memcached ]));
+          # default = pkgs.php.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds memcached ]));
```
savil, line 23: could we create another example so we have:
1. one example with default export
2. one example without a default export

?

Lagoja: I can modify the php-with-extensions to have a default export, and push it to this pr

Lagoja:
The php flake's run_test is broken: 

```
        [1/2] path:my-php-flake#php: Success
        [2/2] path:my-php-flake#hello
        evaluating 'packages.x86_64-linux.hello'...
        [2/2] path:my-php-flake#hello: Success
        $WORK/.devbox/gen/scripts/run_test.sh: 5: [[: not found
        [exit status 127]
        FAIL: /tmp/TestExamples982480092/015/flakes_php_run_test.test.txt:2: unexpected command failure
```

Lagoja (approved):
LGTM, confirmed that it's working on the user example as well

PR #1070, opened 2023-05-31
Update nodejs examples

LucilleH:
## Summary
Update nodejs examples. Add pnpm example as well.

## How was it tested?
devbox shell in each example


LucilleH:
Need to reopen this PR per error on CICD

Issue #687, opened 2023-02-24
[Bug]: devbox shellenv shouldn't prepend to existing path


ipince:
**Current Behavior (bug)**
If you `devbox rm <pkg>` inside a devbox shell, and then do `eval $(devbox shellenv)`, `<pkg>` will still be available because `devbox shellenv` prepends to the existing `PATH`, which still contains a path to `<pkg>`.

**Expected Behavior (fix)**
Instead of prepending to current PATH, we prepend to the PATH as it was when the user entered the devbox shell.

PR #1072, opened 2023-05-31
[shellenv] preserve original XDG_DATA_DIRS during shellenv-ception

savil:
## Summary

shellenv-ception is when we run shellenv within an already enabled shellenv environment.

We would like to preserve idempotency of XDG_DATA_DIRS for shellenv-ception, so 
that we don't recursively keep prepending to the shellenv-enabled XDG_DATA_DIRS
and instead we prepend to the original XDG_DATA_DIRS.

## How was it tested?

same as #1041

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -739,9 +739,8 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	originalXdgDataDirs, ok := env[d.ogXdgDataDirsKey()]
 	if !ok {
 		// if the env does not have XDG_DATA_DIRS then currentXdgDataDirs is empty string
-		currentXdgDataDirs, _ := env["XDG_DATA_DIRS"]
-		env[d.ogXdgDataDirsKey()] = currentXdgDataDirs
-		originalXdgDataDirs = currentXdgDataDirs
+		originalXdgDataDirs = env["XDG_DATA_DIRS"]
+		env[d.ogXdgDataDirsKey()] = originalXdgDataDirs
 	}
 
 	vaf, err := d.nix.PrintDevEnv(ctx, &nix.PrintDevEnvArgs{
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -733,6 +733,17 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 		originalPath = currentEnvPath
 	}
 
+	// We use this XDG_DATA_DIRS by prepending the print-dev-env
+	// XDG_DATA_DIRS to it below. By preserving the original XDG_DATA_DIRS, we
+	// enable this shellenv calculation to be idempotent, if we do shellenv-ception.
+	originalXdgDataDirs, ok := env[d.ogXdgDataDirsKey()]
+	if !ok {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1072** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1072" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1072?utm_source=stack-comment).

mikeland86 (commented):
Instead of saving og values for everything should we just modify `JoinPathLists` to not add stuff that's already there?

savil:
That won't be enough, no?

For example, i'm considering the following scenario:
1. originally in the shell: `XDG_DATA_DIRS=my-path-0`
2. first shellenv produces: `XDG_DATA_DIRS=my-path-1:my-path-0` because some nix package specified that `my-path-1` should be in `XDG_DATA_DIRS`.
3. second shellenv that runs due to some edits to that nix package which changes `my-path-1` to `my-path-2`. So, then we'd want the final env-var to be `XDG_DATA_DIRS=my-path-2:my-path-0`.

Does that make sense?
If so, I believe we do need to preserve the original env-var and not just de-duplicate in JoinPathLists

savil:
Its similar to the motivation for introducing OG_PATH in https://github.com/jetpack-io/devbox/issues/687

mikeland86:
@savil what I had in mind was a bit more involved:

original `XDG_DATA_DIRS=x:y:z`
nix `XDG_DATA_DIRS=a:z`

computed `XDG_DATA_DIRS=a:z:x:y`

This has the following properties:

* nix XDG_DATA_DIRS take precedent.
* No duplicates
* idempotent (because of no-duplicates property)

I think this could be used for all path joining cases (including PATH and LD_LIB) but that would introduce risk to this PR, so maybe use a new `MergePathLists` that joins and dedups?

savil:
@mikeland86 I don't think that removing duplicates will achieve idempotency by itself. 

Continuing on from your example:
- now we have `XDG_DATA_DIRS=a:z:x:y`
- some edit is made, and shellenv is called again, and nix now has `XDG_DATA_DIRS=b:z`

Without the OG_XDG_DATA_DIRS fix, it'll be `XDG_DATA_DIRS=b:a:z:x:y` instead of what we would want: `b:z:x:y`

The PR as is will produce `XDG_DATA_DIRS=b:z:x:y:z`, which is idempotent. Its not "clean" since we have the trailing `:z` at the end, but functionally it won't matter.

savil:
@mikeland86 turns out `JoinPathLists` does de-duplicate. So, it does take care of the scenario you are concerned about.

I still think we also need to preserve the OG_XDG_DATA_DIRS so that we drop any directories that are no longer needed due to any nix edits that are made.

mikeland86:
The result doesn't need to be unchanging to any input, it needs to be unchanging to the _same_ input. 

Specifically, calling shellenv multiple times should produce the same result (which would keep the hash constant). If the input changes then we get a different result.

Using the OG approach makes sense in a single shell world, but not sure it does in a multi-project world. It's also less scalable, we now need to keep a back-up variable for every one of these. I know that's what we already have for PATH, but I think we may want to move towards a deduping with prepending approach eventually.

savil:
Chatted IRL. Gonna punt on this for now since its a theoretical concern (no reports yet), and the pattern could lead to some edge-cases with nested shellenvs.

PR #1074, opened 2023-06-01
Fix template create when ssh key is not setup

LucilleH:
## Summary
For users and Dockerfile that has no ssh key setup, we need to clone by https

## How was it tested?
devbox run build
devbox create --template apache


mikeland86 (approved):
I guess this works without auth because repo is public?

LucilleH:
> I guess this works without auth because repo is public?

yes

Issue #1029, opened 2023-05-19
[Documentation]: VS Code integration with the direnv extension


apgrucza:
The Devbox documentation makes mention of the Devbox VS Code extension and dev containers, but neither currently provide VS Code users with an adequate experience (as described in #1001). In that issue it was identified that the [direnv VS Code extension](https://marketplace.visualstudio.com/items?itemName=mkhl.direnv) provides a more streamlined Devbox experience. I find this approach preferable to dev containers, which negate the benefits of Devbox as a Docker-free solution.

I wish I'd come across the direnv extension earlier, as I spent a lot of time looking for workarounds and nearly gave up on Devbox. Having it in the Devbox documentation could prevent others having the same experience.

Note the [known issues](https://marketplace.visualstudio.com/items?itemName=mkhl.direnv#known-issues) with this extension. There are two other direnv extensions in the marketplace, but I've not tried them.

Lagoja:
@apgrucza we had some struggles with the direnv extension + Devbox in the past, but if it's working well we can consider recommending it. Does your project work well since you installed it?

We've made changes to our direnv integration, so it's possible that we're now at a state where they work well together

Lagoja:
Did some testing with a few example projects, and this seems to work reasonably well if a .envrc file is present. There are some limitations highlighted in the extension's docs that are worth noting. 

- The Direnv Extension doesn't seem to work with tasks of `process` type
- There seem to be reports where the Direnv Extension does not set the environment quickly enough for other extensions to detect and load the new environment

I still think it's worth recommending as an option. Some next steps we could take on this issue: 

1. Add a note about the Direnv extension to the Direnv + VSCode pages in our documentation
2. Test the IntelliJ Direnv extension to see if that also helps with IntelliJ setup

I would also still like to find a way for Devbox to work with VSCode and other editors without having dependencies on direnv + the direnv extension.

apgrucza:
The direnv extension worked well for my Devbox trial, but I didn't do extensive testing (I tried the terminal, a task of type shell, Python debugging and an extension that tries to execute a locally installed tool).

To remove the dependency on direnv, perhaps some functionality could be copied from the direnv extension into the Devbox extension.

[...]

PR #1075, opened 2023-06-01
[vscode-extension] reopen vscode in devbox env using process communication

mohsenari:
## Summary
Added ability to reload a vscode window and re-open it in a devbox environment.
This is done in 4 steps: 
1. Devbox extension invokes devbox CLI (devbox integrate vscode) as a child process.
2. CLI then prepares the devbox environment (computeNixEnv) and sends a message to parent process that preparing new environment is finished and ready.
3. Parent process (vscode window) upon receiving the finish message, closes itself. 
4. CLI will open a new vscode window (cmd.execute) with env variables it has received from setting up devbox environment.

NOTE: This required changes to both CLI and the extension, so the release of extension has to happen after CLI.

Addresses #1029 

## How was it tested?
1. compile CLI and put the binary in an directory with devbox.json (let's call it `vstest/`)
2. specify a package (e.g., hello@latest) and an env variable in devbox.json (e.g., "foo": "bar123")
3. in devbox.ts update `const devbox = 'devbox';` to `const devbox = <absolute path to compiled CLI binary>`
4. `cd devbox/vscode-extension` then package an installable extensinon file: `vsce package`.(this will create a devbox-version.vsix file)
5. install this extension on vscode by `code --install-extension path-to-vsix-file`
6. open `vstest/` in vscode. 
7. In vscode's terminal run `echo $foo` it should be empty. Then type `which hello` and it should point to devbox environment's hello, or say command not found
8. open command palette in vscode (cmd + shift + p) and type "reopen in devbox" and choose the suggested action.
9. once vscode is re-opened, try step 7 and confirm that vscode now has devbox environment setup.


vscode-extension/src/devbox.ts
```diff
@@ -0,0 +1,61 @@
...
+                        await workspace.fs.stat(dotdevbox);
+                    } catch (error) {
+                        //.devbox doesn't exist
+                        // running devbox shellenv to create it
+                        spawnSync('devbox', ['shellenv'], {
+                            cwd: workingDir.path
+                        });
+                    }
+                    // change this to absolute path to custom compiled devbox when testing
+                    const devbox = 'devbox';
```
LucilleH, line 35: this would have to be something else for prod?

LucilleH (approved):
Looks good to me!

vscode-extension/src/devbox.ts
mohsenari: @LucilleH No, for testing locally on a compiled devbox this would have to point to the path to the compiled devbox.
This comment is mostly a note-to-self for future changes.
For example, if I change something in CLI and want to see how it interacts with the extension, I should change this line to `/Users/mohsen/projects/devbox/dist/devbox` so that my changes in compiled devbox CLI gets used instead of the installed devbox. Hope that makes sense.

LucilleH: Ah, `To use a custom compiled devbox when testing, change this to an absolute path.` lol

PR #1077, opened 2023-06-01
[path] Add global path command

mikeland86:
## Summary

Inspired by https://github.com/jetpack-io/devbox/pull/1058

## How was it tested?

```bash
➜  devbox git:(landau/global-path) devbox global path
/Users/mike/.local/share/devbox/global/default
```


savil (approved):
lgtm, modulo @loreto's concerns about the flag?

PR #1081, opened 2023-06-01
fix: don't error when existing symlink points at correct path

bketelsen:
## Summary
Update error handling in wrapnix package to succeed if intended symlinks already exist

## How was it tested?
Manually testing.

### DIFF
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -144,6 +144,18 @@ func createSymlinksForSupportDirs(projectDir string) error {
 		newname := filepath.Join(projectDir, plugin.WrapperPath, dir.Name())
 
 		if err := os.Symlink(oldname, newname); err != nil {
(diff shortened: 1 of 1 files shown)


bketelsen:
confirmed this resolves the issue on macos and linux.

internal/wrapnix/wrapper.go
```diff
@@ -144,6 +144,18 @@ func createSymlinksForSupportDirs(projectDir string) error {
 		newname := filepath.Join(projectDir, plugin.WrapperPath, dir.Name())
 
 		if err := os.Symlink(oldname, newname); err != nil {
+			// ignore if the symlink already exists
+			if errors.Is(err, os.ErrExist) {
```
savil, line 148: Ideally, `fs.ErrExist` since `os.ErrExist` is kinda deprecated, but nbd

savil (approved):
Thanks!

Issue #1049, opened 2023-05-25
[Bug]: Devbox wrappers and the infinite loop of doom


literatesnow:
**Current Behavior (bug)**
Our `devbox.json` has the package `coreutils` (mostly for sed on a mac) and this makes wrappers for standard unix things such as `tr`, `tail`, `[`, `cat`, `dd`, `ls` etc in the `.devbox/virtenv/.wrappers/bin` directory.

1. Open a terminal and `devbox shell` (this is where we run our apps). This terminal is fine as `__DEVBOX_SHELLENV_HASH == <hash>`
2. Open another terminal and `devbox shell` (this is where we run our editor). In this terminal `__DEVBOX_SHELLENV_HASH != <hash>` and it tries to `eval ... devbox shellenv ...` for every wrapped program.

Unfortunately something in `devbox shellenv` calls a standard unix thing (I think it's `tr`), which is wrapped, which in turn tries to `eval devbox shellenv`, which calls another wrapper again, which ... goes on until you run out of processes.

I'm not sure of the purpose of `__DEVBOX_SHELLENV_HASH != <hash>` in the wrapper scripts so this is a bug report instead of a PR.

**Expected Behavior (fix)**
Wrapper scripts don't call themselves recursively. I added a guard in the wrapper but I'm not sure that's the right solution.

```bash
if [[ "$__DEVBOX_SHELLENV_HASH" != "..." ]] && [[ -z "$__DEVBOX_SHELLENV_GUARD" ]]; then
  export __DEVBOX_SHELLENV_GUARD=true
  eval "$(DO_NOT_TRACK=1 devbox shellenv -c ...)"
fi
```

I also tried commenting out the `eval` line and everything seems fine?

**Additional context**
devbox version 0.5.2 darwin_amd64 8f579b2c go1.20.4 launcher_0.2.0

PR #1083, opened 2023-06-01
[shellenv] Fix infinite loop

mikeland86:
## Summary

Fixes https://github.com/jetpack-io/devbox/issues/1049

In certain cases, shellenv might call a wrapped binary causing an infinite loop. This prevents it. 

## How was it tested?

Called a test script and confirmed env var was set.

### DIFF
--- a/internal/wrapnix/wrapper.sh.tmpl
+++ b/internal/wrapnix/wrapper.sh.tmpl
@@ -14,7 +14,7 @@ fi
 {{- end }}
 
 {{/*
-We use ShellEnvHashKey to prevent doing shellenv if the correct envrioment is 
+We use ShellEnvHashKey to prevent doing shellenv if the correct environment is
 already set. (perf optimization)
 
 We use the guard to prevent infinite loop if something in shellenv causes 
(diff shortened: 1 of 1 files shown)


internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -14,14 +14,18 @@ fi
 {{- end }}
 
 {{/*
-We use ShellEnvHashKey to avoid re-sourcing shellenv. Since wrappers
-call other wrappers and potentially modify the environment, we don't want the
-environment to get re-written.
+We use ShellEnvHashKey to prevent doing shellenv if the correct envrioment is
```
savil, line 17: typo: environment

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -14,14 +14,18 @@ fi
...
 {{/*
-We use ShellEnvHashKey to avoid re-sourcing shellenv. Since wrappers
-call other wrappers and potentially modify the environment, we don't want the
-environment to get re-written.
+We use ShellEnvHashKey to prevent doing shellenv if the correct envrioment is 
+already set. (perf optimization)
+
+We use the guard to prevent infinite loop if something in shellenv causes 
+another wrapped binary to be called. The guard is specific to this project so shellenv
+could still cause another project's shellenv to be called.
```
savil, line 22: which is okay, and desirable, right?

savil (approved):

internal/wrapnix/wrapper.sh.tmpl
mikeland86: yep, by design

PR #1084, opened 2023-06-01
[CLI] devbox shell --pure

mohsenari:
## Summary
I added a `--pure` flag for `devbox shell` to create an isolated env without variables leaking from the parent shell.

Q: Should I add a --pure flag for `devbox run` and `devbox shellenv`?
A: Yes
Q2: Because the created shell is "pure" the devbox command is also not available inside the pure shell. Should that be an exception?
A: Yes

Note: in pure mode, the `$HOME` env variable has to leak in because the devbox binary relies on it. 

## How was it tested?
- compile
- `export FOO=bar`
- `./devbox shell --pure`
- `echo $FOO` should show no value
- Also `./devbox run --pure -- echo $FOO` should show no value


internal/boxcli/shell.go
```diff
@@ -36,6 +37,8 @@ func shellCmd() *cobra.Command {
 
 	command.Flags().BoolVar(
 		&flags.PrintEnv, "print-env", false, "print script to setup shell environment")
+	command.Flags().BoolVar(
+		&flags.Pure, "pure", false, "Creates an isolated shell without taking any variables from parent system.")
```
savil, line 41: what do you think about:

> without sharing any environment from the parent system.

?

internal/impl/devbox.go
```diff
@@ -827,16 +830,22 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
-	env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
+	if pure {
+		env["PATH"] = nixEnvPath
+	} else {
+		env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
+	}
```
savil, line 837: I think this if-else branching is not needed, because originalPath on line 744-748 will be empty-string. This could be as before IIUC.

savil (commented):
1. Yes, I do think we need the flag for shellenv and run
2. And yes, I do think we need to ensure devbox binary is included.

Could we add some testscript unit-tests for pure versus impure?

Some prelim comments. Need to go to a meeting....

savil (commented):
nice!


1. Yes, I do think we need the flag for shellenv and run
2. And yes, I do think we need to ensure devbox binary is included.

Could we add some testscript unit-tests for pure versus impure?

Some prelim comments. Need to go to a meeting....

internal/impl/shell.go
```diff
@@ -87,28 +87,34 @@ func NewDevboxShell(nixpkgsCommitHash string, opts ...ShellOption) (*DevboxShell
...
 	var bashNixStorePath string // of the form /nix/store/{hash}-bash-{version}/
 
-	cmd := exec.Command(
-		"nix", "eval", "--raw",
-		fmt.Sprintf("%s#bash", nix.FlakeNixpkgs(nixpkgsCommitHash)),
-	)
+	// construct exec args, add --pure-eval if pure shell
+	execArgs := []string{"eval", "--raw"}
+	if pure {
+		execArgs = append(execArgs, "--pure-eval")
```
savil, line 113: 🤔 I'm not super familiar and didn't find good nix docs about this.

1. Why do we need it? Could you point to some good docs you may have found about it?
2. Should we also add it to `print-dev-env` call inside `nix.computeEnv`? Hmm but that would complicate the printDevEnv cache that we have.

mohsenari: https://docs.haskellstack.org/en/v2.9.1/nix_integration/#pure-and-impure-nix-shells

This is the doc I found. Which aligns with our approach for a pure shell.

savil (commented):
Nice, this is looking more promising. I had a few thoughts inline, and below: 

1.`CreateWrappers` sets up the `wrapper.sh.tmpl` which calls `shellenv`: https://github.com/jetpack-io/devbox/blob/main/internal/wrapnix/wrapper.sh.tmpl#L29. 

I tried reasoning thru whether we need to invoke `--pure` here as well, or not. But this is pretty confusing when I think of various scenarios (`devbox add` within a `pure` shell etc.), but my head is hurting now. 

@mikeland86 thoughts?


2. I think we need the `.envrc` file that is generated for direnv to also incorporate `--pure`? Maybe in a follow up PR...

3. When I start `devbox shell --pure`, I get:
```
 (mohsen--pure-shell)> devbox shell --pure
Starting a devbox shell...

Error: We found a /nix directory but nix binary is not in your PATH and we were not able to find it in the usual locations. Your nix installation might be broken. If restarting your terminal or reinstalling nix doesn't work, please create an issue at https://github.com/jetpack-io/devbox/issues
```

My nix is in `/nix/var/nix/profiles/default/bin` (IIRC its from the default devbox install). Do we need to ensure that this is in PATH? One solution is to make the following edit inside `computeNixEnv` function:
```
        currentEnvPath := env["PATH"]
+       if currentEnvPath == "" {
+               currentEnvPath = "/nix/var/nix/profiles/default/bin"
+       }
```

4. `devbox add` doesn't seem to work for me inside a pure shell:
```
(devbox) Savil-Srivastavas-MacBook-Pro:devbox savil$ devbox add ripgrep

Error: There was an error installing nix packages
source: exec: "git": executable file not found in $PATH

(devbox) Savil-Srivastavas-MacBook-Pro:devbox savil$
```

internal/boxcli/shell.go
mikeland86: This is how nix described pure:


> --pure
If this flag is specified, the environment is almost entirely cleared before the interactive shell is started, so you get an environment that more closely corresponds to the “real” Nix build. A few variables, in particular HOME, USER and DISPLAY, are retained.

devbox.go
```diff
@@ -32,18 +32,18 @@ type Devbox interface {
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
+	PrintEnv(ctx context.Context, includeHooks bool, pure bool) (string, error)
```
mikeland86, line 35: Not a blocker

I understand that adding boolean is the most straight forward way to do this, but boolean params are generally code smell and make code harder to read. Imagine `doStuff(true, false, true)` you need to add comments or check signature every time.

A few alternatives:

* Use named types. `PrintEnv(ctx, false, pure)`
* Use options struct `PrintEnv(ctx, opts{Pure: true})`

internal/impl/devbox.go
```diff
@@ -735,7 +737,11 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 		if ignoreCurrentEnvVar[key] {
 			continue
 		}
-		env[key] = val
+		// Need HOME env variable for pure shell to leak through
+		// otherwise devbox binary won't work
+		if !pure || key == "HOME" {
```
mikeland86, line 742: Should this be `HOME, USER and DISPLAY` to match nix?

internal/impl/devbox.go
```diff
@@ -846,11 +852,13 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 
 	d.setCommonHelperEnvVars(env)
```
mikeland86, line 853: I'm unsure if these should go in pure or not. cc: @gcurtis

internal/impl/shell.go
mikeland86: This is just trying to find the shell to use right? I don't think this needs `pure-eval` (it's unrelated to the shell created for the user)

internal/wrapnix/wrapper.go
```diff
@@ -33,7 +34,7 @@ var wrapper string
 var wrapperTemplate = template.Must(template.New("wrapper").Parse(wrapper))
 
 // CreateWrappers creates wrappers for all the executables in nix paths
-func CreateWrappers(ctx context.Context, devbox devboxer) error {
+func CreateWrappers(ctx context.Context, devbox devboxer, pure bool) error {
```
mikeland86, line 37: Adding `pure` here makes this PR quite a bit more complicated because booleans need to be threaded in more places. It's also not really a "wrapper", but I see why creating it here is simpler.

To keep changes simpler I would suggest always creating the devbox symlink (so get rid of `pure` here and just assume it's always true). There's no downside to having the symlink exist in non-pure environments.

internal/impl/devbox.go
```diff
@@ -280,10 +280,10 @@ func (d *Devbox) RunScript(ctx context.Context, cmdName string, cmdArgs []string
 // creates all wrappers, but does not run init hooks. It is used to power
 // devbox install cli command.
 func (d *Devbox) Install(ctx context.Context) error {
-	if _, err := d.PrintEnv(ctx, false /* run init hooks */); err != nil {
+	if _, err := d.PrintEnv(ctx, false /* run init hooks */, false /* pure */); err != nil {
 		return err
 	}
-	return wrapnix.CreateWrappers(ctx, d)
+	return wrapnix.CreateWrappers(ctx, d, false /* pure */)
```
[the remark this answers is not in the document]
mohsenari: I need this parameter to be passed to `CreateWrappers` because I create the devbox symlink in wrappers bin in there.

mikeland86 (changes requested):
This PR as is makes me a bit nervous. Like @savil said, it's difficult to reason about what needs pure vs impure environments. Additionally adding boolean params everywhere makes code more difficult to read (and 1 boolean leads to 2, 3.....)

Open to ideas, but I think we can make 2 changes that will reduce the size of this PR a bunch.

1) Remove `pure` from CreateWrappers and just create the symlink unconditionally. I'm not sure this is the right place, but it's simple.
2) Change `devbox.Open(dir string, writer io.Writer)` to be `devbox.Open(dir string, writer io.Writer, devbox.Opts{})` and add `pure` as a field in the `devbox` struct. This way you don't need to pass `pure` everywhere. You have it when you need it. The only downside to this is that you can't have a single devbox object produce `pure` and `impure` outcomes, but I think thats OK.

internal/impl/devbox.go
```diff
@@ -320,7 +320,7 @@ func (d *Devbox) PrintEnv(ctx context.Context, includeHooks bool) (string, error
 }
 
 func (d *Devbox) ShellEnvHash(ctx context.Context) (string, error) {
-	envs, err := d.nixEnv(ctx)
+	envs, err := d.nixEnv(ctx, false /* pure */)
```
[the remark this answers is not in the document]
mohsenari: I'm not following

internal/impl/devbox.go
mohsenari: yeah sure, I want to keep it as minimal as possible. But matching nix is ok in this case.

mohsenari:
@mikeland86 @savil 
I made an overhaul and removed `pure` being passed as a function argument. I added it to the `Devbox` struct and set it when `devbox.Open()` is called. 
But that means I had to update all `devbox.Open()` calls, most of them get an empty `devbox.Opts{}`. So I hope that's ok.

In `NewDevboxShell()` I had to pass it as a function argument because of the way `ShellOptions` is implemented.

I also removed `--pure-eval` from `nix eval` since I couldn't find justification to have it added.

internal/impl/shell.go
```diff
@@ -64,14 +64,15 @@ type DevboxShell struct {
...
+	pure           bool
 }
 
 type ShellOption func(*DevboxShell)
 
 // NewDevboxShell initializes the DevboxShell struct so it can be used to start a shell environment
 // for the devbox project.
-func NewDevboxShell(nixpkgsCommitHash string, opts ...ShellOption) (*DevboxShell, error) {
-	shPath, err := shellPath(nixpkgsCommitHash)
+func NewDevboxShell(nixpkgsCommitHash string, pure bool, opts ...ShellOption) (*DevboxShell, error) {
```
mikeland86, line 74: Make `NewDevboxShell` and `shellPath` have devbox as receiver that way you don't need to add `pure` param.

devbox.go
```diff
@@ -55,13 +55,17 @@ type Devbox interface {
...
 
+type Opts struct {
+	Pure bool
+}
+
 // Open opens a devbox by reading the config file in dir.
-func Open(dir string, writer io.Writer) (Devbox, error) {
-	return impl.Open(dir, writer, true)
+func Open(dir string, writer io.Writer, opts *Opts) (Devbox, error) {
+	return impl.Open(dir, writer, true, opts.Pure)
```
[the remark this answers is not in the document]
mohsenari: @mikeland86 that creates a circular import (currently devbox imports impl). We'd have to import impl in devbox package.

internal/impl/devbox.go
```diff
@@ -746,6 +754,9 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	}
 
 	currentEnvPath := env["PATH"]
+	if d.pure { // make nix available inside pure shell - necessary for devbox add to work
+		currentEnvPath = "/nix/var/nix/profiles/default/bin"
```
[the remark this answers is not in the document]
mohsenari: It is! you can test it. In a pure shell, `devbox add` throws this error:
```bash
Error: We found a /nix directory but nix binary is not in your PATH and we were not 
able to find it in the usual locations. Your nix installation might be broken. 
If restarting your terminal or reinstalling nix doesn't work, please create an issue at 
https://github.com/jetpack-io/devbox/issues
```

internal/impl/shell.go
```diff
@@ -199,6 +201,12 @@ func WithShellStartTime(time string) ShellOption {
 	}
 }
 
+func WithPureShell(pure bool) ShellOption {
+	return func(s *DevboxShell) {
+		s.pure = pure
+	}
```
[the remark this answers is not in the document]
mohsenari: Sorry forgot to remove it.

mohsenari:
@mikeland86 Addressed the nit feedback. Not sure about the `SHELL` variable and about `d.setCommonHelperEnvVars(env)`

devbox.go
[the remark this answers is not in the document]
mikeland86: You can create a 3rd package `devopt.Opts` that is accessed from boxcli, devbox, impl

internal/impl/devbox.go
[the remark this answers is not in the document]
mohsenari: Regardless, I removed passing pure as param to functions.

devbox.go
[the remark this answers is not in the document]
mohsenari: Done. It made calling devbox.Open() much clearer 👍

internal/boxcli/shell.go
mohsenari: Updated the flag description to a similar text.

internal/impl/shell.go
```diff
@@ -87,28 +88,29 @@ func NewDevboxShell(nixpkgsCommitHash string, opts ...ShellOption) (*DevboxShell
...
-	if path != "" {
-		debug.Log("Using SHELL env var for shell binary path: %s\n", path)
-		return path, nil
+	if !pure {
+		// First, check the SHELL environment variable.
+		path = os.Getenv(envir.Shell)
+		if path != "" {
+			debug.Log("Using SHELL env var for shell binary path: %s\n", path)
+			return path, nil
+		}
```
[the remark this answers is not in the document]
gcurtis: I think here we should ignore it and always use Nix's bash shell. IMO the shell can have a pretty big impact on the environment.

I agree with the use-case of letting the user override the shell, but it might make more sense to do that through a new devbox.json field. That way it can be a Nix-installed shell that's guaranteed to be the same for everyone.

internal/impl/devbox.go
gcurtis: I think we still want `LD_LIBRARY_PATH` and `LIBRARY_PATH`, but without the current environment's values appended. Since it looks like `env["LIBRARY_PATH"]` and `env["LD_LIBRARY_PATH"]` will always be empty, I think this is okay as-is.

devbox.go
```diff
@@ -55,13 +56,28 @@ type Devbox interface {
...
+		Dir:          opts.Dir,
+		Pure:         opts.Pure,
+		ShowWarnings: true,
+		Writer:       opts.Writer,
+	})
 }
 
-func OpenWithoutWarnings(dir string, writer io.Writer) (Devbox, error) {
-	return impl.Open(dir, writer, false)
+func OpenWithoutWarnings(opts *devopt.Opts) (Devbox, error) {
```
mikeland86, line 73: nit, this can be removed

internal/boxcli/devopt/devboxopts.go
```diff
@@ -0,0 +1,10 @@
+package devopt
```
mikeland86, line 1: nit, move to `impl/devopt`

internal/impl/devbox.go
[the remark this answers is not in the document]
mikeland86: hm, OK. I'm not 100% sure this works in all cases (e.g. single user installs).

I just tested and single user on ubuntu installed:

```bash
$ which nix
/home/mike/.nix-profile/bin/nix
```

See https://github.com/jetpack-io/devbox/blob/main/internal/nix/source.go#L32-L32

For some reason, we are not sourcing nix correctly. In the pure shell, is `__ETC_PROFILE_NIX_SOURCED=1` set?

If that is set, then we may need a different approach.

One possible approach is to look in `$PATH` and try to find:

`/nix/var/nix/profiles/default/bin`
OR
`$HOME/.nix-profile/bin/nix` 

and keep them around. 

There's also :

```bash
xdg.StateSubpath("nix/profile/etc/profile.d/nix.sh"),
xdg.StateSubpath("nix/profiles/profile/etc/profile.d/nix.sh"),
```

I'm not sure if nix also stores bins there. 

------------

So in short, I think you need to look in path for any of these:

```bash
/nix/var/nix/profiles/default/bin
$HOME/.nix-profile/bin
xdg.StateSubpath(/nix/profile/bin)
xdg.StateSubpath(nix/profiles/profile/bin)
```

If any of those are in PATH, then they need to be retained. In practice it will only be one of them.

I'm not 100% sure there are no other nix paths, but this should cover all the cases we currently consider.

internal/wrapnix/wrapper.go
```diff
@@ -91,6 +91,10 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 			return errors.WithStack(err)
 		}
 	}
+	err = createDevboxSymlink(devbox.ProjectDir())
+	if err != nil {
```
mikeland86, line 95: nit, inline error handling.

internal/wrapnix/wrapper.go
```diff
@@ -162,3 +166,20 @@ func createSymlinksForSupportDirs(projectDir string) error {
...
 }
+
+// Creates a symlink for devbox in .devbox/virtenv/.wrappers/bin
+// so that devbox can be available inside a pure shell
+func createDevboxSymlink(projectDir string) error {
+
+	// Get absolute path for where devbox is called
+	devboxPath, err := filepath.Abs(os.Args[0])
+	if err != nil {
+		return errors.Errorf("Failed to create devbox symlink. Devbox command won't be available inside the shell. Error: %v", err)
```
mikeland86, line 177: use `errors.Wrap`, `errors.Wrapf` or `errors.WithStack` instead. `errors.Errorf` is designed to create a new error instead of propagating an existing one.

mikeland86 (approved):
I think all my last round of comments can be done in a follow up. The only semi-urgent one is to search PATH for a few places nix binary can be and not just a single hard coded one. This gives us a better chance of finding it in all types of installations.

Since this PR is already really large, you should merge and do a fast follow with a fix for that. Other nits can be addressed after that.

internal/impl/devbox.go
```diff
@@ -72,8 +74,9 @@ type Devbox struct {
 
 var legacyPackagesWarningHasBeenShown = false
 
-func Open(path string, writer io.Writer, showWarnings bool) (*Devbox, error) {
-	projectDir, err := findProjectDir(path)
+func Open(opts *devopt.Opts) (*Devbox, error) {
+	// func Open(path string, writer io.Writer, showWarnings bool, pure bool) (*Devbox, error) {
```
savil, line 78: rm

internal/boxcli/shellenv.go
```diff
@@ -40,6 +42,9 @@ func shellEnvCmd() *cobra.Command {
 	command.Flags().BoolVar(
 		&flags.install, "install", false, "install packages before exporting shell environment")
 
+	command.Flags().BoolVar(
+		&flags.pure, "pure", false, "If this flag is specified, devbox creates an isolated environment inheriting almost no variables from the current environment. A few variables, in particular HOME, USER and DISPLAY, are retained.")
```
savil, line 46: could make a `const pureFlagMsg = "...."` that is reused in all these places, instead of copy-pasta

savil (approved):
Much nicer!

As some followups, I think we will want to enable a direnv user to opt-into pure mode

savil:
It would also be great to add some testscript unit-tests under `testscripts/` for the --pure case.

internal/boxcli/shellenv.go
mohsenari: Nice idea! The description is slightly different for shell, run and shellenv. So I decided to copy each time for now. In a follow up try making a function that reuses the same variable but maybe differs in description.

savil: oh they're different? nevermind mind me then

mohsenari:
Merging this now, will followup with another PR addressing nix in PATH issue.

PR #1086, opened 2023-06-02
[push] Add overwrite parameter for force push

hezhizhen:
## Summary

1. Add `ctx` to `Push` for later use (we can add trace to every method)
2. Add `overwrite` to `Push` for force push in git


mikeland86 (commented):
@hezhizhen I originally had this, but the way push is implemented `--force` is never needed. (because whenever we push, we always clone the latest repo)

hezhizhen:
> @hezhizhen I originally had this, but the way push is implemented `--force` is never needed. (because whenever we push, we always clone the latest repo)

I see.

PR #1087, opened 2023-06-02
Update nodejs examples

LucilleH:
## Summary
Update nodejs examples. Add pnpm example as well.

## How was it tested?
devbox shell in each example


examples/development/nodejs/nodejs-pnpm/devbox.json
```diff
@@ -0,0 +1,12 @@
...
+    "packages": [
+        "nodejs@18",
+        "nodePackages.pnpm@latest"
+    ],
+    "shell": {
+        "init_hook": [
+            "pnpm install",
+            "pnpm run start"
+        ]
+    }
```
savil, line 11: would you mind adding a `run_test` devbox script?
That way, our `testscripts` framework will run it to ensure we don't regress this functionality.

(optional) In that, you could verify some output if you want, but even just ensuring it "runs" is a good start.

examples/development/nodejs/nodejs-typescript/devbox.json
```diff
@@ -1,9 +1,12 @@
...
-        "nodePackages.typescript"
+        "nodejs@18.16"
     ],
     "shell": {
-        "init_hook": null
+        "init_hook": [
+            "npm install",
+            "npm run build",
+            "npm run start"
+        ]
```
savil, line 10: same request here.

savil (approved):
approving to unblock, but see inline

Lagoja (approved):

PR #1088, opened 2023-06-02
Change error message for devbox add when a package is not found

LucilleH:
## Summary
Change error message for devbox add when a package is not found. To repro:
`devbox add hellow@latest`

Original output:
```
Error: No results found for "hellow@latest".
```

New output:
```
Error: No results found for "hellow@latest".

To search for packages, use the `devbox search` command
```

## How was it tested?
```
devbox run build
devbox add hellow@latest
```

### DIFF
--- a/internal/boxcli/add.go
+++ b/internal/boxcli/add.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
+	"go.jetpack.io/devbox/internal/nix"
 )
 
 const toSearchForPackages = "To search for packages, use the `devbox search` command"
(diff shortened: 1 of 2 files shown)


internal/boxcli/add.go
```diff
@@ -37,7 +36,7 @@ func addCmd() *cobra.Command {
 				return nil
 			}
 			err := addCmdFunc(cmd, args, flags)
-			if errors.Is(err, nix.ErrPackageNotFound) {
+			if err != nil {
```
mikeland86, line 39: This will mask any other error (including not logging them to Sentry)

I think we should change https://github.com/jetpack-io/devbox/blob/main/internal/searcher/client.go#L63-L63 to return `ErrPackageNotFound` (maybe move it to searcher pkg) so we only show this error message for that error. 

There's also this potential error as well: https://github.com/jetpack-io/devbox/blob/main/internal/nix/input.go#L260-L260

not sure if we need any special handling for that.

mikeland86 (changes requested):

internal/boxcli/add.go
LucilleH: Just checked, no need for https://github.com/jetpack-io/devbox/blob/main/internal/nix/input.go#L260-L260 as that only happens when a user does:
```
devbox add nodejs@
```
with the `@` but no version

mikeland86 (approved):

mikeland86 (approved):

PR #1091, opened 2023-06-02
[shellenv] filter out buildInput paths from PATH

### LINKED ISSUE #1060: [Bug]: devbox global rm doesn't remove packages
**Current Behavior (bug)**
```
$ devbox global add stress
$ devbox global rm stress
$ which stress
/home/<user>/.local/share/devbox/global/default/.devbox/nix/profile/default/bin/stress
```
This behaviour still exists even after I add devbox shell 

**Expected Behavior (fix)**
The command should not be installed/available in the terminal anymore

**Additional context**

```
Version:     0.5.2
Platform:    linux_amd64
Commit:      8f579b2c45c730742cbf9b7c400f247abf902775
Commit Time: 2023-05-18T16:18:52Z
Go Version:  go1.20.4
Launcher:    0.2.0
```

devbox.json (global)
```
{
  "packages": [
    "ripgrep",
    "git",
    "neovim",
    "bat",
    "direnv",
    "fd",
    "mosh",
    "lazygit",
    "byobu",
    "cargo",
    "nodejs-slim",
    "nodePackages_latest.npm",
    "python310Packages.pip",
    "python310"
  ],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
  }
}
```

savil:
## Summary

**problem**
Doing `devbox rm <package>` within a devbox shell will usually still have the package's binary accessible in PATH. This is confusing and unexpected.

**cause**

In shellenv's `PATH`, we now have 3 sources of paths:
1. `.devbox/virtenv/.wrappers/bin`
2. `.devbox/nix/profile/default/bin`
3. `/nix/store/<hash>-<package>-<version>/bin`

When we do `devbox rm <package>`, the first two (wrappers and profile bins) are removed, but the `/nix/store` remains in PATH. 

The user would have to do `eval $(devbox shellenv)` to remove it. But that is not a good UX.

**background context**

Heres’ why we need/don’t need each one:
1. bin-wrapper dir: is our path for binaries. Binaries here are wrapped to ensure the correct environment is sourced. “Support files” are also sym linked here to ensure that binaries that require relative dependencies work as well (e.g. mariadb)
2. profile bin path: This is a fallback hack for certain edge cases (e.g. curl). If a package uses propagateBuildInputs then we may not correctly create a bin wrapper for it. Currently we use $buildInputs to determine binaries that need to be wrapped, but this does not contain propagated inputs, only explicit inputs. 
3. nix store paths: these are included by nix from `nix print-dev-env`. We've never removed them, but need to ensure nothing breaks.

Thanks to @mikeland86 for explaining this. I've copy-pasted this from a chat conversation.

**this PR**
This PR removes from PATH the subset of `/nix/store` paths that belong to the buildInputs, during `shellenv`. 

Fixes #1060 

TODO:
- [x] do more extensive testing

## How was it tested?

```
> devbox add stress
> devbox shell
(devbox)> which stress
# printed wrappers/bin path
(devbox)> devbox rm stress
(devbox)> which stress
# empty
# BEFORE: this would print the /nix/store path
```

- [x] ensure testscripts pass.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -841,6 +841,11 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
+	nixEnvPath = filterPathList(nixEnvPath, func(path string) bool {
+		return !strings.HasPrefix(path, "/nix/store")
+	})
+	debug.Log("PATH after removing /nix/store paths is: %s", nixEnvPath)
+
 	env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
 	debug.Log("computed environment PATH is: %s", env["PATH"])
 
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -841,6 +841,10 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
+	// We filter out nix store paths so that if a user removes a package from their devbox
+	// it no longer is available in their environment. This is needed because nix may keep
+	// packages around if they are used somewhere else or if the user hasn't triggered
+	// garbage collection.
 	nixEnvPath = filterPathList(nixEnvPath, func(path string) bool {
 		return !strings.HasPrefix(path, "/nix/store")
 	})
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -841,14 +841,21 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
-	// We filter out nix store paths so that if a user removes a package from their devbox
-	// it no longer is available in their environment. This is needed because nix may keep
-	// packages around if they are used somewhere else or if the user hasn't triggered
-	// garbage collection.
+	// We filter out nix store paths of devbox-packages (represented here as buildInputs).
+	// Motivation: if a user removes a package from their devbox it should no longer
+	// be available in their environment.
+	buildInputs := strings.Split(env["buildInputs"], " ")
 	nixEnvPath = filterPathList(nixEnvPath, func(path string) bool {
-		return !strings.HasPrefix(path, "/nix/store")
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #1091** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1091" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1091?utm_source=stack-comment).

internal/impl/devbox.go
```diff
@@ -827,6 +827,11 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
+	nixEnvPath = filterPathList(nixEnvPath, func(path string) bool {
```
mikeland86, line 830: Add comment:

```golang
// We filter out nix store paths so that if a user removes a package from their devbox 
// it no longer is available in their environment. This is needed because nix may keep
// packages around if they are used somewhere else or if the user hasn't triggered 
// garbage collection
```

mikeland86 (approved):

internal/impl/devbox.go
```diff
@@ -827,6 +827,15 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	nixEnvPath := env["PATH"]
 	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
 
+	// We filter out nix store paths so that if a user removes a package from their devbox
+	// it no longer is available in their environment. This is needed because nix may keep
+	// packages around if they are used somewhere else or if the user hasn't triggered
+	// garbage collection.
+	nixEnvPath = filterPathList(nixEnvPath, func(path string) bool {
+		return !strings.HasPrefix(path, "/nix/store")
```
savil, line 835: A problem with filtering _all_ the `/nix/store` paths is that we also remove many utilities that are automatically included by `pkgs.mkShell` in the generated `flake.nix`.

As a result, basic functionality breaks, like invoking `mkdir`.

@mikeland86 thoughts on fixing this?

A (bad) idea is to generate `print-dev-env` PATH without any packages in the flake.nix. Keep the `/nix/store/` paths from this, while removing any new /nix/store paths added from print-dev-env using the generated flake.nix. BUT this would involve calling `print-dev-env` twice, which is slow.

mikeland86: Can you filter out `$BuildInputs` only?

savil: yep, this works I think. PTAL!

savil:
@mikeland86 going to land, since the change is as we'd discussed.

PR #1096, opened 2023-06-03
[push] Simplify git clone for push

hezhizhen:
## Summary

1. clone `.git` only for push (See https://stackoverflow.com/questions/38999901/clone-only-the-git-directory-of-a-git-repo for more information)
2. add `CreateDevboxTempDIr` to `fileutil` and reuse it in more cases

### DIFF
--- a/internal/boxcli/push.go
+++ b/internal/boxcli/push.go
@@ -6,6 +6,7 @@ package boxcli
 import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
 	"go.jetpack.io/devbox"
 )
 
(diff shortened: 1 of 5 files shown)


hezhizhen:
@mikeland86 PTAL

mikeland86 (approved):
Nice! Awesome improvement.

mikeland86:
Small request, in the future please include a test plan (can be unit test or even just some bash commands). This provides move confidence that code was tested and gives reviewers a way to test it as well.

hezhizhen:
> Small request, in the future please include a test plan (can be unit test or even just some bash commands). This provides move confidence that code was tested and gives reviewers a way to test it as well.

My mistake. I tried it locally, but I didn't provide test steps.

PR #1097, opened 2023-06-03
Keep file mode the same as it is cleared

hezhizhen:
## Summary

1. Get the filemode of the directory before everything inside is deleted and use that mode for re-creation
2. Reuse the `ClearDir` in gen-docs

### DIFF
--- a/internal/fileutil/dir.go
+++ b/internal/fileutil/dir.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+
 	"go.jetpack.io/devbox/internal/cmdutil"
 )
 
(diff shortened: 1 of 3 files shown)


hezhizhen:
@mikeland86 @savil

internal/fileutil/dir.go
```diff
@@ -26,8 +27,14 @@ func CopyAll(src, dst string) error {
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	mode := f.Mode()
+
 	if err := os.RemoveAll(dir); err != nil {
 		return errors.WithStack(err)
 	}
-	return errors.WithStack(os.MkdirAll(dir, 0755))
+	return errors.WithStack(os.MkdirAll(dir, mode))
```
savil, line 39: what do you think of changing this function to not error if the directory doesn't exist?

In that case (i.e. not exist), we can `os.MkdirAll(dir, 0755)`

savil (commented):
Overall, I like the idea

savil (approved):

internal/fileutil/dir.go
hezhizhen: I didn't think about it.
I should check if the `dir` exists first. If it doesn't, set mode to 0755. This check should be under L34.
I'll create a PR later to fix it.
Thx @savil

hezhizhen: See #1103

PR #1098, opened 2023-06-03
[trace] Add trace for Devbox methods

hezhizhen:
## Summary

1. Add `ctx` parameter for most Devbox methods
2. Add trace tasks for Devbox methods
3. Add regions for sub-methods

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -25,11 +25,11 @@ type Devbox interface {
 	ProjectDir() string
 	// Generate creates the directory of Nix files and the Dockerfile that define
 	// the devbox environment.
-	Generate() error
-	GenerateDevcontainer(force bool) error
-	GenerateDockerfile(force bool) error
-	GenerateEnvrcFile(force bool) error
-	Info(pkg string, markdown bool) error
+	Generate(ctx context.Context) error
+	GenerateDevcontainer(ctx context.Context, force bool) error
+	GenerateDockerfile(ctx context.Context, force bool) error
+	GenerateEnvrcFile(ctx context.Context, force bool) error
+	Info(ctx context.Context, pkg string, markdown bool) error
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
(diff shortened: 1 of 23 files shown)


hezhizhen:
The PR is still working in progress. Is it a good idea to trace them all? @savil @mikeland86

savil:
@gcurtis do you have thoughts on this? I know you'd added the initial tracing infra, and dashboards.

My concern would be that adding to each method may get overwhelming to track in the dashboards, but i'm not sure where to look for them.

hezhizhen:
> @gcurtis do you have thoughts on this? I know you'd added the initial tracing infra, and dashboards.
> 
> My concern would be that adding to each method may get overwhelming to track in the dashboards, but i'm not sure where to look for them.

Any updates? @savil @gcurtis

gcurtis (commented):
This looks good to me. There aren't any dashboards for the CLI, it's just for `devbox -trace` which generates a `go tool trace` file to help track down performance issues.

hezhizhen:
The rest methods are so simple that I don't think it's necessary to add trace. PTAL @gcurtis

hezhizhen:
@gcurtis @savil Any comments?

internal/impl/update.go
```diff
@@ -1,6 +1,9 @@
 // Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
 // Use of this source code is governed by the license in the LICENSE file.
 
+// Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
```
gcurtis, line 6: Duplicate?

gcurtis (approved):
Thanks for cleaning all this up!

savil:
@hezhizhen thanks! if you could resolve the conflcts, I can help merge.

hezhizhen:
> @hezhizhen thanks! if you could resolve the conflcts, I can help merge.

Done. PTAL @savil

internal/impl/update.go
hezhizhen: Yeah. That's weird. I removed it.

savil (approved):

PR #1101, opened 2023-06-05
Hide devbox cloud cli command for now in favor of devbox.sh

LucilleH:
## Summary
Hide devbox cloud cli command for now as we decided that devbox.sh will be the primary way to create a vm, and a user can ssh into that vm directly with local credentials

## How was it tested?
devbox run build
devbox help


loreto:
LGTM

loreto (approved):

PR #1103, opened 2023-06-06
[fileutil] use 0755 as default filemode to create dir

hezhizhen:
## Summary

Use 0755 as default filemode to create dir. If dir exists, use its own filemode to recreate.

### DIFF
--- a/internal/fileutil/dir.go
+++ b/internal/fileutil/dir.go
@@ -4,6 +4,7 @@
 package fileutil
 
 import (
+	"io/fs"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


internal/fileutil/dir.go
```diff
@@ -27,11 +27,13 @@ func CopyAll(src, dst string) error {
 }
 
 func ClearDir(dir string) error {
+	// if the dir doesn't exist, use default filemode 0755 to create it
+	// if the dir exists, use its own filemode to re-create it
+	var mode os.FileMode = 0755
 	f, err := os.Stat(dir)
-	if err != nil {
-		return errors.WithStack(err)
+	if err == nil {
```
savil, line 34: could we preserve returning an error unless it is fs.ErrNotExist?
```
var mode os.FileMode
if f, err := os.Stat(dir); if err == nil {
      mode = f.Mode()
} else if errors.Is(err, fs.ErrNotExist) {
         mode = 0755
else {
 	return errors.WithStack(err)
}
```

savil (approved):
thanks!

PR #1105, opened 2023-06-06
[versioned] If user adds new version, remove previous versions of canonical

mikeland86:
## Summary

If there's an existing package with same canonical, remove it. If there's more than one do nothing.

## How was it tested?

Need to test this a bit more to ensure no weird side-effects of doing a `remove` inside an `add`

On this repo did:

```bash
cat devbox.json
devbox add go@1.19
cat devbox.json
devbox add go@1.20
cat devbox.json
```

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,8 +1,8 @@
 {
   "packages": [
+    "go@1.20",
     "actionlint@1.6.23",
-    "golangci-lint@1.52.2",
-    "go@1.20"
+    "golangci-lint@1.52.2"
   ],
   "env": {
     "PATH": "$PATH:$PWD/dist"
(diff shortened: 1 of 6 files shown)


loreto (approved):
LGTM

savil:
You could add a testscript unit-test similar to the test plan.

May need to add a custom function (https://github.com/jetpack-io/devbox/blob/8b4a1050ef730cedc400c62a6bf3f71adac9bbb6/testscripts/testrunner/testrunner.go#L70) called `json.list.contains` and use it as `json.list.contains devbox.json packages go@1.20` and `json.list.contains.not devbox.json packages go@1.19`

savil:
thanks for adding the testscript!

PR #1106, opened 2023-06-06
Change caddy to run on non privileged ports

LucilleH:
## Summary
Change caddy to run on non privileged ports. I added the following script to the caddy example:
- devbox run start

Since I have a global caddy running on port `2019` for the admin endpoint already, I need to run this caddy under a separate admin port.

## Issue I still need help with (resolved. See commit "fix process compose")
Running `devbox services up` has the following error
```
[caddy_1	] Error: loading initial config: loading new config: starting caddy administration endpoint: listen tcp 127.0.0.1:2019: bind: address already in use
```
Where as `devbox run start` is successful 🤷‍♀️ Are we doing something special with process compose? Seems like for some reason, it is loading some initial config, then reload it with the new config. And because the initial config has a port conflict, everything fails.

To reproduce, go to [https://devbox.sh/github.com/jetpack-io/devbox?folder=examples/servers/caddy&branch=lucille--caddy-on-vm](https://devbox.sh/github.com/jetpack-io/devbox?folder=examples/servers/caddy&branch=lucille--caddy-on-vm)

## How was it tested?
devbox.sh

### DIFF
--- a/plugins/caddy/Caddyfile
+++ b/plugins/caddy/Caddyfile
@@ -1,9 +1,15 @@
 # See https://caddyserver.com/docs/caddyfile for more details
+{
+	admin 0.0.0.0:2020
+	auto_https disable_certs
+	http_port 8800
+	https_port 4443
+}
 
-localhost:2020 {
-        root * {$CADDY_ROOT_DIR}
-        log {
-            output file {$CADDY_LOG_DIR}/caddy.log
-        }
-        file_server
+:8080 {
+	root * {$CADDY_ROOT_DIR}
+	log {
+		output file {$CADDY_LOG_DIR}/caddy.log
+	}
+	file_server
 }
--- a/plugins/caddy/Caddyfile
+++ b/plugins/caddy/Caddyfile
@@ -6,7 +6,7 @@
 	https_port 4443
 }
 
-:8080 {
+:8082 {
 	root * {$CADDY_ROOT_DIR}
 	log {
 		output file {$CADDY_LOG_DIR}/caddy.log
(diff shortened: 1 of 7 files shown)


plugins/caddy/Caddyfile
```diff
@@ -1,9 +1,15 @@
 # See https://caddyserver.com/docs/caddyfile for more details
+{
+	admin 0.0.0.0:2020
+	auto_https disable_certs
+	http_port 8800
+	https_port 4443
+}
```
mikeland86, line 7: What's this for?

mikeland86 (approved):
Small nit, can you bump plugin version. We don't use it right now, but it gets saved in the lock file and in the future we can use it to suggest updates that can fix overwrite and fix config files.

Should we choose different ports for different services? (e.g. Apache already uses 8080).

Reason to use same port: projects rarely need more than 1 webserver.
Reason not to: People can accidentally trigger conflicts between different projects and it's a bit hard to debug.

mikeland86 (approved):
Small nit, can you bump plugin version. We don't use it right now, but it gets saved in the lock file and in the future we can use it to suggest updates that can fix overwrite and fix config files.

Should we choose different ports for different services? (e.g. Apache already uses 8080).

Reason to use same port: projects rarely need more than 1 webserver.
Reason not to: People can accidentally trigger conflicts between different projects and it's a bit hard to debug.

Lagoja (approved):

plugins/caddy/Caddyfile
LucilleH: This is to change the default caddy http `80` and https port `443` port to be non privileged port `8800` and `4443`

PR #1109, opened 2023-06-07
[lockfile] add lockfile.Tidy and call from ensurePackagesAreInstalled

### LINKED ISSUE #1095: [Bug]: Changing a package version keeps the old package version in the `devbox.lock` file
**Current Behavior (bug)**

if I change `devbox.json` from `go@1.19` to `go@1.20`, and do `devbox install`, I still see the `go@1.19` in the devbox.lock (in addition to the new `go@1.20`)

**Expected Behavior (fix)**
Changing a package version should drop the old package version from `devbox.lock` file

savil:
## Summary

Problem:
if I change devbox.json from go@1.19 to go@1.20, and do devbox install, I still see the go@1.19 in the devbox.lock (in addition to the new go@1.20)

This PR ensures that when we update installed packages, we also update the lockfile
to have the set of packages that match those of the devbox.json config.

implementation note: I needed to rename lockfile.Packages to a different name
since devboxProject is embedded in the lockfile.File struct and we have a name
conflict.

Fixes #1095

## How was it tested?

- [x] will add a testscript unit test that matches the steps below:

```
# initial devbox.lock file
(devbox) savil@Savil-Srivastavas-MacBook-Pro ~/c/j/devbox (savil/update-lockfile-v2)> cat devbox.lock
{
  "lockfile_version": "1",
  "packages": {
    "actionlint@1.6.23": {
      "last_modified": "2023-03-31T22:52:29Z",
      "resolved": "github:NixOS/nixpkgs/242246ee1e58f54d2322227fc5eef53b4a616a31#actionlint",
      "version": "1.6.23"
    },
    "go@1.20": {
      "last_modified": "2023-05-01T16:53:22Z",
      "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#go",
      "version": "1.20.3"
    },
    "golangci-lint@1.52.2": {
      "last_modified": "2023-05-01T16:53:22Z",
      "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#golangci-lint",
      "version": "1.52.2"
    }
  }
}

# edit the go@1.20 to be go@1.19⏎
(devbox) savil@Savil-Srivastavas-MacBook-Pro ~/c/j/devbox (savil/update-lockfile-v2)> vim devbox.json
(devbox) savil@Savil-Srivastavas-MacBook-Pro ~/c/j/devbox (savil/update-lockfile-v2)> devbox install
Finished installing packages.


(devbox) savil@Savil-Srivastavas-MacBook-Pro ~/c/j/devbox (savil/update-lockfile-v2)> cat devbox.lock
{
  "lockfile_version": "1",
  "packages": {
    "actionlint@1.6.23": {
      "last_modified": "2023-03-31T22:52:29Z",
      "resolved": "github:NixOS/nixpkgs/242246ee1e58f54d2322227fc5eef53b4a616a31#actionlint",
      "version": "1.6.23"
    },
    "go@1.19": {
      "last_modified": "2023-05-01T16:53:22Z",
      "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#go_1_19",
      "version": "1.19.8"
    },
    "golangci-lint@1.52.2": {
      "last_modified": "2023-05-01T16:53:22Z",
      "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#golangci-lint",
      "version": "1.52.2"
    }
  }
}⏎
# Notice that the older go@1.20 is now gone
```

### DIFF
--- a/internal/lock/lockfile.go
+++ b/internal/lock/lockfile.go
@@ -23,7 +23,7 @@ type File struct {
 	resolver
 
 	LockFileVersion string              `json:"lockfile_version"`
-	PackageMap      map[string]*Package `json:"packages"`
+	Packages        map[string]*Package `json:"packages"`
 }
 
 type Package struct {
@@ -39,7 +39,7 @@ func GetFile(project devboxProject, resolver resolver) (*File, error) {
 		resolver:      resolver,
 
 		LockFileVersion: lockFileVersion,
-		PackageMap:      map[string]*Package{},
+		Packages:        map[string]*Package{},
 	}
 	err := cuecfg.ParseFile(lockFilePath(project), lockFile)
 	if errors.Is(err, fs.ErrNotExist) {
@@ -62,7 +62,7 @@ func (l *File) Add(pkgs ...string) error {
 
 func (l *File) Remove(pkgs ...string) error {
 	for _, p := range pkgs {
-		delete(l.PackageMap, p)
+		delete(l.Packages, p)
 	}
 	return l.Save()
 }
@@ -70,7 +70,7 @@ func (l *File) Remove(pkgs ...string) error {
 // Resolve updates the in memory copy for performance but does not write to disk
 // This avoids writing values that may need to be removed in case of error.
 func (l *File) Resolve(pkg string) (*Package, error) {
-	if entry, ok := l.PackageMap[pkg]; !ok || entry.Resolved == "" {
+	if entry, ok := l.Packages[pkg]; !ok || entry.Resolved == "" {
 		locked := &Package{}
 		var err error
 		if IsVersionedPackage(pkg) {
@@ -83,14 +83,14 @@ func (l *File) Resolve(pkg string) (*Package, error) {
 			// whatever hash is in the devbox.json
 			locked = &Package{Resolved: l.LegacyNixpkgsPath(pkg)}
 		}
-		l.PackageMap[pkg] = locked
+		l.Packages[pkg] = locked
 	}
 
-	return l.PackageMap[pkg], nil
+	return l.Packages[pkg], nil
 }
 
 func (l *File) ForceResolve(pkg string) (*Package, error) {
-	delete(l.PackageMap, pkg)
+	delete(l.Packages, pkg)
 	return l.Resolve(pkg)
 }
 
@@ -100,7 +100,7 @@ func (l *File) Save() error {
 		return nil
 	}
 
-	return cuecfg.WriteFile(lockFilePath(l), l)
+	return cuecfg.WriteFile(lockFilePath(l.devboxProject), l)
 }
 
 func (l *File) LegacyNixpkgsPath(pkg string) string {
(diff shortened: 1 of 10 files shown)


savil:
* main
  * **PR #1109** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1109" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1109?utm_source=stack-comment).

internal/lock/lockfile.go
```diff
@@ -127,6 +128,10 @@ func IsLegacyPackage(pkg string) bool {
 		!strings.HasPrefix(pkg, "/")
 }
 
+func (l *File) Tidy(project devboxProject) {
+	l.PackageMap = lo.PickByKeys(l.PackageMap, project.Packages())
```
savil, line 132: @mikeland86 not 100% confident that the keys in the lockfile are _intended_ to match the package-names in devbox.json. Can you confirm?

mikeland86, line 132: This doesn't need a parameter. The lockFile struct has access to devbox struct (i.e. `l.devboxProject.Packages()`)

mikeland86: Yeah they should always match. Right now `nix.Input` uses `Input.String()` instead of `Input.Raw` as a key (raw is the package name in devbox.json) but they always match for devbox packages. I think we should probably change those callsites to use `input.Raw` just to be sure (since for flakes `Raw` and `String()` might not be the same)

But for this use case this is safe.

internal/lock/lockfile.go
```diff
@@ -22,7 +23,7 @@ type File struct {
 	resolver
 
 	LockFileVersion string              `json:"lockfile_version"`
-	Packages        map[string]*Package `json:"packages"`
+	PackageMap      map[string]*Package `json:"packages"`
```
mikeland86, line 26: You don't need to rename this. See this draft https://github.com/jetpack-io/devbox/pull/1113 

You can access conflicting functions on structs by specifying the embedded struct explicitly (e.g. lock.devboxProject.Packages())

That said, when embedded structs get this complicated it's probably better to get rid of them (and just make them explicit fields) (not something that needs to happen in this PR though).

mikeland86 (approved):
LGTM, pending revert of `Packages` name change. You can also remove the `Packages()` function

internal/lock/lockfile.go
savil: oh good point

internal/lock/lockfile.go
savil: that draft was useful, thanks!

> That said, when embedded structs get this complicated it's probably better to get rid of them (and just make them explicit fields) (not something that needs to happen in this PR though).

yeah, that's what `savil/update-lockfile-v1` was about. But I thought it may receive pushback so took this more conservative approach. See #1120

internal/lock/lockfile.go
savil: oh I see. We should probably change that sooner than later...

PR #1114, opened 2023-06-07
[auth] Implement device flow auth

mikeland86:
## Summary

This adds a new hidden top level command `devbox auth`. It adds a few other commands:

```
devbox auth login
devbox auth logout
devbox auth refresh
devbox auth whoami
```

This infrastructure will be used to power `devbox global push/pull`. It is customizable so that a user could specify own auth0 tenant. 

## How was it tested?

Changed refresh token expiration to be 5 seconds so I could test out auto-refresh.

```
devbox auth login
devbox auth logout
devbox auth refresh
devbox auth whoami
```


loreto:
Questions on the UX and not the implementation:
1. Do we need `refresh`? Seems like an implementation detail that we shouldn't expose to users.
2. I don't remember: but are the login flows and signup flows the same? If not, would it be worth it to also have `devbox auth signup`

mikeland86:
`refresh` is only for debugging.  Trying to retrieve the user does refresh automatically.

`login` does both login and registration. This is easiest approach.

internal/auth/authenticator.go
```diff
@@ -0,0 +1,94 @@
...
+// tenancy.
+func NewAuthenticator() *Authenticator {
+	return &Authenticator{
+		ClientID: envir.GetValueOrDefault(
+			"DEVBOX_AUTH_CLIENT_ID",
+			"5PusB4fMm6BQ8WbTFObkTI0JUDi9ahPC",
+		),
+		Domain: envir.GetValueOrDefault(
+			"DEVBOX_AUTH_DOMAIN",
+			"auth.jetpack.io",
```
LucilleH, line 36: Does that mean if we ship this cli version, we'd have to use `auth.jetpack.io` even if we choose a different domain later on?

mikeland86: If we ever change auth domains, we would need to maintain this one as long as users are on this version of the CLI.

loreto: This code won't follow redirects?

internal/auth/auth.go
```diff
@@ -0,0 +1,305 @@
...
+		return 0, usererr.New("Access was denied")
+	}
+
+	// Unknown error
+	return 0, usererr.New("Unable to login")
+}
+
+func getAuthFilePath() string {
+	return xdg.StateSubpath(filepath.FromSlash("devbox/auth.json"))
+}
```
savil, line 305: is this mostly taken from the `jetpack` tool login code? what changes were made?

internal/auth/user.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, err
+		}
+	}
+
+	return &user{filesystemTokens: filesystemTokens, idToken: IDToken}, nil
+}
+
+func (u *user) String() string {
+	return u.Email()
+}
```
savil, line 53: I think this is fine for now, but eventually we'd need to move away from email to our own user-id as the true-identifier of a user (since a user may change their email)

savil (approved):
nice!

LucilleH:
> I think this is fine for now, but eventually we'd need to move away from email to our own user-id as the true-identifier of a user (since a user may change their email)

I find myself feeling a little uneasy with this PR, because we are shipping it as part of a CLI binary release, it is hard to change once we shipped it (verses something like devbox.sh, where we can change things as we see fit).

Once we ship this, does that mean we are locked to use auth0 as the auth provider? Is there a way to abstract this layer to the server side so we can easily swap things out? 🤔

internal/auth/auth.go
mikeland86: this stuff is new (launchpad doesn't store auth in state XDG I don't think, it also has a custom struct instead of just storing the auth0 response). The device code polling is inspired by launchpad but heavily modified in order to be simpler and not require a lot of copy pasting or importing (i.e. protos from that repo)

internal/auth/user.go
mikeland86: Yeah, this is mostly for printing info, not meant to be a canonical ID or anything

mikeland86:
As discussed with @loreto and @LucilleH this will go behind feature flag initially to give us a chance to test it out.

alvarogonzalez-packlink:
Is this Devbox Cloud specific, or cloud agnostic?

loreto:
> Is this Devbox Cloud specific, or cloud agnostic?
> 
> Edit, to clarify: is the idea I can, for example, push/pull config from a custom bucket my team owns, or just using Devbox Cloud?

@alvarogonzalez-packlink Great question.

This feature has nothing to do with our "remote developer environments", it's a different cloud feature to upload and share devbox environments with others. To upload to a cloud store you need to be authenticated. The plan is to build it in a cloud agnostic way; similar to how docker can talk to many different container registries. We're iterating on the feature behind a feature flag, so that we can iron out how to improve it and continue to make it agnostic.

Let me know if you have more thoughts or concerns: would love to hear them so we continue to improve the product!

alvarogonzalez-packlink:
Sorry @loreto , I think I was filtering merged PRs and missed your question.

I was checking this and #1128 , and being a bit confused by the "auth.jetpack.io" workflow, so that's why I asked if this was around Jetpack.io services only, or available for self-hosting.

I haven't checked much the auth workflow, and I don't know much Go and my AWS is a bit rusty 🤣 so I don't know exactly who's doing the auth. By now (and I understand you're still testing) it feels like Jetpack is somehow giving permissions to users to certain buckets, and maybe through the web login process handing the client some authentication key, and then the client uses that key to download, directly from the AWS bucket, the profile.

If I were doing simple self-hosting support, I would do a bunch of things:

- Let the user manage authz/authn (that is, they're in charge of creating users, buckets, roles, groups on their cloud platform)
- Maybe use a multi-platform / vendor-neutral object storage library to avoid importing a new cloud library each time I want to support a new cloud. Probably something like [go-storage](https://github.com/beyondstorage/go-storage). But, for example, the Terraform CLI I mention later, uses the official SDK for every storage backend it supports 🤷 
- Let that library use the default credentials in the environment, or specify a credential file

So configuration would be a bit like when you define a state backend for Terraform :

> **Note**
>  If you're not familiar with Terraform, the _state_ contains all the information about the status of your infrastructure the last time you run the Terraform tool. This can be stored in the Terraform Cloud service, or self managed by you in object storage or local path.

You're using Terraform's managed service? You [declare that in a file](https://developer.hashicorp.com/terraform/language/settings/terraform-cloud#usage-example) and then do `terraform login` (I think this is similar to your workflow):

```hcl
terraform {
  cloud {
    organization = "example_corp"
    ## Required for Terraform Enterprise; Defaults to app.terraform.io for Terraform Cloud
    hostname = "app.terraform.io"

    workspaces {
      tags = ["app"]
    }
  }
}
```

You self-manage? You [declare that bucket in a file](https://developer.hashicorp.com/terraform/language/settings/backends/s3#example-configuration), and then by yourself manage whatever permissions you need in that bucket:

```hcl
terraform {
  backend "s3" {
    bucket = "mybucket"
    key    = "path/to/my/key"
    region = "us-east-1"
  }
}
```

PR #1115, opened 2023-06-07
[update] update description for update command

hezhizhen:
## Summary

1. update description for `update` command
2. add copyright automatically

## How was it tested?

devbox update --help
devbox global update --help

### DIFF
--- a/internal/boxcli/update.go
+++ b/internal/boxcli/update.go
@@ -6,6 +6,7 @@ package boxcli
 import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
 	"go.jetpack.io/devbox"
 )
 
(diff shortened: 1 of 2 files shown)


internal/boxcli/update.go
```diff
@@ -18,10 +19,10 @@ func updateCmd() *cobra.Command {
 
 	command := &cobra.Command{
 		Use:   "update [pkg]...",
-		Short: "Updates packages in your devbox",
-		Long: "Updates one, many, or all packages in your devbox. " +
+		Short: "Update packages in your devbox",
+		Long: "Update one, many, or all packages in your devbox. " +
 			"If no packages are specified, all packages will be updated. " +
-			"Only updates versioned packages (e.g. `python@3.11`), not packages that are pinned to a nix channel (e.g. `python3`)",
+			"Only update versioned packages (e.g. `python@3.11`), not packages that are pinned to a nix channel (e.g. `python3`)",
```
mikeland86, line 25: updates is a bit more accurate because it is describing what it does instead of being imperative. (the first 2 uses can be imperative so `update` is ok.)

Either way, this functionality is actually changing in https://github.com/jetpack-io/devbox/pull/1107 so I'll update it there.

mikeland86 (approved):

internal/boxcli/update.go
hezhizhen: Then I think it should be `Only versioned packages are updated ...`

PR #1116, opened 2023-06-07
Combine checks and parsing of versioned packages

hezhizhen:
## Summary

1. Move `IsVersionedPackage` from `lock` to `fileutil`
2. Change `IsVersionedPackage` to `ParseVersionedPackage` to combine checking and parsing of packages (return name and version and a boolean value indicating if the package is versioned)
3. Reuse `ParseVersionedPackage` in versioned packages checking and parsing cases.

## How was it tested?

devbox update python
devbox update python@3.10


hezhizhen:
Wait https://github.com/jetpack-io/devbox/pull/1115 to be merged first.

internal/fileutil/fileutil.go
```diff
@@ -55,3 +55,10 @@ func FileContains(path string, substring string) (bool, error) {
 	}
 	return strings.Contains(string(data), substring), nil
 }
+
+// ParseVersionedPackage checks if the given package is a versioned package (`python@3.10`)
+// and returns its name and version
+func ParseVersionedPackage(pkg string) (string, string, bool) {
```
mikeland86, line 61: fileutil doesn't seem like the right place for this. Maybe `input.go` in nix package? (I think this creates a dependency cycle though) Or we can create a new `devpkg` (I've been meaning to move `nix.Input` to own package, but that's out of scope for this PR)

mikeland86 (changes requested):

internal/fileutil/fileutil.go
hezhizhen: That's right. I'll create a new `devpkg` first and then move `nix.Input` to that package and rename that package.

hezhizhen: I'll learn more about input and then think about adding a package for nix packages.

hezhizhen:
@mikeland86 PTAL

mikeland86 (approved):
Nice!

savil (approved):

PR #1117, opened 2023-06-07
[searcher] Merge Search and SearchVersion

hezhizhen:
## Summary

1. Merge `Search` and `SearchVersion` into `Search` (if version is not provided, do not use `v=<version>` in search
2. Remove unused `Exists` function

## How was it tested?

devbox search python

### DIFF
--- a/internal/searcher/client.go
+++ b/internal/searcher/client.go
@@ -32,6 +32,7 @@ var cachedClient *client
 func Client() *client {
 	if cachedClient == nil {
 		endpoint, _ := url.JoinPath(searchHost(), "search")
+		// TODO: how to handle error
 		cachedClient = &client{
 			endpoint: endpoint,
 		}
@@ -39,16 +40,24 @@ func Client() *client {
 	return cachedClient
 }
 
-func (c *client) Search(query string) (*SearchResult, error) {
-	return execSearch(c.endpoint + "?q=" + url.QueryEscape(query))
+func (c *client) Search(query string, options ...func() string) (*SearchResult, error) {
+	if query == "" {
+		return nil, fmt.Errorf("query should not be empty")
+	}
+
+	searchURL := c.endpoint + "?q=" + url.QueryEscape(query)
+
+	for _, op := range options {
+		searchURL += op()
+	}
+
+	return execSearch(searchURL)
 }
 
-func (c *client) SearchVersion(query, version string) (*SearchResult, error) {
-	return execSearch(
-		c.endpoint +
-			"?q=" + url.QueryEscape(query) +
-			"&v=" + url.QueryEscape(version),
-	)
+func WithVersion(version string) func() string {
+	return func() string {
+		return "&v=" + url.QueryEscape(version)
+	}
 }
 
 func (c *client) Resolve(pkg string) (*lock.Package, error) {
@@ -56,7 +65,7 @@ func (c *client) Resolve(pkg string) (*lock.Package, error) {
 	if version == "" {
 		return nil, usererr.New("No version specified for %q.", name)
 	}
-	result, err := c.SearchVersion(name, version)
+	result, err := c.Search(name, WithVersion(version))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


internal/searcher/searcher.go
```diff
@@ -15,9 +15,9 @@ import (
 
 func SearchAndPrint(w io.Writer, query string) error {
 	c := Client()
-	result, err := c.Search(query)
+	result, err := c.Search(query, "")
```
mikeland86, line 18: I split them up mostly to void this ^ 

passing in empty args hurts readability (especially strings and booleans) because the code reader now needs to click in to the function to understand what params are. 

One pattern that avoids this but uses a lot of boilerplate is "WithOptions":

```golang
result, err := c.Search(query)
result, err := c.Search(query, WithVersion(version))
```

mikeland86 (changes requested):

hezhizhen:
@mikeland86 PTAL

hezhizhen:
@mikeland86 Any comments?

internal/searcher/client.go
```diff
@@ -32,31 +32,40 @@ var cachedClient *client
 func Client() *client {
 	if cachedClient == nil {
 		endpoint, _ := url.JoinPath(searchHost(), "search")
+		// TODO: how to handle error
```
mikeland86, line 35: just panic is probably fine since this error can't happen on the default host. It can only happen on a custom host. I guess we could also return an error here, but that's a bigger change.

internal/searcher/client.go
```diff
@@ -32,31 +32,40 @@ var cachedClient *client
...
+	return execSearch(searchURL)
 }
 
-func (c *client) SearchVersion(query, version string) (*SearchResult, error) {
-	return execSearch(
-		c.endpoint +
-			"?q=" + url.QueryEscape(query) +
-			"&v=" + url.QueryEscape(version),
-	)
+func WithVersion(version string) func() string {
```
savil, line 57: I think defining a `type SearchOption func() string` would help here.

As-is, `searcher.WithVersion` is a ambiguous function since its not clear what context it needs to be used in. By adding this type definition, readers can see the return type `SearchOption` and then look for functions that accept `SearchOption` parameters.

mikeland86 (approved):
LGTM

cc: @gcurtis 

I'm away from keyboard so not gonna merge but @savil @gcurtis feel free to merge.

internal/searcher/client.go
```diff
@@ -32,31 +32,40 @@ var cachedClient *client
...
-	return execSearch(c.endpoint + "?q=" + url.QueryEscape(query))
+func (c *client) Search(query string, options ...func() string) (*SearchResult, error) {
+	if query == "" {
+		return nil, fmt.Errorf("query should not be empty")
+	}
+
+	searchURL := c.endpoint + "?q=" + url.QueryEscape(query)
+
+	for _, op := range options {
+		searchURL += op()
```
mikeland86, line 51: Another possible signature for the option function is a function that takes the endpoint, modifies it and returns the new one. But this is OK as well. If we ever add more options we can generalize if needed

hezhizhen: > Another possible signature for the option function is a function that takes the endpoint, modifies it and returns the new one. But this is OK as well. If we ever add more options we can generalize if needed

Yeah, I first came up with this one. But here's a problem: each option needs to handle some similar cases for 0 or more queries. So I finally decided to return the query string in the option to  be appended to the endpoint.

If we have more options, we can see which will be better.

internal/searcher/client.go
hezhizhen: Then I think we can return a user error to warn users for invalid custom hosts?

internal/searcher/client.go
hezhizhen: Exactly.

hezhizhen: Now that this PR is merged, I'll create a new PR to update it.

internal/searcher/client.go
hezhizhen: https://github.com/jetpack-io/devbox/pull/1173

internal/searcher/client.go
hezhizhen: https://github.com/jetpack-io/devbox/pull/1173

PR #1125, opened 2023-06-09
Fix stacks examples

LucilleH:
## Summary
Cleaning up the stacks examples as some of them are out-of-date.

## How was it tested?
on devbox.sh


examples/stacks/lapp-stack/my_app/config.php
```diff
@@ -1,5 +1,5 @@
 <?php
-	$db_hostname	= "127.0.0.1";
+	$db_hostname	= "0.0.0.0";
```
LucilleH, line 2: unclear if this works, but hardcoding it to `127.0.0.1` certainly does not work in cloud environments

Lagoja (approved):

PR #1127, opened 2023-06-09
Use `devbox.lock` instead of `Devbox.lock`

mikenikles:
## Summary

Minor copy change to use the correct file name.

## How was it tested?

Manually read the content.


savil (approved):
Thanks!

PR #1128, opened 2023-06-09
[cloud-profile] Implement push/pull of cloud profile

mikeland73:
## Summary

This is stacked on https://github.com/jetpack-io/devbox/pull/1114

This implements cloud based global profiles using s3, our existing auth tokens, and identity federation.

When user is logged in, we save an IDToken. Tokens are then used to assume an AWS role that allows the user to push and pull data to a path reserved to them.

The two commands that do this right now are:

```bash
devbox global push
devbox global pull
```

(without a url/path)

It also improves how we overwrite existing profiles by checking for a profile before downloading the new profile.

## How was it tested?

```
devbox global push
devbox global pull
```

(inspected bucket).

* Also tried pushing/pulling to path that user did not have access to and it failed.
* If you push/pull without being logged in you get an error message.


LucilleH:
This also makes me feel uneasy like the other PR (See comment https://github.com/jetpack-io/devbox/pull/1114#issuecomment-1585286800)

I wonder if there's a way to abstract this aws layer out to the server side, so that the devbox CLI stay as dumb as possible?

mikeland86:
@LucilleH following today's conversation, I'll put this behind feature flag.

FWIW, I think there's 2 paths we can take to make this more generic:

* You should be able to enable S3 pushing to your own with a env var (e.g. `DEVBOX_S3_BUCKET`). 
* We can and should eventually move the STS token creation server side to simplify client code, reduce binary size, and allow for new features (e.g. shared team profiles)

I think we can do both of these as follow ups.

loreto (approved):
LGTM

PR #1129, opened 2023-06-09
[update] UX fixes to update path

mikeland86:
## Summary

Fixes/refactors:

* If legacy package is global, show correct update message
* If a legacy package is updated, don't try to update it again (which will show warning because it is not versioned)
* Made `ShowWarnings` -> `IgnoreWarnings` because we ignore in the vast majority of cases. Fixed a few cases that were accidentally ignored.

## How was it tested?

* Modified global config to have "curl" pacakge.
* Saw `devbox global update` and ran it.
* Inspected config.

Repeated process for non-global as well.

### DIFF
--- a/internal/impl/devbox_test.go
+++ b/internal/impl/devbox_test.go
@@ -43,10 +43,9 @@ func testShellPlan(t *testing.T, testPath string) {
 		assert := assert.New(t)
 
 		_, err := Open(&devopt.Opts{
-			Dir:          baseDir,
-			Writer:       os.Stdout,
-			ShowWarnings: true,
-			Pure:         false,
+			Dir:    baseDir,
+			Writer: os.Stdout,
+			Pure:   false,
 		})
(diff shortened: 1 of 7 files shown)


savil (approved):

internal/impl/devbox_test.go
```diff
@@ -43,10 +43,9 @@ func testShellPlan(t *testing.T, testPath string) {
 		assert := assert.New(t)
 
 		_, err := Open(&devopt.Opts{
-			Dir:          baseDir,
-			Writer:       os.Stdout,
-			ShowWarnings: true,
```
mohsenari, line 48: We don't need `ignoreWarnings: false` for tests?

mohsenari (approved):

internal/impl/devbox_test.go
mikeland86: false is default

PR #1139, opened 2023-06-12
[ux] Add FNotice

hezhizhen:
## Summary

1. add `FNotice` and reuse it

## How was it tested?

1. Install an old version of devbox
2. Run `devbox` to trigger version checking
3. See the notice message

### DIFF
--- a/internal/ux/messages.go
+++ b/internal/ux/messages.go
@@ -15,6 +15,11 @@ func Fsuccess(w io.Writer, format string, a ...any) {
 	fmt.Fprintf(w, format, a...)
 }
 
+func Finfo(w io.Writer, format string, a ...any) {
+	color.New(color.FgYellow).Fprint(w, "Info: ")
+	fmt.Fprintf(w, format, a...)
+}
+
 func Fwarning(w io.Writer, format string, a ...any) {
 	color.New(color.FgHiYellow).Fprint(w, "Warning: ")
 	fmt.Fprintf(w, format, a...)
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/MakeNowJust/heredoc/v2 v2.0.1
 	github.com/MicahParks/keyfunc/v2 v2.1.0
-	github.com/a8m/envsubst v1.4.2
 	github.com/alessio/shellescape v1.4.1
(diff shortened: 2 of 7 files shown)


hezhizhen:
@savil @mikeland86 Any comments?

internal/ux/messages.go
```diff
@@ -10,17 +10,22 @@ import (
...
 	fmt.Fprintf(w, format, a...)
 }
 
-func Ferror(w io.Writer, format string, a ...any) {
+func FError(w io.Writer, format string, a ...any) {
 	color.New(color.FgHiRed).Fprint(w, "Error: ")
 	fmt.Fprintf(w, format, a...)
 }
+
+func FNotice(w io.Writer, format string, a ...any) {
```
hezhizhen, line 28: @savil WDYT?

internal/boxcli/create.go
```diff
@@ -71,7 +72,7 @@ func runCreateCmd(
 		return err
 	}
 
-	ux.Fsuccess(
+	ux.FSuccess(
```
savil, line 75: 🤔 isn't the golang naming convention better followed by the previous code? For example, consider `fmt.Fprintf` and `fmt.Sprintf` and so on...

internal/ux/messages.go
```diff
@@ -10,17 +10,22 @@ import (
...
+	// TODO: Add a notice prefix?
+	color.New(color.FgYellow).Fprintf(w, format, a...)
```
savil, line 30: 🤔 Yeah, I think we should change this to have a prefix It'll fit the convention of the rest of our messages.

Thinking a bit more, I have a slight preference for changing "notice" to "info" since its easier to read at a quick glance. 

Could we rename this to `Finfo` (lowercase i) and have a yellow "Info: " prefix?

savil (commented):
@hezhizhen thanks for your attention on this!

I had a couple of comments inline:

internal/boxcli/create.go
hezhizhen: > 🤔 isn't the golang naming convention better followed by the previous code? For example, consider `fmt.Fprintf` and `fmt.Sprintf` and so on...

Oh, that makes sense.

internal/ux/messages.go
hezhizhen: Done. PTAL

savil (approved):
Thanks!

PR #1140, opened 2023-06-12
[nix] Check if nix binary exists after installing it

hezhizhen:
## Summary

1. Add some comments
2. Re-check after installing nix

## How was it tested?

Run any commands of devbox


hezhizhen:
Not necessary.

PR #1143, opened 2023-06-12
[pure-shell] making sure nix is in path in pure mode

mohsenari:
## Summary
This is a follow up PR from as a result of [this discussion](https://github.com/jetpack-io/devbox/pull/1084#discussion_r1224750558). 
Note, xdg dirs don't keep any nix executables as far as I know. Single user and multi user nix installation put the nix executable in `~/.nix-profile/bin/` and `/nix/var/nix/profiles/default/bin` respectively.

Also added tests for devbox run pure mode. For `devbox shellenv`, pure mode doesn't really make sense to write tests for, and for `devbox shell` it's not possible to write testscript tests.

## How was it tested?
- in a single user nix installed env, run `./devbox shell --pure` 
- `devbox add which`
- `which nix` should point to `~/.nix-profile/bin/nix`
- in a multi-user nix, same test should point to `/nix/var/nix/profiles/default/bin/nix`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -1188,8 +1188,9 @@ func (d *Devbox) convertEnvToMap(currentEnv []string) (map[string]string, error)
 			continue
 		}
 		// handling special cases to for pure shell
-		// Passing HOME USER and DISTPLAY for pure shell to leak through
+		// Passing HOME USER and DISPLAY for pure shell to leak through
 		// otherwise devbox binary won't work - this matches nix
+		// We include PATH to find the nix installation. It is cleaned for pure mode below
 		if !d.pure || key == "HOME" || key == "USER" || key == "DISPLAY" || key == "PATH" {
 			env[key] = val
 		}
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -746,22 +746,9 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 			continue
 		}
 		// handling special cases to for pure shell
-		if d.pure {
-			// Finding nix executables in path and passing it through
-			// Needed for devbox commands inside pure shell to work
-			if key == "PATH" {
-				nixInPath, err := findNixInPATH(env)
-				if err != nil {
-					return nil, err
-				}
-				env[key] = nixInPath
-			}
-			// Passing HOME USER and DISTPLAY for pure shell to leak through
-			// otherwise devbox binary won't work - this matches nix
-			if key == "HOME" || key == "USER" || key == "DISPLAY" {
-				env[key] = val
-			}
-		} else {
+		// Passing HOME USER and DISTPLAY for pure shell to leak through
+		// otherwise devbox binary won't work - this matches nix
+		if !d.pure || key == "HOME" || key == "USER" || key == "DISPLAY" || key == "PATH" {
 			env[key] = val
 		}
 	}
@@ -774,6 +761,16 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	}
 
 	currentEnvPath := env["PATH"]
+	if d.pure {
+		// Finding nix executables in path and passing it through
+		// Needed for devbox commands inside pure shell to work
+		nixInPath, err := findNixInPATH(env)
+		if err != nil {
+			return nil, err
+		}
+		currentEnvPath = nixInPath
+	}
+
 	debug.Log("current environment PATH is: %s", currentEnvPath)
 	// Use the original path, if available. If not available, set it for future calls.
 	// See https://github.com/jetpack-io/devbox/issues/687
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -745,9 +745,10 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 		if ignoreCurrentEnvVar[key] {
 			continue
 		}
-		// Passing HOME USER and DISTPLAY for pure shell to leak through
-		// otherwise devbox binary won't work - this matches nix
-		if !d.pure {
+		// handling special cases to for pure shell
+		if d.pure {
+			// Finding nix executables in path and passing it through
+			// Needed for devbox commands inside pure shell to work
 			if key == "PATH" {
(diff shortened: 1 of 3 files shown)


internal/impl/devbox.go
```diff
@@ -760,8 +760,15 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
...
-		currentEnvPath = "/nix/var/nix/profiles/default/bin"
+	if d.pure { // make nix available inside pure shell - necessary for devbox commands to work
+		singleUserNixBin := fmt.Sprintf("%s/.nix-profile/bin", env["HOME"])
+		multiUserNixBin := "/nix/var/nix/profiles/default/bin"
+		// ~/.nix-profile/bin takes precedent because of this from Nix reference manual:
+		// "You generally wouldn’t have /nix/var/nix/profiles/some-profile/bin in your PATH.
+		// Rather, there is a symlink ~/.nix-profile that points to your current profile.
+		// This means that you should put ~/.nix-profile/bin in your PATH
+		// (and indeed, that’s what the initialisation script /nix/etc/profile.d/nix.sh does)."
+		currentEnvPath = fmt.Sprintf("%s:%s", singleUserNixBin, multiUserNixBin)
```
mikeland86, line 771: From offline chat: IMO we should only keep stuff in path if it's already there and not add it if it's missing to avoid polluting the path.

mikeland86 (commented):
@mohsenari see nix XDG code: https://github.com/NixOS/nix/commit/2384d360839e27edb3c928da858ec911415c8b4d

(I have not tested it, but I assume the goal was to put the nix binaries in the custom XDG dir)

internal/impl/devbox.go
mohsenari: Fair enough. I updated the PR after reading more about xdg stuff and tested the PR in a fresh ubuntu install with XDG_STATE_HOME set to an arbitrary dir, and then installing nix.

internal/impl/devbox.go
```diff
@@ -1190,3 +1176,34 @@ func (d *Devbox) addHashToEnv(env map[string]string) error {
...
+		if !found {
+			return nil, errors.Errorf("expected \"=\" in keyval: %s", kv)
+		}
+		if ignoreCurrentEnvVar[key] {
+			continue
+		}
+		// handling special cases to for pure shell
+		// Passing HOME USER and DISTPLAY for pure shell to leak through
+		// otherwise devbox binary won't work - this matches nix
+		if !d.pure || key == "HOME" || key == "USER" || key == "DISPLAY" || key == "PATH" {
```
savil, line 1193: What's the motivation for including `PATH` even in `--pure` mode?

PATH seems to be one of the critical env-vars that we definitely want to omit from the external environment for pure mode, because then all sorts of external programs become callable inside the pure shell.

savil: Oh I see. Could you add a small comment like:
```
// We include PATH to find the nix installation. It is cleaned for pure mode below.
```
?

internal/impl/devbox.go
```diff
@@ -1190,3 +1176,34 @@ func (d *Devbox) addHashToEnv(env map[string]string) error {
...
+	for _, kv := range currentEnv {
+		key, val, found := strings.Cut(kv, "=")
+		if !found {
+			return nil, errors.Errorf("expected \"=\" in keyval: %s", kv)
+		}
+		if ignoreCurrentEnvVar[key] {
+			continue
+		}
+		// handling special cases to for pure shell
+		// Passing HOME USER and DISTPLAY for pure shell to leak through
```
savil, line 1191: typo: DISPLAY

savil (approved):
cool! thanks for adding the test

internal/impl/devbox.go
```diff
@@ -1190,3 +1176,35 @@ func (d *Devbox) addHashToEnv(env map[string]string) error {
 	}
 	return err
 }
+
+func (d *Devbox) convertEnvToMap(currentEnv []string) (map[string]string, error) {
```
mikeland86, line 1180: This function is doing much more than converting env to map. We should name it accordingly. It sounds like a generic transformation. Always err on the side of over-specifying.

internal/impl/shell.go
```diff
@@ -411,3 +411,21 @@ func filterPathList(pathList string, keep func(string) bool) string {
 	}
 	return strings.Join(filtered, string(filepath.ListSeparator))
 }
+
+func findNixInPATH(env map[string]string) (string, error) {
+	defaultSingleUserNixBin := fmt.Sprintf("%s/.nix-profile/bin", env["HOME"])
+	defaultMultiUserNixBin := "/nix/var/nix/profiles/default/bin"
+	xdgNixBin := xdg.StateSubpath("/nix/profile/bin")
+	pathElements := strings.Split(env["PATH"], ":")
```
mikeland86, line 419: Use `filepath.SplitList`

internal/impl/shell.go
```diff
@@ -411,3 +411,21 @@ func filterPathList(pathList string, keep func(string) bool) string {
 	}
 	return strings.Join(filtered, string(filepath.ListSeparator))
 }
+
+func findNixInPATH(env map[string]string) (string, error) {
+	defaultSingleUserNixBin := fmt.Sprintf("%s/.nix-profile/bin", env["HOME"])
```
mikeland86, line 416: This should use `os.UserHomeDir`

internal/impl/shell.go
```diff
@@ -411,3 +411,21 @@ func filterPathList(pathList string, keep func(string) bool) string {
...
+	defaultMultiUserNixBin := "/nix/var/nix/profiles/default/bin"
+	xdgNixBin := xdg.StateSubpath("/nix/profile/bin")
+	pathElements := strings.Split(env["PATH"], ":")
+	debug.Log("path elements: %v", pathElements)
+	for _, el := range pathElements {
+		if el == xdgNixBin ||
+			el == defaultSingleUserNixBin ||
+			el == defaultMultiUserNixBin {
+			return el, nil
+		}
```
mikeland86, line 426: I sort of think we should include all that are specified instead of just the first. It's not a big deal, but it would be more defensive code.

mohsenari: @mike isn't your suggestion contradicting your comment [here](https://github.com/jetpack-io/devbox/pull/1143#discussion_r1227248076)?

PR #1146, opened 2023-06-13
fix: pid file was set to log file path

JoyceBabu:
## Summary

The PID file was set to the log file path.


ipince (approved):
Thank you for your first PR!

PR #1148, opened 2023-06-13
Remove virtualenv data xdg directory symlinks

LucilleH:
## Summary
Remove virtualenv data xdg directory symlinks. The current state is quite hard to reason about for the following reason:

### Diverging source of truth
We have `.devbox/virtualenv` which supposedly should be the source of truth for plugins. However, upon inspection, it looks like the paths and variables set for `{{ .Virtualenv }}` is pointing at my XDG directory. 

Because of such divergence, any operations that directly modifies the XDG directory is not reflected in the `.devbox/virtualenv` directory. 

However, the rest of the devbox CLI regards `.devbox/virtualenv` as the source of truth.

This is the root cause of quite a few plugin errors such as mariadb, drupal, and python venv.

### Obscure XDG path
The reason we have a hash directory in our XDG path is to avoid path collision. But since I use devbox so often, my XDG paths become a clown show:
Screenshot 2023-06-13 at 12 18 10 PM

This makes it impossible to debug plugins.

To get around the paths too long error described in https://github.com/jetpack-io/devbox/pull/874, I override the env variable that points to the `.sock` path in our `run_test` scripts.

### TODO

- [x]  Clean up symlink creation functions
- [x] Remove symlink tests

## How was it tested?
```
devbox run build
cd examples/databases/mariadb
devbox install
devbox services up
```


examples/databases/postgres/devbox.json
```diff
@@ -1,8 +1,8 @@
 {
   "packages": [
-    "postgresql"
+    "postgresql@latest"
```
savil, line 3: I will be surprised this works in testscripts in our CICD without needing to set PGHOST explicitly to a short directory

internal/plugin/plugin.go
```diff
@@ -95,7 +86,7 @@ func (m *Manager) create(
 	name := pkg.CanonicalName()
 
 	// Always create this dir because some plugins depend on it.
-	if err = createDir(filepath.Join(m.ProjectDir(), VirtenvPath, name)); err != nil {
+	if err = createDir(filepath.Join(virtenvPath, name)); err != nil {
```
savil, line 89: unrelated to this line, but github won't let me comment on other lines.

We could remove this line `strings.HasPrefix(filePath, virtenvPath)` in `shouldCreateFile`, and drop the `virtenvPath` parameter:
https://github.com/jetpack-io/devbox/blob/aff03a352fc674ac1b0d68ed78f68ddb24c68612/internal/plugin/plugin.go#L273

savil (approved):
Thanks @LucilleH !

examples/databases/postgres/devbox.json
savil: (not this line specifically, I meant more generally that i'll be surprised the postgres example passes.

examples/databases/mariadb/devbox.json
```diff
@@ -9,6 +9,8 @@
         "mysql -u devbox_user -p -D devbox_lamp"
       ],
       "test_db_setup": [
+        "mkdir -p /tmp/devbox/mariadb/run",
+        "export MYSQL_UNIX_PORT=/tmp/devbox/mariadb/run/mysql.sock",
```
mikeland86, line 13: love it

mikeland86 (approved):
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

examples/databases/postgres/devbox.json
LucilleH: You are right! I need to overwrite it.

examples/databases/mariadb/devbox.json
savil: I had to hold my nose on this to unblock Lucille, because these examples should show the ideal code a user would write, and this is a workaround we're implementing for our specific testscripts CICD tests.

PR #1149, opened 2023-06-13
[CLI] Removed DISPLAY & USER to leak to pure shell

mohsenari:
## Summary
TSIA

## How was it tested?
- compile
- `./devbox shell --pure`
- `echo $USER`
- `echo $DISPLAY`

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -1193,8 +1193,7 @@ func (d *Devbox) convertEnvToMap(currentEnv []string) (map[string]string, error)
 			continue
 		}
 		// handling special cases to for pure shell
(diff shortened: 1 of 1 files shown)


internal/impl/devbox.go
```diff
@@ -1193,10 +1193,10 @@ func (d *Devbox) convertEnvToMap(currentEnv []string) (map[string]string, error)
 			continue
 		}
 		// handling special cases to for pure shell
-		// Passing HOME USER and DISPLAY for pure shell to leak through
+		// Passing HOME for pure shell to leak through
```
savil, line 1196: lets drop the `this matches nix` comment below?

savil (approved):
thanks!

PR #1150, opened 2023-06-13
[bugfix, cleanup] Finish moving plugin services to process compose

ipince:
## Summary
Removes `services` stanza from the plugin's json definition, in favor of using `process-compose.yaml`.

* Fixes bug where, in doing `devbox services ls`, we would only return the _first_ service defined in a plugin, and would call it the name of the plugin (instead of the name of the service). This had some downstream effects as well.
* Removes creation of wrappers of the form `<svc>-start|stop-service`. They were using the json-based definitions of the services, but `devbox service start <svc>` was using the process-compose definition, so they could diverge. Some wrappers were not being generated either, due to the bug above.
* Deletes obsolete code.
* Updates plugins that define services and bump their versions. Since plugins are packaged in the CLI, and the process-compose yamls were already being written before this change, I think this change is backwards-compatible when it comes to plugins.

Reviewers, please advise on whether this change would affect service port-forwarding.

## How was it tested?
Ran `devbox services ls|start|stop|up` with different plugins installed and verified they worked as expected.

## Backwards-compatible?
NO, because we're removing the service wrappers (e.g. `redis-start-service`). I don't think they're being used, but @mikeland86 please confirm?  If needed, I can replace them with `devbox services start redis` or something like that.

### DIFF
--- a/internal/services/services.go
+++ b/internal/services/services.go
@@ -1,105 +1,11 @@
 // Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
 // Use of this source code is governed by the license in the LICENSE file.
 
-//lint:file-ignore U1000 Ignore unused function temporarily for debugging
 package services
 
-import (
-	"encoding/json"
-	"fmt"
-	"io"
-	"os"
-
-	"github.com/a8m/envsubst"
-	"github.com/fatih/color"
-	"github.com/pkg/errors"
-
-	"go.jetpack.io/devbox/internal/envir"
-)
-
-type Services map[string]Service
+type Services map[string]Service // name -> Service
 
 type Service struct {
-	Name               string            `json:"name"`
-	Env                map[string]string `json:"-"`
-	RawPort            string            `json:"port"`
-	Start              string            `json:"start"`
-	Stop               string            `json:"stop"`
+	Name               string
 	ProcessComposePath string
-}
-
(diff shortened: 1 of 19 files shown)


Lagoja (approved):
LGTM, but i'd check with Mike on the bits that were removed

internal/services/services.go
```diff
@@ -1,105 +1,11 @@
...
-// TODO: (john) Since moving to process-compose, our services no longer use the old `toggleServices` function. We'll need to clean a lot of this up in a later PR.
-
-type serviceAction int
-
-const (
-	startService serviceAction = iota
-	stopService
-)
-
-func printProxyURL(w io.Writer, services Services) error { // TODO: remove it?
```
mikeland86, line 40: The callsite to this was accidentally removed in https://github.com/jetpack-io/devbox/pull/836/files (I looked it up because I was surprised this was unused)

The purpose of this endpoint is to print the URL the user can use to proxy to their devbox.sh service. I think it's an important part of making devbox cloud usable. 

That said, currently only a single plugin declares a port, so we may want to just show the generic message and figure out how to show the port later.

mikeland86 (approved):
This is great! See my comment about `printProxyURL` which was accidentally removed a while back. I think we should add it in even if just generic.

We can figure out a different way of finding the port.

internal/services/services.go
ipince: I'll address this in a separate PR.

PR #1093, opened 2023-06-03
[update] call createWrappers

savil:
## Summary

`devbox update` should ensure the wrappers point to the latest binaries of any updated packages.

## How was it tested?

- take a devbox.json having go@1.20 package  (such as the devbox-repo!)
- do `devbox shell` 
- confirm go version 1.20 via `go version`
- manually set a package to go@1.19
- do `devbox update`
- run `go version` and see the updated version

BEFORE: would continue to show `go 1.20` version
AFTER: shows `go 1.19` version

PR #1151, opened 2023-06-13
[bin wrappers] add --omit-wrappers-from-path to shellenv to prevent a binary from invoking a wrapped-binary.

### LINKED ISSUE #1142: [Bug]: Wrappers interfere with `iex -s mix`
**Current Behavior (bug)**

Trying to run `iex -s mix <target>` raises the following exception:

```
iex -S mix phx.server
Erlang/OTP 25 [erts-13.1.4] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

** (SyntaxError) .devbox/virtenv/.wrappers/bin/mix:10:17: invalid character "w" after number 78. If you intended to write a number, make sure to separate the number from the character (using comma, space, etc). If you meant to write a function name or a variable, note that identifiers in Elixir cannot start with numbers. Unexpected token: w
    |
 10 | exec /nix/store/78wdiq951nz7hkhsxnsrmcjrwsvw3ydn-elixir-1.14.3/bin/mix "$@"
    |                 ^
    (elixir 1.14.3) lib/code.ex:1260: Code.require_file/2
```

This is caused by the `iex` parsing the Devbox wrapped mix binary as Elixir code, and erroring out when it encounters the Nix hash. The line above works when the wrapper is bypassed and the store path is referenced directly

`iex -S /nix/store/78wdiq951nz7hkhsxnsrmcjrwsvw3ydn-elixir-1.14.3/bin/mix phx.server`

**Expected Behavior (fix)**
Wrappers should not interfere with binary execution. This bug could affect other situations besides `iex`, so a more general solution for wrappers may be needed

savil:
## Summary

**Motivation**
In elixir projects, `iex -S mix <script>` has the `iex` interpreter invoking the
`mix` build tool. However, the `mix` build tool has been wrapped by devbox in a bash script.
And this leads to a failure as `iex` is unable to interpret a bash script.

**Fix**
Binaries should be able to directly invoke other installed binaries.

We implemented the binary-wrappers so that the shell-environment is correctly 
configured prior to the installed binary being invoked. In this case, the `iex`
binary wrapper will run thereby correctly configuring the environment. Any subsequent
binary invoked in the same sub-shell will inherit this up-to-date environment. So, 
the other binaries can be directly invoked (instead of the binary-wrapper scripts).

**Implementation**
In this PR, we change the `PATH` that is set within a binary-wrapper.

Specifically, the `PATH` has the form:
```
<project-dir>/.devbox/virtenv/.wrappers/bin : <project-dir>/.devbox/nix/profile/default/bin
```
Within the binary-wrapper, we now omit the `.wrappers/bin` directory from PATH
by using a new hidden flag `--omit-wrappers-from-path` on `devbox shellenv`.

**perf analysis**

I expect the perf impact to be minimal.

1. Speed up: Binaries invoking other binaries will be a bit faster since they no longer first execute the binary-wrapper.

2. Slowdown: we remove one layer of caching. Previously, we'd check the shellenv-hash in the binary wrapper, and skip invoking `devbox shellenv` if the hash was up-to-date. Now, on every invocation of a binary-wrapper we'll call `devbox shellenv`. This should still be fast due to the internal caching we do within `devbox shellenv`.

Fixes #1142

## How was it tested?

BEFORE: doing `devbox run -- iex -S mix` would result in the error in the linked issue.

AFTER: 
```
> devbox run -- iex -S mix
* creating .nix-mix/archives/hex-2.0.6
* creating .nix-mix/elixir/1-14/rebar3
Resolving Hex dependencies...
Resolution completed in 0.015s
Unchanged:
  cowboy 2.9.0
  cowlib 2.11.0
  ranch 1.8.0
All dependencies are up to date
Erlang/OTP 25 [erts-13.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

Hello World!
Interactive Elixir (1.14.4) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Goodbye World

15:14:23.204 [notice] Application elixir_hello exited: shutdown

BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
```

Also - did test plan of https://github.com/jetpack-io/devbox/pull/1093 to verify `devbox update` continues working. A sanity check since we dropped ShellEnvHash from the wrappers and I wanted to ensure this works as expected.

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -13,10 +13,11 @@ import (
 
 type shellEnvCmdFlags struct {
 	config               configFlags
-	runInitHook          bool
 	install              bool
-	useCachedPrintDevEnv bool
+	omitWrappersFromPath bool
 	pure                 bool
+	runInitHook          bool
+	useCachedPrintDevEnv bool
 }
 
 func shellEnvCmd() *cobra.Command {
@@ -44,6 +45,12 @@ func shellEnvCmd() *cobra.Command {
 
 	command.Flags().BoolVar(
 		&flags.pure, "pure", false, "If this flag is specified, devbox creates an isolated environment inheriting almost no variables from the current environment. A few variables, in particular HOME, USER and DISPLAY, are retained.")
+
+	// This flag is to be used by our generated bin-wrappers shell script.
+	command.Flags().BoolVar(
+		&flags.omitWrappersFromPath, "omit-wrappers-from-path", false, "If this flag is specified, "+
+			"the PATH from shellenv will not include the binary wrappers")
+	command.Flag("omit-wrappers-from-path").Hidden = true
 
 	// This is no longer used. Remove after 0.4.8 is released.
 	command.Flags().BoolVar(
@@ -74,7 +81,7 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 		}
 	}
 
-	envStr, err := box.PrintEnv(cmd.Context(), flags.runInitHook)
+	envStr, err := box.PrintEnv(cmd.Context(), flags.runInitHook, flags.omitWrappersFromPath)
 	if err != nil {
 		return "", err
 	}
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -81,7 +81,12 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 		}
 	}
 
-	envStr, err := box.PrintEnv(cmd.Context(), flags.runInitHook, flags.omitWrappersFromPath)
+	opts := &devopt.PrintEnv{
+		Ctx:                  cmd.Context(),
+		IncludeHooks:         flags.runInitHook,
+		OmitWrappersFromPath: flags.omitWrappersFromPath,
+	}
+	envStr, err := box.PrintEnv(opts)
 	if err != nil {
 		return "", err
 	}
--- a/internal/wrapnix/wrapper.sh.tmpl
+++ b/internal/wrapnix/wrapper.sh.tmpl
@@ -14,19 +14,21 @@ fi
 {{- end }}
 
 {{/*
-We use ShellEnvHashKey to prevent doing shellenv if the correct environment is
-already set. (perf optimization)
 
 We use the guard to prevent infinite loop if something in shellenv causes 
 another wrapped binary to be called. The guard is specific to this project so shellenv
 could still cause another project's shellenv to be called.
 
 DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
+
+--omit-wrappers-from-path so that we do not invoke other bin-wrappers from
+this bin-wrapper. Instead, we directly invoke the binary from the nix store, which
+should be in PATH.
 */ -}}
 
-if [[ "${{ .ShellEnvHashKey }}" != "{{ .ShellEnvHash }}" ]] && [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
+if [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
 export {{ .ShellEnvHashKey }}_GUARD=true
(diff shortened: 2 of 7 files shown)


savil:
* main
  * **PR #1151** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1151" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1151?utm_source=stack-comment).

savil:
gah, I have a lot of local changes that didn't get added to the commit here. Updating!

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -14,19 +14,21 @@ fi
...
 
 DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
+
+--omit-wrappers-from-path so that we do not invoke other bin-wrappers from
+this bin-wrapper. Instead, we directly invoke the binary from the nix store, which
+should be in PATH.
 */ -}}
 
-if [[ "${{ .ShellEnvHashKey }}" != "{{ .ShellEnvHash }}" ]] && [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
+if [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
```
savil, line 29: I should also delete ShellEnvHash from the data inserted into the template

mikeland86, line 29: I think the guard can be removed (the guard is in place to avoid infinite wrapper loop but since we're removing it from path it can't happen. 

This is hard to test so we may be better off leaving as is for now. Maybe add a comment

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -14,19 +14,21 @@ fi
...
 could still cause another project's shellenv to be called.
 
 DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
+
+--omit-wrappers-from-path so that we do not invoke other bin-wrappers from
+this bin-wrapper. Instead, we directly invoke the binary from the nix store, which
+should be in PATH.
 */ -}}
 
-if [[ "${{ .ShellEnvHashKey }}" != "{{ .ShellEnvHash }}" ]] && [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
```
mikeland86, line 27: This makes me sad. It makes the happy (common) path slower. Thoughts on skipping wrapper creation if hash matches? That's a risky change, so I guess it also depends on how soon we think we'll have the prompt solution.

internal/boxcli/shellenv.go
```diff
@@ -74,7 +81,7 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 		}
 	}
 
-	envStr, err := box.PrintEnv(cmd.Context(), flags.runInitHook)
+	envStr, err := box.PrintEnv(cmd.Context(), flags.runInitHook, flags.omitWrappersFromPath)
```
mikeland86, line 84: When adding a second boolean is a good time to switch to a struct :P

mikeland86 (approved):
no blocking comments

internal/wrapnix/wrapper.sh.tmpl
savil: yeah, I'll leave a comment. Would rather be super conservative with these changes.

internal/wrapnix/wrapper.sh.tmpl
savil: its very marginally slower, not perceptible, and not a huge concern IMO. We'll be running this same `devbox shellenv` on each prompt-display in the alternate solution.

I'm not keen on skipping wrapper creation if hash matches, since it feels like a riskier change to make.

internal/boxcli/shellenv.go
savil: ah, I knew you'd say this :p

PR #1152, opened 2023-06-13
[cleanup] remove launched flags for UnifiedEnv and Lockfile; remove Disabled method

savil:
## Summary

1. This PR removes the Disabled method. I had added it a while back after missing a `not` operator in `!featureflag.<flag name>.Enabled()`
and wanted to make it explicit. @mikeland86 objected saying he prefers "one way to do things" over "more natural language".
I no longer care to argue this case, so removing.

2. We've launched UnifiedEnv and Lockfile, so can remove these flags.

## How was it tested?

did not test except to see that it compiles.


savil:
* main
  * **PR #1152** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1152" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1152?utm_source=stack-comment).

internal/boxcli/featureflag/feature.go
```diff
@@ -54,10 +54,6 @@ func (f *feature) Enabled() bool {
 	return f.enabled
 }
 
-func (f *feature) Disabled() bool {
```
mohsenari, line 57: nit: There are also lowercase `disabled()` and `enabled()`. Based on their code, it seems like they mark a feature enabled or disabled. So they "enable" or "disable" a feature flag. 

I think it's better if we change them to that.

mohsenari (approved):
LGTM, just resolve the conflict and a small somewhat related nit that can be addressed in a future PR.

internal/boxcli/featureflag/feature.go
savil: done!

Issue #1051, opened 2023-05-26
[Bug]: Local flakes are not copied when generating a Dockerfile


mdwn:
**Current Behavior (bug)**
When generating a Dockerfile for a devbox, included local flakes will not be copied along with the devbox.json. This makes `devbox install` fail during the Dockerfile build.

**Expected Behavior (fix)**
`devbox install` should succeed on the generated Dockerfile.

**Additional context**
Please include the output of `devbox version -v` and
a copy of your `devbox.json` file.

```
$ devbox version -v
Version:     0.5.2
Platform:    darwin_arm64
Commit:      8f579b2c45c730742cbf9b7c400f247abf902775
Commit Time: 2023-05-18T16:18:52Z
Go Version:  go1.20.4
Launcher:    0.2.0
```

devbox.json:
```json
{
  "packages": [
    "bash@5.2-p15",
    "github:nixos/nixpkgs/757a0d107c238d031652a8c09d1f6bf1b6f523a3#go",
    "nodejs@16.18.1",
    "git@2.40.1",
    "gcc@12.2.0",
    "libiconvReal@1.16",
    "libfido2@1.13.0",
    "gci@0.9.1",
    "yamllint@1.28.0",
    "addlicense@1.0.0",
    "shellcheck@0.9.0",
    "python@3.11.2",
    "yarn@1.22.19",
    "protobuf3_20@3.20.3",
    "protoc-gen-go@1.28.1",
    "patchelf@0.15.0",
    "path:build.assets/flake"
  ],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
    "commit": "f91ee3065de91a3531329a674a45ddcb3467a650"
  }
}
```

Lagoja:
Looks like we need to ensure that the subfolder is copied in along with the rest of the project. Not sure why this is happening specifically with this Flake subfolder.

mohsenari:
@mdwn thanks for reporting this and apologies for late response. @Lagoja The fix for this is not super simple. It requires going through the declared packages and recognizing which ones are local flakes then adding them to Dockerfile before generating it. I'll probably work on it on Monday and ping here once the PR is out.

mohsenari:
@mdwn I have a PR that addresses this issue. Once merged and a new release issued, I'll ping you here to test again.

mohsenari:
With the release of 0.5.5, this issue should be fixed. There is another task somewhat related, which is in progress. That is, copying the whole project with dockerignore. See #1164

PR #1154, opened 2023-06-14
[dockerfile] Copying local flakes in dockerfile

mohsenari:
## Summary
This PR addresses the issue in generated dockerfiles by `devbox generate dockerfile` and `devbox generate devcontainer` in which packages mentioned as local flakes are not copied over in dockerfile and user has to manually do that.
Addresses: #1051 

## How was it tested?
- compile
- add a local flake (example: https://github.com/NixOS/templates/blob/master/go-hello/) 
- add local flake to devbox
- ./devbox generate dockerfile
- check the contents of generated dockerfile and confirm there is a Step 6 comment and COPY commands for each local flake.

### DIFF
--- a/internal/boxcli/generate/devcontainer_util.go
+++ b/internal/boxcli/generate/devcontainer_util.go
@@ -38,10 +38,11 @@ type vscode struct {
 
 type dockerfileData struct {
 	IsDevcontainer bool
+	LocalFlakeDirs []string
 }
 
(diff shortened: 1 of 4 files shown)


Lagoja:
This Dockerfile doesn't copy any of my code, is that expected?

Lagoja (approved):

PR #1156, opened 2023-06-14
[direnv] copy bash, fish, zsh shell code from direnv

savil:
## Summary

**Overall Motivation**
We've been having issues with the bin-wrappers, and want to explore an alternate approach. Bin-wrappers ensure that the shell environment is completely updated prior to invoking the binary, via running `devbox shellenv` if needed.

This alternate approach takes inspiration from direnv. It will seek to register a prompt `hook` in the shell. The hook will execute a shell function that will `export` the shell environment to update it, prior to displaying the shell prompt. 

To punt on how to transition existing devbox commands, we're introducing `devbox hook` and `devbox export` as distinct commands, analogous to `direnv hook` and `direnv export`.

**This PR's motivation**

As part of implementing this, we'd like to make use of direnv's battle-tested shell code where we can. Direnv has a nice shell abstraction for handling the hook and environment-variables. We could adopt this abstraction to better support shell code we generate beyond this immediate use-case of prompt-hooks.

So, we:
1. Copy bash, fish, zsh `Shell` code from direnv. Thank you 🙏 
2. Add ksh, posix, unknownSh structs. I filled in the struct functions as placeholders for now.

I call the package `shenv` since it handles the "shell environment". I stayed away from "shellenv" to avoid confusion with our existing "shellenv" command's operation.

## How was it tested?

Have _not_ tested. Subsequent PR will test and correct as needed.


savil:
* main
  * **PR #1156** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1156" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1156?utm_source=stack-comment).

loreto (approved):
LGTM although I am having trouble figuring out which files came from direnv, and which ones did you add.

Is ksh the only one you added?  Why not bring over `tcsh`, `elvish` and `gha` as well?

As mentioned during standup ... direnv *must* support ksh, right? Is it possible we just need to map ksh to one of the pre-existing ones? Maybe the posix one? (Wikipedia says KornShell complies with POSIX.2)

Are you sure you don't want to also copy the `env` related files direnv has? Or at least the env struct along with `GetEnv`, `Copy`, `ToGoEnv`, `ToShell`, `Fetch`

I think I would also copy `shell.go` along with the `DetectShell`, it handles figuring out the shell from the path if necessary: `DetectShell("-/usr/local/bin/bash")`

savil:
> LGTM although I am having trouble figuring out which files came from direnv, and which ones did you add.

The PR summary mentions:

So, we:
1. Copy bash, fish, zsh Shell code from direnv. Thank you 🙏
2. Add ksh, posix, unknownSh structs. I filled in the struct functions as placeholders for now.

> Is ksh the only one you added? 

see above

> Why not bring over `tcsh`, `elvish` and `gha` as well?

Our current `devbox shell` doesn't support those. So, I was limiting scope for now. We can add them later if we need.


> 
> As mentioned during standup ... direnv _must_ support ksh, right? Is it possible we just need to map ksh to one of the pre-existing ones? Maybe the posix one? (Wikipedia says KornShell complies with POSIX.2)

Yeah, I didn't get a chance to investigate that. Will do, and if needed will drop the `ksh` shell from this place.


> Are you sure you don't want to also copy the `env` related files direnv has? Or at least the env struct along with `GetEnv`, `Copy`, `ToGoEnv`, `ToShell`, `Fetch`

For now, I was focussed purely on getting the `hook` and `export` commands working. In a follow up PR, I want to investigate whether we can do the "environment diffing" that direnv does, and will likely incorporate direnv's `Env` utilities once I know how we can use them. 

> I think I would also copy `shell.go` along with the `DetectShell`, it handles figuring out the shell from the path if necessary: `DetectShell("-/usr/local/bin/bash")`

Ah, I do this: the `shenv.go` code is copied from the `shell.go` code. I did modify `DetectShell` just a bit to expunge the figuring out shell from path, since I didn't see how it would be used in our integration. As needed, I can add it back.

PR #1157, opened 2023-06-14
[workaround] add PGHOST at top-level env

savil:
## Summary

This simple fix moves the `export PGHOST` in the lapp-stack and lepp-stack examples to the `env` field of the devbox.jsons

Upside:
1. Much simpler fix compared to #1158 , and more maintainable.

Downside:
1. These are reused as templates for users, and are not the ideal code. 

I'm inclined to land this for now, and figure out a better solution without adding the bash code of #1158

### DIFF
--- a/examples/stacks/lapp-stack/devbox.json
+++ b/examples/stacks/lapp-stack/devbox.json
@@ -7,7 +7,8 @@
     "apache@2.4"
   ],
   "env": {
-    "PGPORT": "5432"
+    "PGPORT": "5432",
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1157** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1157" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1157?utm_source=stack-comment).

PR #1158, opened 2023-06-14
filter PATH for virtenv/.wrappers/bin in wrapper bash

savil:
## Summary

This PR is an alternative approach to #1151.

Instead of having the wrapper always call `devbox shellenv`, we just edit the 
PATH to remove `.virtenv/wrappers/bin` in bash.

Upside: 
1. this should fix the issue with the failing lapp-stack and lepp-stack tests
because the env-vars exported in the `run_test` will no longer get overwritten
by the plugin's env-var when the wrapper does `eval $(devbox shellenv)`.
2. In contrast to #1157, we keep the `export PGHOST` out of the `env` of devbox.json
which is better for templates.

Downside:
1. adding bash logic, so much less maintainable and higher risk of breaking users.

**Implementation**
1. revert #1151
2. implement wrappers.sh.tmpl changes to filterPath from wrappers/bin

## How was it tested?

testscripts should pass.

[ ] - in elixir example, I can do `devbox run -- iex -S mix`

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -33,7 +33,7 @@ type Devbox interface {
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	PrintEnv(opts *devopt.PrintEnv) (string, error)
+	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
 	PrintGlobalList() error
 	Pull(ctx context.Context, overwrite bool, path string) error
 	Push(url string) error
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #1158** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1158" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1158?utm_source=stack-comment).

PR #1159, opened 2023-06-15
Revert "[bin wrappers] add --omit-wrappers-from-path to shellenv to prevent a binary from invoking a wrapped-binary. (#1151)"

savil:
## Summary

Reverting due to regression:
Within a devbox script if any environment variable has been exported, and
if a devbox-plugin also sets that env-var, then the devbox-plugin env-var
will be set for the wrapped-binary.

Specifically, in the `lapp-stack/devbox.json`, we have:
```
      "run_test": [
        "mkdir -p /tmp/devbox/lapp",
        "export PGHOST=/tmp/devbox/lapp",
        "initdb",
```

but the PGHOST that `initdb` sees is a different value i.e. the one set by the devbox plugin.

In the code being reverted, the wrapped-bin's bash script would always invoke
`eval $(devbox shellenv --omit-wrappers-from-path=true)` in order to omit wrappers
from PATH. A side-effect is that the PGHOST would also get overridden.

## How was it tested?

no testing done.
straight revert, tests should pass.

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -33,7 +33,7 @@ type Devbox interface {
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	PrintEnv(opts *devopt.PrintEnv) (string, error)
+	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
 	PrintGlobalList() error
 	Pull(ctx context.Context, overwrite bool, path string) error
 	Push(url string) error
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #1159** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1159" target="_blank">Graphite</a>  👈
    * **PR #1160** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1160" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1159?utm_source=stack-comment).

LucilleH (approved):

PR #1160, opened 2023-06-15
wrappers] ensure bin-wrappers invoke other binaries and not bin-wrappers

### LINKED ISSUE #1142: [Bug]: Wrappers interfere with `iex -s mix`

savil:
## Summary

This is an alternate to #1151 (see explanation in #1159)

In this approach, the wrapped-binary always calls:
`eval $(devbox shellenv only-path-without-wrappers)` 
to ensure that the PATH is set without the `wrappers/bin` directory.

In contrast to #1151, we do not always set any other environment variables

Fixes #1142

## How was it tested?

issue from #1142 is fixed:
```
> devbox run -- iex -S mix
* creating .nix-mix/archives/hex-2.0.6
* creating .nix-mix/elixir/1-14/rebar3
Resolving Hex dependencies...
Resolution completed in 0.016s
Unchanged:
  cowboy 2.9.0
  cowlib 2.11.0
  ranch 1.8.0
All dependencies are up to date
Erlang/OTP 25 [erts-13.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

Hello World!
Interactive Elixir (1.14.4) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Goodbye World

17:18:09.349 [notice] Application elixir_hello exited: shutdown

BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
```

- [ ] testscripts should pass

### DIFF
--- a/internal/boxcli/shell.go
+++ b/internal/boxcli/shell.go
@@ -59,7 +59,7 @@ func runShellCmd(cmd *cobra.Command, flags shellCmdFlags) error {
 	if flags.printEnv {
 		// false for includeHooks is because init hooks is not compatible with .envrc files generated
 		// by versions older than 0.4.6
-		script, err := box.PrintEnv(cmd.Context(), false /*includeHooks*/)
+		script, err := box.PrintEnv(&devopt.PrintEnv{Ctx: cmd.Context()})
 		if err != nil {
 			return err
 		}
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -55,6 +55,9 @@ func shellEnvCmd() *cobra.Command {
 	// This is used by bin wrappers and not meant for end users.
 	command.Flag("use-cached-print-dev-env").Hidden = true
 	flags.config.register(command)
+
+	command.AddCommand(shellEnvOnlyPathWithoutWrappersCmd())
+
 	return command
 }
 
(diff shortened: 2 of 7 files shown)


savil:
* main
  * **PR #1159** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1159" target="_blank">Graphite</a> 
    * **PR #1160** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1160" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1160?utm_source=stack-comment).

LucilleH (approved):
LGTM

In this case shellenv will be invoked on every binary call, which is quite SAD.

As a follow up, perhaps we can generate this PATH env variable and store it in a file. In every wrapped binary, we simply source that file. 

The file only needs to be re-generated if `__DEVBOX_SHELLENV_HASH_` does not match.

internal/boxcli/shellenv.go
```diff
@@ -74,7 +77,60 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
...
+}
+
+type shellEnvOnlyPathWithoutWrappersCmdFlags struct {
+	config configFlags
+}
+
+func shellEnvOnlyPathWithoutWrappersCmd() *cobra.Command {
+	flags := shellEnvOnlyPathWithoutWrappersCmdFlags{}
+	command := &cobra.Command{
+		Use:     "only-path-without-wrappers",
```
mohsenari, line 99: Nit, drop the only and let the command be "path-without-wrappers"

mohsenari (approved):
Nice. Thank you. Just one nit comment

internal/boxcli/shellenv.go
savil: 🤔 I'd prefer leaving it in to make it clear that this is not also returning other env-vars of shellenv.

savil:
> In this case shellenv will be invoked on every binary call, which will impact performance.

That was the case with the fix in #1151 as well.

I'm not very concerned because a cached `devbox shellenv` is pretty fast (30 milliseconds if I do `time eval $(devbox shellenv)`), and it will 100% be cached in this case. The direnv-inspired solution we are exploring will call `devbox shellenv` on every shell prompt, which is okay, because its fast enough.

internal/boxcli/shell.go
```diff
@@ -59,7 +59,7 @@ func runShellCmd(cmd *cobra.Command, flags shellCmdFlags) error {
 	if flags.printEnv {
 		// false for includeHooks is because init hooks is not compatible with .envrc files generated
 		// by versions older than 0.4.6
-		script, err := box.PrintEnv(cmd.Context(), false /*includeHooks*/)
+		script, err := box.PrintEnv(&devopt.PrintEnv{Ctx: cmd.Context()})
```
mikeland86, line 62: super golang nit, generally context and struct are passed as 2 args. The reason is that storing context in a struct is risky because we may accidentally keep it in state somewhere.

internal/impl/devbox.go
```diff
@@ -322,9 +322,23 @@ func (d *Devbox) PrintEnv(ctx context.Context, includeHooks bool) (string, error
 		return "", err
 	}
 
+	if opts.OnlyPathWithoutWrappers {
+		path := []string{}
+		for _, p := range strings.Split(envs["PATH"], string(filepath.ListSeparator)) {
```
mikeland86, line 327: `filepath.SplitList` ?

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -29,4 +29,9 @@ export {{ .ShellEnvHashKey }}_GUARD=true
 eval "$(DO_NOT_TRACK=1 devbox shellenv -c {{ .ProjectDir }})"
 fi
 
+# So that we do not invoke other bin-wrappers from
+# this bin-wrapper. Instead, we directly invoke the binary from the nix store, which
```
mikeland86, line 33: nit, it's the binary of the profile (which I guess points to the nix store path)

internal/boxcli/shell.go
savil: good point, this has been undone in a future PR already!

internal/wrapnix/wrapper.sh.tmpl
savil: the emphasis here is on _binary_ as opposed to _wrapper_

PR #1163, opened 2023-06-15
[bin wrappers] fixes for only-path-without-wrappers call

savil:
## Summary

follow up to #1160 

correctness:
- consolidate some code into calling `wrapnix.GetWrapperBinPath`
- the earlier code had a bug where I forgot to call `-c <project dir>`, but that is moot now (see below change)

perf:
- call DO_NOT_TRACK=1  to minimize segment perf impact
- simplify OnlyPathWithoutWrappers to directly call `os.Getenv("PATH")`, remove WrapperBins, and exportify it. This function is used in a very specific context, so we can _know_ we can  skip the nix.ComputeEnv calculation.
- Doing `time devbox shellenv only-path-without-wrappers` is now consistently ~ 18-25 milliseconds. 

I have a slight preference for this over writing PATH to a file and sourcing it, since we may run into correctness issues with that file getting outdated, and it'll introduce perf regression on every shellenv call to write the file.

## How was it tested?

- [x] testscripts pass
- [x] `iex -S mix` works

visual inspection:
```
    1 #!/nix/store/w849dr5qcbrrnxv69sy7kdnifa9jpjyf-bash-5.2-p15/bin/bash
  2
  3
  4
  5 if [[ "$__DEVBOX_SHELLENV_HASH_ac6a60d54a9d1ba6618cce8d40bfdfb50059d814edfe5f49bca3168c5c34e970" != "b1a4903de70aa4d610ecb00ffb3dd755913b4505651e6bebbf41571a406d415c" ]] &    & [[ -z "$__DEVBOX_SHELLENV_HASH_ac6a60d54a9d1ba6618cce8d40bfdfb50059d814edfe5f49bca3168c5c34e970_GUARD" ]]; then
  6 export __DEVBOX_SHELLENV_HASH_ac6a60d54a9d1ba6618cce8d40bfdfb50059d814edfe5f49bca3168c5c34e970_GUARD=true
  7 eval "$(DO_NOT_TRACK=1 devbox shellenv -c /Users/savil/code/jetpack/devbox/examples/development/elixir/elixir_hello)"
  8 fi
  9
 10 eval "$(DO_NOT_TRACK=1 devbox shellenv only-path-without-wrappers -c /Users/savil/code/jetpack/devbox/examples/development/elixir/elixir_hello)"
 11
 12 exec /nix/store/47whvyx9x48ishndcjw71xs7l0a6v6sd-elixir-1.14.4/bin/iex "$@"
```

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -33,7 +33,7 @@ type Devbox interface {
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	PrintEnv(opts *devopt.PrintEnv) (string, error)
+	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
 	PrintEnvVars(ctx context.Context) ([]string, error)
 	PrintGlobalList() error
 	Pull(ctx context.Context, overwrite bool, path string) error
@@ -73,4 +73,8 @@ func GlobalDataPath() (string, error) {
 
 func PrintEnvrcContent(w io.Writer) error {
 	return impl.PrintEnvrcContent(w)
+}
+
+func OnlyPathWithoutWrappers() string {
+	return impl.OnlyPathWithoutWrappers()
 }
--- a/devbox.go
+++ b/devbox.go
@@ -75,6 +75,11 @@ func PrintEnvrcContent(w io.Writer) error {
 	return impl.PrintEnvrcContent(w)
 }
 
-func OnlyPathWithoutWrappers() string {
-	return impl.OnlyPathWithoutWrappers()
+// ExportifySystemPathWithoutWrappers reads $PATH, removes `virtenv/.wrappers/bin` paths,
+// and returns a string of the form `export PATH=....`
+//
+// This small utility function could have been inlined in the boxcli caller, but
+// needed the impl.exportify functionality. It does not depend on core-devbox.
+func ExportifySystemPathWithoutWrappers() string {
+	return impl.ExportifySystemPathWithoutWrappers()
 }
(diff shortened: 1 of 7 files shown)


devbox.go
```diff
@@ -73,3 +73,7 @@ func GlobalDataPath() (string, error) {
 func PrintEnvrcContent(w io.Writer) error {
 	return impl.PrintEnvrcContent(w)
 }
+
+func OnlyPathWithoutWrappers() string {
```
mohsenari, line 77: What was the reason for skipping adding this to Devbox interface?

internal/impl/devbox.go
```diff
@@ -322,30 +322,35 @@ func (d *Devbox) PrintEnv(opts *devopt.PrintEnv) (string, error) {
...
 	}
 
 	return envStr, nil
 }
 
+// OnlyPathWithoutWrappers is a small utility to filter WrapperBin paths from PATH
+func OnlyPathWithoutWrappers() string {
+
+	path := []string{}
+	for _, p := range strings.Split(os.Getenv("PATH"), string(filepath.ListSeparator)) {
```
mohsenari, line 339: os.Getenv("PATH") works because OnlyPathWithoutWrappers is used only in the context of direnv enabled or inside devbox shell. So the PATH is guaranteed to be a devbox computed PATH.
But it won't work outside that context, if you just run the command you'll get PATH env variable (assuming you don't have devbox global).
So I worry this might be used in future in a different context with the assumption that it will compute a devbox PATH and exclude wrappers bin and introduce a bug.

savil: 1. The `devbox shellenv only-path-without-wrappers` command is Hidden, and I can add more comments to clarify its usage. It'll only be used by someone developing devbox code. So, this is low risk.

2. The internal function could be renamed to `ExportSystemPathWithoutWrappers`. WDYT?

devbox.go
savil: Its a small utility function which doesn't have any dependency on the core "devbox" struct.

I was going to inline it in `boxcli/shellenv.go`, but needed the `exportify` function so added it here.

mohsenari (approved):

LucilleH (approved):
This solution is much better than the one I suggested in the previous PR!

Issue #1136, opened 2023-06-10
[Feature Request] Generate Makefile for scripts


ibuildthecloud:
Make is so common as the entrypoint to projects that it would be nice if I could generate a Makefile from devbox that just adds targets for `scripts`.  So basically `devbox generate scripts-makefile` would create/add to makefile the following

```makefile
foo:
     devbox run foo
```
Where foo is a script defined.  You could obvioulsy make this fancier and just put in some make-fu so that targets are dynamically generated from devbox so that you don't have to keep adding them.  I'll personally try to figure that out for myself.
```[tasklist]
### Tasks
- [ ] https://github.com/jetpack-io/devbox/issues/1141
```

ibuildthecloud:
So I got the following working
```makefile
devboxtargets := $(shell devbox run targets)

.PHONY: $(devboxtargets)
.DEFAULT_GOAL := targets
    

$(devboxtargets):
    devbox run $@
```

So the interesting thing that came from this fun exercise is that it would be a lot easier if there was a devbox command to just list the available scripts.  Right now I'm just doing
```json
"scripts": {
      "targets": "jq -r '.shell.scripts | keys | join(\" \")' devbox.json"
    }
```

ibuildthecloud:
I could be convinced that this is easy enough to do that it doesn't need to be a feature, but conceptually it's the same as the `.envrc` integration which is almost as trivial under the hood.

Lagoja:
Listing the scripts would be very helpful, and it is implemented for `devbox global run`, so it should be fairly easy to do. We could have them list if you run `devbox run` without passing an argument

ibuildthecloud:
Oh I guess that sorta is there already
 ```$ devbox run
Available scripts:
* run
* targets
* dev
```
The only thing is I need a raw output that is just the list, no other test.  Maybe `-q` could be abused for that `devbox run -q`

PR #1167, opened 2023-06-16
[run] Add list flag to list all scripts

hezhizhen:
## Summary

Add `list` flag to `devbox run` to list all scripts.

## How was it tested?

`devbox run --list`
`devbox run -l`

### DIFF
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -16,8 +16,9 @@ import (
 )
 
 type runCmdFlags struct {
-	config configFlags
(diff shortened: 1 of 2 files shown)


ipince (approved):
Code LGTM.

I wonder if anyone has strong opinions on whether this is needed though

Lagoja:
> Code LGTM.
> 
> I wonder if anyone has strong opinions on whether this is needed though

This was related to a request in #1136. @loreto are we ok with this UI?

PR #1170, opened 2023-06-16
[pure shell] Fixed devbox not being in path

mohsenari:
## Summary
Fixed an issue which caused devbox to disappear from PATH in pure shell mode.

NOTE: Moved the location of the symlink from `.devbox/virtenv/bin` to `.devbox/bin`. I'm open to changing the location but since it's only in pure mode and the location doesn't really affect the user (they can just use `$(which devbox)` inside the shell to get the location dynamically) I'm eager to not have the location of devbox symlink be a blocking issue.

## How was it tested?
- compile
- ./devbox shell --pure
- which devbox
- which nix

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -1218,12 +1218,14 @@ func (d *Devbox) convertEnvToMap(currentEnv []string) (map[string]string, error)
 	// handling special case for PATH
 	if d.pure {
 		// Finding nix executables in path and passing it through
-		// Needed for devbox commands inside pure shell to work
-		nixInPath, err := findNixInPATH(env)
+		// As well as adding devbox itself to PATH
+		// Both are needed for devbox commands inside pure shell to work
+		includedInPath, err := findNixInPATH(env)
 		if err != nil {
 			return nil, err
 		}
-		env["PATH"] = nixInPath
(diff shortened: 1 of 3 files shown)


Lagoja (approved):
Tested and it seems to work now, adding Nix + Devbox to path in the shell makes sense to me

savil (approved):
Awesome!

For a possible followup PR, what do you think about moving `createDevboxSymlink` out of `wrapnix` since it no longer has to do with wrappers?

We could put it inside `PrintEnv` innards somewhere appropriate. Maybe `convertEnvToMap` but seems odd to have a "symlink created" side-effect in a function with that name. Maybe one of its callers...

mohsenari:
@savil 

> Awesome!
> 
> For a possible followup PR, what do you think about moving `createDevboxSymlink` out of `wrapnix` since it no longer has to do with wrappers?
> 
> We could put it inside `PrintEnv` innards somewhere appropriate. Maybe `convertEnvToMap` but seems odd to have a "symlink created" side-effect in a function with that name. Maybe one of its callers...

that's a good suggestion, @savil Mike also had a comment on the name `convertEnvToMap` is a bit misleading and I agree. When I defined that function, I only wanted to convert env slice to map but I later had to add more logic to it. 
So I can make a follow up PR updating that function's name and also moving the createDevboxSymlink logic to it as well.

PR #1172, opened 2023-06-16
[direnv-inspired] add export and hook commands, and hook up shell, global and direnv

savil:


savil:
* main
  * **PR #1156** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1156" target="_blank">Graphite</a> 
    * **PR #1172** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1172" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1172?utm_source=stack-comment).

internal/boxcli/hook.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type hookFlags struct {
+	config configFlags
+}
+
+func hookCmd() *cobra.Command {
+	flags := hookFlags{}
+	cmd := &cobra.Command{
+		Use:   "hook [shell]",
```
loreto, line 18: Another possible name for this command is `activate` (it's what https://github.com/jdxcode/rtx uses)

Leave as hook, and let's get the PR in, but I wanna get John's thoughts on which one is the better name, before we remove the feature flag and start transitioning people.

loreto (approved):
LGTM

internal/boxcli/hook.go
savil: yes, agree on `activate`. Please see my RFC proposal in the PR summary for reasons why I like it...

PR #1173, opened 2023-06-17
[searcher] Add SearchOption and handle host error

hezhizhen:
## Summary

1. Add `SearchOption` for better declaration
2. Handle the error from parsing hosts

## How was it tested?

DEVBOX_SEARCH_HOST=foo devbox search hello


hezhizhen:
@savil

PR #1178, opened 2023-06-19
[auth] use time.Duration instead of int for intervals

hezhizhen:
## Summary

I think it's clearer to use `time.Duration` instead of `int` for intervals because the former contains the quantity and the unit while the latter only contains the quantity.

I wonder if it's possible to convert the `Interval` field in `codeResponse` to `time.Duration`.

## How was it tested?

devbox auth whoami


hezhizhen:
@mikeland73

mikeland73 (changes requested):
Duration changes look good but please revert the formatting changes, including extra spaces (unless they were caused by gofmt). This helps keep git blame clean.

We don't enforce a line length, but I usually aim for 80. We generally don't comment on line length in PRs, but in this case they are reverting a conscious decision and not related to the main purpose. That said, happy to reconsider if folks think we should allow longer lines, just think we should do that separately before reformating code.

hezhizhen:
> Duration changes look good but please revert the formatting changes, including extra spaces (unless they were caused by gofmt). This helps keep git blame clean.
> 
> We don't enforce a line length, but I usually aim for 80. We generally don't comment on line length in PRs, but in this case they are reverting a conscious decision and not related to the main purpose. That said, happy to reconsider if folks think we should allow longer lines, just think we should do that separately before reformating code.

Sorry for the late reply, and apologize first.

I prefer short lines and try to keep them in 80 characters. But I also try to keep it in one line if it only exceeds a little (shorter than 120 characters).

I'll create a new PR with the main purpose only.

Issue #1094, opened 2023-06-03
[Bug]: Devbox Update does not update `github` based flakes


Lagoja:
**Current Behavior (bug)**
Right now, running `devbox update` when a `github` based flake is in the packages list will return the following, and not update the package

```
Skipping github:goreleaser/nur#goreleaser-pro because it is not a versioned package
```

**Expected Behavior (fix)**
Running `devbox update` should update the flake to the latest version of the referenced branch or tag, similar to how `nix flake update` works. 

For example -- the flake input above references the main branch of the `goreleaser/nur` repo, and it should update the flake to the latest commit on main when the user runs `devbox update`

PR #1180, opened 2023-06-19
[update] Update flakes

mikeland73:
## Summary

Update flakes via `nix profile upgrade`.

Purposefully forgiving.

Note: local flakes don't need updating because they automatically use latest version of file.

fixes https://github.com/jetpack-io/devbox/issues/1094

## How was it tested?

```bash
devbox add github:F1bonacc1/process-compose 
devbox update github:F1bonacc1/process-compose
```

This works, but but need to test something that will actually update.


savil (approved):
if it works, it works?

PR #1194, opened 2023-06-21
(cli) Don't check version when updating devbox

mauricioabreu:
## Summary

I don't think it is necessary to warn users about a fresh new version if they are updating devbox

## How was it tested?

`go run main.go update`


Lagoja:
Seems reasonable to me, not sure why there's the extra `PrintLn(args)` though? @mikeland73

mauricioabreu:
@Lagoja good catch! Removed `Println`

Lagoja (approved):

savil:
Thank you :)

PR #1195, opened 2023-06-21
[global] Better error message when profile doesn't exist

mikeland73:
## Summary

If path doesn't exist, S3 returns access denied. Show a better error message. This can be improved by checking if object exists (But that would slow things down, so maybe only check if the initial pull failed so we can show a better error message).

## How was it tested?

* deleted my profile manually
* Ran `devbox global pull`


loreto (approved):
LGTM

PR #1199, opened 2023-06-22
[vscode-extension] fix vscode terminal PATH 

mohsenari:
## Summary
Fixes the issue where terminal's $PATH env variable wouldn't get updated in integrated terminal.

## How was it tested?
buckle up, the steps are long:

- `cd vscode-extension`
- `vsce package` (install vsce via `npm i -g vsce`)
- `code --install-extension devbox-0.0.7.vsix`
- setup a devbox.json in a sample directory with a couple of packages (e.g., `hello`, `go@1.18`)
- open the sample directory in vscode
- cmd+shift+p and run `Devbox: Reopen in Devbox shell environment`
- after vscode reopens, echo $PATH in terminal

### DIFF
--- a/vscode-extension/src/devbox.ts
+++ b/vscode-extension/src/devbox.ts
@@ -82,14 +82,32 @@ async function setupDotDevbox(workingDir: Uri, dotdevbox: Uri) {
 }
 
 function updateVSCodeConf() {
-    const shell = process.env["SHELL"] ?? "/bin/zsh";
-    const devboxCompatibleShell = {
-        "devboxCompatibleShell": {
-            "path": shell,
-            "args": []
-        }
-    };
     if (process.platform === 'darwin') {
+        const shell = process.env["SHELL"] ?? "/bin/zsh";
+        const shellArgsMap = (shellType: string) => {
+            switch (shellType) {
+                case "fish":
+                    // We special case fish here because fish's `fish_add_path` function
+                    // tends to prepend to PATH by default, hence sourcing the fish config after
+                    // vscode reopens in devbox environment, overwrites devbox packages and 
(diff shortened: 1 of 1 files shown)


Lagoja:
It looks like it's working!!! I've been messing with my settings a bit so it would be good to get others on the team to test this, but it's worked on Go, Elixir, and Python projects so far.

vscode-extension/src/devbox.ts
```diff
@@ -21,16 +22,17 @@ export async function devboxReopen() {
...
-                            cwd: workingDir.path
-                        });
-                    }
+                    progress.report({ message: 'Installing devbox packages...', increment: 25 });
+                    await setupDotDevbox(workingDir, dotdevbox);
+
+                    // setup required vscode settings
+                    progress.report({ message: 'Updating configurations...', increment: 50 });
+                    updateVSCodeConf();
+                    setTimeout(() => { }, 1000);
```
LucilleH, line 31: Why do we need the `setTimeout`?

vscode-extension/src/devbox.ts
```diff
@@ -21,16 +22,17 @@ export async function devboxReopen() {
...
+
+                    // Calling CLI to compute devbox env
+                    progress.report({ message: 'Calling Devbox to setup environment...', increment: 80 });
+                    // callDevboxIntegrate(workingDir, progress.report, resolve, reject);
```
LucilleH, line 35: ✂️

LucilleH (approved):
🎉

vscode-extension/src/devbox.ts
mohsenari: ugh, that was from debugging, I'll remove it.

PR #1203, opened 2023-06-22
[nix input] Add comments, and renames for clarity

savil:
## Summary

I find `nix.Input` fairly hard to understand. I make a first pass at adding
comments and renaming functions and variables for clarity. Made it about
halfway through the `nix/input.go` file.

**nomenclature**: I have tried to ensure consistency around:
1.  `inputs` referring to `nix.Input`
2. `flakeInputs` referring to `shellgen.flakeInput`
3. `packages` referring to raw strings that are in the packages list in `devbox.json`.

cc @mikeland73 

## How was it tested?

Have not tested. Please review carefully.

relying on testscripts passing.

### DIFF
--- a/internal/nix/input.go
+++ b/internal/nix/input.go
@@ -59,21 +59,21 @@ func InputsFromStrings(rawNames []string, l lock.Locker) []*Input {
 // The raw name corresponds to a devbox package from the devbox.json config.
 func InputFromString(raw string, locker lock.Locker) *Input {
 	// We ignore the error because... TODO @mikeland why?
-	u, _ := url.Parse(raw)
+	inputURL, _ := url.Parse(raw)
 
 	// This handles local flakes in a relative path.
 	// `raw` will be of the form `path:./local_flake_subdir#myPackage`
 	// for which path:<empty>, opaque:./local_subdir, and scheme:path
-	if u.Path == "" && u.Opaque != "" && u.Scheme == "path" {
+	if inputURL.Path == "" && inputURL.Opaque != "" && inputURL.Scheme == "path" {
 		// This normalizes url paths to be absolute. It also ensures all
 		// path urls have a single slash (instead of possibly 3 slashes)
-		normalizedURL := "path:" + filepath.Join(locker.ProjectDir(), u.Opaque)
-		if u.Fragment != "" {
-			normalizedURL += "#" + u.Fragment
+		normalizedURL := "path:" + filepath.Join(locker.ProjectDir(), inputURL.Opaque)
+		if inputURL.Fragment != "" {
+			normalizedURL += "#" + inputURL.Fragment
 		}
-		u, _ = url.Parse(normalizedURL)
+		inputURL, _ = url.Parse(normalizedURL)
 	}
-	return &Input{*u, locker, raw, ""}
+	return &Input{*inputURL, locker, raw, ""}
 }
 
 // InputFromProfileItem sets the raw Input as the `item`'s unlockedReference i.e.
--- a/internal/nix/input.go
+++ b/internal/nix/input.go
@@ -20,24 +20,50 @@ import (
 	"go.jetpack.io/devbox/internal/lock"
 )
 
+// Input represents a "package" added to the devbox.json config.
+// The word "input" is used because it will be referenced in a generated flake.nix.
+// A unique feature of flakes is that they have well-defined "inputs" and "outputs".
+// This Input will be aggregated into a specific "flake input" (see shellgen.flakeInput).
 type Input struct {
 	url.URL
 	lockfile lock.Locker
-	Raw      string
+
+	// Raw is the devbox package name from the devbox.json config.
+	// Raw has a few forms:
+	// 1. Devbox Packages
+	//    a. versioned packages
+	//       examples:  go@1.20, python@latest
+	//    b. any others?
+	// 2. Local
+	//    flakes in a relative sub-directory
+	//    example: ./local_flake_subdir#myPackage
+	// 3. Github
+	//    remote flakes with raw name starting with `Github:`
+	//    example: github:nixos/nixpkgs/5233fd2ba76a3accb5aaa999c00509a11fd0793c#hello
+	Raw string
 
 	normalizedPackageAttributePathCache string // memoized value from normalizedPackageAttributePath()
 }
 
-func InputsFromStrings(names []string, l lock.Locker) []*Input {
+// InputsFromStrings constructs Input from the list of package names provided.
+// These names correspond to devbox packages from the devbox.json config.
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #1203** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1203?utm_source=stack-comment).

internal/nix/input.go
```diff
@@ -50,46 +76,62 @@ func InputFromString(raw string, locker lock.Locker) *Input {
...
+
+	// replace all non-alphanumeric with dashes
 	return inputNameRegex.ReplaceAllString(result, "-")
 }
 
-func (i *Input) URLForInput() string {
-	if i.IsDevboxPackage() {
+// URLForFlake is the url to be used as the input in the generated flake.nix
+func (i *Input) URLForFlake() string {
+	if i.isDevboxPackage() {
```
mikeland73, line 134: I'm not sure this is an improvement. This is literally the url that is used as an input in a flake.

Assume this struct is getting renamed to Package which will make the existing name less redundant

internal/nix/input.go
```diff
@@ -20,24 +20,50 @@ import (
 	"go.jetpack.io/devbox/internal/lock"
 )
 
+// Input represents a "package" added to the devbox.json config.
+// The word "input" is used because it will be referenced in a generated flake.nix.
```
mikeland73, line 24: We should rename this struct to package so not sure if this comment will be accurate

mikeland73 (commented):
I think there's good stuff here but some doc blocks are not accurate and a few just feel somewhat redundant (don't feel strongly against including but not sure they add, and adding text makes code less readable)

In general if a function is hard to understand I think renaming can be better solution than adding comments (though obviously not mutually exclusive)

internal/nix/input.go
```diff
@@ -50,46 +76,62 @@ func InputFromString(raw string, locker lock.Locker) *Input {
...
+// isLocal specifies whether this input is a local flake.
+// Usually, this is of the form: `path:./local_flake_subdir#myPackage`
+func (i *Input) isLocal() bool {
 	// Technically flakes allows omitting the scheme for local absolute paths, but
 	// we don't support that (yet).
 	return i.Scheme == "path"
 }
 
-func (i *Input) IsDevboxPackage() bool {
+// isDevboxPackage specifies whether this input is a nix package defined in a devbox.json config.
```
mikeland73, line 93: This is not specific enough. This specifically checks if it's a devbox package as defined by our own syntax (canonical@version). A "nix" package can be a flake or a legacy attribute path.

mikeland73 (commented):
Also, this is renaming some variable names from packages to inputs but I think we should go in opposite direction. We should rename this struct and and inputs generally to be packages.

Note that initially this struct was only inputs but now it more generally represents packages.

internal/shellgen/flake_input.go
```diff
@@ -67,10 +67,12 @@ func (f *flakeInput) BuildInputs() []string {
...
 	defer trace.StartRegion(ctx, "flakeInputs").End()
 
-	inputs := map[string]*flakeInput{}
+	// Use the verbose name flakeInputs to distinguish from `inputs`
+	// which refer to `nix.Input` in most of the codebase.
+	flakeInputs := map[string]*flakeInput{}
 
-	userPackages := devbox.PackagesAsInputs()
-	pluginPackages, err := devbox.PluginManager().PluginPackages(userPackages)
+	userInputs := devbox.PackagesAsInputs()
```
mikeland73, line 74: I think this is regressing. Moving towards naming things packages will be easier to understand imo

internal/plugin/manager.go
```diff
@@ -45,16 +45,16 @@ func (m *Manager) ApplyOptions(opts ...managerOption) {
 	}
 }
 
-func (m *Manager) PluginPackages(inputs []*nix.Input) ([]*nix.Input, error) {
-	pkgs := []*nix.Input{}
+func (m *Manager) PluginInputs(inputs []*nix.Input) ([]*nix.Input, error) {
```
mikeland73, line 48: Same here

mikeland73 (commented):
@savil let's chat about this? 

I was hoping to move more away from the word "input" but this changes more stuff to be "input".

Inputs should be narrowly understood to be what is passed into a flake. This struct is more general and used to represent a package and possible plugin.

internal/shellgen/flake_input.go
savil: From my PR summary:

> nomenclature: I have tried to ensure consistency around:
- inputs referring to nix.Input
- flakeInputs referring to shellgen.flakeInput
- packages referring to raw strings that are in the packages list in devbox.json.

internal/plugin/manager.go
savil: > nomenclature: I have tried to ensure consistency around:
inputs referring to nix.Input
flakeInputs referring to shellgen.flakeInput
packages referring to raw strings that are in the packages list in devbox.json.

internal/nix/input.go
savil: The old name `URLForInput` conveys very little information, and is redundant. IMO the new name is much better.

internal/nix/input.go
savil: sounds good, I'll update.

internal/nix/input.go
savil: Whoever does future renames should come here to update the comment then?

savil:
@mikeland73 I'm happy to chat. When are you available?

Why are we integrating all these concepts (flake inputs, plugins?) into one struct?

savil:
My main pushback is it is better to be consistent, than pick the ideal name. Right now the code mixes packages and inputs. This makes it challenging to follow. I'd prefer getting to a consistent state.

internal/nix/input.go
savil: sounds good

internal/nix/input.go
```diff
@@ -20,24 +20,50 @@ import (
...
+	for _, rawName := range rawNames {
+		inputs = append(inputs, InputFromString(rawName, l))
 	}
 	return inputs
 }
 
+// InputsFromStrings constructs Input from the raw name provided.
+// The raw name corresponds to a devbox package from the devbox.json config.
 func InputFromString(raw string, locker lock.Locker) *Input {
+	// We ignore the error because: TODO why?
```
savil, line 61: @mikeland73 why do we ignore the error?

mikeland73: The input is pretty validated so this would be a very rare failure, but this was mostly to avoid a large cascading change when this struct was introduced.

PR #1208, opened 2023-06-22
[remove nixpkgs] part 1: write info for nix store paths in lockfile

savil:
## Summary

**Motivation For "Remove Nixpkgs"**
Today, when adding a versioned package, Devbox will look up the versioned package’s metadata by mapping the user-specified version to a nixpkgs commit hash and an attribute path. This `nixpkgs/<commit hash>#attributePath` is used in the generated flake.nix. 

We want to remove the dependency on the nixpkgs commit hash, so that the user need not download that nixpkgs artifact, and spend time resolving it.

**Essence of the approach**
Today, when adding a versioned package, Devbox will look up the versioned package’s metadata by mapping the user-specified version to a nixpkgs commit hash and an attribute path. This `nixpkgs/<commit hash>#attributePath` is used in the generated flake.nix. 
```
builtins.fetchClosure {
  fromStore = "https://cache.nixos.org";
  fromPath = /nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1;
  toPath = /nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1;
}
```

if we can map `git@2.33.1` to the `/nix/store` paths, then we need not download nixpkgs.

**Implementation in this PR**

This PR focusses on the resolving packages to the "locked" representation in `devbox.lock`. Anytime a package is added or removed, we check if the `devbox.lock` needs additional `SystemInfo` for each package. This `SystemInfo` can be used to derive `fromPath` and `toPath` in `builtins.fetchClosure`.

Next PR will read this information from the lockfile to generate a `flake.nix` that uses `builtins.fetchClosure`.

cc @mikeland73 

## How was it tested?

Using the devbox repository, did `devbox add hello`.

```
diff --git a/devbox.json b/devbox.json
index 7b263cb4..55ee3f8a 100644
--- a/devbox.json
+++ b/devbox.json
@@ -2,7 +2,8 @@
   "packages": [
     "go@1.20",
     "actionlint@1.6.23",
-    "golangci-lint@1.52.2"
+    "golangci-lint@1.52.2",
+    "hello@latest"
   ],
   "env": {
     "PATH": "$PATH:$PWD/dist"
diff --git a/devbox.lock b/devbox.lock
index 7dffa054..3baae253 100644
--- a/devbox.lock
+++ b/devbox.lock
@@ -4,17 +4,78 @@
     "actionlint@1.6.23": {
       "last_modified": "2023-03-31T22:52:29Z",
[...]

PR #1209, opened 2023-06-23
[remove nixpkgs] part 2: generate flake.nix using builtins.fetchClosure

savil:
## Summary

See #1208 for motivation.
That previous PR handled the "writing the system info to the lockfile". In this PR: 

1. Generate the `flake.nix` from `flake_remove_nixpkgs.nix.tmpl` template.
   - this new template uses `builtins.fetchClosure` to download the package's binary directly from the binary-cache-store (like `cache.nixos.org`). This lets us bypass `nixpkgs` for nix packages in the binary store.
2. The template data is `shellgen.flakePlan`.
3. In the `flakePlan`, we:
   a.  filter `flakeInputs` to be those _not_ in the binary store (like remote or local flakes)
   b. add a new field `Packages []*nix.Package`.
   
In `nix.Package`, we have a few new methods specific to the BinaryStore, like `IsInBinaryStore` and `PathInBinaryStore`.

**Testing note:**
I was stymied in extensively testing this by:
- search index lacking x86_64-darwin system info (and I have an intel mac)
- devbox cloud having limitations (will share details elsewhere)
- the simple test plan below took me many hours to complete.

cc @mikeland73 

## How was it tested?

Tested with `examples/development/go/hello-world`.

For a more complex example, with a flake, the `examples/flakes/remote` produces this `flake.nix`:
```
{
   description = "A devbox shell";

   inputs = {
     nixpkgs.url = "http://lax.devbox-nixed.internal/nixos/nixpkgs/archive/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62.tar.gz";
     gh-nixos-nixpkgs-5233fd2ba76a3accb5aaa999c00509a11fd0793c.url = "github:nixos/nixpkgs/5233fd2ba76a3accb5aaa999c00509a11fd0793c";
     gh-F1bonacc1-process-compose.url = "github:F1bonacc1/process-compose";
   };

   outputs = {
     self,
     nixpkgs,
     gh-nixos-nixpkgs-5233fd2ba76a3accb5aaa999c00509a11fd0793c,
     gh-F1bonacc1-process-compose,
   }:
       let
          pkgs = nixpkgs.legacyPackages.x86_64-linux;
       in
       {
         devShells.x86_64-linux.default = pkgs.mkShell {
           buildInputs = [
             gh-nixos-nixpkgs-5233fd2ba76a3accb5aaa999c00509a11fd0793c.legacyPackages.x86_64-linux.hello
             gh-nixos-nixpkgs-5233fd2ba76a3accb5aaa999c00509a11fd0793c.legacyPackages.x86_64-linux.cowsay
             gh-F1bonacc1-process-compose.packages.x86_64-linux.process-compose
           ];
         };
       };
 }
```


savil:
* main
  * **PR #1208** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1208" target="_blank">Graphite</a> 
    * **PR #1209** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1209" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1209?utm_source=stack-comment).

gcurtis:
Sorry for the delay - I just remembered that I never reviewed this. I'll take a look tomorrow morning.

internal/shellgen/flake_package.go
```diff
@@ -0,0 +1,66 @@
...
+	if sysInfo == nil {
+		return nil, nil
+	}
+
+	attributePath, err := pkg.PackageAttributePath()
+	if err != nil {
+		return nil, err
+	}
+
+	storeDir := strings.Join([]string{sysInfo.FromHash, sysInfo.StoreName, sysInfo.StoreVersion}, "-")
```
gcurtis, line 32: The store version might be empty, in which case omit the trailing -.

internal/shellgen/flake_package.go
```diff
@@ -0,0 +1,66 @@
...
+	return &flakePackage{
+		Name:      attributePath,
+		FromStore: nixosCacheURL,
+		FromPath:  filepath.Join("/nix/store", storeDir),
+		// TODO add ToPath:
+	}, nil
+}
+
+func flakePackages(devbox devboxer) ([]*flakePackage, error) {
```
gcurtis, line 41: I know this wasn't added in your PR, but the number of interfaces we have is making the code really hard to follow.

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -0,0 +1,42 @@
...
+     {{- end }}
+   }:
+       let
+          pkgs = nixpkgs.legacyPackages.{{ .System }};
+          {{- range $_, $pkg := .Packages }}
+          {{ $pkg.Name }}.{{ $.System }} = builtins.fetchClosure{
+             fromStore = "{{ $pkg.FromStore }}";
+             fromPath = "{{ $pkg.FromPath }}";
+          };
+          {{- end }}
```
gcurtis, line 26: I think the attribute path order should be flipped to `{{ $pkg.System }}.{{ $.Name }}` to follow the flake convention. The name should also be quoted in case it has special characters.

Another option is to just inline this directly in the `builtInputs` list with something like:

```nix
[
   {{- range .Packages }}
   (builtins.fetchClosure{
      fromStore = "{{ .FromStore }}";
      fromPath = "{{ .FromPath }}";
   })
   {{- end }}
]
```

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -0,0 +1,42 @@
...
+         devShells.{{ .System }}.default = pkgs.mkShell {
+           buildInputs = [
+             {{- range $_, $pkg := .Packages }}
+             {{ $pkg.Name }}.{{ $.System }}
+             {{- end }}
+             {{- range $_, $flake := .FlakeInputs }}
+             {{- range $flake.BuildInputs }}
+             {{.}}
+             {{- end }}
+             {{- end }}
```
gcurtis, line 38: Can be simplified to:

```suggestion
             {{- range .FlakeInputs }}
             {{- range .BuildInputs }}
             {{.}}
             {{- end }}
             {{- end }}
```

internal/shellgen/flake_plan.go
```diff
@@ -3,13 +3,17 @@ package shellgen
...
+
+	"go.jetpack.io/devbox/internal/nix"
 )
 
 // flakePlan contains the data to populate the top level flake.nix file
 // that builds the devbox environment
 type flakePlan struct {
 	NixpkgsInfo *NixpkgsInfo
 	FlakeInputs []*flakeInput
+	Packages    []*flakePackage
```
gcurtis, line 15: Could this be a slice of values instead? It doesn't look like they need to be pointers.

internal/shellgen/flake_package.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+
+	"go.jetpack.io/devbox/internal/boxcli/featureflag"
+	"go.jetpack.io/devbox/internal/nix"
+)
+
+type flakePackage struct {
```
gcurtis, line 11: I'm not sure if `flakePackage` is an accurate name. It's fetching a store path and its dependencies, which may or may not have been built from a flake.

internal/shellgen/flake_package.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	storeDir := strings.Join([]string{sysInfo.FromHash, sysInfo.StoreName, sysInfo.StoreVersion}, "-")
+	return &flakePackage{
+		Name:      attributePath,
+		FromStore: nixosCacheURL,
+		FromPath:  filepath.Join("/nix/store", storeDir),
+		// TODO add ToPath:
```
gcurtis, line 37: Have you looked at the [`builtins.storePath` function](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-storePath)? If we're running in impure mode, then it might make things a little easier. It doesn't require a content-addressed hash and it will automatically use the Nix-configured binary caches (aka substituters). That'll make this work automatically with nixed in Devbox Cloud.

gcurtis (approved):
I had some comments, but approving to keep the train moving.

internal/shellgen/flake_input.go
```diff
@@ -77,11 +78,18 @@ func flakeInputs(ctx context.Context, devbox devboxer) ([]*flakeInput, error) {
...
-	order := []string{}
 	// We prioritize plugin packages so that the php plugin works. Not sure
 	// if this is behavior we want for user plugins. We may need to add an optional
 	// priority field to the config.
-	for _, input := range append(pluginInputs, userInputs...) {
+	allInputs := append(pluginInputs, userInputs...)
+	allInputs = lo.Filter(allInputs, func(item *nix.Package, _ int) bool {
+		// This is temporary, so that we can support the existing flake.nix.tmpl
+		// as well as the new flake_remove_nixpkgs.nix.tmpl.
+		return !featureflag.RemoveNixpkgs.Enabled() || item.SystemInfo() == nil
```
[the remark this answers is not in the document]
savil: This will continue to include packages (like local or remote flakes) that don't have system info

internal/shellgen/flake_package.go
mikeland73: I'm a bit confused, why do we need this instead of using `nix.Package` directly? It feels like just a another struct with the same data and function?

internal/shellgen/flake_plan.go
mikeland73: Can it be `nix.Packages` instead?

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -0,0 +1,42 @@
+{
+   description = "A devbox shell";
+
+   inputs = {
+     {{/* nixpkgs could be removed once we turn php/haskell planners into flakes */ -}}
```
mikeland73, line 5: rm this comment.

mikeland73 (commented):
Not blocking, but I would really want to avoid adding a new `Package` struct. I also think system info should be part of resolve. Ideally system info can be abstracted away so that the whatever generates the flake can just get the data if it's present.

btw, you should assume that we will not be able to avoid nixpkgs for certain packages for the short and medium term, so we need to structure this in a clean way where both types can coexist.

savil:
@gcurtis addressed comments and made updates

@mikeland73 Folded `flakePackages` into `nix.Package`. Addressed the concern around calling `lockfile.Resolve` in the getter for `FromPath` in `nix.Package`.

internal/nix/input.go
```diff
@@ -73,7 +73,15 @@ func PackageFromString(raw string, locker lock.Locker) *Package {
...
-	return &Package{*pkgURL, locker, raw, ""}
+
+	pkg := &Package{
+		URL:                                 *pkgURL,
+		lockfile:                            locker,
+		Raw:                                 raw,
+		normalizedPackageAttributePathCache: "",
+	}
+
+	return pkg
```
mikeland73, line 84: nit, you can revert this if you want. You can also omit `normalizedPackageAttributePathCache` if you used name fields.

internal/nix/input.go
```diff
@@ -407,6 +415,40 @@ func (p *Package) hashFromNixPkgsURL() string {
 	return HashFromNixPkgsURL(p.URLForFlakeInput())
 }
 
+// BinaryCacheStore is the store from which to fetch this package's binaries.
+// It is used as FromStore in builtins.fetchClosure.
+const BinaryCacheStore = "https://cache.nixos.org"
+
+func (p *Package) IsInBinaryStore() bool {
+	return p.isVersioned()
+}
```
mikeland73, line 424: Is there a better way to determine this? For example only true if lock file data has system info?

internal/shellgen/flake_plan.go
```diff
@@ -35,9 +40,17 @@ func newFlakePlan(ctx context.Context, devbox devboxer) (*flakePlan, error) {
 		}
 	}
 
-	shellPlan := &flakePlan{}
-	var err error
-	shellPlan.FlakeInputs, err = flakeInputs(ctx, devbox)
+	userPackages := devbox.PackagesAsInputs()
+	pluginPackages, err := devbox.PluginManager().PluginInputs(userPackages)
```
mikeland73, line 44: I'll rename `PluginInputs` and `PackagesAsInputs` in a follow up.

Also, I'm thinking of changing plugins to encapsulate it more. Having plugins show up in shellgen package is not idael

mikeland73 (approved):
This is great! I really like how this ended up.

internal/nix/input.go
savil: 🤔 yeah, we could add that check too. In my previous iteraiton of this PR, I was checking both `p.isVersioned()` and "has system info". 

I can add that here. It feels weird to possibly make network-calls in an if-statement but I understand the architecture better now so lets roll with it.

PR #1219, opened 2023-06-26
[vscode-extension] Adding missing commands to devbox vscode extension - v0.1.1

mohsenari:
## Summary
- Fixed small issue in readme from last release
- added missing commands from CLI to vscode extension's command palette (install, update, search)

Q: should I add more devbox commands? e.g., global, info, version, auth, etc.

## How was it tested?
- vsce package
- code --install <generated-devbox>.vsix
- open a project with devbox.json
- cmd+shift+p and type devbox
- confirm install, update and search exists in the list of available commands and try them.


LucilleH (approved):
Personal opinion: those three commands make sense. I'm not sure about the rest.

PR #1220, opened 2023-06-26
MYSQL Plugin

Lagoja:
## Summary

Adds a plugin for MySQL DB. Implementation is very similar to MariaDB plugin, except: 

1. init_hook is changed to use `mysqld --initialize-insecure`
2. removed unnecessary symlink wrappers from flake.nix
3. renamed services

## How was it tested?

`devbox shell` using the example in this PR, then: 
* `devbox services up` -- should show mysql + mysql_logs 
* `devbox run mysql` -- should give you a mysql console


Lagoja:
Some thing weird happened with this, let me update and re-push

PR #1221, opened 2023-06-26
Mysql Plugin v1

Lagoja:
## Summary

Adds a plugin for MySQL DB. Implementation is very similar to MariaDB plugin, except: 

1. init_hook is changed to use `mysqld --initialize-insecure`
2. removed unnecessary symlink wrappers from flake.nix
3. renamed services

## How was it tested?

`devbox shell` using the example in this PR, then: 
* `devbox services up` -- should show mysql + mysql_logs 
* `devbox run mysql` -- should give you a mysql console

### DIFF
--- a/plugins/mysql.json
+++ b/plugins/mysql.json
@@ -1,6 +1,6 @@
 {
     "name": "mysql",
-    "version": "0.0.2",
+    "version": "0.0.1",
     "match": "^mysql?[0-9]*$",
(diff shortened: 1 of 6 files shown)


plugins/mysql.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "mysql",
+    "version": "0.0.2",
```
mikeland73, line 3: v1

mikeland73 (approved):

PR #1222, opened 2023-06-27
telemetry: report segment events asynchronously

### LINKED ISSUE #1003: [Bug]: telemetry makes the cli tool very slow
**Current Behavior (bug)**

```zsh
gelus ➜  ~ REPORTTIME=1
gelus ➜  ~ devbox global shellenv
export PATH="${XDG_DATA_HOME:-$HOME/.local/share}/devbox/global/current/bin${PATH+:$PATH}";
hash -r
devbox global shellenv  10.39s user 0.65s system 128% cpu 8.575 total
gelus ➜  ~ DO_NOT_TRACK=1 devbox global shellenv     # executes instantly
export PATH="${XDG_DATA_HOME:-$HOME/.local/share}/devbox/global/current/bin${PATH+:$PATH}";
hash -r
```

**Expected Behavior (fix)**
Telemetry doesn't slowdown the tool.

From the source code, looks like an event is sent to segment for every invocation of the command line tool, `cd`-ing with direnv and opening a new shell are unusably slow. Thanks for keeping the DO_NOT_TRACK flag to disable it!

**Additional context**
```
gelus ➜  ~ devbox version -v
Version:     0.4.9
Platform:    linux_amd64
Commit:      b03f0d26da3df4318f2a8d70c0531c0c77d0d33c
Commit Time: 2023-05-02T23:26:52Z
Go Version:  go1.20.3
Launcher:    0.2.0
```

gcurtis:
Instead of blocking the process from exiting, buffer Segment events to disk and start another process to upload them. With this change all telemetry should be non-blocking.

Most of the Segment event logic now lives in the telemetry package instead of the cobra middleware. A bunch of the unused logic around tracking CLI SSH sessions has also been removed. The telemetry package API now looks something like:

```go
type EventName int

const (
	EventCommandSuccess EventName = iota
	EventShellInteractive
	EventShellReady
)

func Error(err error, meta Metadata)
func Event(e EventName, meta Metadata)
func Start()
func Stop()
func Upload()
```

Fixes #1003.

### DIFF
--- a/internal/envir/util.go
+++ b/internal/envir/util.go
@@ -6,7 +6,6 @@ package envir
 import (
 	"os"
 	"strconv"
-	"time"
 )
 
 func IsDevboxCloud() bool {
@@ -37,28 +36,6 @@ func IsInBrowser() bool { // TODO: a better name
 func IsCI() bool {
 	ci, err := strconv.ParseBool(os.Getenv("CI"))
 	return ci && err == nil
-}
-
-func ShellStart() time.Time {
-	return ParseShellStart(os.Getenv(DevboxShellStartTime))
-}
-
-func FormatShellStart(t time.Time) string {
-	if t.IsZero() {
-		return ""
-	}
-	return strconv.FormatInt(t.Unix(), 10)
-}
-
-func ParseShellStart(s string) time.Time {
-	if s == "" {
-		return time.Time{}
-	}
-	unix, err := strconv.ParseInt(s, 10, 64)
-	if err != nil {
-		return time.Time{}
-	}
-	return time.Unix(unix, 0)
 }
 
 // GetValueOrDefault gets the value of an environment variable.
--- a/internal/envir/util.go
+++ b/internal/envir/util.go
@@ -6,12 +6,8 @@ package envir
 import (
 	"os"
 	"strconv"
+	"time"
 )
-
-func IsCLICloudShell() bool { // TODO: not used any more
-	cliCloudShell, _ := strconv.ParseBool(os.Getenv(devboxCLICloudShell))
-	return cliCloudShell
-}
 
 func IsDevboxCloud() bool {
 	return os.Getenv(DevboxRegion) != ""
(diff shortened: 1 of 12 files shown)


internal/envir/util.go
```diff
@@ -43,6 +39,28 @@ func IsCI() bool {
...
+func ParseShellStart(s string) time.Time {
+	if s == "" {
+		return time.Time{}
+	}
+	unix, err := strconv.ParseInt(s, 10, 64)
+	if err != nil {
+		return time.Time{}
+	}
+	return time.Unix(unix, 0)
+}
```
savil, line 62: I'd prefer keeping this logic out of the `envir` package to keep it streamlined. 

I'd suggest it is better encapsulated in the telemetry package, or even a more focussed `shtelemetry` package

internal/telemetry/telemetry.go
```diff
@@ -4,33 +4,326 @@
...
+// Stop stops gathering telemetry and flushes buffered events to disk.
+func Stop() {
+	if !started || !needsFlush.Load() {
+		return
+	}
+
+	// Report errors in a separate process so we don't block exiting.
+	exe, err := os.Executable()
+	if err == nil {
+		_ = exec.Command(exe, "bug").Start()
```
savil, line 83: could we rename `bug` to be more self-explanatory like `upload-telemetry` ?

savil (approved):
awesome!

ibuildthecloud:
FWIW, this completely breaks in my env :)

I wanted to test this because the telemetry is completely killing me.  You never knew how much a couple 100ms can drive you nuts.

Anyhow.... This hangs forever

```
 cd -
/home/darren/src/oss/devbox
direnv: loading ~/src/oss/devbox/.envrc
direnv: using devbox
direnv: ([/home/linuxbrew/.linuxbrew/Cellar/direnv/2.32.3/bin/direnv export bash]) is taking a while to execute. Use CTRL-C to give up.
```
I'm debugging why.  It's odd because if I run things manually it seems all happy.  To be continued....

ibuildthecloud:
This seems suspicious... I have a bunch of `/home/darren/src/oss/devbox/dist/devbox bug` processes laying around.  :thinking:

ibuildthecloud:
And this commit definitely breaks it.  I just tested if this is a just a regression else  so I rebased on main.

ibuildthecloud:
Ok, it's definitely that.  The `devbox bug` command that is getting spawned is blocking.  I'll let you debug why, I think just doing any `direnv` will reproduce this.  If I have to guess why this is a happening I would say that the `devbox bug` process is stuck on reading stdin or writing stdout/stderr since it's being ran as a subshell in direnv and direnv probably blocks all I/O interaction from the shell.

gcurtis:
@ibuildthecloud thanks for catching that. It turns out this was happening because a goroutine was returning without calling `WaitGroup.Done` when the telemetry key wasn't set (which usually only happens in dev builds). I fixed that bug and added a fallback that nukes the process after 5 seconds to make sure we don't leave processes laying around.

ibuildthecloud:
Cool this works now (sorta/maybe).  I just now realized that dev build probably don't have telemetry enabled so I don't really know if this is now faster.

Issue #1193, opened 2023-06-21
[Bug]: Loss of interactive Bash functionality in Pure Mode


apgrucza:
**Current Behavior (bug)**

When using Bash as a shell and running `devbox shell --pure`, the following is observed:

- The arrow keys input `^[[A` `^[[B` `^[[C` `^[[D` instead of moving the cursor and navigating through the history
- The tab key does not perform command completion

These errors (from an unmodified `.bashrc`) are also shown:

- `bash: shopt: progcomp: invalid shell option name`
- `bash: shopt: hostcomplete: invalid shell option name`
- `bash: complete: command not found`

**Expected Behavior (fix)**

If I add the `bashInteractive` package to my devbox and then type `bash` from inside the pure shell, the problems listed above go away. So, the fix seems to be that Devbox should use the `bashInteractive` package for pure mode, rather than the normal `bash` package.

**Additional context**

Output of `devbox version -v`:

```
Version:     0.5.5
Platform:    linux_amd64
Commit:      891e8de7a8ab7a3fec42fad069326d68cb84603e
Commit Time: 2023-06-19T19:18:00Z
Go Version:  go1.20.5
Launcher:    0.2.0
```

Lagoja:
Arrow keys and tabs definitely look like a bug. Good catch on the bashInteractive package, we can try using that to create the shell to confirm if that fixes things. 

The bashrc errors seem related to your issue #1196. The missing packages make sense (in pure mode, you wouldn't have access to those packages unless you installed them in Devbox), but we should probably not source the `.bashrc` as part of creating the shell

apgrucza:
FYI the errors listed above do go away when using `bashInteractive` as they relate to [Bash Builtin Commands](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html) rather than separate binaries. Issue #1196 covers different errors.

mohsenari:
@apgrucza thanks for reporting this. Will look into this and fix it for next release, if the fix doesn't introduce more complications.
We wanted the pure shell have minimal impact from the system for use cases where absolute reproducibility is crucial. If the only solution is to add a `bashInteractive` package on user's behalf, I'd rather leave that as a suggestion when pure shell is activated.

mohsenari:
This issue should be fixed in 0.5.6

PR #1223, opened 2023-06-28
[CLI] Changed default nix bash to bashInteractive

mohsenari:
## Summary
In `shellPath()` function where we decide which type of shell to use when dropping user in devbox shell, if we can't figure out the shell, we drop the user in the default `bash` that nix installs in store. But this bash doesn't handle arrow keys and keyboard combinations very well, so it creates a frustrating experience for users (including pure shell users).  `bashInteractive` package has all these capabilities and makes the experience better for a 'default' shell. 

So this PR changes this:
- `nix eval <bash>` to get store path and use to create default shell
to this:
- `nix eval <bashInteractive>` to get store path
- nix build <generated store path>
- use store path to create default shell

Addresses #1193

## How was it tested?
- compile
- ./devbox shell --pure
OR 
- unset $SHELL
- ./devbox shell

### DIFF
--- a/internal/impl/shell.go
+++ b/internal/impl/shell.go
@@ -111,7 +111,7 @@ func shellPath(devbox *Devbox) (path string, err error) {
 
 	cmd := exec.Command(
 		"nix", "eval", "--raw",
-		fmt.Sprintf("%s#bash", nix.FlakeNixpkgs(devbox.cfg.NixPkgsCommitHash())),
+		fmt.Sprintf("%s#bashInteractive", nix.FlakeNixpkgs(devbox.cfg.NixPkgsCommitHash())),
 	)
 	cmd.Args = append(cmd.Args, nix.ExperimentalFlags()...)
 	out, err := cmd.Output()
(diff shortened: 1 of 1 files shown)


internal/impl/shell.go
```diff
@@ -120,6 +120,14 @@ func shellPath(devbox *Devbox) (path string, err error) {
 	}
 	bashNixStorePath = string(out)
 
+	// install bashInteractive in nix/store without creating a symlink to local directory (--no-link)
+	cmd = exec.Command("nix", "build", bashNixStorePath, "--no-link")
+	cmd.Args = append(cmd.Args, nix.ExperimentalFlags()...)
+	err = cmd.Run()
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
```
mikeland73, line 129: Thoughts on using `devbox.addDevboxUtilityPackage` instead? It installs in a specific profile. (not sure if that works with `bashNixStorePath`)

mikeland73 (approved):

PR #1228, opened 2023-06-28
Fix broken envrc template

ibuildthecloud:
## How was it tested?

I ran it and it didn't explode with 
```
$ devbox generate direnv --print-envrc
Error: template: envrcContent.tmpl:5: unexpected {{else}}
```


savil (approved):
Ugh 😩 thank you 🙏

Issue #1200, opened 2023-06-22
[Bug]: Bash prompt not colored in pure mode


apgrucza:
**Current Behavior (bug)**
When using Bash as a shell and running `devbox shell`, the prompt is colored (green and blue). But when running `devbox shell --pure`, the prompt is not colored (white).

**Expected Behavior (fix)**
The prompt should be colored in pure mode. This should just be a matter of adding `TERM` to the list of environment variables that are passed through from the parent shell.

**Additional context**
Output of `devbox version -v`:
```
Version:     0.5.5
Platform:    linux_amd64
Commit:      891e8de7a8ab7a3fec42fad069326d68cb84603e
Commit Time: 2023-06-19T19:18:00Z
Go Version:  go1.20.5
Launcher:    0.2.0
```

mohsenari:
This issue should be fixed in v0.5.6

PR #1230, opened 2023-06-28
[CLI] Added TERM as leaking env vars in pure shell

mohsenari:
## Summary
Added `$TERM` to the list of env vars that leak to pure shell. Reason being, Pure shell doesn't allow colorful text in prompt because TERM variable is unset. 
Adding the TERM allows pure shell to know if terminal supports color and enable it.

Addresses #1200

## How was it tested?
- compile
- `./devbox shell --pure`
- `ls -a`
- `ls -a --color`
- confirm directories and executables show in different colors when `--color` is passed.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -1103,7 +1103,8 @@ func (d *Devbox) parseEnvAndExcludeSpecialCases(currentEnv []string) (map[string
 		// handling special cases to for pure shell
 		// Passing HOME for pure shell to leak through otherwise devbox binary won't work
 		// We also include PATH to find the nix installation. It is cleaned for pure mode below
-		if !d.pure || key == "HOME" || key == "PATH" {
+		// TERM leaking through is to enable colored text in the pure shell
+		if !d.pure || key == "HOME" || key == "PATH" || key == "TERM" {
 			env[key] = val
 		}
 	}


savil (approved):
Makes sense to me. I remember needing to add `TERM` when we were exploring using `nix develop` as part of adopting flakes.

gcurtis (approved):
Are `HOME`, `PATH`, and `TERM` the only env vars that we let leak in pure mode? I think we need to allow quite a few more to prevent things from breaking. This is the list of environment variables we would allow back when Devbox's shell was originally pure by default:

- `__ETC_PROFILE_NIX_SOURCED`
- `all_proxy`
- `DEVBOX_REGION`
- `DEVBOX_USE_VERSION`
- `DISPLAY`
- `ftp_proxy`
- `HOME`
- `http_proxy`
- `https_proxy`
- `IN_NIX_SHELL`
- `LANG`
- `LC_ALL`
- `LC_COLLATE`
- `LC_CTYPE`
- `LC_MESSAGES`
- `LC_MONETARY`
- `LC_NUMERIC`
- `LC_TIME`
- `LOGNAME`
- `NIX_SSL_CERT_FILE`
- `NIXPKGS_ALLOW_UNFREE`
- `no_proxy`
- `OLDPWD`
- `PAGER`
- `PARENT_PATH`
- `PWD`
- `SECURITYSESSIONID`
- `SHELL_SESSIONS_DISABLE`
- `SHLVL`
- `SHLVL`
- `SSH_TTY`
- `SSL_CERT_FILE`
- `TERM`
- `TERM_PROGRAM`
- `TERM_PROGRAM_VERSION`
- `TERM_SESSION_ID`
- `TZ`
- `USER`
- `WAYLAND_DISPLAY`
- `WAYLAND_SOCKET`
- `XDG_RUNTIME_DIR`

A bunch of those are probably still relevant.

See:

- https://github.com/jetpack-io/devbox/commit/184a5a20461049464446aa30c02de7f951cdf913#diff-6ab333856edad4c76174d5d4b57d5b538f722f53dd8113d64139eda9f6e739bbL578-L636
- https://github.com/NixOS/nix/blob/b8bb8026d22dac8e86a7ff8135e281b6ebef08f4/src/nix-build/nix-build.cc#L103-L109

mohsenari:
> Are `HOME`, `PATH`, and `TERM` the only env vars that we let leak in pure mode? I think we need to allow quite a few more to prevent things from breaking. This is the list of environment variables we would allow back when Devbox's shell was originally pure by default...

@gcurtis you might be right, but I'd rather go down the path of least number of leaked env vars as much as possible. Unless we have user issues, or cases an env var not leaking breaks things.

gcurtis:
I was just mentioning it because a lot of these did break users before. We could say that if you're choosing pure mode you need to set these yourself, but I'm not sure if most users will know that. There are also some that are user-specific (i.e., they can't be added to devbox.json). Some of the issues we saw before were: 

- `NIX_SSL_CERT_FILE` not being set will break some Nix packages that use HTTPS
- Not setting locales (`LC_*`) can break text encoding.
- `LOGNAME` is required by POSIX so a lot of things might assume it's set.

I think we're going to end up repeating a lot of the same bugs if we don't go through and see which need to be set.

mohsenari:
@gcurtis what if we add this list to the docs page explaining that excluding these env vars (aka using pure mode) might cause these known side effects.

PR #1233, opened 2023-06-29
Revert "[RFC][direnv-inspired] introduce export and hook commands, use in devbox shell, global and direnv (#1172)"

savil:
## Summary

Reverts #1172.

While we feel architecturally this is a better approach, it would introduce some
non-trivial risk.

For the moment, bin-wrappers seem to have stabilized so we'll punt on this, and
circle back to it later.

## How was it tested?

compiles

- [ ] testscripts should pass


loreto (approved):
LGTM

PR #1236, opened 2023-06-29
[remove nixpkgs] enable fallback for packages not in binary store

savil:
## Summary

Many testscript unit-tests were failing because they use unversioned legacy packages like `"hello"`, as opposed to `hello@latest`, or `hello@<version>`.

These legacy packages do not use the Binary Cache Store.

I considered whether we should treat them automatically as upgraded to `@latest`, but decided against:
1. We'll need this logic in the flake template regardless. There will be some packages that are not in binary store.
2.  This also keeps the `Package.IsInBinaryStore` easier to reason about.

We should instead encourage users to `devbox update` their packages: which our messaging already does. 

The remaining three testscript unit-tests are failing due to "unable to find nix executable" errors. Will fix in followup.

## How was it tested?

enabled feature-flag and ran testscripts: https://github.com/jetpack-io/devbox/actions/runs/5417704606/jobs/9848964717?pr=1236

### DIFF
--- a/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
+++ b/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
@@ -17,6 +17,14 @@
    }:
       let
         pkgs = nixpkgs.legacyPackages.{{ .System }};
+        {{- range .FlakeInputs }}
+        {{- if .IsNixpkgs }}
+        {{.PkgImportName}} = (import {{.Name}} {
+          system = "{{ $.System }}";
+          config.allowUnfree = true;
+        });
+        {{- end }}
+        {{- end }}
       in
       {
         devShells.{{ .System }}.default = pkgs.mkShell {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1236** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1236" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1236?utm_source=stack-comment).

savil:
@mikeland73 ping

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -17,6 +17,14 @@
...
       let
         pkgs = nixpkgs.legacyPackages.{{ .System }};
+        {{- range .FlakeInputs }}
+        {{- if .IsNixpkgs }}
+        {{.PkgImportName}} = (import {{.Name}} {
+          system = "{{ $.System }}";
+          config.allowUnfree = true;
+        });
+        {{- end }}
+        {{- end }}
```
savil, line 27: these lines are the same as in the current `flake.nix`:
https://github.com/jetpack-io/devbox/blob/main/internal/shellgen/tmpl/flake.nix.tmpl#L27-L34

LucilleH (approved):
This is only for cases where the `RemoveNixpkgs` feature flag is on?

savil:
> This is only for cases where the `RemoveNixpkgs` feature flag is on?

yep

PR #1238, opened 2023-06-30
[refactor] intro nixprofile package

savil:
## Summary

There are three motivations for introducing the `nixprofile` package:

1. We should keep `nix` package focussed on executing core nix apis. This would
reduce its dependencies on other packages, enabling more packages to call `nix`
functions without hitting cycle errors.

`nixprofile` has a lot of devbox-specific logic. So, it should be its own package that calls `nix` apis (which themselves can be improved)

2. I tried moving `nix.Package` to `devpkg` but it was extremely difficult to avoid
new cyclic import errors. This PR will help avoid those.

3.  This will also help break the import cycle for calling `nix.System` from `lock.Lockfile`.

## How was it tested?

compiles

- [x] testscript should pass


savil:
* main
  * **PR #1238** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1238" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1238?utm_source=stack-comment).

internal/nix/nixprofile/profile.go
```diff
@@ -189,6 +190,11 @@ func (item *NixProfileListItem) AttributePath() (string, error) {
 	return attrPath, nil
 }
 
+// PackageFromProfileItem constructs a nix.Package using the unlocked reference
+func (item *NixProfileListItem) PackageFromProfileItem(locker lock.Locker) *nix.Package {
```
mikeland73, line 194: Name is redundant. Can just be `Package()`

Don't love that it requires a lockfile but I see that it's necessary.

mikeland73 (approved):
This seems generally good, except I would keep nix commands all in the nix package. Specifically, I would use nixprofile to turn devbox data into a specific nix call and then have nixprofile call the nix package. That way all nix API access is in a single package. Not a blocker.

internal/nix/nixprofile/profile.go
savil: done

savil:
> This seems generally good, except I would keep nix commands all in the nix package. Specifically, I would use nixprofile to turn devbox data into a specific nix call and then have nixprofile call the nix package. That way all nix API access is in a single package. Not a blocker.

Yes, 100% aligned. Its directionally what I had in mind with "can be improved" in the PR Summary: "So, it should be its own package that calls nix apis (which themselves can be improved)".

I'd like to circle back to it in a future PR. I wanna finish moving `nix.Input` to `devpkg`.

savil:
Soo, I decided to do it in this PR itself. So, now `nixprofile` calls `nix` apis. @mikeland73 PTAL

There is still "stutter" in naming of `nixprofile` functions but I'm punting on that for now.

Unfortunately, the `git cp` no longer carried over from `nix/profile.go` to `nixprofile/profile.go`, and so the git blame is a bit broken...

mikeland73 (approved):

PR #1239, opened 2023-06-30
[refactor] move nix.Package to devpkg.Package

savil:
## Summary

This PR moves `nix.Package` to `devpkg` package.

`devpkg.Package` depends on `nix` and `lock`, which is okay. 

I needed to introduce a small package called `devpkgutil` to resolve some gnarly cycles. 
- `lock/lockfile.go` calls these utilities.
- `nix/search.go` calls these utilities.

This `devpkgutil` is definitely a workaround, and I spent a lot of time trying to make alternate paths work:
- tried refactoring `lockfile.Resolve` to not depend on `devpkg.ParseVersionedPackage`
- tried refactoring `nix/search.go` to a `nixsearch` package (similar to `nixprofile`), but that wouldn't help in this case.

 Open to hear alternative solutions.

UPDATE:
- moved 2 of the 3 `devpkgutil` functions to `nix` package in `nix/nixpkgs.go`
- one function remains: `ParseVersionedPackage`. Will tackle in follow up to remove `devpkgsutil`


savil:
Making draft.

I will make two changes:
1. intro `nixsearch` package so `nix` package can be kept with minimal dependencies (as in #1238)
2. fold `devpkgutils` into `devpkg` since (1) will remove the cycles. This is a prime example of why we should make `nix` have minimal dependencies.

internal/lock/lockfile.go
```diff
@@ -15,6 +15,7 @@ import (
 	"go.jetpack.io/devbox/internal/devpkg/devpkgutil"
 
 	"go.jetpack.io/devbox/internal/cuecfg"
+	_ "go.jetpack.io/devbox/internal/nix"
```
mikeland73, line 18: This is just to test no cycles exist, ensure it builds. Can remove

mikeland73 (approved):

PR #1240, opened 2023-06-30
 intro nixsearch package so we can call nix.System inside lock.Lockfile

savil:
## Summary

The goal was to enable calling `nix.System` inside `lock.Lockfile`.

Unfortunately, `nix/search.go` was dependent on the `lock` package and so we 
had a circular dependency.

I resolved this by introducing a `nixsearch` package. The motivation is similar to #1238, though the custom devbox-specific logic is a bit less than `nixprofile`. Again, open to alternatives.

**Implementation note**
To maximally preserve git blame, I did 
1. `git mv nix/search.go nixsearch/search.go`
2. `git add nix/searching.go` a new file

## How was it tested?

compiles

- [ ] testscripts should pass


savil:
* main
  * **PR #1238** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1238" target="_blank">Graphite</a> 
    * **PR #1239** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1239" target="_blank">Graphite</a> 
      * **PR #1240** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1240" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1240?utm_source=stack-comment).

mikeland73 (changes requested):
Is this needed after https://github.com/jetpack-io/devbox/pull/1239? 

I don't think it's a good idea to create so many mini packages.

savil:
too many merge-conflcts... will redo.

PR #1245, opened 2023-06-30
call nix.System directly from lock.lockfile

savil:
## Summary

Removes the workaround hack I had made

## How was it tested?

compiles

PR #1247, opened 2023-07-01
move nix.System call inside featureflag.RemoveNixpkgs if-condition

savil:
## Summary

I'm not yet entirely sure why, but #1245 led to some testscript error.
For now, I'm hiding the nix.System call inside the feature flag's if-condition
to get the testscripts status to green again on `main`.

## How was it tested?

testscripts should pass


savil:
* main
  * **PR #1247** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1247" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1247?utm_source=stack-comment).

mikeland73 (approved):

PR #1249, opened 2023-07-03
[search] Update search endpoints

mikeland73:
## Summary

This updates search endpoints to use new `v1` endpoints. It's a fairly small change, but required modernizing the format we use to unmarshal results. I took advantage to make a few low risk refactoring changes:

* Removed lock dependency from search. This allows lock to have an actual searcher instance and not an interface.
* Removed `devpkgutil` package and incorporated to searcher.
* Improved search command to truncate results/versions to improve ui/ux. Added flag to show all commands.
* Improved search command by allowing version to be specified to get a preview of what it will resolve to. This can be improved by doing server side filtering and using search instead of resolve.
* Fixed `devbox update` bug where updating was causing lock files entries to get replaced even if version had not changed. (This causes nixpkgs to change which slows stuff down)

## How was it tested?

* Ran `devbox update`
* Tested `devbox add python@3.10`
* Tested `devbox search` with many values, with and without the `--show-all` flag

image


internal/lock/resolve.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	packageVersion, err := searcher.Client().Resolve(name, version)
+	if err != nil {
+		return nil, errors.Wrapf(nix.ErrPackageNotFound, "%s@%s", name, version)
+	}
+
+	sysInfos := map[string]*SystemInfo{}
+	if featureflag.RemoveNixpkgs.Enabled() {
+		sysInfos = buildLockSystemInfos(*packageVersion)
```
savil, line 35: why dereference, instead of passing the pointer in?

internal/searcher/client.go
```diff
@@ -32,100 +26,37 @@ func Client() *client {
...
+	return execGet[SearchResults](searchURL)
 }
 
-// SearchOption returns a string for query to be appended to the endpoint
-type SearchOption func() string
-
-func WithVersion(version string) SearchOption {
-	return func() string {
-		return "&v=" + url.QueryEscape(version)
+func (c *client) Resolve(name, version string) (*PackageVersion, error) {
```
savil, line 43: could you add a doc-block comment for what resolve does?

internal/searcher/client.go
```diff
@@ -32,100 +26,37 @@ func Client() *client {
...
-			StoreName:    sysInfo.StoreName,
-			StoreVersion: sysInfo.StoreVersion,
-		}
-	}
-	return systemInfos, nil
+	return execGet[PackageVersion](searchURL)
 }
 
-func execSearch(url string) (*SearchResult, error) {
+func execGet[T any](url string) (*T, error) {
```
savil, line 59: 👍🏾 I feel my programming brain has gotten too comfortable with golang when a simple thing like templates feels novel again :p

internal/lock/resolve.go
```diff
@@ -0,0 +1,88 @@
...
+			Format(time.RFC3339),
+		Resolved: fmt.Sprintf(
+			"github:NixOS/nixpkgs/%s#%s",
+			packageInfo.CommitHash,
+			packageInfo.AttrPaths[0],
+		),
+		Version: packageInfo.Version,
+		Systems: sysInfos,
+	}, nil
+}
```
savil, line 57: okay, wow, this is much cleaner than the old code!

Also, a 100 bonus points for having the `searcher` functions return `searcher` structs, and transforming them into `lock` structs here. I was also thinking of doing that.

savil (approved):
This is awesome! Please land asap so I can build on it :)

Issue #1248, opened 2023-07-03
[Feature]: Add trailing newline to Devbox files


maxdeviant:
**Is your feature request related to a problem you're trying to solve? Please describe.**

Currently Devbox does not write trailing newlines to its various files (`devbox.json`, `devbox.lock`, etc).

This results in a visual indicator in Git diffs showing the lack of a trailing newline:

<img width="564" alt="Screenshot 2023-07-03 at 5 19 01 PM" src="https://github.com/jetpack-io/devbox/assets/1486634/377b3602-ce64-4f4a-86d6-97768a5030b2">

<img width="851" alt="Screenshot 2023-07-03 at 5 19 08 PM" src="https://github.com/jetpack-io/devbox/assets/1486634/af4a3f56-3006-4ac1-ad79-77469ecaad59">

**Describe the solution you'd like**

I would like it if Devbox did any one of the following:
1. Wrote trailing newlines by default
2. Had a flag/setting to write trailing newlines
3. Retained trailing newlines added by hand

**Describe alternatives you've considered**

I don't see a reason why Devbox shouldn't support trailing newlines.

savil:
@maxdeviant thanks for filing this!

Lagoja:
Hey @maxdeviant, this should be in the latest 0.5.7 Release. Please let us know if this is not resolved for you

maxdeviant:
> Hey @maxdeviant, this should be in the latest 0.5.7 Release. Please let us know if this is not resolved for you

Thanks so much!

PR #1251, opened 2023-07-05
[CLI] Added a "\n" to end of writefile

mohsenari:
## Summary
Added '\n' to end of `WriteFile()` in `cuecfg` package. This resolves an issue of devbox.json unnecessarily showing up in PR diffs because the newline at the end of the file is removed by CLI and was added by IDE formatters.

However, this function is used to write the config to a file as well as a couple of other cases such as in `auth` package. 
I'm guessing the `\n` for cases other than writing config won't cause an issue. But I haven't tested those cases.

Addresses #1248 

## How was it tested?
- compile
- `./devbox init` has a newline at the end of devbox.json and devbox.lock
- `./devbox add hello` keeps the newline at the end of devbox.json and devbox.lock

### DIFF
--- a/internal/cuecfg/cuecfg.go
+++ b/internal/cuecfg/cuecfg.go
@@ -82,7 +82,7 @@ func WriteFile(path string, value any) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	data = append(data, "\n"...)
+	data = append(data, '\n')
 	return errors.WithStack(os.WriteFile(path, data, 0644))
 }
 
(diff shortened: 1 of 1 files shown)


internal/cuecfg/cuecfg.go
```diff
@@ -82,7 +82,7 @@ func WriteFile(path string, value any) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-
+	data = append(data, "\n"...)
```
mikeland73, line 85: nit, newline is a single character:
`data = append(data, '\n')`

mikeland73 (approved):
This has been annoying me for a while, thanks!

PR #1252, opened 2023-07-05
Fix script gen clean up failure

LucilleH:
## Summary
Getting the following error when running any script
```
[DEBUG] 2023/07/05 21:06:23 go.jetpack.io/devbox/internal/shellgen/scripts.go:73: failed to clean up script file build.sh, error = remove ~/devbox/.devbox/gen/scripts/build.sh.sh: no such file or directory
```

It looks like `.sh` is wrongly appended to the existing files under the scripts directory. This PR fixes that.

## How was it tested?
devbox run build


savil (approved):
Thanks!

PR #1255, opened 2023-07-06
[remove nixpkgs] Enable installing and removing packages

savil:
## Summary

builds on top of #1236

Goal: 
We should be able to install the package without downloading nixpkgs.

Changes made to:
- `nixprofile.ProfileInstall`: use `/nix/store/<hash>` as `installable`
- `nixprofile.ProfileListIndex`: return index when `item.nixStorePath == pkg.PathInStore`
- `devpkgs.Package.ValidateExists`: `true` if `IsInBinaryStore`
- `devpkgs.Package.EnsureNixpkgsPrefetched`: skip if `IsInBinaryStore`.

## How was it tested?

`devbox add go@1.20.3` and `devbox rm go@1.20.3`
- [x] nixpkgs not downloaded

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
+	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/cuecfg"
 	"go.jetpack.io/devbox/internal/lock"
@@ -324,6 +325,15 @@ func (p *Package) ValidateExists() (bool, error) {
 	if p.isVersioned() && p.version() == "" {
 		return false, usererr.New("No version specified for %q.", p.Path)
 	}
+
+	isInStore, err := p.IsInBinaryStore()
+	if err != nil {
+		return false, err
+	}
+	if isInStore {
+		return true, nil
+	}
+
 	info, err := p.NormalizedPackageAttributePath()
 	return info != "", err
 }
@@ -378,6 +388,17 @@ func (p *Package) LegacyToVersioned() string {
 }
 
 func (p *Package) EnsureNixpkgsPrefetched(w io.Writer) error {
+
+	isInStore, err := p.IsInBinaryStore()
+	if err != nil {
+		return err
+	}
+	if isInStore {
+		// We can skip prefetching nixpkgs, if this package is in the binary
+		// cache store.
+		return nil
+	}
+
 	hash := p.HashFromNixPkgsURL()
 	if hash == "" {
 		return nil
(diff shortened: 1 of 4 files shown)


internal/devpkg/package.go
```diff
@@ -378,6 +388,17 @@ func (p *Package) LegacyToVersioned() string {
...
 func (p *Package) EnsureNixpkgsPrefetched(w io.Writer) error {
+
+	isInStore, err := p.IsInBinaryStore()
+	if err != nil {
+		return err
+	}
+	if isInStore {
+		// We can skip prefetching nixpkgs, if this package is in the binary
+		// cache store.
+		return nil
```
LucilleH, line 399: COol!

internal/nix/nixprofile/profile.go
```diff
@@ -69,6 +69,23 @@ func ProfileListIndex(args *ProfileListIndexArgs) (int, error) {
...
+		return -1, err
+	}
+	if inStore {
+		pathInStore, err := args.Input.PathInBinaryStore()
+		if err != nil {
+			return -1, err
+		}
+		for _, item := range list {
+			if pathInStore == item.nixStorePath {
+				return item.index, nil
```
LucilleH, line 83: What's index again? Priority of the ordering of package items?

internal/nix/nixprofile/profile.go
```diff
@@ -204,12 +227,24 @@ func ProfileInstall(args *ProfileInstallArgs) error {
...
+		if err != nil {
+			return err
+		}
+	} else if installable == "" {
+		// fallback if PathInBinaryStore is not applicable.
+		var err error
+		installable, err = input.URLForInstall()
+		if err != nil {
+			return err
+		}
```
LucilleH, line 244: Can we extract line 230 to 244 in it's own separate function that simply returns `(installable, error)`? It is getting a bit hard to read/keep track of

LucilleH (approved):

internal/nix/nixprofile/profile.go
savil: index is the number of the items printed when running `nix profile list`.

For this example output, the index is the first column:
```
> nix profile list
0 - - /nix/store/k5jplv7zj17bh58wjah949p30cn9rm92-devenv
1 - - /nix/store/l7x6hdaml5g158d2lsrr5jl3nmjksvj6-user-environment
2 - - /nix/store/2piikn5jm7mbsp1j6nw9n0yzfl6yy56x-sqlite-3.39.2-bin
```

We can use the index to do `nix profile remove <index>` (or something like that)

internal/nix/nixprofile/profile.go
```diff
@@ -234,3 +257,28 @@ func ProfileRemoveItems(profilePath string, items []*NixProfileListItem) error {
 	}
 	return nix.ProfileRemove(profilePath, indexes)
 }
+
+// installableForPackage determines how nix profile should install this package.
+// Keeping in `nixprofile` package since its specific to how nix profile works,
+// rather than a general property of devpkg.Package
+func installableForPackage(pkg *devpkg.Package) (string, error) {
```
mikeland73, line 264: Should this function just be part of devpkg.Package so that `URLForInstall` always returns the correct value?

We can rename `URLForInstall` if needed

PR #1256, opened 2023-07-07
[remove nixpkgs] add toPath, and edit devbox.lock only on update command

savil:
## Summary

This PR rolls in a few fixes:
1. Add `toPath` to `builtins.fetchClosure`
    - lets us remove `--impure` flag in `nix print-dev-env`
2. Add `ca_store_path` to `devbox.lock`
    - Because this is system-dependent, _and_ is calculated locally, a `devbox update` will only update the user's system's `ca_store_path`. 
    - This is good for team-mates on the same system, and to avoid the slow `nix store make-content-addressed <path>` call on each `devbox shell/run/shellenv`.
3. Only update `devbox.lock` during `devbox update`. 
    - Previously, it would update system info on `devbox shell/run/shellenv`.
4. Simplify `devbox.lock` schema's `SystemInfo` to have `StorePath` and `CAStorePath`. These are used as `fromPath` and `toPath` in `builtins.fetchClosure`.

Miscellaneous changes:
1. I remove `actionlint` from the `devbox.json`. It is not core to our workflows, and I am working with very minimal space on devbox cloud VM. This reduces "out of space errors" marginally.

## How was it tested?

In `examples/development/go/hello-world`:
`devbox update` and `devbox shell`

The `devbox.lock` diff is now 
```
❯ git diff
diff --git a/examples/development/go/hello-world/devbox.lock b/examples/development/go/hello-world/
devbox.lock
index 34556d3f..168874db 100644
--- a/examples/development/go/hello-world/devbox.lock
+++ b/examples/development/go/hello-world/devbox.lock
@@ -2,9 +2,18 @@
   "lockfile_version": "1",
   "packages": {
     "go@1.19.8": {
-      "last_modified": "2023-05-01T16:53:22Z",
-      "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#go_1_19",
-      "version": "1.19.8"
+      "last_modified": "2023-05-03T02:55:54Z",
+      "resolved": "github:NixOS/nixpkgs/0d373d5af960504dd60c3d06c65e553b36ef29d8#go_1_19",

+      "version": "1.19.8",
+      "systems": {
+        "aarch64-darwin": {
+          "store_path": "/nix/store/7m99ip3616l3z670y83p34r3plwd4iq1-go-1.19.8"
+        },
+        "x86_64-linux": {
+          "store_path": "/nix/store/r0x0agq0vwn0p6z99vkkvn8l8a8idzsb-go-1.19.8",
+          "ca_store_path": "/nix/store/ns557l24yf1f16f9jvbmxv57qdkkn2mj-go-1.19.8"
+        }
+      }
     }
   }
-}
\ No newline at end of file
+}
```

### DIFF
--- a/internal/impl/update.go
+++ b/internal/impl/update.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/devpkg"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/nix/nixprofile"
@@ -90,7 +91,13 @@ func (d *Devbox) updateDevboxPackage(
 	if err != nil {
 		return err
 	}
-	if existing != nil && existing.Version != newEntry.Version {
+	if existing == nil {
+		ux.Finfo(d.writer, "Resolved %s to %[1]s %[2]s\n", pkg, newEntry.Resolved)
+		d.lockfile.Packages[pkg.Raw] = newEntry
+		return nil
+	}
+
+	if existing.Version != newEntry.Version {
 		ux.Finfo(d.writer, "Updating %s %s -> %s\n", pkg, existing.Version, newEntry.Version)
 		if err := d.removePackagesFromProfile(ctx, []string{pkg.Raw}); err != nil {
 			// Warn but continue. TODO(landau): ensurePackagesAreInstalled should
@@ -98,13 +105,38 @@ func (d *Devbox) updateDevboxPackage(
 			ux.Fwarning(d.writer, "Failed to remove %s from profile: %s\n", pkg, err)
 		}
 		d.lockfile.Packages[pkg.Raw] = newEntry
-	} else if existing == nil {
-		ux.Finfo(d.writer, "Resolved %s to %[1]s %[2]s\n", pkg, newEntry.Resolved)
-		d.lockfile.Packages[pkg.Raw] = newEntry
-	} else {
-		ux.Finfo(d.writer, "Already up-to-date %s %s\n", pkg, existing.Version)
+		return nil
 	}
 
+	// Check if the package's system info is missing, or not complete.
+	userSystem, err := nix.System()
+	if err != nil {
+		return err
+	}
+	var needsSysInfo bool
+	var needsLocalStorePath bool
+	if featureflag.RemoveNixpkgs.Enabled() {
+		userSysInfo := d.lockfile.Packages[pkg.Raw].Systems[userSystem]
+		needsSysInfo = userSysInfo == nil
+		if !needsSysInfo {
+			// Check if the LocalStorePath is missing for the user's system.
+			// Since any one user cannot add this field for all systems, 
+			// we'll need to progressively add it to a project's lockfile.
+			needsLocalStorePath = userSysInfo.LocalStorePath == ""
+		}
+	}
+	if needsSysInfo {
+		d.lockfile.Packages[pkg.Raw] = newEntry
+		ux.Finfo(d.writer, "Updated system information for %s\n", pkg)
+		return nil
+	} else if needsLocalStorePath {
+		// Update the LocalStorePath for the user's system
+		d.lockfile.Packages[pkg.Raw].Systems[userSystem].LocalStorePath = newEntry.Systems[userSystem].LocalStorePath
+		ux.Finfo(d.writer, "Updated system information for %s\n", pkg)
+		return nil
+	}
+
+	ux.Finfo(d.writer, "Already up-to-date %s %s\n", pkg, existing.Version)
 	return nil
 }
 
--- a/internal/impl/update.go
+++ b/internal/impl/update.go
@@ -109,31 +109,29 @@ func (d *Devbox) updateDevboxPackage(
 	}
 
 	// Check if the package's system info is missing, or not complete.
-	userSystem, err := nix.System()
-	if err != nil {
-		return err
-	}
-	var needsSysInfo bool
-	var needsLocalStorePath bool
 	if featureflag.RemoveNixpkgs.Enabled() {
-		userSysInfo := d.lockfile.Packages[pkg.Raw].Systems[userSystem]
-		needsSysInfo = userSysInfo == nil
-		if !needsSysInfo {
-			// Check if the LocalStorePath is missing for the user's system.
-			// Since any one user cannot add this field for all systems, 
-			// we'll need to progressively add it to a project's lockfile.
-			needsLocalStorePath = userSysInfo.LocalStorePath == ""
+		userSystem, err := nix.System()
+		if err != nil {
+			return err
 		}
-	}
-	if needsSysInfo {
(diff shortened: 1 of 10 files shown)


LucilleH (approved):

internal/nix/store.go
```diff
@@ -0,0 +1,42 @@
...
+	return filepath.Join("/nix/store", storeDir)
+}
+
+// contentAddressedRegex matches the output of `nix store make-content-addressed`.
+// It is used to select the content-addressed store path (the second one in the example below).
+//
+// Example:
+// > nix store make-content-addressed  /nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1
+// rewrote '/nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1' to '/nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1'
+var contentAddressedRegex = regexp.MustCompile(`rewrote\s'[\/a-z0-9-\.]+'\sto\s'([a-z0-9-\/\.]+)'`)
```
gcurtis, line 26: Does `nix store make-content-addressed --json` work?

internal/nix/store_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestContentAddressedPath(t *testing.T) {
+
+	testCases := []struct {
+		storePath string
+		expected  string
+	}{
+		{
+			"/nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1",
+			"/nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1",
```
gcurtis, line 16: Does Nix automatically download these if they're not in the store yet?

internal/nix/store_test.go
```diff
@@ -0,0 +1,32 @@
+package nix
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestContentAddressedPath(t *testing.T) {
```
gcurtis, line 8: 👍🏻

gcurtis (commented):
I'm finding some of the new terms a bit confusing. Could we make them match what Nix calls them?

- `BinaryStorePath` - this can just be `StorePath`. There isn't any difference between a path in a remote binary cache and a local store.
- `LocalStorePath` - it looks like this is just the content-addressed version of the store path. Could we call it `CAStorePath` (or `ContentAddressedStorePath` if CA sounds like certificate)?
- `BinaryStore` → `BinaryCache`

internal/nix/store_test.go
savil: yep

internal/nix/store.go
savil: oh yeah, good call. Will try that...

savil:
> I'm finding some of the new terms a bit confusing. Could we make them match what Nix calls them?
> 
> * `BinaryStorePath` - this can just be `StorePath`. There isn't any difference between a path in a remote binary cache and a local store.
> * `LocalStorePath` - it looks like this is just the content-addressed version of the store path. Could we call it `CAStorePath` (or `ContentAddressedStorePath` if CA sounds like certificate)?
> * `BinaryStore` → `BinaryCache`

Sounds good to me. Don't feel strongly about the current names (except I didn't want it to be "fromPath" and "toPath", since those don't make sense outside the context of `fetchClosure`)

internal/lock/lockfile.go
```diff
@@ -40,12 +38,13 @@ type Package struct {
...
-	StoreName    string `json:"store_name,omitempty"`
-	StoreVersion string `json:"store_version,omitempty"`
-	ToHash       string `json:"to_hash,omitempty"`
+	// StorePath is the cache key in the Binary Cache Store (cache.nixos.org)
+	// It is of the form <hash>-<name>-<version>
+	// <name> may be different from the canonicalName so we store the full store path.
+	StorePath string `json:"store_path,omitempty"`
+	// CAStorePath is the content-addressed path for the nix package in /nix/store
+	// It is of the form <hash>-<name>-<version>
+	CAStorePath string `json:"local_store_path,omitempty"`
```
savil, line 47: oops, forgot to rename the key to `ca_store_path`

savil (commented):
All CICD has passed. Going to make a small "rename" I missed out and then land...

internal/impl/update.go
```diff
@@ -90,21 +91,50 @@ func (d *Devbox) updateDevboxPackage(
...
+	if featureflag.RemoveNixpkgs.Enabled() {
+		userSystem, err := nix.System()
+		if err != nil {
+			return err
+		}
+
+		// Check if the system info is missing for the user's system.
+		sysInfo := d.lockfile.Packages[pkg.Raw].Systems[userSystem]
+		if sysInfo == nil {
+			d.lockfile.Packages[pkg.Raw] = newEntry
```
mikeland73, line 121: Wait, could this override an existing (different) system that already has a CA path?

internal/impl/update.go
```diff
@@ -90,21 +91,50 @@ func (d *Devbox) updateDevboxPackage(
...
-	} else if existing == nil {
-		ux.Finfo(d.writer, "Resolved %s to %[1]s %[2]s\n", pkg, newEntry.Resolved)
-		d.lockfile.Packages[pkg.Raw] = newEntry
-	} else {
-		ux.Finfo(d.writer, "Already up-to-date %s %s\n", pkg, existing.Version)
+		return nil
+	}
+
+	// Check if the package's system info is missing, or not complete.
+	if featureflag.RemoveNixpkgs.Enabled() {
```
mikeland73, line 112: Most of the logic in this block could be encapsulated in lock file. Maybe:
```golang
if existing.CanBeUpdated(newEntry) {
  d.lockfile.Update(existing, newEntry)
}
``` 

or something like that. It could deal with version changes and also missing system stuff.

That way we could make `Package.Systems`  semi-private. (It still needs to be exported because it is part of the JSON)

mikeland73: Also, is this logic block missing profile install/uninstall? e.g. if a package previously required nixpkgs and we replace it with one that does not require it, do we want to update the profile?

savil: yes, I agree this could likely be moved into the `lock` package. In general, I think quite a bit of the update code (beyond this block) could be better encapsulated. I may not have time to get to it though...

internal/impl/update.go
savil: You are right! Much obliged 🙏🏾 

sending fix: https://github.com/jetpack-io/devbox/pull/1259

internal/impl/update.go
savil: > Also, is this logic block missing profile install/uninstall? e.g. if a package previously required nixpkgs and we replace it with one that does not require it, do we want to update the profile?

🤔 yeah, this one I am unsure about. In theory, the package's binary should be the exact same. 

Its also why I think it was okay and/or put a question mark after using `InputAddressedPath` in the `Package.Installable` [here](https://github.com/jetpack-io/devbox/blob/main/internal/devpkg/package.go#L157-L158).

@gcurtis would you have an opinion?

PR #1257, opened 2023-07-07
[remove nixpkgs] Naming: update Package methods to InputAddressedPath and ContentAddressedPath, and s/BinaryStore/BinaryCache

savil:
## Summary

This PR is a follow up to this request from @gcurtis:
https://github.com/jetpack-io/devbox/pull/1256#pullrequestreview-1517592912

BUT having `NixStorePath` and `ContentAddressedPath` seemed odd since content-addressed paths
are also nix store paths. So, in this PR, I'm explicit about `NixStorePath` being
`InputAddressedPath`.

Also - changed `BinaryStore` to `BinaryCache`.

**RFC**:
I didn't change the `lock.SystemInfo` struct, I've left it as:
```
type SystemInfo struct {
	// StorePath is the input-addressed path for the nix package in /nix/store
	// It is the cache key in the Binary Cache Store (cache.nixos.org)
	// It is of the form <hash>-<name>-<version>
	// <name> may be different from the canonicalName so we store the full store path.
	StorePath string `json:"store_path,omitempty"`
	// CAStorePath is the content-addressed path for the nix package in /nix/store
	// It is of the form <hash>-<name>-<version>
	CAStorePath string `json:"ca_store_path,omitempty"`
}
```

I think this could be okay, because the default `store_path` for nix is input-addressed, for now at least.

If you'd like I can change `StorePath` to be explicitly:
```
InputAddrStorePath string `json:"input_addr_path, omitempty"
ContentAddrStorePath string `json:"content_addr_path, omitempty"
```

Let me know!

## How was it tested?

compiles

### DIFF
--- a/internal/nix/nixprofile/profile.go
+++ b/internal/nix/nixprofile/profile.go
@@ -31,7 +31,7 @@ func ProfileListItems(
 	}
 
 	// The `line` output is of the form:
-	// <index> <UnlockedReference> <LockedReference> <NixStorePath>
+	// <index> <UnlockedReference> <LockedReference> <InputAddressedPath>
 	//
 	// Using an example:
 	// 0 github:NixOS/nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04#legacyPackages.x86_64-darwin.go_1_19 github:NixOS/nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04#legacyPackages.x86_64-darwin.go_1_19 /nix/store/w0lyimyyxxfl3gw40n46rpn1yjrl3q85-go-1.19.3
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #1257** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1257" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1257?utm_source=stack-comment).

LucilleH (approved):
oh this is much clearer!

mikeland73 (approved):

gcurtis (approved):
Thanks! This made things a lot easier to understand.

internal/nix/nixprofile/profile.go
```diff
@@ -31,7 +31,7 @@ func ProfileListItems(
 	}
 
 	// The `line` output is of the form:
-	// <index> <UnlockedReference> <LockedReference> <NixStorePath>
+	// <index> <UnlockedReference> <LockedReference> <InputAddressedPath>
```
savil, line 34: will revert this line

PR #1259, opened 2023-07-07
[remove nixpkgs] bug fix: update only the user's sysInfo, rather than for all systems

savil:
## Summary

@mikeland73 points out that this line was erroneously updating all SystemInfos:
https://github.com/jetpack-io/devbox/pull/1256/files#r1256408509

Instead, we should only update the missing SystemInfo of the user's current system.

## How was it tested?

- [x] do `devbox update` of a lockfile having another system's SystemInfo

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -509,8 +509,8 @@ func (p *Package) ContentAddressedPath() (string, error) {
 
 	ux.Fwarning(
 		os.Stderr,
-		"calculating local_store_path. This may be slow.\n"+
-			"Run `devbox update` to speed this up for next time.",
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1259** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1259" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1259?utm_source=stack-comment).

mikeland73 (approved):

savil:
will land once tests pass

Issue #881, opened 2023-04-08
[Bug]: devbox shell startup occassionally ~20-30s 


altano:
**Current Behavior (bug)**
I frequently get `devbox shell` times of 20-30s. MOST of the time startup is very short, e.g. I just ran  `time devbox run echo hello` and got 0.1s. But occasionally, usually after not using my computer for a day or two, my first `devbox shell` will take a really long time.

```
➜  devbox shell
Ensuring packages are installed.
Starting a devbox shell...
dockedirenv: loading ~/src/docker/.envrc
direnv: using devbox
direnv: using nix .devbox/gen/shell.nix
r compose up -d --build
direnv: ([/opt/homebrew/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
```
... ~20-30 second pause ...

```
direnv: export +NIX_BUILD_TOP +TEMP +TEMPDIR +TMP ~LD_LIBRARY_PATH ~LIBRARY_PATH ~TMPDIR ~buildPhase
direnv: loading ~/src/docker/everace/.envrc
direnv: export +DOCKER_CONTEXT +DOCKER_HOST
```

**Expected Behavior (fix)**
<1s shell startup, consistently

**Additional context**
```
(devbox) ➜  devbox version -v
Version:     0.4.6
Platform:    darwin_arm64
Commit:      2ddf1856034dd544c5ff966e20e7bef1f3897fbb
Commit Time: 2023-03-31T19:14:57Z
Go Version:  go1.19.7
```
devbox.json
```
{
  "packages": ["docker-client"],
  "shell": {
    "init_hook": null
  },
  "nixpkgs": {
    "commit": "f80ac848e3d6f0c12c52758c0f25c10c97ca3b62"
  }
}
```

.envrc
```
# Automatically sets up your devbox environment whenever you cd into this
# directory via our direnv integration:

use_devbox() {
    watch_file devbox.json
    if [ -f .devbox/gen/shell.nix ]; then
        DEVBOX_SHELL_ENABLED_BACKUP=$DEVBOX_SHELL_ENABLED
        use nix .devbox/gen/shell.nix
        eval $(devbox shell --print-env)
        export DEVBOX_SHELL_ENABLED=$DEVBOX_SHELL_ENABLED_BACKUP
    fi
}
use devbox

# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
# for more details
```

savil:
[...]

PR #1260, opened 2023-07-07
[bug] Added abs path ref. to devbox binary in wrappers

mohsenari:
## Summary
This fixes an issue where if you install `coreutils` devbox shell in some cases gets stuck in infinite loop due this sequence of events:
1. devbox shell creates wrappers
2. wrappers try to omit wrappers from PATH and call devbox shellenv but it's unsuccessful 
3. devbox call in wrappers uses a couple of util tools (like `tr`) that now have their own wrappers because of coreutils.
4. the process is never able to skip the wrappers hence showing `fork: Resource Temporarily Unavailable`

I think it addresses #881 or at least a portion of it.

The fix:
in step 3 instead of calling devbox (launcher script in /usr/local/bin) we get the absolute path to the devbox binary and reference that directly in wrappers script.
This way the launcher script is not called more than once and avoids getting stuck in infinite loop.

## How was it tested?
- compile -> compiled binary is now in <path to devbox>/dist/devbox
- update line 413 of launcher (/usr/local/devbox) to `local -r bin="<path to devbox>/dist/devbox"`
- in an empty directory run devbox init
- add `coreutils@latest` to devbox.json
- run `devbox shell`

### DIFF
--- a/internal/wrapnix/wrapper.sh.tmpl
+++ b/internal/wrapnix/wrapper.sh.tmpl
@@ -19,7 +19,7 @@ DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
 
 if [[ "${{ .ShellEnvHashKey }}" != "{{ .ShellEnvHash }}" ]] && [[ -z "${{ .ShellEnvHashKey }}_GUARD" ]]; then
 export {{ .ShellEnvHashKey }}_GUARD=true
-eval "$(DO_NOT_TRACK=1 devbox shellenv -c {{ .ProjectDir }})"
+eval "$(DO_NOT_TRACK=1 {{ .DevboxBinaryPath }} shellenv -c {{ .ProjectDir }})"
 fi
 
 {{/*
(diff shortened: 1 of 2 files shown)


internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -29,6 +29,6 @@ should be in PATH.
 
 DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
 */ -}}
-eval "$(DO_NOT_TRACK=1 devbox shellenv only-path-without-wrappers)"
+eval "$(DO_NOT_TRACK=1 {{ .DevboxBinaryPath }} shellenv only-path-without-wrappers)"
```
gcurtis, line 32:
```suggestion
eval "$(DO_NOT_TRACK=1 '{{ .DevboxBinaryPath }}' shellenv only-path-without-wrappers)"
```

gcurtis (approved):

PR #1261, opened 2023-07-08
Add a Java Spring Boot Example

Lagoja:
## Summary

Example that uses Java Spring Boot + MySQL 

## How was it tested?

`devbox install`
`devbox run setup_db`
`devbox run runBoot`


mohsenari (approved):

examples/stacks/spring/src/main/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+spring.jpa.hibernate.ddl-auto=update
+spring.datasource.url=jdbc:mysql://${MYSQL_HOST:localhost}:3306/db_example
+spring.datasource.username=springuser
+spring.datasource.password=password
```
mohsenari, line 4: nit: put a comment here to remind people to update this username/password and also exclude it from git tracking for production usecases.

examples/stacks/spring/gradle/wrapper/gradle-wrapper.jar
mohsenari, line 1: I think this jar file isn't needed and can be deleted

PR #1264, opened 2023-07-10
devbox add fallthrough and fallback for packages not handled by search

LucilleH:
## Summary
`devbox add` fallthrough and fallback for packages not handled by search.

This PR does not do the following:
1. Print a warning that a package is in a fallthrough case (I'm unclear if a warning is needed?)
2. Accept an alternative nixpkg commit hash for that specific package (I'm unclear if we are deprecating `nixpkg` commit field in `devbox.json` immediately?)

I'd like to address those two things above if needed in a separate PR as a follow up.

Example lockfile:
```
{
  "lockfile_version": "1",
  "packages": {
    "python310": {
      "plugin_version": "0.0.1",
      "resolved": "github:NixOS/nixpkgs/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62#python310",
      "source": "nixpkg"
    },
    "python310Packages.pip@latest": {
      "last_modified": "2023-05-06T16:57:53Z",
      "plugin_version": "0.0.1",
      "resolved": "github:NixOS/nixpkgs/16b3b0c53b1ee8936739f8c588544e7fcec3fc60#python310Packages.pip",
      "source": "devbox-search",
      "version": "23.0.1"
    },
    "stdenv.cc.cc.lib": {
      "resolved": "github:NixOS/nixpkgs/f80ac848e3d6f0c12c52758c0f25c10c97ca3b62#stdenv.cc.cc.lib",
      "source": "nixpkg"
    }
  }
}

```

## How was it tested?
In the Tensorflow example:
```
devbox add stdenv.cc.cc.lib
devbox add nodejs-16_x
devbox update
```

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -403,7 +403,7 @@ func (p *Package) Versioned() string {
 }
 
 func (p *Package) IsLegacy() bool {
-	return p.isDevboxPackage() && !p.isVersioned()
+	return p.isDevboxPackage() && !p.isVersioned() && p.lockfile.Source(p.Raw) == ""
 }
 
 func (p *Package) LegacyToVersioned() string {
--- a/internal/devpkg/package_test.go
+++ b/internal/devpkg/package_test.go
@@ -119,6 +119,10 @@ func (l *lockfile) LegacyNixpkgsPath(pkg string) string {
 	)
 }
 
+func (l *lockfile) Source(pkg string) string {
+	return ""
+}
+
 func (l *lockfile) Resolve(pkg string) (*lock.Package, error) {
 	switch {
 	case strings.Contains(pkg, "path:"):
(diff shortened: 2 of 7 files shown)


mikeland73 (commented):
This is a great and simple start, but I think there a few issues we may need to solve before we merge (or ship?)

* Adding fall-through package causes the warning `Warning: Your devbox.json contains packages in legacy format. Please run devbox update to update your devbox.json.` to be shown on subsequent commands. 
* Running `devbox update` will add `@latest` to the packages in `devbox.json` which will cause them to fail on the next run/shell. (Adding `@latest` was our upgrade path for old configs)

Possible solution:

* Add "source" field to lock file packages. Source can be devbox-search, nixpkgs. etc. Only show legacy warning is source is unknown and package is in legacy format.
* When doing `devbox update` don't update stuff with nixpkgs source. If lockfile/source is missing then check if package exists in search before appending the `@latest`.

internal/devpkg/package.go
```diff
@@ -403,7 +403,7 @@ func (p *Package) Versioned() string {
 }
 
 func (p *Package) IsLegacy() bool {
-	return p.isDevboxPackage() && !p.isVersioned()
+	return p.isDevboxPackage() && !p.isVersioned() && p.lockfile.Source(p.Raw) == ""
```
mikeland73, line 406: Another possible API could be

`p.lockfile.Get(p.Raw).Source()`  to have lock packages themselves respond to source()

(and just default Source() to be empty string if not set or if package is nil)

mikeland73 (approved):
Nice! This is super clean

PR #1266, opened 2023-07-11
Update lockfile api for package source

LucilleH:
## Summary
Update lockfile api for package source as suggested in https://github.com/jetpack-io/devbox/pull/1264#discussion_r1259057306

## How was it tested?
devbox run build
devbox update

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -403,7 +403,12 @@ func (p *Package) Versioned() string {
 }
 
 func (p *Package) IsLegacy() bool {
-	return p.isDevboxPackage() && !p.isVersioned() && p.lockfile.Source(p.Raw) == ""
+	var packageSource = ""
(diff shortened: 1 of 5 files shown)


internal/devpkg/package.go
```diff
@@ -403,7 +403,12 @@ func (p *Package) Versioned() string {
 }
 
 func (p *Package) IsLegacy() bool {
-	return p.isDevboxPackage() && !p.isVersioned() && p.lockfile.Source(p.Raw) == ""
+	var packageSource = ""
+	if p.lockfile.Get(p.Raw) != nil {
```
mikeland73, line 407: golang hack

in lock package:

```golang
func (p *Package) GetSource() string {
  if p == nil {
    return ""
  }
  return p.Source
}
```

mikeland73 (approved):

PR #1279, opened 2023-07-17
[errors] Show nicer error when installing a package that is not supported by current system

mikeland73:
## Summary

Shows better error messages in following cases:
* When package doesn't exist on search or in nixpkgs as attribute path
* When package is not supported by current platform

## How was it tested?

`devbox add glibcLocales@2.37-8`
`devbox add fuse3`

image


internal/nix/eval.go
```diff
@@ -46,6 +46,21 @@ func PackageKnownVulnerabilities(path string) []string {
 	return vulnerabilities
 }
 
+func PackagePlatforms(path string) []string {
+	cmd := command("eval", path+".meta.platforms", "--json")
```
gcurtis, line 50: What if instead of searching and eval'ing the platforms we just eval the derivation's outPath and check the output? For example:

```
# Bad
$ nix eval --raw nixpkgs#fuse3
error: Package ‘fuse-3.11.0’ in /nix/store/b63zicy7mkwlp4vmca056kawsrmxfr36-source/pkgs/os-specific/linux/fuse/common.nix:92 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-apple-darwin"
         package.meta.platforms = [
...
```

vs.

```
# Good
$ nix eval --raw nixpkgs#fuse 
/nix/store/8pqmbsp8zkq8656k5yaravbaxixaaa3c-macfuse-stubs-4.4.1
```

gcurtis (commented):
I think this is pretty close, but I think if we just try evaluating the derivation we could catch more errors. Would that work?

mikeland73:
@gcurtis PTAL

gcurtis (approved):

PR #1281, opened 2023-07-18
[plugins] Implements github plugins

mikeland73:
## Summary

This implements github plugins. They use the following syntax in `devbox.json`:

```json
"include": [
    "github:jetpack-io/devbox-plugin-example/6ea0ef9e12ab58dbbb145ca8f3a465611cb603a9"
  ]
```

revision is optional and defaults to master/main.

Repos must have `devbox.json` at top level.

TODO:

- [x] Better error message for missing devbox.json and other files.
- [ ] Add to lockfile? The are currently locked using optional revision
- [ ] Updates: They are currently don't update unless change is made that busts local.lock cache.

## How was it tested?

image


loreto (approved):
LGTM

PR #1291, opened 2023-07-20
[bugfix] Write back lockfile after devbox update

ipince:
## Summary

After calling `devbox update`, we were not writing back to `devbox.lock`. `ensurePkgsAreInstalled` would then return early because the local lock file hadn't changed (since it re-reads `devbox.lock` from disk, rather than using the updated in-memory lockfile), so it would skip writing the lockfile as well.

This also makes the lockfile a bit more self-healing, since it will overwrite the current user's sysInfo entirely, rather than just adding the CAStorePath. This ensures that the StorePath and CAStorePath are written together.

## How was it tested?
Manually, but plan to add testscripts in separate PR. I tested a few different scenarios, notably:

1. `devbox update` adds system info for current user without changing system info for other systems.
2. If current system info is missing something (either CAStorePath or Path), `devbox update` will restore it.
3. If, say, CAStorePath is missing for the current system, and the user calls `devbox shell`, then we'll print a warning instructing the user to run `devbox update`. Calling it will then actually fix the lockfile and the warning will not be reprinted. It's worth mentioning that the warning will only be shown once to the user, because then they'll cache print-dev-env, so we won't re-exercise the path that computes CAStorePath. Not super ideal, but good enough I think.
4. If current system info is missing entirely, it is added.
5. If current system info exists and matches new one, nothing is done and we print "Already up-to-date"

### DIFF
--- a/internal/impl/update.go
+++ b/internal/impl/update.go
@@ -54,6 +54,10 @@ func (d *Devbox) Update(ctx context.Context, pkgs ...string) error {
 		}
 	}
 
+	if err := d.lockfile.Save(); err != nil {
+		return err
+	}
+
 	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internal/lock/package.go
```diff
@@ -43,3 +43,15 @@ func (p *Package) IsAllowInsecure() bool {
...
 	return p.AllowInsecure
 }
+
+func (i *SystemInfo) Equals(other *SystemInfo) bool {
+	if i == nil {
+		return other == nil
+	}
+	if other == nil {
+		return false
+	}
```
mikeland73, line 53:
```golang
if i == nil || other == nil {
  return i == other
}
return *i == *other
```

internal/impl/update.go
```diff
@@ -54,6 +54,10 @@ func (d *Devbox) Update(ctx context.Context, pkgs ...string) error {
 		}
 	}
 
+	if err := d.lockfile.Save(); err != nil {
```
mikeland73, line 57: One downside to this approach is that if anything breaks while installing packages (which is fairly common due to stuff like insecure packages) then we leave the lockfile in a bad state that doesn't represent reality. This would have happened while updating `nodejs@16`

An alternative approach is to have `local.IsUpToDate` take a lock file instead of reading from disk. This would guarantee not running into this issue again. Similarly, `localLock.Update()` would take a lock file.

Not for this PR, but I think we could improve the local lock file API. Specifically I would make it private and just have the lock struct interact with it. e.g.

`d.lockfile.LocalIsUpToDate()`

And privatize `local.Update` (just have it be called by `lock.Save()` since they are both called together anyway.

This would make all the local stuff private.

mikeland73 (approved):

internal/impl/update.go
ipince: > An alternative approach is to have local.IsUpToDate take a lock file instead of reading from disk. 

Yeah.. I have an implementation of this, but I prefer to get this in and later think about how to refactor the local lock stuff, since it may have larger implications.

PR #1302, opened 2023-07-24
[tests] Fix docs about example tests

ipince:
## Summary
We were abusing the "example test" term to really mean "run `devbox run run_test` on any devbox.json". Updated code and docs to reflect reality, since it was confusing while I was debugging a failed test.

## How was it tested?
`DEVBOX_RUN_DEVBOX_JSON_TESTS=1 go test testscripts [-run TestExamples/flake_remote_run_test`

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -19,7 +19,7 @@ on:
         type: boolean
   workflow_dispatch:
     inputs:
-      run-example-tests:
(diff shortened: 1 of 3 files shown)


ipince:
Note: the failing test is unrelated to this change (due to broken `process-compose` flake)

savil (approved):
ah, yeah, I guess this got generalized when `testscripts/shellenv` was added but the names were never updated properly...

PR #1305, opened 2023-07-25
Update the Yarn overlay example and make sure it works

Lagoja:
## Summary

- Fixes the yarn overlay example so that the nodejs16 version matches the one installed by devbox
- Updates the lockfile

## How was it tested?
`devbox shell`
`yarn versions`

### DIFF
--- a/examples/flakes/overlay/yarn-overlay/flake.nix
+++ b/examples/flakes/overlay/yarn-overlay/flake.nix
@@ -3,7 +3,7 @@
     "This flake outputs a modified version of Yarn that uses NodeJS 16";
 
   inputs = {
-    nixpkgs.url = "nixpkgs/nixos-21.11";
+    nixpkgs.url = "nixpkgs/fc3de6da83863f8f36fdcac1c199c6066a6a0378";
     flake-utils.url = "github:numtide/flake-utils";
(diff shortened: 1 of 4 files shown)


examples/flakes/overlay/yarn-overlay/flake.nix
```diff
@@ -3,7 +3,7 @@
     "This flake outputs a modified version of Yarn that uses NodeJS 16";
 
   inputs = {
-    nixpkgs.url = "nixpkgs/nixos-21.11";
+    nixpkgs.url = "nixpkgs/fc3de6da83863f8f36fdcac1c199c6066a6a0378";
```
mikeland73, line 6: This is a bit fragile, but not sure there another way beyond hardcoding the hash in devbox.json

mikeland73 (approved):

examples/flakes/overlay/yarn-overlay/flake.nix
Lagoja: They'll diverge if we delete or update the lockfile, but I'm not sure there's a better way to do it. I think if we had something like devbox.nix we could use the project details in the user flakes

PR #1307, opened 2023-07-25
[easy][minor fix] add error object to error message's format string inputs

savil:
## Summary

*Motivation*
Seeing this output:
```
error parsing "nix profile list" output: %!w(string=Index:              0): %!s(MISSING)
go.jetpack.io/devbox/internal/nix/nixprofile.parseNixProfileListItem
	/home/runner/work/devbox/devbox/internal/nix/nixprofile/profile.go:138
```
in failing CICD: https://github.com/jetpack-io/devbox/actions/runs/5658605381/job/15330281379

This fixes the error message, not the underlying cause

*Fix*
%w format string in golang is for errors. Docs: https://pkg.go.dev/fmt#Errorf

PR #1308, opened 2023-07-25
[nix profile list] parse using --json in new nix version

savil:
## Summary

Problem:
Latest nix version of 2.17 changes the output format of `nix profile list` to be:
```
Index:              0
Flake attribute:    legacyPackages.aarch64-linux.curl
Original flake URL: github:NixOS/nixpkgs/3c614fbc76fc152f3e1bc4b2263da6d90adf80fb
Locked flake URL:   github:NixOS/nixpkgs/3c614fbc76fc152f3e1bc4b2263da6d90adf80fb
Store paths:        /nix/store/3qcvhxxmgcdj6izfs8d3m8csdlsj92ng-curl-8.1.1-bin /nix/store/6skgivna073ziw95xvbksvd5c8n5vi5w-curl-8.1.1-man
```
Previously, the "keys" didn't exist in the output. This broke how we were parsing the output.

Fix in pseudo code:
```
try invoke: nix profile list --json
if error:
  do old code: nix profile list
```

the older nix versions don't support --json

## How was it tested?

in some devbox projects:
```
# clear old state
> rm -rf .devbox

> devbox shell

# run some commands specific to that project
```

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -144,6 +144,24 @@ func System() (string, error) {
 	return cachedSystem, nil
 }
 
+// versionRegex parses the output of `nix --version` to capture the semver version string
+// example output: nix (Nix) 2.13.3
+var versionRegex = regexp.MustCompile(`nix\s\(Nix\)\s([0-9]+.[0-9]+.[0-9]+)`)
+
+func Version() (string, error) {
+	cmd := command("--version")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	matches := versionRegex.FindStringSubmatch(string(out))
+	if matches == nil || len(matches) < 2 {
+		return "", errors.New("failed to parse nix version")
+	}
+	return matches[1], nil
+}
+
 // Warning: be careful using the bins in default/bin, they won't always match bins
 // produced by the flakes.nix. Use devbox.NixBins() instead.
 func ProfileBinPath(projectDir string) string {
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -144,22 +144,19 @@ func System() (string, error) {
 	return cachedSystem, nil
 }
 
-// versionRegex parses the output of `nix --version` to capture the semver version string
-// example output: nix (Nix) 2.13.3
-var versionRegex = regexp.MustCompile(`nix\s\(Nix\)\s([0-9]+.[0-9]+.[0-9]+)`)
-
 func Version() (string, error) {
 	cmd := command("--version")
 	out, err := cmd.Output()
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
+	// example output: nix (Nix) 2.13.3
 
-	matches := versionRegex.FindStringSubmatch(string(out))
-	if matches == nil || len(matches) < 2 {
+	if !strings.HasPrefix(string(out), "nix (Nix)") {
 		return "", errors.New("failed to parse nix version")
 	}
-	return matches[1], nil
+
+	return strings.TrimSpace(strings.TrimPrefix(string(out), "nix (Nix)")), nil
 }
 
 // Warning: be careful using the bins in default/bin, they won't always match bins
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -144,21 +144,6 @@ func System() (string, error) {
 	return cachedSystem, nil
 }
 
-func Version() (string, error) {
-	cmd := command("--version")
-	out, err := cmd.Output()
-	if err != nil {
-		return "", errors.WithStack(err)
-	}
-	// example output: nix (Nix) 2.13.3
-
-	if !strings.HasPrefix(string(out), "nix (Nix)") {
-		return "", errors.New("failed to parse nix version")
-	}
-
-	return strings.TrimSpace(strings.TrimPrefix(string(out), "nix (Nix)")), nil
-}
-
 // Warning: be careful using the bins in default/bin, they won't always match bins
 // produced by the flakes.nix. Use devbox.NixBins() instead.
 func ProfileBinPath(projectDir string) string {
--- a/internal/nix/nixprofile/profile.go
+++ b/internal/nix/nixprofile/profile.go
@@ -42,7 +42,6 @@ func ProfileListItems(
 	}
 
 	if useJSON {
-		fmt.Printf("nix profile list json output:\n%v\n", output)
 		type ProfileListElement struct {
 			Active      bool     `json:"active"`
 			AttrPath    string   `json:"attrPath"`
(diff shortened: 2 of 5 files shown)


savil:
* main
  * **PR #1307** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1307" target="_blank">Graphite</a> 
    * **PR #1308** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1308" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1308?utm_source=stack-comment).

internal/nix/nix.go
```diff
@@ -144,6 +144,24 @@ func System() (string, error) {
 	return cachedSystem, nil
 }
 
+// versionRegex parses the output of `nix --version` to capture the semver version string
+// example output: nix (Nix) 2.13.3
+var versionRegex = regexp.MustCompile(`nix\s\(Nix\)\s([0-9]+.[0-9]+.[0-9]+)`)
```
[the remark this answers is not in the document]
mikeland73: And then you can use `strings.TrimPrefix("nix (Nix)")` to make this code simpler.

internal/nix/nixprofile/profile.go
```diff
@@ -26,18 +28,62 @@ func ProfileListItems(
 	profileDir string,
 ) (map[string]*NixProfileListItem, error) {
 
-	lines, err := nix.ProfileList(writer, profileDir)
+	version, err := nix.Version()
```
mikeland73, line 31: Don't return error here. That puts us at a risk of failing even when we would do correct thing.

internal/nix/nixprofile/profile.go
```diff
@@ -26,18 +28,62 @@ func ProfileListItems(
...
 	if err != nil {
 		return nil, err
 	}
 
+	// if version is >= 2.17, we can use the json output
+	useJSON := vercheck.SemverCompare(version, "2.17") >= 0
```
mikeland73, line 37: Use `2.17.0` (2.17 is technically not valid semver) 

This defaults to false if version is not semver. Is that OK?  (I think probably yes)

internal/nix/nixprofile/profile.go
mikeland73: (e.g. what if in some past version `nix --version` returned something weird, we're better off just trying with `useJSON = false`

internal/nix/nix.go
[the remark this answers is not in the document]
savil: yeah, I'll switch to TrimPrefix

internal/nix/nixprofile/profile.go
savil: not sure I follow. What do you mean "failing even when we do correct thing?"

In general, I'd prefer erroring here if we fail to get the nix version properly. Otherwise, we'll error below with a more obscure error...

internal/nix/nixprofile/profile.go
savil: ugh... good point

let me check the behavior of this `vercheck.SemverCompare` for `2.17` like output.

btw, do we know if `nix --version` prints `2.17` or `2.17.0`?

internal/nix/nixprofile/profile.go
```diff
@@ -26,11 +28,52 @@ func ProfileListItems(
...
+		if err := json.Unmarshal([]byte(output), &structOutput); err != nil {
+			return nil, err
+		}
+
+		result := map[string]*NixProfileListItem{}
+		for index, element := range structOutput.Elements {
+			result[element.OriginalURL] = &NixProfileListItem{
+				index:             index,
+				unlockedReference: element.OriginalURL,
+				lockedReference:   element.URL,
```
savil, line 68: @mikeland73 do we think I got this mapping right?

internal/nix/nixprofile/profile.go
mikeland73: If we fail to read the version and ignore the error, we can continue doing what we were doing before. Otherwise we'll have introduced a new error and potentially break old nix installations that were already working.

internal/nix/nixprofile/profile.go
```diff
@@ -26,11 +28,52 @@ func ProfileListItems(
...
 ) (map[string]*NixProfileListItem, error) {
 
-	lines, err := nix.ProfileList(writer, profileDir)
+	version, err := nix.Version()
 	if err != nil {
 		return nil, err
 	}
 
+	// if version is >= 2.17.0, we can use the json output
+	useJSON := vercheck.SemverCompare(version, "2.17.0") >= 0
```
savil, line 37: I am somewhat concerned about this failing in an unexpected way. The underlying function call will return:
```
// An invalid semantic version string is considered less than a valid one.
// All invalid semantic version strings compare equal to each other.
```

According to https://lazamar.co.uk/nix-versions/?channel=nixpkgs-unstable&package=nix, there may be some prerelease versions of nix that have non-standard semver versions.

However, these prerelease versions also appear to be rather old, and not present in newer releases of nix. So, maybe this is okay?

internal/nix/nixprofile/profile.go
gcurtis: I'd just try running `nix profile list --json` and if it fails try `nix profile list` (without the flag). Then you don't need to do the version comparison stuff. Example with old Nix:

```
$ nix profile list --json
error: unrecognised flag '--json'
Try 'nix --help' for more information.
$ nix profile list       
0 flake:nixpkgs#legacyPackages.aarch64-darwin.fuse github:NixOS/nixpkgs/d0f2758381caca8b4fb4a6cac61721cc9de06bd9#legacyPackages.aarch64-darwin.fuse /nix/store/8pqmbsp8zkq8656k5yaravbaxixaaa3c-macfuse-stubs-4.4.1
```

internal/nix/nixprofile/profile.go
mikeland73: 2.17.0

image

`vercheck.SemverCompare` seems to work because `semver.parse` can read versions missing the patch and defaults them to `0`. But I'm not sure that a missing patch number is generally considered semver.

internal/nix/nixprofile/profile.go
mikeland73: I think this is OK. non semver versions will be considered older which defaults to the non --json.

I think as a follow up after this PR we could:

(only use the `--json` flag if we're sure version is higher (e.g. both are semver)) and then try to parse things both ways (try one and if it fails try the other)

gcurtis (approved):
Looks good. If we can get away with not needing to parse the version, then that would be nice.

internal/nix/nixprofile/profile.go
mikeland73: Here's a concrete example:

Imagine that in nix v2.1.0 running `nix --version` returns `2.1.0` without the `nix (Nix)` prefix. Then this function would return an error and we would stop execution. But in this case, `version` would be `""` which would currently set `useJSON` to false. So it's a case where ignoring the error would cause correct execution but returning an error would cause execution to break. 

We don't want to break any old versions of nix, so I think just assuming useJSON is false if we can't parse the version is the correct way to go.

savil:
@gcurtis @mikeland73 PTAL

gcurtis (approved):

internal/nix/nixprofile/profile.go
mikeland73: No, it's not 1 to 1 on either locked or unlocked. I'm taking a look

internal/nix/nixprofile/profile.go
```diff
@@ -26,11 +26,45 @@ func ProfileListItems(
...
+				unlockedReference: element.OriginalURL,
+				lockedReference:   element.URL,
+				nixStorePath:      element.StorePaths[0],
+			}
+		}
+		return result, nil
+	}
+
+	output, err = nix.ProfileList(writer, profileDir, false /*useJSON*/)
+	if err == nil {
```
savil, line 62: ugh, this should be `err != nil`

mikeland73 (approved):

PR #1309, opened 2023-07-25
[update] Add simple unit test for lockfile merge logic

ipince:
## Summary

I'm sorry if this steps a bit on your toes :( . I just figured I wanted some unit test before I made the changes to the merging logic, and noticed that if I just refactored a function a little bit, then we could add some unit tests easily at the function level.

I just added 1 unit test here to demonstrate, and will add more later as I implement the bugfix.

## How was it tested?
Unit tests


savil (approved):
Thanks! This looks great, and nope, this didn't step on my toes at all. Thanks for adding it!

I'll focus on adding testscript tests which are more end-to-end than these golang unit-tests (to minimize conflicts)

PR #1263, opened 2023-07-09
[nix] enable `auto-allocate-uids` experimental flag

tsirysndr:
Screenshot 2023-07-09 4 27 28 PM

## Summary

In this Pull Request, I enabled the experimental flag `auto-allocate-uids` to address an issue related to the installation of Nix packages. The error occurred when executing the command `/nix/var/nix/profiles/default/bin/nix print-dev-env /root/.local/share/devbox/global/default/.devbox/gen/flake/flake.nix --extra-experimental-features ca-derivations --option experimental-features nix-command flakes --json, which resulted in an exit status of 1.`

To resolve this issue, I modified the `ExperimentalFlags` function by adding the `"auto-allocate-uids"` option to the list of experimental flags. 

This change is specifically intended for a Linux environment with Nix version 2.15, installed using the instructions provided at https://github.com/DeterminateSystems/nix-installer

## How was it tested?
You can retry the package installation by running the command `devbox global add gh`.

PR #1311, opened 2023-07-25
nix: don't pass --experimental-features to nix commands

gcurtis:
When running nix commands, devbox passes two flags to enable experimental features:

1. `nix --extra-experimental-features ...`
2. `nix --option experimental-features ...`

The second flag causes issues because it overwrites the user's nix.conf. It doesn't seem to be doing anything, so remove it and just use `--extra-experimental-features`. There's also some cleanup to switch all callers of `nix.ExperimentalFlags` to use `nix.Command` instead. This function sets the correct flags and forwards SIGINTs to Nix instead of just killing the process when devbox gets interrupted.

Details
-------

A specific bug caused by this flag is https://github.com/jetpack-io/devbox/pull/1263. The DeterminateSystems installer configures Nix without a build users group and instead enables the `auto-allocate-uids` setting in `/etc/nix/nix.conf` on Linux. If this setting gets disabled by devbox, then the Nix daemon has no way of building flakes and fails.

The reason why `--option experimental-features ...` was added is because we were still seeing errors about ca-derivations not being enabled even though the `--extra-experimental-features ca-derivations` flag was set. Overwriting the config setting seemed to fix it, but that might've been a red herring.

The ca-derivations feature is unique in that it must be set in the Nix daemon, not just the client. This is because it changes the schema of the Nix database, requiring a restart of the daemon. The only reliable way to enable it is to add it to `/etc/nix/nix.conf` and restart the nix-daemon process.

Given that this flag is currently breaking some users and doesn't seem to be having the intended effect, I think we should remove it and figure out another way to ensure ca-derivations is enabled. Switching to using the DeterminateSystems installer and printing errors telling the user how to enable it are probably the best bet.

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -31,6 +31,12 @@ permissions:
   contents: read
   pull-requests: read
 
+defaults:
+  run:
+    # Explicitly setting the shell to bash runs commands with
+    # `bash --noprofile --norc -eo pipefail` instead of `bash -e`.
+    shell: bash
+
 env:
   HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}"
   HOMEBREW_NO_ANALYTICS: 1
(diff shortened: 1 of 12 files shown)


savil:
> Switching to using the DeterminateSystems installer and printing errors telling the user how to enable it are probably the best bet.

yup, lets do this!

savil:
The code looks good but there's quite a few failing cicd tests. I tried looking into "why" but am failing.

gcurtis:
The macOS auto-install test is still failing, so I'm going to postpone this until next cycle. I think this PR will require a couple other changes:

- Make devbox check that `experimental-features = ca-derivations ...` is set in `nix show-config`.
- If it isn't, add it (will require sudo) and restart the daemon (if it's a multi-user install).

I'm really not sure why `-o experimental-features ...` made things work. I'm wondering if it's a bug in the CLI. Either way, we'll need to remove it so we don't break with Nix Linux installs.

PR #1313, opened 2023-07-25
[tests] Add test for old nix versions

mikeland73:
## Summary

Since we saw CLI format change (See https://github.com/jetpack-io/devbox/pull/1308) we should test against many nix versions.

Possible followup: In `main` should we run all nix versions?

## How was it tested?
CICD


.github/workflows/cli-tests.yaml
```diff
@@ -139,3 +139,27 @@ jobs:
           echo "access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}" > ~/.config/nix/nix.conf
 
           devbox run echo "Installing packages..."
+  
+  old-nix-version:
```
savil, line 143: lets call it `install-old-nix-version`

savil (approved):

PR #1314, opened 2023-07-26
[lock] Refactor local lock

mikeland73:
## Summary

Follow up to https://github.com/jetpack-io/devbox/pull/1291.

This simplified the lockfile interface and mostly removes "local" lock from public view.

It addresses this comment: https://github.com/jetpack-io/devbox/pull/1291#discussion_r1270102373

## How was it tested?

`devbox update` on a package that needs system info downloaded.

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -10,12 +10,7 @@
     "golangci-lint@1.52.2": {
       "last_modified": "2023-05-01T16:53:22Z",
       "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#golangci-lint",
-      "version": "1.52.2",
-      "systems": {
-        "aarch64-darwin": {
-          "store_path": "/nix/store/rp380cz8gxv8pw5zq985yblym19vr9g8-golangci-lint-1.52.2",
-          "ca_store_path": "/nix/store/2lm73605ppky81a387rwxi1clzqd5133-golangci-lint-1.52.2"
-        }
+      "version": "1.52.2"
       }
     }
   }
--- a/devbox.lock
+++ b/devbox.lock
@@ -11,7 +11,6 @@
       "last_modified": "2023-05-01T16:53:22Z",
       "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#golangci-lint",
       "version": "1.52.2"
-      }
     }
   }
 }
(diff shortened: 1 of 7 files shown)


internal/lock/lockfile.go
```diff
@@ -137,8 +140,34 @@ func IsLegacyPackage(pkg string) bool {
...
+func (f *File) Tidy() {
+	f.Packages = lo.PickByKeys(f.Packages, f.devboxProject.Packages())
+}
+
+// IsUpToDateAndInstalled returns true if the lockfile is up to date and the
+// local hashes match, which generally indicates all packages are correctly
+// installed and print-dev-env has been computed and cached.
+func (f *File) IsUpToDateAndInstalled() (bool, error) {
+	if dirty, err := f.isDirty(); err != nil || dirty {
+		return !dirty, err
```
ipince, line 152: can we separate the `err != nil` and the `err == nil && dirty == true` paths?

Otherwise when we do have an error, we'll return `true, error`, which is fine assuming the caller always checks the error, but feels a bit weird.

devbox.lock
```diff
@@ -2,14 +2,21 @@
...
+      "resolved": "github:NixOS/nixpkgs/3c614fbc76fc152f3e1bc4b2263da6d90adf80fb#go",
+      "source": "devbox-search",
+      "version": "1.20.5"
     },
     "golangci-lint@1.52.2": {
       "last_modified": "2023-05-01T16:53:22Z",
       "resolved": "github:NixOS/nixpkgs/8670e496ffd093b60e74e7fa53526aa5920d09eb#golangci-lint",
-      "version": "1.52.2"
+      "version": "1.52.2",
+      "systems": {
```
ipince, line 14: remove these? i assume you were testing with the feature enabled?

ipince (approved):
LGTM!

internal/lock/local.go
```diff
@@ -34,16 +34,24 @@ func (l *localLockFile) equals(other *localLockFile) bool {
...
+	if err != nil {
+		return false, err
+	}
+	newLock, err := forProject(project)
 	if err != nil {
 		return false, err
 	}
 
-	return l.equals(newLock), nil
+	return filesystemLocal.equals(newLock), nil
```
savil, line 47: this is confusing to read. Could we call them either:
`filesystemLock.equals(newLock)`
or
`filesystemLocal.equals(newLocal)`

since they are conceptually the same kind of object

PR #1315, opened 2023-07-26
[bugs]: add other sysinfos, but respect existing ca paths

ipince:
## Summary
This should fix the two remaining bugs in the lockfile merging logic:

1. If we have new system infos for a system that is not our own, add them anyway.
2. If we have a system info whose StorePath does not match the existing info's StorePath, then replace the info. This is to ensure correctness--all infos should come from the same resolved package version.

## How was it tested?
Added unit tests!


savil (approved):
awesome!

PR #1316, opened 2023-07-26
ci: upgrade nix install action, use it in nix-test-versions

gcurtis:
Changes extracted from #1311 so they can be merged sooner.

- Upgrade the Nix installer action to v4 and configure the experimental features Devbox needs.
- Print out the Nix version, `/etc/nix/nix.conf`, and the resolved Nix config to make debugging easier.
- Do the same for `test-nix-versions` and add the last 3 Nix releases to the matrix. Nix 2.15.1 is the current stable version and 2.17.0 is the latest.
- Update `TestContentAddressedPath` to accept hashes from Nix <2.17 and 2.17+.

### DIFF
--- a/internal/nix/store_test.go
+++ b/internal/nix/store_test.go
@@ -2,18 +2,23 @@ package nix
 
 import (
 	"fmt"
+	"strings"
 	"testing"
+
+	"golang.org/x/exp/slices"
 )
 
 func TestContentAddressedPath(t *testing.T) {
-
 	testCases := []struct {
 		storePath string
-		expected  string
+		expected  []string
 	}{
 		{
 			"/nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1",
-			"/nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1",
+			[]string{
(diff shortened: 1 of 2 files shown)


internal/nix/store_test.go
```diff
@@ -2,18 +2,23 @@ package nix
...
 		storePath string
-		expected  string
+		expected  []string
 	}{
 		{
 			"/nix/store/r2jd6ygnmirm2g803mksqqjm4y39yi6i-git-2.33.1",
-			"/nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1",
+			[]string{
+				"/nix/store/ldbhlwhh39wha58rm61bkiiwm6j7211j-git-2.33.1",
+				"/nix/store/d49wyvsz5nkqa23qp4p0ikr04mw9n4h9-git-2.33.1",
```
savil, line 20: a comment here explaining that one is for content-addressed-paths before nix 2.17, and the other one for after 2.17 would be helpful context

savil (approved):

mikeland73:
Should we move these to main only? They are a bit flaky.

gcurtis:
@mikeland73 I'm okay with that, but this wasn't adding any new tests. It just runs the existing ones on a couple more Nix versions. I could take a look at the flakey jobs.

PR #1318, opened 2023-07-27
[rm nixpkgs] make HEAD request to BinaryCache to ensure binary is cached

savil:
## Summary

This PR makes a HTTP HEAD request to `https://cache.nixos.org/<hash>.narinfo` to verify that the binary is indeed cached.

If not, we fallback to the legacy path that is dependent on downloading nixpkgs.

**Implementation note**
In `IsInBinaryCache`, I wanted to insert the HEAD request. However, this function can be called in a loop for all packages, which is bad for perf.

Instead, in this design, a new function `FillNarInfoCache(packageList)` makes the HEAD requests concurrently, storing results in an in-memory cache. 

We _require_ that callers invoke `FillNarInfoCache(packageList)` prior to invoking `IsInBinaryCache`. Failing this, `IsInBinaryCache` will return an error. Not thrilled with this design, since it may lead to inadvertent errors, but this code is pretty core and our test coverage should expose any codepaths that call `IsInBinaryCache` without `FillNarInfoCache`.

## How was it tested?

I could do `devbox shell` in a project that has `hello` package. I inserted some print statements to verify that the HEAD 
response was 200 and that the caching for the boolean value of isNarInfoInCache was working.


internal/devpkg/package.go
```diff
@@ -546,3 +580,30 @@ func (p *Package) StoreName() (string, error) {
...
+// /nix/store/<hash>-<name>-<version>
+//
+// This is a helper struct for analyzing the string representation
+type storePathParts struct {
+	hash    string
+	name    string
+	version string
+}
+
+func newStorePathParts(path string) *storePathParts {
```
gcurtis, line 594: Some names have multiple dashes. You can copy https://github.com/jetpack-io/axiom/blob/main/devbox-cloud/search/internal/nixpkgs/eval.go#L584 which tries to mimic the exact Nix behavior.

internal/devpkg/package.go
```diff
@@ -462,8 +470,34 @@ func (p *Package) IsInBinaryCache() (bool, error) {
...
-	_, ok := entry.Systems[userSystem]
-	return ok, nil
+	sysInfo, ok := entry.Systems[userSystem]
+	if !ok {
+		return false, nil
+	}
+
+	// Check if the narinfo is present in the binary cache
+	if exists, ok := isNarInfoInCache[p.Raw]; ok {
+		fmt.Printf("narInfo cache hit: %v\n", exists)
```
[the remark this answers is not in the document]
savil: oh I'll be deleting the Printfs prior to landing! Just have them there to ensure this code works as I want. (this entire section is also gated by the feature-flag)

internal/devpkg/package.go
savil: Good call! will switch

internal/devpkg/package.go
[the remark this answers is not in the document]
savil: I'll remove this after I test the final version of this PR on devbox-cloud

internal/devpkg/package.go
```diff
@@ -462,8 +470,34 @@ func (p *Package) IsInBinaryCache() (bool, error) {
...
+		return exists, nil
+	}
+
+	httpClient := &http.Client{
+		Timeout: time.Second * 5,
+	}
+	pathParts := newStorePathParts(sysInfo.StorePath)
+	reqURL := BinaryCache + "/" + pathParts.hash + ".narinfo"
+	res, err := httpClient.Head(reqURL)
```
mikeland73, line 490: Adding an HTTP request in `IsInBinaryCache` will likely introduce N+1s. I generally try to avoid network requests in struct methods for this reason. 

In practice this means we will make 1 sequential network call per package any time local.lock is out of date. A few alternatives:

* Only check on failure. If nix profile install or print-dev-env fail, then check.
* Make http calls in parallel. This may break if cache has any sort of rate limit (though if it's just S3 I doubt it).
* If cache is never evicted, we can check on installation instead. If it's missing from cache, mark as missing in lockfile.

mikeland73 (commented):
I'm afraid this approach will introduce a perf regression which scales O(n) with number of packages. Thoughts on alternatives in comment?

internal/devpkg/package.go
savil: Speaking with @gcurtis it appears there's some likelihood that the nix maintainers start evicting old packages from the cache.

So, given that, option 2 seems the better path to pursue. A rate-limit doesn't seem to be a big concern: nix itself does a lot of requests to the cache, and when populating the search index @gcurtis reports doing many requests to the cache in parallel without getting rate-limited.

option 1 of checking upon failure I am ruling out because it seems quite hard to implement.

savil:
needs revision so marking as Draft.

savil:
* main
  * **PR #1318** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1318" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1318?utm_source=stack-comment).

internal/devpkg/package.go
```diff
@@ -556,3 +655,57 @@ func (p *Package) EnsureUninstallableIsInLockfile() error {
...
+	for i, r := range name {
+		if dashIndex != 0 && !unicode.IsLetter(r) {
+			return &storePathParts{hash: hash, name: name[:dashIndex], version: name[i:]}
+		}
+		dashIndex = 0
+		if r == '-' {
+			dashIndex = i
+		}
+	}
+	return &storePathParts{hash, name, "" /*version*/}
```
gcurtis, line 688: `storePathParts{Version: ""}` instead of a comment.

internal/devpkg/package.go
```diff
@@ -556,3 +655,57 @@ func (p *Package) EnsureUninstallableIsInLockfile() error {
...
+}
+
+func (p *storePathParts) Equal(other *storePathParts) bool {
```
gcurtis, line 691: If you use `storePathParts` as a value then you can just compare with `==`.

internal/devpkg/package.go
```diff
@@ -556,3 +655,57 @@ func (p *Package) EnsureUninstallableIsInLockfile() error {
...
+}
+
+// FillNarInfoCache checks the remote binary cache for the narinfo of each
+// package in the list, and caches the result.
+// Callers of IsInBinaryCache must call this function first.
+func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
+	group, _ := errgroup.WithContext(ctx)
+	for _, p := range packages {
+		// If the package's NarInfo status is already known, skip it
+		if _, ok := isNarInfoInCache[p.Raw]; ok {
```
gcurtis, line 702:
```suggestion
		if isNarInfoInCache[p.Raw] {
```

gcurtis (approved):
> We require that callers invoke FillNarInfoCache(packageList) prior to invoking IsInBinaryCache. Failing this, IsInBinaryCache will return an error. Not thrilled with this design, since it may lead to inadvertent errors, but this code is pretty core and our test coverage should expose any codepaths that call IsInBinaryCache without FillNarInfoCache.

If you're looking for an alternate approach, each package could have a `sync.Once` that makes the HEAD request. The `PackageFromString(s)` functions would call `go pkg.checkCache.Do(...)` to kick off the requests in the background. The `Package.IsInBinaryCache` method would use the same once to block until the network request has finished.

internal/devpkg/package.go
```diff
@@ -556,3 +655,57 @@ func (p *Package) EnsureUninstallableIsInLockfile() error {
...
+			continue
+		}
+		pkg := p
+		group.Go(func() error {
+			return pkg.fillNarInfoCache()
```
gcurtis, line 707: I don't think `pkg.fillNarInfoCache` is safe to run concurrently. The map, Nix system/version, etc. all need locking.

gcurtis (changes requested):
I think (correct me if I'm wrong) that `fillNarInfoCache` is being run concurrently for each package, but it calls a bunch of stuff that isn't safe (writing to `isNarInfoInCache`, `nix.Version`, etc.).

internal/devpkg/package.go
savil: oh good catch! I can ensure the `nix.System` value is cached prior to executing this, but may need to add some locking for the map.

savil (commented):
@gcurtis updated the code without doing the `sync.Once`.

> If you're looking for an alternate approach, each package could have a sync.Once that makes the HEAD request. The PackageFromString(s) functions would call go pkg.checkCache.Do(...) to kick off the requests in the background. The Package.IsInBinaryCache method would use the same once to block until the network request has finished.

Ooh I like this approach. It would simplify the code, and ensure it works by default: The current pattern of "ensure you call X before calling Y method" is really ugly :(

The reason I didn't directly kick off a HEAD request in one of the package constructors is:
1. Unfortunately, we recreate the Package structs for the same packages in different places. The structs are not reused. This is why I have a `devpkg` variable `isNarInfoInCache`, and not a struct field.
2. We may create a Package struct which is not used for `IsBinaryInCache`, and so we'd do wasted work.

One idea is to modify your idea from the PackageFromString constructor to the IsInBinaryCache where we read the result (the modification is: lazily initialize it):
1. Inside `pkg.IsInBinaryCache()`, we can do `go sync.Once(pkg.fillNarInfoCache)`. And call that `sync.Once(pkg.fillNarInfoCache)` again to block for the return value. 
 - If its the first time, it'll kick off a goroutine in the background.
 
~Let me try in a stacked PR.... so as not to block this PR but also try to improve it.~ EDIT: nope, this doesn't work: it'll be block other packages.

savil:
Hmm, I need to make some fixes...moving to draft.

savil:
@gcurtis PTAL

internal/devpkg/package.go
```diff
@@ -462,37 +499,113 @@ func (p *Package) HashFromNixPkgsURL() string {
...
+// fillNarInfoCache fills the cache value for the narinfo of this package,
+// if it is eligible for the binary cache.
+func (p *Package) fillNarInfoCache() error {
+	if eligible, err := p.isEligibleForBinaryCache(); err != nil {
+		return err
+	} else if !eligible {
+		return nil
+	}
+
+	sysInfo, err := p.sysInfoIfExists()
```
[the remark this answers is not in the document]
savil: Gah, you are right.

internal/devpkg/package.go
```diff
@@ -462,37 +499,113 @@ func (p *Package) HashFromNixPkgsURL() string {
...
+				" the package is eligible for binary cache",
+		)
 	}
 
-	// enable for nix >= 2.17
-	return vercheck.SemverCompare(version, "2.17.0") >= 0, nil
+	pathParts := newStorePathParts(sysInfo.StorePath)
+	reqURL := BinaryCache + "/" + pathParts.hash + ".narinfo"
+	//nolint:govet
+	ctx, _ := context.WithDeadline(context.Background(), time.Now().Add(5*time.Second))
```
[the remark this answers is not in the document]
savil: will fix

internal/devpkg/package.go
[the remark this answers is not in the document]
savil: Thought: Instead of sprinkling locks everywhere, I could ensure these are invoked for all packages prior to starting the go-routines. Then within the goroutines, they'll be in read-only mode. Would that work?

savil:
Gah, I really dislike the implementation in this PR. It makes the code brittle and harder to reason about. Inclined to try other approaches.

internal/devpkg/package.go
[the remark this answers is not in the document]
gcurtis: I think that would work. It might be easier to just split apart the HTTP request from everything else. Then you can calculate the store hash and check the map synchronously.

mikeland73 (commented):
@savil this is great! 

I wonder if we can split out much of the nar cache stuff in devpkg into own file to keep it a bit more separate.

internal/devpkg/package.go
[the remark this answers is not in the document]
savil: nix.Version is cached already

PR #1324, opened 2023-07-27
[Bug fix] Ensure bin-wrappers use latest devbox binary to prevent false update notifications

savil:
## Summary

**Motivation**

In #1260, the bin-wrappers were changed to invoke the devbox-binary, instead of the devbox-launcher. However, when devbox has been updated, these bin-wrappers have the older devbox-binary's path hardcoded. So, they call the older devbox-binary leading to two issues:
1. runs older functionality of the older devbox binary
2. re-prints the notice about "New devbox available. Run .... to update", despite the user already having updated their devbox.

**Implementation**

1. Create a symlink that always points to the "current" devbox binary. 
    - I use "current" instead of "latest" to be consistent with the Launcher's usage of these terms.
2. In the bin-wrappers, we prepend the directory containing this devbox-symlink to PATH. The bin-wrappers revert back to directly invoking devbox, as in `devbox shellenv`.
    - This is done for robustness: if the symlink is missing, then the bin-wrappers do still invoke `devbox` via the launcher. This is problematic for users who have installed `coreutils` via devbox, but for most users this fallback would work.
3. We ensure the bin-wrappers are created in `ensurePackagesAreInstalled` that runs in `devbox.PrintEnv` during `devbox shellenv`. 
    - This is needed because we need existing users that have the bin-wrappers with the hardcoded DevboxExecutablePath to refresh their bin-wrappers. When they open a new terminal, `devbox global shellenv` would run, refreshing these bin-wrappers.
4. Finally, we create the Devbox symlink during `devbox version update`, so that it points to the _new_ Devbox binary.
     

**Drawbacks**
Switching between devbox versions (for testing and development) may lead to subtle unexpected behavior: The devbox-symlink will point to the version that created the bin-wrappers, rather than respecting the devbox of the PATH. We need to ensure we run `devbox install` or similar to re-create the bin-wrappers.

## How was it tested?

1. Sanity check to do `devbox shell` in `examples/development/go/hello-world`. 
2. Sanity check to do `devbox run build` using the devbox binary from this PR.
3. Deleted devbox-symlink, and ran `devbox version -v` to ensure that it was re-created.

### DIFF
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -82,14 +82,14 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 //
 //     So, the bin-wrappers need to directly invoke the devbox binary.
 //
-//  2. The devbox binary's path will change when devbox is updated. This means
-//     that using absolute paths to the devbox binaries in the bin-wrappers
+//  2. The devbox binary's path will change when devbox is updated. Hence
+//     using absolute paths to the devbox binaries in the bin-wrappers
 //     will result in bin-wrappers invoking older devbox binaries.
 //
 //     So, the bin-wrappers need to use a symlink to the latest devbox binary. This
 //     symlink is updated when devbox is updated.
 func CreateDevboxSymlink() (string, error) {
-	curDir := xdg.CacheSubpath(filepath.Join("devbox", "current"))
+	curDir := xdg.CacheSubpath(filepath.Join("devbox", "bin", "current"))
 	if err := fileutil.EnsureDirExists(curDir, 0755, false /*chmod*/); err != nil {
 		return "", err
 	}
@@ -105,9 +105,11 @@ func CreateDevboxSymlink() (string, error) {
 		return "", errors.WithStack(err)
 	}
 
-	if err := os.Symlink(devboxBinaryPath, currentDevboxSymlinkPath); err != nil {
+	// Don't return error if error is os.ErrExist to protect against race conditions
+	if err := os.Symlink(devboxBinaryPath, currentDevboxSymlinkPath); err != nil && !errors.Is(err, os.ErrExist) {
 		return "", errors.WithStack(err)
 	}
+
 	return currentDevboxSymlinkPath, nil
 }
 
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -14,8 +14,10 @@ import (
 
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/cmdutil"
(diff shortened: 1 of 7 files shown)


savil:
* main
  * **PR #1324** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1324" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1324?utm_source=stack-comment).

internal/wrapnix/wrapper.go
```diff
@@ -72,12 +72,50 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
...
+//     So, the bin-wrappers need to directly invoke the devbox binary.
+//
+//  2. The devbox binary's path will change when devbox is updated. This means
+//     that bin-wrappers may invoke older devbox binaries, which have different
+//     functionality. Annoyingly, they will also print "New devbox available" messages,
+//     even though the user has just updated devbox.
+//     So, the bin-wrappers need to use a symlink to the latest devbox binary. This
+//     symlink is updated when devbox is updated.
+func CreateDevboxSymlink() (string, error) {
+	curDir := xdg.CacheSubpath(filepath.Join("devbox", "current"))
```
mikeland73, line 92: thoughts on making this `devbox/bin/current` instead? that way we keep it with all the other bins.

internal/wrapnix/wrapper.go
```diff
@@ -72,12 +72,51 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
...
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	// We will always re-create this symlink to ensure correctness.
+	if err := os.Remove(currentDevboxSymlinkPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return "", errors.WithStack(err)
+	}
+
+	if err := os.Symlink(devboxBinaryPath, currentDevboxSymlinkPath); err != nil {
```
mikeland73, line 108: Don't return error if error is `os.ErrExist` (to prevent against race conditions)

mikeland73 (approved):
perfect 👍

internal/wrapnix/wrapper.go
savil: yeah, I like that. Will update.

internal/wrapnix/wrapper.go
savil: good call.

savil:
I have a small commit locally to fix the code review feedback, but its not showing up on Github due to their incident:
https://www.githubstatus.com/incidents/xfmyxxhcknvj

Issue #1122, opened 2023-06-08
[Bug]: Dev Containers - Python virtual environment not activated when `postCreateCommand` runs


apgrucza:
**Current Behavior (bug)**

The Python virtual environment is not activated by the time that VS Code runs the `postCreateCommand` action. This is because the Python virtual environment is created on demand, rather than during the Docker build phase.

To reproduce:

1. Run `devbox create --template python-pip`
2. Run `cd python-pip`
3. Run `devbox generate devcontainer`
4. Add the following to `.devcontainer/devcontainer.json`: `"postCreateCommand": "which python"`
5. Run `code .`
6. In the popup window, click **Reopen in Container**

When the **Configuring...** terminal window appears, the path shown is not to the Python virtual environment:
`/code/.devbox/virtenv/.wrappers/bin/python`

**Expected Behavior (fix)**

The terminal window should show this path:
`/home/devbox/.local/state/devbox/v-933c2/python310Packages.pip/.venv/bin/python`

Currently `devbox install` does not create the Python virtual environment, but `devbox run` does. So, a workaround is to modify the Dockerfile by changing
```dockerfile
RUN devbox install
```
to
````dockerfile
RUN devbox run -- echo "Installing packages"
````

A proper solution would be to change `devbox install` to (optionally?) do the extra initialisation required, or to provide a separate `devbox` command to do this.

**Additional context**
Output of `devbox version -v`:
```
Version:     0.5.4
Platform:    linux_amd64
Commit:      92bb293af46fe6869648c20e4a6476dc83df0092
Commit Time: 2023-06-01T21:35:06Z
Go Version:  go1.20.4
Launcher:    0.2.0
```

mohsenari:
Apologies for late response on this and thanks for reporing it. This is an interesting case. I think the fix can be as simple as changing what happens in the CLI when `devbox install` is called to make sure the python plugin that sets up virtualenv gets called as well. 
But that might defeat the purpose of an "install" command. I'll try reproducing this and will make a PR to fix the use case.

mohsenari:
with the release of 0.5.10 I think this fix is in, and the template is changes to `devbox run -- "installed package"` instead of `devbox install`

PR #1325, opened 2023-07-27
[devcontainer] Changed to single user nix in devcontainer

mohsenari:
## Summary
Multi user nix installation with creating a devbox user in dockerfile causes an issue in devcontainer with Nix due to permission issues with subdirectories nix creates in `/tmp/`. Changing to a single user installation and handling everything under root user in docker seems to fix the issue.

## How was it tested?
- `devbox init && devbox add hello`
- `devbox generate devcontainer`
- open vscode: `code .`
- in vscode command palette (cmd + shift + p) run `Dev Containers: Rebuild Without Cache and Reopen in Container`
- confirm that devcontainer gets created successfully and vscode switches to remote devcontainer environment.

### DIFF
--- a/internal/impl/generate/tmpl/devcontainerDockerfile.tmpl
+++ b/internal/impl/generate/tmpl/devcontainerDockerfile.tmpl
@@ -4,27 +4,16 @@ FROM debian:stable-slim
 RUN apt-get update
 RUN apt-get -y install bash binutils git{{if .IsDevcontainer}} gnupg2{{- end}} xz-utils wget sudo
 
-# Step 2: Setting up devbox user
-ENV DEVBOX_USER=devbox
-RUN adduser $DEVBOX_USER
-RUN usermod -aG sudo $DEVBOX_USER
-RUN echo "devbox ALL=(ALL:ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/$DEVBOX_USER
-
-USER $DEVBOX_USER
+# Step 2: Installing Nix
+RUN wget --output-document=/dev/stdout https://nixos.org/nix/install | sh -s -- --daemon
+RUN . ~/.nix-profile/etc/profile.d/nix.sh
+ENV PATH="/root/.nix-profile/bin:$PATH"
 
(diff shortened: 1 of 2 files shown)


mikeland73:
cc: @gcurtis do you see any issues doing everything as root single user?

gcurtis:
As long as the root user is also the user that will be using Nix in the container, then I think it's okay.

internal/impl/generate/tmpl/devcontainerDockerfile.tmpl
```diff
@@ -4,37 +4,26 @@ FROM debian:stable-slim
...
-RUN sudo chown $DEVBOX_USER:root /code
 COPY devbox.json devbox.json
 {{if len .LocalFlakeDirs}}
 # Step 6: Copying local flakes directories
 {{- end}}
 {{range $i, $element := .LocalFlakeDirs -}}
 COPY {{$element}} {{$element}}
 {{end}}
-RUN devbox install
+RUN devbox run -- echo "Installed Packages."
```
LucilleH, line 24: Can this go back to `RUN devbox install`?

LucilleH (approved):

internal/impl/generate/tmpl/devcontainerDockerfile.tmpl
mohsenari: that change is because of this bug which I've confirmed doesn't occur when using `devbox run` 
#1122

mohsenari:
> As long as the root user is also the user that will be using Nix in the container, then I think it's okay.

yes that is indeed the case. Everything the container does, it does under the root user.

apgrucza:
@mohsenari A side effect of this change is that any file created within the bind-mounted workspace will be owned by `root` on the host. If the user then begins to work outside the dev container, they will get permission errors when writing to this file. For this reason, I believe it's usual to not use `root` in dev containers.

Is there another way to resolve the permission issues with the subdirectories that Nix creates in `/tmp/` without resorting to using `root`? Are you able to elaborate on the problem encountered?

PR #1330, opened 2023-07-28
[bin wrappers] Fixes: don't export PATH, and eval the symlink of the binary

savil:
## Summary

This PR tries to address some concerns with #1324.

1. We drop `export` from `PATH` in the bin-wrappers. This would have modified
PATH for all child programs, which we need not do. 

2. When creating the symlink, we ensure the target value is passed through `EvalSymlink`.
Previously, it was set as the result of `os.Executable`, which may still have been
a symlink.

## How was it tested?

did a basic sanity test of opening a devbox shell in a golang project

### DIFF
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -106,16 +106,19 @@ func CreateDevboxSymlink() (string, error) {
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
-	devboxBinaryPath, err := filepath.EvalSymlinks(execPath)
-	if err != nil {
-		// This may return an error due to symlink loops. In that case, we
-		// return an empty string, and the bin-wrapper should handle it accordingly.
-		return "", nil
-	}
+	devboxBinaryPath, evalSymlinkErr := filepath.EvalSymlinks(execPath)
+	// we check the error below, because we always want to remove the symlink
 
 	// We will always re-create this symlink to ensure correctness.
 	if err := os.Remove(currentDevboxSymlinkPath); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return "", errors.WithStack(err)
+	}
+
+	if evalSymlinkErr != nil {
+		// This may return an error due to symlink loops. In that case, we
+		// return an empty string, and the bin-wrapper should handle it accordingly.
+		// nolint:nilerr
+		return "", nil
 	}
 
 	// Don't return error if error is os.ErrExist to protect against race conditions.
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -72,7 +72,8 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 	return createSymlinksForSupportDirs(devbox.ProjectDir())
 }
 
-// CreateDevboxSymlink creates a symlink to the devbox binary.
+// CreateDevboxSymlink creates a symlink to the devbox binary. It may return
+// no error and the symlink path, even if it fails to create the actual symlink.
 //
 // Needed because:
 //
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -55,15 +55,19 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 	if err != nil {
 		return err
 	}
+	devboxSymlinkDir := ""
+	if devboxSymlinkPath != "" { // may be empty if not possible to create symlink
+		devboxSymlinkDir = filepath.Dir(devboxSymlinkPath)
+	}
 
 	for _, bin := range bins {
 		if err = createWrapper(&createWrapperArgs{
-			devboxer:         devbox,
-			BashPath:         bashPath,
-			Command:          bin,
-			ShellEnvHash:     shellEnvHash,
-			DevboxSymlinkDir: filepath.Dir(devboxSymlinkPath),
-			destPath:         filepath.Join(destPath, filepath.Base(bin)),
+			devboxer:                 devbox,
+			BashPath:                 bashPath,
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #1330** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1330" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1330?utm_source=stack-comment).

internal/wrapnix/wrapper.go
```diff
@@ -89,13 +89,19 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
...
 	}
 	currentDevboxSymlinkPath := filepath.Join(curDir, "devbox")
 
-	devboxBinaryPath, err := os.Executable()
+	// Get the path to the devbox binary.
+	execPath, err := os.Executable()
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	devboxBinaryPath, err := filepath.EvalSymlinks(execPath)
```
savil, line 104: let me harden this, so if an error is returned, then we do NOT error here. And then we skip modifying the PATH in the bin-wrappers.

mikeland73, line 104: This will fail if existing symlink is already a loop. I think in that case we should just remove the symlink and not create a new one.

mikeland73: I think if this errors, you can simply delete the symlink and not recreate it. The rest can be the same.

savil: I didn't remove the symlink: i'm a little concerned that if it was previously created and was working fine, then by removing the symlink we may break some bin-wrappers (that are from a different project) have the symlinkDir in their PATH

internal/wrapnix/wrapper.go
```diff
@@ -89,22 +94,33 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
...
 	// We will always re-create this symlink to ensure correctness.
 	if err := os.Remove(currentDevboxSymlinkPath); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return "", errors.WithStack(err)
 	}
 
+	if evalSymlinkErr != nil {
+		// This may return an error due to symlink loops. In that case, we
+		// return an empty string, and the bin-wrapper should handle it accordingly.
+		// nolint:nilerr
+		return "", nil
```
mikeland73, line 121: return symlink dir

internal/wrapnix/wrapper.go
```diff
@@ -89,22 +94,33 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 //     So, the bin-wrappers need to use a symlink to the latest devbox binary. This
 //     symlink is updated when devbox is updated.
 func CreateDevboxSymlink() (string, error) {
```
mikeland73, line 96: change definition of this function to return sym link dir.

internal/wrapnix/wrapper.go
```diff
@@ -55,15 +55,19 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
...
-			BashPath:         bashPath,
-			Command:          bin,
-			ShellEnvHash:     shellEnvHash,
-			DevboxSymlinkDir: filepath.Dir(devboxSymlinkPath),
-			destPath:         filepath.Join(destPath, filepath.Base(bin)),
+			devboxer:                 devbox,
+			BashPath:                 bashPath,
+			Command:                  bin,
+			ShellEnvHash:             shellEnvHash,
+			DevboxSymlinkDirIfExists: devboxSymlinkDir,
```
mikeland73, line 69: revert

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -1,6 +1,8 @@
 #!{{ .BashPath }}
 
-export PATH="{{ .DevboxSymlinkDir }}:$PATH"
+{{ if ne .DevboxSymlinkDirIfExists "" }}
+PATH="{{ .DevboxSymlinkDirIfExists }}:$PATH"
+{{ end }}
```
mikeland73, line 5: revert

mikeland73 (commented):
I think we should always add the symlink dir to path even if the sym link was not created. The reason is that those binwrappers will work correctly when we are able to create the sym link.

Note that adding an empty dir to PATH does no harm.

Otherwise, we end up with a bunch of bin wrappers that need to be recreated later.

savil:
> I think we should always add the symlink dir to path even if the sym link was not created. The reason is that those binwrappers will work correctly when we are able to create the sym link.

hm fair point

internal/wrapnix/wrapper.go
savil: I'd prefer not doing that, since I'd expect a function named `CreateDevboxSymlink` to return the path to the symlink, rather than its directory. Changing the name to `CreateDevboxSymlinkAndGetDir` feels verbose. I think its okay for the handling logic to do `filepath.Dir`

savil: but yeah, let me know if you feel strongly, and I can make the change.

mikeland73 (approved):
OK, promise last comments.

Change `CreateDevboxSymlink` to return `error` only. Rename it to `CreateDevboxSymlinkIfPossible`. it returns nil if `evalSymlinkErr != nil`

create package variable
`var devboxSymlinkDir = xdg.CacheSubpath(filepath.Join("devbox", "bin", "current"))` 

use that variable in `CreateDevboxSymlinkIfPossible` and also in `CreateWrappers`

savil:
@mikeland73 cool, PTAL

mikeland73 (approved):

PR #1333, opened 2023-07-29
[docs] sidebars.js, "guides/plugins" section is duplicated

angelside:
"guides/plugins" was duplicated

## Summary

image

image

## How was it tested?
Not tested


LucilleH (approved):
Ooops! Good find thank you!

PR #1344, opened 2023-08-03
[easy] copy devbox.lock in generated Dockerfile

savil:
## Summary

Mohsen has a PR that will eventually copy all the code of the project into
the dockerfile, but I can see why that may be complicated.

For now, sending a quick fix to just copy the `devbox.lock` file 
so that our users get the exact packages they expect. This line can be removed
when the aforementioned PR is ready.

## How was it tested?

ran `devbox generate dockerfile` and saw the `COPY devbox.lock devbox.lock` 
line was present


mohsenari (approved):
thanks for sending this 👍

PR #1345, opened 2023-08-03
fix --process-compose-file issue

Lagoja:
## Summary

Fix the issue where passing a custom process compose file with `--process-compose-file` failed to load custom services

## How was it tested?

1. Add a process-compose.yaml with a service in a custom path
2. Load the file with `devbox services up --process-compose-file path/to/file
3. Verify that the service starts

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -658,6 +658,19 @@ func (d *Devbox) StartProcessManager(
 	background bool,
 	processComposeFileOrDir string,
 ) error {
+
+	if !d.IsEnvEnabled() {
+		args := []string{"services", "up"}
+		args = append(args, requestedServices...)
+		if processComposeFileOrDir != "" {
+			args = append(args, "--process-compose-file", processComposeFileOrDir)
+		}
+		if background {
+			args = append(args, "--background")
+		}
+		return d.RunScript(ctx, "devbox", args)
+	}
+
 	svcs, err := d.Services()
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


internal/impl/devbox.go
```diff
@@ -631,7 +631,7 @@ func (d *Devbox) RestartServices(ctx context.Context, serviceNames ...string) er
 
 	// TODO: Restart with no services should restart the _currently running_ services. This means we should get the list of running services from the process-compose, then restart them all.
 
-	svcSet, err := d.Services()
+	svcSet, err := d.Services("")
```
savil, line 634: The empty string can be hard to understand for a reader.

I'd suggest one of:
1. `d.Services("" /*userProcessCompose*/)`
2. define a separate function called `d.ServicesWithUserProcessCompose()`, and then `d.Services()` calls that with an empty-string.

savil (approved):
nice!

internal/impl/devbox.go
mikeland73: Option 3, make it an option for devbox.Open:

In internal/boxcli/services.go:

```golang
func startProcessManager(
	cmd *cobra.Command,
	args []string,
	servicesFlags servicesCmdFlags,
	flags serviceUpFlags,
) error {
	box, err := devbox.Open(&devopt.Opts{
		Dir:    servicesFlags.config.path,
                CustomProcessComposeFile: flags.processComposeFile,
		Writer: cmd.ErrOrStderr(),
	})
	if err != nil {
		return errors.WithStack(err)
	}

	return box.StartProcessManager(cmd.Context(), args, flags.background, flags.processComposeFile)
}
```

Another benefit of this method is we can eventually easily add this flag to other services commands (services ls, stop, etc)

PR #1346, opened 2023-08-03
[CLI] Added http url handling to create template

mohsenari:
## Summary
Added arbitrary git repo url handling to `devbox create` by adding `--repo` and `--subdir` flags. The behavior is as follows:
Either `--template` or `--repo` are required flags.
When `--template` is provided, we clone from devbox and use the mapping (no change)
When `--repo` is provided we clone that repo and if `--subdir` is provided, we copy over only the contents of the `subdir` otherwise, the whole repo is copied over.

Update: marked the 2 new flags (`--repo` and `--subdir`) hidden.

## How was it tested?
See unit tests

### DIFF
--- a/internal/boxcli/create.go
+++ b/internal/boxcli/create.go
@@ -17,19 +17,21 @@ import (
 type createCmdFlags struct {
 	showAll  bool
 	template string
+	repo     string
+	subdir   string
 }
 
 func createCmd() *cobra.Command {
 	flags := &createCmdFlags{}
 	command := &cobra.Command{
-		Use:   "create [dir] --template <template>",
+		Use:   "create [dir] --template <template> | --repo <repo URL> --subdir <subdirectory>",
 		Short: "Initialize a directory as a devbox project using a template",
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if flags.template == "" {
+			if flags.template == "" && flags.repo == "" {
 				fmt.Fprintf(
 					cmd.ErrOrStderr(),
-					"Usage: devbox create [dir] --template <template>\n\n",
+					"Usage: devbox create [dir] --template <template> | --repo <repo URL> --subdir <subdirectory>\n\n",
 				)
 				templates.List(cmd.ErrOrStderr(), flags.showAll)
 				if !flags.showAll {
@@ -52,6 +54,14 @@ func createCmd() *cobra.Command {
 		&flags.showAll, "show-all", false,
 		"show all available templates",
 	)
+	command.Flags().StringVarP(
+		&flags.repo, "repo", "r", "",
+		"Git repository HTTPS URL to import template files from. Example: https://github.com/jetpack-io/devbox",
+	)
+	command.Flags().StringVarP(
+		&flags.subdir, "subdir", "s", "",
+		"Subdirectory of the Git repository in which the template files reside. Example: examples/tutorial",
+	)
 
 	return command
 }
@@ -63,7 +73,7 @@ func runCreateCmd(cmd *cobra.Command, args []string, flags *createCmdFlags) erro
 		path = filepath.Join(wd, flags.template)
 	}
 
-	err := templates.Init(cmd.ErrOrStderr(), flags.template, path)
+	err := templates.Init(cmd.ErrOrStderr(), flags.template, flags.repo, flags.subdir, path)
 	if err != nil {
 		return err
 	}
--- a/internal/boxcli/create.go
+++ b/internal/boxcli/create.go
@@ -58,12 +58,13 @@ func createCmd() *cobra.Command {
 		&flags.repo, "repo", "r", "",
 		"Git repository HTTPS URL to import template files from. Example: https://github.com/jetpack-io/devbox",
 	)
-	command.Flags().MarkHidden("repo")
 	command.Flags().StringVarP(
 		&flags.subdir, "subdir", "s", "",
 		"Subdirectory of the Git repository in which the template files reside. Example: examples/tutorial",
 	)
-	command.Flags().MarkHidden("subdir")
+	// this command marks a flag as hidden. Error handling for it is not necessary.
+	_ = command.Flags().MarkHidden("repo")
+	_ = command.Flags().MarkHidden("subdir")
 
 	return command
 }
--- a/internal/boxcli/create.go
+++ b/internal/boxcli/create.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/templates"
 	"go.jetpack.io/devbox/internal/ux"
 )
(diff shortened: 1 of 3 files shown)


internal/templates/template.go
```diff
@@ -23,16 +25,23 @@ func Init(w io.Writer, template, dir string) error {
...
+		cloneURL = template
 	}
 
 	tmp, err := os.MkdirTemp("", "devbox-template")
 	if err != nil {
 		return errors.WithStack(err)
 	}
 	cmd := exec.Command(
-		"git", "clone", "https://github.com/jetpack-io/devbox.git", tmp,
+		fmt.Sprintf("git clone %s %s", cloneURL, tmp),
```
mikeland73, line 44: Any reason to wrap all of these in a single string? It may work, but generally you want to pass each a an arg. Keep in mind that exec.Command treats the first arg differently from the rest (e.g. it will wrap in double quotes etc). So if  cloneURL of temp have any crazy characters this may fail.

internal/templates/template.go
```diff
@@ -23,16 +25,23 @@ func Init(w io.Writer, template, dir string) error {
...
+	cloneURL := "https://github.com/jetpack-io/devbox.git"
 	if !ok {
-		return usererr.New("unknown template %q", template)
+		u, err := url.Parse(template)
+		splits := strings.Split(u.Path, "/")
+		if err != nil || u.Scheme != "https" || len(splits) == 0 {
+			return usererr.New("unknown template name or format %q", template)
+		}
+		templatePath = splits[len(splits)-1]
+		cloneURL = template
```
mikeland73, line 36: I would to a tiny bit of validation here. Something like:

```golang
if !ok {
  cloneURL, templatePath, err := parseGithubURL(template)
  if err != nil {
    return err
  }
}

func parseGithubURL(u string) (string, string, error) {
  // validate u 
  // parse 
}

File: internal/templates/template.go [hunk 3]
```diff
@@ -23,16 +25,23 @@ func Init(w io.Writer, template, dir string) error {
 	}
 
 	templatePath, ok := templates[template]
+	cloneURL := "https://github.com/jetpack-io/devbox.git"
 	if !ok {
-		return usererr.New("unknown template %q", template)
+		u, err := url.Parse(template)
+		splits := strings.Split(u.Path, "/")
+		if err != nil || u.Scheme != "https" || len(splits) == 0 {
```
mikeland73, line 32: splits needs to have more than zero no? only the org/repo name are 2 parts. Also, FYI this will return empty parts as well because Path starts with a "/"

internal/templates/template.go
```diff
@@ -23,16 +25,23 @@ func Init(w io.Writer, template, dir string) error {
...
+			return usererr.New("unknown template name or format %q", template)
+		}
+		templatePath = splits[len(splits)-1]
```
mikeland73, line 35: This won't work if directory is not provided and also ton't work if directory is nested (multiple slashes)

internal/templates/template.go
```diff
@@ -23,16 +25,23 @@ func Init(w io.Writer, template, dir string) error {
 	}
 
 	templatePath, ok := templates[template]
+	cloneURL := "https://github.com/jetpack-io/devbox.git"
```
mikeland73, line 28: alternative name `repoURL`

internal/templates/template.go
mikeland73: Putting `parseGithubURL` let's you write a few unit tests as well to ensure you are getting what you expect.

mikeland73 (changes requested):
Ugh, one thing we didn't think about is that github urls for subdirectories are not just base url plus directory. They usually include branch as well, e.g.

https://github.com/jetpack-io/devbox/tree/main/plugins

So https://github.com/jetpack-io/devbox/plugins doesn't really point to anything.

We could try to be smart and remove the "/tree/<branch>" when parsing.

LucilleH:
Would it make things easier if we pass in `--repo` and `--folder` as separate args?

mohsenari:
> Would it make things easier if we pass in `--repo` and `--folder` as separate args?

I'm coming to the same conclusion, because I don't wanna restrict it only to github and then parse `/tree/branch` because branch might also include `/` in its name. So it makes it clearer to use `--repo` and `--subdirectory` or `--folder` to be clear and also skip the hacky URL path parsing.

internal/templates/template.go
mohsenari: This was a mistake, I reverted it.

internal/boxcli/create.go
```diff
@@ -63,7 +76,7 @@ func runCreateCmd(cmd *cobra.Command, args []string, flags *createCmdFlags) erro
 		path = filepath.Join(wd, flags.template)
 	}
 
-	err := templates.Init(cmd.ErrOrStderr(), flags.template, path)
+	err := templates.Init(cmd.ErrOrStderr(), flags.template, flags.repo, flags.subdir, path)
```
mikeland73, line 79: if flags.template {
  err = templates.InitFromName
} else {
  err = templates.InitFromRepo
}

internal/boxcli/create.go
```diff
@@ -63,7 +76,7 @@ func runCreateCmd(cmd *cobra.Command, args []string, flags *createCmdFlags) erro
 		path = filepath.Join(wd, flags.template)
```
mikeland73, line 76: if `flags.template` is not set, the second part of filepath.Join() should be:

* filepath.Base(subdir) if subdir is set.
* filepath.Base(repo) otherwise.

internal/templates/template.go
```diff
@@ -80,3 +79,27 @@ func createDirAndEnsureEmpty(dir string) error {
...
+		subdirPath = tPath
+	} else if repo != "" {
+		_, err := url.Parse(template)
+		if err != nil {
+			return "", "", usererr.New("Invalid URL format for --repo %s", repo)
+		}
+		subdirPath = subdir
+		// this is to handle cases where user puts repo url with .git at the end
+		// like: https://github.com/jetpack-io/devbox.git
+		repoURL, _ = strings.CutSuffix(repo, ".git")
```
mikeland73, line 100: TrimSuifix

internal/templates/template.go
```diff
@@ -17,23 +19,20 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 )
 
-func Init(w io.Writer, template, dir string) error {
+func Init(w io.Writer, template, repo, subdir, dir string) error {
```
mikeland73, line 22: Consider breaking this into `InitFromName` and `InitFromRepo` where `InitFromName` calls `InitFromRepo`. Also consider renaming variables to make them less ambiguous (dir, subdir, subdirPath, is pretty confusing. For example existing `dir` can be `target` or `targetDir`)

This would help simplify the logic and prevent all the duplicate variables you have here (e.g. subdir and subdirPath, repo and repoURL).

e.g.

```
# InitFromRepo is similar to existing `Init` but removes template map logic.
# it calls `strings.TrimSuifix` and can validate repo parses correctly.
func InitFromRepo(repo, subdir, target string) {
  ..
}

func InitFromName(w io.Writer, template, target string) {
  templatePath, ok := templates[template]
  if !ok {
    return usererr.New("unknown template %q", template)
  }
  return InitFromRepo("https://github.com/jetpack-io/devbox", templatePath, target)
}
```

mikeland73 (approved):
non-blocking style suggestions. 

I would say getting fixing the path assignment is important but we can always just use it only with a path for now.

PR #1349, opened 2023-08-07
remove cuego.Complete from cuecfg.Marshal

savil:
## Summary

`cuego.Complete` leads to non-trivial complexity when dealing with custom json
marshalling. In particular, it does a round-trip JSON marshal/unmarshal and fields
that are not handled by golang's json-framework can be lost. This interferes
with custom marshalling/unmarshalling.

`cuego.Complete` is only needed if we rely on the `cue:` annotations to autocomplete some fields (I mean, assign default values respecting cue constraints). We don't rely on this, so, right now it is superfluous.

I also remove the `cue:` annotations in a couple of structs since we do not
rely on them.

## How was it tested?

- [x] `devbox init` should produce the expected output
- [x] `devbox add` 
- [x] `devbox rm`

testscript unit tests must pass

### DIFF
--- a/internal/cuecfg/cuecfg.go
+++ b/internal/cuecfg/cuecfg.go
@@ -8,18 +8,12 @@ import (
 	"os"
 	"path/filepath"
 
-	"cuelang.org/go/cuego"
 	"github.com/pkg/errors"
 )
 
 // TODO: add support for .cue
 
 func Marshal(valuePtr any, extension string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


mikeland73 (approved):
I had no idea what this was for. cc: @loreto

loreto:
LGTM

I don't know if there's other places where `cue` is handling the config, but at this point, you might as well remove the dependency completely.

`cuecfg` is a misnomer now, you might want to rename the package.

PR #1351, opened 2023-08-07
[Config Packages] Enable optional map spec

savil:
## Summary

**Motivation**
User feedback is leading us to add more parameters besides version to the packages config.
For instance, some packages may only need to be added on certain platforms.

Previously, packages were specified as `[]string`. For example:
```
{
  "packages": [
    "python@3.9",
    "hello",
    "cowsay@latest"
  ],
  ...
}
```

In the new format, they can also be specified as:
```
{
   "packages": {
      "python": "3.9",
      "hello": "",
      "cowsay": {
        "version": "latest",
      }
   }
}
```
NOTE: `python` is an inline-version, while `cowsay` has a struct value which can
accomodate other fields like `platforms`.

**Approach**
For now, we enable both packages as List and Map to work. The default continues to be List.
We do this by adding a `Packages` struct and `Package` struct with custom JSON marshalling and unmarshalling.

**Future PRs**
1. Introduce `Packages.platforms` and/or `Packages.excludedPlatforms` fields. Use them when installing to nix profile, and generating the flake.nix.
2. Introduce `devbox add` and `devbox rm` flags for platform-specific packages.
3. Detect package usage failures when due to platform mis-match and suggest using CLI commands to run to make platform-specific.

## How was it tested?

Added `TestJsonifyConfigPackages` and other unit-tests

testscripts pass

`devbox init` produces the same output as before.

### DIFF
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -189,13 +189,10 @@ func (p Package) MarshalJSON() ([]byte, error) {
 func parseVersionedName(versionedName string) (name, version string) {
 	var found bool
 	name, version, found = searcher.ParseVersionedPackage(versionedName)
-	if found {
-		return name, version
-	} else {
+	if !found {
 		// Case without any @version in the versionedName
-		name = versionedName
 		// We deliberately do not set version to `latest`
-		version = ""
+		return versionedName, "" /*version*/
 	}
 	return name, version
 }
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -1,0 +1,213 @@
+package devconfig
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/iancoleman/orderedmap"
+	"github.com/pkg/errors"
+)
+
+type jsonKind int
+
+const (
+	// jsonList is the legacy format for packages
+	jsonList jsonKind = iota
+	// jsonMap is the new format for packages
+	jsonMap jsonKind = iota
+)
+
+type Packages struct {
+	jsonKind jsonKind
+
+	// Collection contains the set of package definitions
+	// We don't want this key to be serialized automatically, hence the "key" in json is "-"
+	Collection []Package `json:"-,omitempty"`
+}
+
+// VersionedNames returns a list of package names with versions.
+// NOTE: if the package is unversioned, the version will be omitted (doesn't default to @latest).
+//
+// example:
+// ["package1", "package2@latest", "package3@1.20"]
+func (pkgs *Packages) VersionedNames() []string {
+	result := []string{}
+	for _, p := range pkgs.Collection {
+		name := p.name
+		if p.Version != "" {
+			name += "@" + p.Version
+		}
+		result = append(result, name)
+	}
+	return result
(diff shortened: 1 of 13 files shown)


savil:
* main
  * **PR #1349** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1349" target="_blank">Graphite</a> 
    * **PR #1351** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1351" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1351?utm_source=stack-comment).

internal/devconfig/packages.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func (p Package) MarshalJSON() ([]byte, error) {
+	if p.kind == versionOnly {
+		return json.Marshal(p.Version)
+	}
+
+	// If we have a regular package, we want to marshal the entire struct:
+	type Alias Package // Use an alias-type to avoid infinite recursion
```
mikeland73, line 188: This is cool! I've gotten into infinite recursion issue when implementing custom marshaling and I don't remember the solution being this clean.

nit, make lower case

internal/devconfig/packages.go
```diff
@@ -0,0 +1,226 @@
...
+	return json.Marshal((Alias)(p))
+}
+
+// parseVersionedName parses the name and version from package@version representation
+func parseVersionedName(versionedName string) (name, version string) {
```
mikeland73, line 193: see `searcher.ParseVersionedPackage`

this does a little more, but hopefully you can use it to prevent some code duplication.

internal/devconfig/packages.go
```diff
@@ -0,0 +1,226 @@
...
+	var version string
+	if err := json.Unmarshal(data, &version); err == nil {
+		p.kind = versionOnly
+		p.Version = version
+		return nil
+	}
+
+	// Second, attempt to unmarshal as a Package struct
+	type Alias Package // Use an alias-type to avoid infinite recursion
+	alias := &Alias{}
```
mikeland73, line 171: nit lower case, maybe

```golang
type packageAlias Package // Use an alias-type to avoid infinite recursion
alias := &packageAlias{}
```

mikeland73 (approved):
This is really nice implementation! great encapsulation, super easy to read.

internal/devconfig/packages.go
```diff
@@ -0,0 +1,226 @@
...
+	if err := json.Unmarshal(data, alias); err != nil {
+		return errors.WithStack(err)
+	}
+
+	// more robust way to copy all fields from alias?
```
gcurtis, line 176: Try `*p = Package(*alias)`

gcurtis (approved):
Thanks for the tests! I also like the idea of aliasing the type for (un)marshalling.

The map order mattering is unfortunate. Maybe we can eventually stop relying on that.

savil (commented):
Thanks for the comments! responses inline, and updates made.

Issue #1215, opened 2023-06-24
[Bug]: Can't compile go programs in debug mode


ibuildthecloud:
**Current Behavior (bug)**
Ironically I'm trying to hack on devbox inside of devbox and I hit this issue.

```
$ ~/go/bin/dlv debug ./cmd/devbox/
# runtime/cgo
In file included from /nix/store/bhhb91angi23wsr42apz1in1n9b9ilg3-glibc-2.37-8-dev/include/bits/libc-header-start.h:33,
                 from /nix/store/bhhb91angi23wsr42apz1in1n9b9ilg3-glibc-2.37-8-dev/include/stdlib.h:26,
                 from _cgo_export.c:3:
/nix/store/bhhb91angi23wsr42apz1in1n9b9ilg3-glibc-2.37-8-dev/include/features.h:413:4: error: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Werror=cpp]
  413 | #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
      |    ^~~~~~~
cc1: all warnings being treated as errors
exit status 1
```

I know this is really a nix issue and not devbox directly, but being that this configuration of running `go` sdk inside devbox is a primary use case I would assume this would work.

**Additional context**
Please include the output of `devbox version -v` and
```
$ devbox version -v
Version:     0.5.5
Platform:    linux_amd64
Commit:      891e8de7a8ab7a3fec42fad069326d68cb84603e
Commit Time: 2023-06-19T19:18:00Z
Go Version:  go1.20.5
Launcher:    0.2.0
```
```
$ cat /home/darren/.local/share/devbox/global/default/devbox.json
{
  "packages": [
    "vim-full@latest",
    "cmake@latest",
    "python@3.8",
    "gcc@latest",
    "libxcrypt@latest",
    "go@1.20",
    "gh@latest",
    "k3d@latest",
    "xdg-utils@latest",
    "xfce.xfce4-terminal@latest",
    "ubuntu_font_family@latest",
    "hicolor-icon-theme@latest",
    "fontconfig@latest",
    "xorg.xorgserver@latest",
    "xterm@latest",
    "jetbrains.goland@latest",
    "fuse3@latest",
    "fuse@latest",
    "strace@latest",
    "git@latest",
    "procps@latest"
  ],
  "shell": {
    "init_hook": [
      "echo 'Welcome to devbox!' > /dev/null"
    ],
    "scripts": {
      "test": [
[...]

PR #1353, opened 2023-08-08
[env] Add --env and --env-file flags

mikeland73:
## Summary

Adds `--env` and `--env-file` flags that can be used with `run` `shell` and `services`

If both flags are specified they are merged with `--env` taking precedence. 

environment variables passed in with flags take the highest level of precedence, replacing any env variables set by nix, devbox.json, or plugins.

TODO: The flag code is a bit brittle and repetitive, I think we can consolidate the common flag registration.

## How was it tested?

```
devbox run -e FOO=BAR echo \$FOO
devbox services up -e HTTPD_PORT=8083
```


mohsenari (approved):
We should add in our docs that if bith env and env-file are specified env takes priority

Lagoja:
Should also help with #1215, since we can set the NIX_HARDENING_ENABLE variable with the -e flag

PR #1354, opened 2023-08-09
[env] Add env flags to direnv commands

mikeland73:
## Summary

This allows users to generate .envrc files that use custom env variables like so:

```bash
devbox generate direnv -e foo=bar -e foo1=bar2
devbox generate direnv --env-file .env
```

Users can also manually modify `.envrc` files to add additional flags:

`devbox generate direnv --print-envrc -e foo=bar`

## How was it tested?

Generated .envrc file with custom env flags. Then used direnv and verified environment was setup as expected.

## Backward compatibility

Newly generated `.envrc` files that contain `--env` or `--env-file` flags will not be compatible with previous versions of devbox.

### DIFF
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -61,7 +61,7 @@ func CreateWrappers(ctx context.Context, devbox devboxer) error {
 	}
 
 	for _, bin := range bins {
-		if err = createWrapper(&createWrapperArgs{
+		if err = createWrapperIfNeeded(&createWrapperArgs{
 			devboxer:         devbox,
 			BashPath:         bashPath,
 			Command:          bin,
@@ -141,7 +141,11 @@ type createWrapperArgs struct {
 	destPath         string
 }
 
-func createWrapper(args *createWrapperArgs) error {
+func createWrapperIfNeeded(args *createWrapperArgs) error {
+	// If hash is already set, the env is up to date, no need for wrappers.
+	if args.ShellEnvHash == os.Getenv(args.ShellEnvHashKey()) {
+		return nil
+	}
 	buf := &bytes.Buffer{}
 	if err := wrapperTemplate.Execute(buf, args); err != nil {
 		return errors.WithStack(err)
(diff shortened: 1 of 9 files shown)


Lagoja:
Just tested this. One issue with how this works is that the command will fail if the file passed with `--env-file` is not present. Since the .envrc will probably be checked into source control, but the .env file won't, this will cause the envrc to fail to load if the file is missing:

```
direnv allow
direnv: loading ~/src/devbox/.envrc
direnv: using devbox
Error: open .env.ghost: no such file or directory
```

A better solution might be to implement a "load if file exists" functionality, similar to direnv's `dotenv_if_exists` function: 

https://direnv.net/man/direnv-stdlib.1.html#codedotenvifexists-ltdotenvpathgtcode

savil:
Discussion clarified offline that the behavior won't break if the env-file is missing. Well, it won't break for direnv (`devbox generate direnv --env-file fo`), but will break if env file is missing for `devbox run --env-file foo`

internal/wrapnix/wrapper.go
```diff
@@ -141,7 +141,11 @@ type createWrapperArgs struct {
 	destPath         string
 }
 
-func createWrapper(args *createWrapperArgs) error {
+func createWrapperIfNeeded(args *createWrapperArgs) error {
+	// If hash is already set, the env is up to date, no need for wrappers.
+	if args.ShellEnvHash == os.Getenv(args.ShellEnvHashKey()) {
+		return nil
+	}
```
savil, line 148: wait, this change is quite unrelated to this PR. Can we move it out?

I don't think this makes sense. We need the wrappers because the shellenv environment _may_ change after the wrappers have been generated.

savil (commented):
@mikeland73 where is the logic that makes it forgiving if `--env-file` file is missing?

Also - I don't understand why the shellenv hash change is related, and how it works...

internal/wrapnix/wrapper.go
mikeland73: this is a bad change that snuck in.

savil (approved):
acceping to unblock, assuming:
1. the change to make `--env-file` forgiving for direnv is added
2. the wrapper change is reverted

savil (approved):
nice!

PR #1357, opened 2023-08-09
[Packages] Introduce the concept of Installable packages

savil:
## Summary

This PR lays some groundwork for introducing platform-specific packages in a later PR (#1358)

This PR makes the following changes to `devbox` struct:
1. Replace `devbox.Packages` with `ConfigPackageNames()` and `InstallablePackageNames()`. 
2. Replace `devbox.PackagesAsInputs` with `ConfigPackages()` and `InstallablePackages()`. 
Updates callers to call the appropriate one.

**devpkg.Package changes**
I also introduce a new method `devpkg.PackagesFromConfig`, which enables us to add an `isInstallable bool` field on the `devpkg.Package`. I needed this because it affects the implementation of certain methods (such as `IsLegacy` changed in #1358).

One can still construct via `PackagesFromStrings` which default to `isInstallable = true` (as before). I audited the callers.

## How was it tested?

Did adhoc testing.
Testscript unit tests should pass.

### DIFF
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -36,18 +36,9 @@ type Packages struct {
 func (pkgs *Packages) VersionedNames() []string {
 	result := make([]string, 0, len(pkgs.Collection))
 	for _, p := range pkgs.Collection {
-		name := p.name
-		if p.Version != "" {
-			name += "@" + p.Version
-		}
-		result = append(result, name)
+		result = append(result, p.VersionedName())
 	}
 	return result
-}
-
-func (pkgs *Packages) VersionedNamesForPlatform() ([]string, error) {
-	// TODO savil. Next PR will update this implementation
-	return pkgs.VersionedNames(), nil
 }
 
 // Add adds a package to the list of packages
@@ -157,6 +148,19 @@ func NewPackage(name string, values map[string]any) Package {
 		name:    name,
 		Version: version.(string),
 	}
+}
+
+func (p *Package) VersionedName() string {
+	name := p.name
+	if p.Version != "" {
+		name += "@" + p.Version
+	}
+	return name
+}
+
+func (p *Package) IsEnabledOnPlatform() (bool, error) {
+	// TODO savil. Next PR will update this implementation
+	return true, nil
 }
 
 func (p *Package) UnmarshalJSON(data []byte) error {
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -158,9 +158,9 @@ func (p *Package) VersionedName() string {
 	return name
 }
 
-func (p *Package) IsEnabledOnPlatform() (bool, error) {
+func (p *Package) IsEnabledOnPlatform() bool {
 	// TODO savil. Next PR will update this implementation
-	return true, nil
+	return true
 }
 
 func (p *Package) UnmarshalJSON(data []byte) error {
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -45,9 +45,9 @@ func (pkgs *Packages) VersionedNames() []string {
 	return result
 }
 
-func (pkgs *Packages) VersionedNamesForPlatform() []string {
+func (pkgs *Packages) VersionedNamesForPlatform() ([]string, error) {
 	// TODO savil. Next PR will update this implementation
-	return pkgs.VersionedNames()
+	return pkgs.VersionedNames(), nil
 }
 
 // Add adds a package to the list of packages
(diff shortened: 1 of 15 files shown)


savil:
* main
  * **PR #1357** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1357" target="_blank">Graphite</a>  👈
    * **PR #1358** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1358" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1357?utm_source=stack-comment).

internal/devconfig/packages.go
```diff
@@ -154,6 +150,19 @@ func NewPackage(name string, values map[string]any) Package {
...
 
+func (p *Package) VersionedName() string {
+	name := p.name
+	if p.Version != "" {
+		name += "@" + p.Version
+	}
+	return name
+}
+
+func (p *Package) IsEnabledOnPlatform() (bool, error) {
```
mikeland73, line 161: Making this return `(bool, error)` causes large changes in the code base (and generally makes things  bit more difficult to read). 

If seems like in the follow up PR you implement this using `nix.System`

Suggestion:

Have `nix.EnsureNixInstalled` call `nix.ComputeSystem` which stores the system in a package variable.

nix.System returns it (and panics if not set).

I know this has a bunch of stuff stacked on it so I don't know how easy it would be to make this change, but it would prevent a good amount of code/blame churn.

internal/impl/devbox.go
```diff
@@ -926,13 +941,43 @@ func (d *Devbox) nixFlakesFilePath() string {
...
+func (d *Devbox) ConfigPackageNames() []string {
+	// TODO savil: centralize implementation by calling d.ConfigPackages and getting pkg.Raw
+	// Skipping for now to avoid propagating the error value.
 	return d.cfg.Packages.VersionedNames()
 }
 
-func (d *Devbox) PackagesAsInputs() []*devpkg.Package {
-	return devpkg.PackageFromStrings(d.Packages(), d.lockfile)
+// InstallablePackageNames returns the names of packages that are to be installed
+func (d *Devbox) InstallablePackageNames() ([]string, error) {
```
mikeland73, line 952: I'm not sure this function needs to exist see other comment in plugin.

internal/plugin/plugin.go
```diff
@@ -157,13 +157,18 @@ func (m *Manager) createFile(
 		urlForInput = pkg.URLForFlakeInput()
 	}
 
+	packages, err := m.InstallablePackageNames()
```
mikeland73, line 160: I think it's fine for this to be all package names. Specifically, this is used so that plugin templates know all packages, it is currently used by php and haskell plugins.

If anything, in a follow up we could change this to be Package structs instead of strings. (and it would be backward compatible because the struct has a `String()` method.

This would allow us to remove InstallablePackageNames and just keep a single `PackageNames`.

internal/impl/devbox.go
```diff
@@ -926,13 +941,43 @@ func (d *Devbox) nixFlakesFilePath() string {
...
+		return nil, err
+	}
+	return lo.Map(pkgs, func(pkg *devpkg.Package, _ int) string {
+		return pkg.Raw
+	}), nil
+}
+
+// ConfigPackages returns the packages that are defined in devbox.json
+// NOTE: the return type is different from devconfig.Packages
+func (d *Devbox) ConfigPackages() ([]*devpkg.Package, error) {
```
mikeland73, line 964: This should be a private method.

mikeland73 (approved):
I have 2 main concerns:

* Changing the signature of packages function to return an error causes significant code churn and can be avoided by pre-computing the system earlier in the call.
* Having 2 similar functions (installable vs config) increases cognitive load on future work. A partial solution is (a) Remove `installablePackageNames` because I think it is not needed. (b) Make configPackages private.  

I think longer term we could also get rid of `ConfigPackageNames` and just end up with:

* InstallablePackages (public)
* configPackages (private)

internal/devconfig/packages.go
savil: I added `nix.MustGetSystem`, since `nix.System` currently has an error return refactoring which would introduce a lot of unrelated changes. Can cleanup in a followup PR.

internal/plugin/plugin.go
savil: > If anything, in a follow up we could change this to be Package structs instead of strings. (and it would be backward compatible because the struct has a String() method.

Before we rely on `.String()`, I'd prefer making Package structs not embed url.URL, so we can properly implement `String()` to do what is intended.

PR #1359, opened 2023-08-09
Added --root-user flag to devbox generate

mohsenari:
## Summary
follow up for #1325 
This also potentially fixes an issue reported for users having permission issues with `.devbox/` directory inside devcontainer because of `root` user inside the container.

Added a `--root-user` flag to allow both single user installation and multi user installation for nix as well as setup for Dockerfile and devcontainer.json.

This allows users to choose either `devbox` as a user in the container environment (default) or `root` as a user in the container environment (passing `--root-user` flag).

## How was it tested?
`devbox generate dockerfile -f -r`
compare content of Dockerfile with 
`devbox generate dockerfile -f`

### DIFF
--- a/internal/boxcli/generate.go
+++ b/internal/boxcli/generate.go
@@ -18,6 +18,7 @@ type generateCmdFlags struct {
 	force             bool
 	printEnvrcContent bool
 	githubUsername    string
+	rootUser          bool
 }
 
 func generateCmd() *cobra.Command {
@@ -64,6 +65,8 @@ func devcontainerCmd() *cobra.Command {
 	}
 	command.Flags().BoolVarP(
 		&flags.force, "force", "f", false, "force overwrite on existing files")
+	command.Flags().BoolVarP(
+		&flags.rootUser, "root-user", "r", false, "Use root as default user inside the container")
 	return command
 }
 
@@ -81,6 +84,8 @@ func dockerfileCmd() *cobra.Command {
 	}
 	command.Flags().BoolVarP(
 		&flags.force, "force", "f", false, "force overwrite existing files")
+	command.Flags().BoolVarP(
+		&flags.rootUser, "root-user", "r", false, "Use root as default user inside the container")
 	flags.config.register(command)
 	return command
 }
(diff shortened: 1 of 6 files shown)


internal/impl/generate/devcontainer_util.go
```diff
@@ -23,6 +23,15 @@ import (
...
 var tmplFS embed.FS
 
+type Generate struct {
+	Ctx            context.Context
+	Path           string
+	RootUser       bool
+	IsDevcontainer bool
+	Pkgs           []string
+	LocalFlakeDirs []string
+}
```
mikeland73, line 33: A few issues:

* Not sure about the name (`generate.Generate` is repetitive and doesn't convey what it does).
* As currently used it does not need to be exported.
* It should not contain context. (In general context should never be saved in a struct).

Proposal:

* Name it `Options` instead.
* Remove context.
* Remove Open

Code would look like:

```golang
g := &generate.Options{...}
g.CreateDockerfile(ctx)
```

internal/impl/generate/devcontainer_util.go
```diff
@@ -128,7 +158,10 @@ func getDevcontainerContent(pkgs []string) *devcontainerObject {
 			},
 		},
 		// Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
```
mikeland73, line 160: is this comment still relevant?

mikeland73 (changes requested):
I'm not super familiar with dockerfile changes. Some comments on golang changes.

mohsenari:
@mikeland73 I updated the PR based on the feedback. PTAL

internal/boxcli/generate.go
```diff
@@ -139,13 +144,17 @@ func runGenerateCmd(cmd *cobra.Command, flags *generateCmdFlags) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	generateOpts := devopt.GenerateOpts{
```
mikeland73, line 147: Hack, if you embed `devopt.GenerateOpts` to the flag struct you can pass `flags.GenerateOpts` into the functions below.

mikeland73 (approved):
Looks good!

I'm less confident about the dockerfile stuff so make sure that's well tested.

PR #1362, opened 2023-08-10
Update node example to use the prebuilt binary

LucilleH:
## Summary
Unfortunately I can't tell if the version for `nodePackages.pnpm` is prebuilt or not. Once we have that information in https://nixhub.io I will update the examples again.

## How was it tested?
devbox install

### DIFF
--- a/examples/development/nodejs/nodejs-pnpm/devbox.json
+++ b/examples/development/nodejs/nodejs-pnpm/devbox.json
@@ -1,7 +1,7 @@
 {
     "packages": [
         "nodejs@18",
-        "nodePackages.pnpm@latest"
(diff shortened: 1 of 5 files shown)


gcurtis (approved):

gcurtis:
This is the command to check if a package is in the cache:

```
nix path-info --store 'https://cache.nixos.org' --json nixpkgs/<commit>#<attr_path>
```

For example:

```
nix path-info --store 'https://cache.nixos.org' --json nixpkgs/3c614fbc76fc152f3e1bc4b2263da6d90adf80fb#nodejs_18
```

mohsenari (approved):

PR #1367, opened 2023-08-11
[plugins] Ensure plugin packages are in nix profile

mikeland73:
## Summary

While working on https://github.com/jetpack-io/devbox/pull/1361 I noticed that we were not installing plugin packages (usually just custom flakes) to profile. This means in cases like the PHP plugin, the binary int he profile was wrong because it did not include extensions. Everything was still working as expected because bin wrappers would point to the correct binary but this is bad for 2 reasons:

* We want to get rid of bin wrappers
* It is quite fragile for binaries in profile to be wrong binaries

This PR fixes it by ensuring plugin packages are installed. I also refactored some function names that were out of date.

TODO: We currently output text when installing plugin packages that is not very user friendly. We could improve it.

## How was it tested?

```bash
➜  devbox git:(landau/install-plugin-packages-to-profile) devbox run -c examples/development/php/php8.1 php -m | grep imagick
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Generating autoload files
imagick
```


savil (commented):
could we revert the function renames? My PR from yesterday is already renaming them:
https://github.com/jetpack-io/devbox/pull/1357

savil (approved):
nice! would be nice to revert the function renames to minimize merge-conflicts with my PR...

PR #1373, opened 2023-08-14
[per-OS Packages] Error if adding --platform, when `excluded_platforms` is defined

savil:
## Summary

Background:
for a package, `platforms` and `excluded_platforms` should be defined mutually-exclusively.

Previously, we were being "clever" by dropping "excluded_platforms" when a user
does `devbox add <package> --platform <platform>`. But I think we should:
1. Ensure the user explicitly makes the change to their devbox config.
2. Be consistent with how the inverse is handled: using `--exclude-platform` when `platform` is defined.

## How was it tested?

added testscript unit-tests

### DIFF
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -87,7 +87,13 @@ func (pkgs *Packages) AddPlatforms(versionedname string, platforms []string) err
 
 			// Adding any platform will restrict installation to it, so
 			// the ExcludedPlatforms are no longer needed
-			pkg.ExcludedPlatforms = nil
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1373** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1373" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1373?utm_source=stack-comment).

mikeland73 (approved):

PR #1374, opened 2023-08-14
[nix.System] call EnsureNixInstalled from devbox.Open, and cleanup function API

savil:
## Summary

`nix.System` is widely used. So, in #1357 we wanted to change its API to just
return `string` instead of `string, error`.

To do so, we cache the result in a `nix` package variable. And populate it
during `EnsureNixInstalled`.

However, `EnsureNixInstalled` was only called from Cobra command functions. But
`devbox` as a library also now needs to call it. IMO, it should always have called
it since we do rely on `nix` being installed. This PR adds this.

It also fixes up the API to:
`System: string` and `ComputeSystem: (string, error)`.

NOTE: inside `System`, I still do a best-effort to avoid panic by calling
`ComputeSystem`. This can happen in edge-cases like some tests `generate_test/TestWriteFromTemplate`
that exercise internal functions reliant on `nix.System` without calling `devbox.Open`.

## How was it tested?

tests should pass

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -118,14 +118,13 @@ func ExperimentalFlags() []string {
 	}
 }
 
-// TODO: rename to System
-func MustGetSystem() string {
+func System() string {
 	if cachedSystem == "" {
-		// For internal calls (during tests), this may not be initialized properly
-		// so do a best-effort attempt to initialize it.
-		_, err := System()
+		// While this should have been initialized, we do a best-effort to avoid
+		// a panic.
+		_, err := ComputeSystem()
 		if err != nil {
-			panic("MustGetSystem called before being initialized by System")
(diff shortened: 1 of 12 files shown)


internal/nix/nix.go
```diff
@@ -118,23 +118,21 @@ func ExperimentalFlags() []string {
...
 		}
 	}
 	return cachedSystem
 }
 
 var cachedSystem string
 
-// TODO: rename to ComputeSystem or ComputePlatform?
-func System() (string, error) {
+func ComputeSystem() (string, error) {
```
mikeland73, line 135: ComputeSystem should only return error (we always discard the string)

mikeland73 (approved):

savil:
I had to bump the test-timeout to land this. We needed this PR for correctness. But I will pursue a fix to improve the test times.

PR #1378, opened 2023-08-15
change nix.System back to returning error

savil:
## Summary

Returns `nix.System` error, rather than caching it in `ensureNixInstalled` which
was either error prone, OR had a perf regression.

## How was it tested?

will view testscripts outcome...


savil:
* main
  * **PR #1378** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1378" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1378?utm_source=stack-comment).

PR #1383, opened 2023-08-18
fix: typo in mysql process-compose.yml

kei01234kei:
## Summary

Fix typo in `plugins/mysql/process-compose.yaml` .

### DIFF
--- a/plugins/mysql/process-compose.yaml
+++ b/plugins/mysql/process-compose.yaml
@@ -2,7 +2,7 @@ version: "0.5"
 
 processes:
   mysql:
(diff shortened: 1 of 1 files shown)


LucilleH (approved):
TY!

kei01234kei:
Idk why my code doesn't pass ci though I changed only comment

LucilleH:
@kei01234kei I think we had changes the cicd tests in the past 4 days, and the code here is not rebased yet. But that's ok I think this is safe enough to just merge and ship 👍

PR #1408, opened 2023-08-18
[cli-tests] fix by pinning dependencies in rails examples

savil:
## Summary

We pin yarn to a version that has pre-built binaries.
We also update the nodejs version so it is pinned to a nixpkgs hash commit that has pre-built binaries.

From my tests (see PRs stacked from #1388), I think this should fix the cli-tests that were timing out for nix 2.17

## How was it tested?

tested locally via running `devbox shell`.

### DIFF
--- a/examples/stacks/rails/blog/Gemfile.lock
+++ b/examples/stacks/rails/blog/Gemfile.lock
@@ -127,6 +127,8 @@ GEM
     nio4r (2.5.9)
     nokogiri (1.14.3-arm64-darwin)
       racc (~> 1.4)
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #1408** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1408" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1408?utm_source=stack-comment).

LucilleH (approved):

PR #1416, opened 2023-08-23
[rm-nixpkgs] Fix profile list bug for packages added by store path

ipince:
## Summary
When a package is added by store path, then when we query `nix profile list`, it'll look like this:
```
> nix profile list --profile .devbox/nix/profile/default
Index:              0
Store paths:        /nix/store/yjkac3v6c1v129fph0gvy9c7ndnx738b-cowsay-3.04
```

instead of this:
```
> nix profile list --profile .devbox/nix/profile/default
Index:              0
Flake attribute:    legacyPackages.x86_64-darwin.hello
Original flake URL: github:NixOS/nixpkgs/3007746b3f5bfcb49e102b517bca891822a41b31
Locked flake URL:   github:NixOS/nixpkgs/3007746b3f5bfcb49e102b517bca891822a41b31
Store paths:        /nix/store/gyh1bri77vlcs1pwxyg4x3z06aql2c6p-hello-2.12.1
```

Thus, when we read the profile items and try to build a map out of the item's unlockedRef, we were always ending up with a map of size 1.

To fix, I changed the map to a slice, and changed the callsites to handle the slice instead.

I made a few other code organization changes that are no-ops.

## How was it tested?
Tested by calling `devbox install` and inspecting the profile with `nix profile list`, with and without the rm-nixpkgs feature enabled. I attempted to add unit tests but it's not as easy as I had thought.

### DIFF
--- a/internal/nix/nixprofile/item.go
+++ b/internal/nix/nixprofile/item.go
@@ -1,0 +1,67 @@
+package nixprofile
+
+import (
+	"fmt"
+	"strings"
+
+	"go.jetpack.io/devbox/internal/devpkg"
+	"go.jetpack.io/devbox/internal/lock"
+	"go.jetpack.io/devbox/internal/redact"
+)
+
+// NixProfileListItem is a go-struct of a line of printed output from `nix profile list`
+// docs: https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-profile-list.html
+type NixProfileListItem struct {
+	// An integer that can be used to unambiguously identify the package in
+	// invocations of nix profile remove and nix profile upgrade.
+	index int
+
+	// The original ("unlocked") flake reference and output attribute path used at installation time.
+	// NOTE that this can be "#" if the package was added to the nix profile via store path.
+	unlockedReference string
+
+	// The locked flake reference to which the unlocked flake reference was resolved.
+	// NOTE that this can be "#" if the package was added to the nix profile via store path.
+	lockedReference string
+
+	// The store path(s) of the package.
+	nixStorePath []string
+}
+
+// AttributePath parses the package attribute from the NixProfileListItem.lockedReference
+//
+// For example:
+// if NixProfileListItem.lockedReference = github:NixOS/nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04#legacyPackages.x86_64-darwin.go_1_19
+// then AttributePath = legacyPackages.x86_64-darwin.go_1_19
+func (item *NixProfileListItem) AttributePath() (string, error) {
+	// lockedReference example:
+	// github:NixOS/nixpkgs/52e3e80afff4b16ccb7c52e9f0f5220552f03d04#legacyPackages.x86_64-darwin.go_1_19
+
+	// AttributePath example:
+	// legacyPackages.x86_64.go_1_19
+	_ /*nixpkgs*/, attrPath, found := strings.Cut(item.lockedReference, "#")
+	if !found {
+		return "", redact.Errorf(
+			"expected to find # in lockedReference: %s from NixProfileListItem: %s",
+			redact.Safe(item.lockedReference),
+			item,
+		)
+	}
+	return attrPath, nil
+}
+
+// ToPackage constructs a nix.Package using the unlocked reference.
+// TODO: this doesn't handle well the case where item.unlockedReference == "#"
+func (item *NixProfileListItem) ToPackage(locker lock.Locker) *devpkg.Package {
+	return devpkg.PackageFromString(item.unlockedReference, locker)
+}
+
+// String serializes the NixProfileListItem back into the format printed by `nix profile list`
+func (item *NixProfileListItem) String() string {
+	return fmt.Sprintf("%d %s %s %s",
+		item.index,
+		item.unlockedReference,
+		item.lockedReference,
+		item.nixStorePath,
+	)
+}
--- a/internal/nix/nixprofile/item.go
+++ b/internal/nix/nixprofile/item.go
@@ -25,7 +25,7 @@ type NixProfileListItem struct {
 	lockedReference string
 
 	// The store path(s) of the package.
-	nixStorePath []string
+	nixStorePath string // TODO: change to slice
 }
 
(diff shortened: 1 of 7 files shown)


internal/nix/nixprofile/profile.go
```diff
@@ -88,83 +93,75 @@ func ProfileListItems(
...
 	}
 	return items, nil
 }
 
 type ProfileListIndexArgs struct {
-	// For performance you can reuse the same list in multiple operations if you
+	// For performance, you can reuse the same list in multiple operations if you
 	// are confident index has not changed.
-	List       map[string]*NixProfileListItem
+	Items      []*NixProfileListItem
```
ipince, line 104: this is the main change

internal/nix/nixprofile/profile.go
```diff
@@ -88,83 +93,75 @@ func ProfileListItems(
...
-			return -1, err
+			return -1, errors.Wrapf(err, "failed to get installable for %s", args.DevPkg.String())
 		}
-		for _, item := range list {
+		for _, item := range items {
 			if pathInStore == item.nixStorePath {
 				return item.index, nil
 			}
 		}
+		return -1, errors.Wrap(nix.ErrPackageNotFound, args.DevPkg.String())
```
ipince, line 139: this is a somewhat subtle change that I thought I'd highlight.

If a package is inCache, then it *must* have been added by store path. So it follows that we *must* find a store path match in the items, and if not, then the package is Not Found--that is, there's no need to fall back to the other checks below.

Is this a safe assumption? I think so, because the cache doesn't change much and a package is either in it or not.. right? Please let me know if this isn't as safe an assumption as I think

internal/nix/nixprofile/item.go
```diff
@@ -0,0 +1,67 @@
...
+			item,
+		)
+	}
+	return attrPath, nil
+}
+
+// ToPackage constructs a nix.Package using the unlocked reference.
+// TODO: this doesn't handle well the case where item.unlockedReference == "#"
+func (item *NixProfileListItem) ToPackage(locker lock.Locker) *devpkg.Package {
+	return devpkg.PackageFromString(item.unlockedReference, locker)
```
savil, line 56: hmm this could be problematic for Items created via store-paths! 😬

savil: Seems to be used in:

1. https://github.com/jetpack-io/devbox/blob/main/internal/nix/nixprofile/profile.go#L144
This call is okay, because we have a separate if-condition that handles BinaryCache (i.e. store-paths) installs and is checked before this call.

2. https://github.com/jetpack-io/devbox/blob/main/internal/impl/packages.go#L481
This one in `extraPackagesInProfile` does feel problematic. Because the `item.UnlockedReference` here will be `#` for items-from-store-paths.

Hmm not sure how to fix (2).

savil: I'm re-thinking installing via store-path in the profile. Did we go down a wrong path by installing by store-paths in the nix profile?

Downside:
1. we cannot recover a `devpkg.Package` from these item-from-store-paths

Since we switched from content-addressed paths back to input-addressed-paths, I'm wondering if we should continue with the old way of installing packages in profiles. What do we gain (to remind me) from installing by store-path? In theory, it should be the same I believe.

WDYT?

internal/nix/nixprofile/profile.go
savil: I think that makes sense.

One scenario I was thinking about is (sequentially):
1. Existing devbox user has installed a package.
2. We ship a Devbox update that has Remove Nixpkgs feature enabled.
3. This code here runs. The Package happens to be in the binary cache, BUT it had previously been installed the "old way" in the nix profile.

I think this code should still work, since the store path is input-addressed and would match here.

internal/nix/nixprofile/item.go
savil: Okay, looking at the PR where I made install happen via nix store path: https://github.com/jetpack-io/devbox/pull/1255

the reason we must do it is to avoid downloading nixpkgs. Cool, we need it!

savil: Okay, to fix (2), I think we can do something like:
```
	extras := []*nixprofile.NixProfileListItem{}
	// Note: because nix.Input uses memoization when normalizing attribute paths (slow operation),
	// and since we're reusing the Input objects, this O(n*m) loop becomes O(n+m) wrt the slow operation.
outer:
	for _, item := range profileItems {
		profileInput := item.ToPackage(d.lockfile)
		for _, devboxInput := range devboxInputs {
		       // assuming devpkg.FillNarInfoCache() is already called
		       if devboxInput.IsInBinaryCache() {
		          for _, path := range item.storePaths {
		            if devboxInput.InputAddressedPath() == path {
		                // mark as present
		            }
		          }
		       }
		
			if profileInput.Equals(devboxInput) {
				continue outer
			}
		}
		extras = append(extras, item)
	}
```



WDYT?

savil (approved):
LGTM!

I think we do need a fix in `extraPackagesInProfile`. See my discussion below...

internal/nix/nixprofile/profile.go
ipince: I just tested a few different scenarios and verified this is safe. Most importantly, I tested:

1. User has installed a package in the "old" way. Then the user upgrades devbox and runs a devbox command that triggers the installation path. The package is now going to be installed by store-path. Because the store-paths match, then we do nothing. That is, we keep the "old-style" package installed in the profile, because we found it.  This is OK.

2. Same as above, but the package store path does NOT match. In this case, devbox will end up removing the old package and adding the new package, which is good.

internal/nix/nixprofile/item.go
ipince: While testing the upgrade path, I concluded that this is an important bug to fix (in separate PR). That's because `tidyProfile` will run the pairwise checks to match profile items with devbox packages every time the number of items in the profile differs from the number of items in devbox.json. And thus, every single time we run into this condition (which I think will be often), we'll end up _removing all store path-installed items from the profile_, because none of them will ever match with what's in devbox.json

PR #1417, opened 2023-08-23
Read all store paths from profile item

ipince:
instead of arbitrarily picking the first one

## How was it tested?
CI tests

### DIFF
--- a/internal/nix/nixprofile/profile.go
+++ b/internal/nix/nixprofile/profile.go
@@ -58,7 +58,7 @@ func ProfileListItems(
 			index:             index,
 			unlockedReference: element.OriginalURL + "#" + element.AttrPath,
 			lockedReference:   element.URL + "#" + element.AttrPath,
-			nixStorePath:      element.StorePaths[0],
+			nixStorePath:      element.StorePaths,
 		})
 	}
 	return items, nil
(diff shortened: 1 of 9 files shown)


internal/nix/nixprofile/profile.go
```diff
@@ -133,7 +133,8 @@ func ProfileListIndex(args *ProfileListIndexArgs) (int, error) {
 			return -1, errors.Wrapf(err, "failed to get installable for %s", args.DevPkg.String())
 		}
 		for _, item := range items {
-			if pathInStore == item.nixStorePath {
+			if len(item.nixStorePaths) == 1 && // this should always be true
+				pathInStore == item.nixStorePaths[0] {
```
savil, line 137: Thisshould be  true, yes. 

Should we assert that by erroring if len of item.nixStorePaths > 1
OR just iterate over the item.nixStorePaths so we don't mind if the assumption is violated and things keep working...

internal/nix/nixprofile/profile.go
```diff
@@ -191,15 +192,15 @@ func parseNixProfileListItem(line string) (*NixProfileListItem, error) {
 	lockedReference := scanner.Text()
 
 	if !scanner.Scan() {
-		return nil, redact.Errorf("error parsing \"nix profile list\" output: line is missing nixStorePath: %s", line)
+		return nil, redact.Errorf("error parsing \"nix profile list\" output: line is missing nixStorePaths: %s", line)
 	}
-	nixStorePath := scanner.Text()
+	nixStorePaths := strings.Fields(scanner.Text())
```
savil, line 197: ooh TIL about `strings.Fields`

savil (approved):
Thanks!

PR #1418, opened 2023-08-25
Add missing READMEs to templates, fix lint

Lagoja:
## Summary

Adds READMEs to each example, so they show up in Templates
   Content should match what's in the Docs
Lints some of the markdown so it passes CI

## How was it tested?
 Localhost


LucilleH (approved):
Whoa!

PR #1421, opened 2023-08-28
[dockerhub] Added action for publishing docker image

mohsenari:
## Summary
added an action to build and publish an image containing devbox and nix installations to dockerhub

## How was it tested?
Not tested yet


.github/workflows/docker-image-release.yml
```diff
@@ -0,0 +1,27 @@
+name: docker-image-release
+
+on: workflow_dispatch
```
LucilleH, line 3: You can test this in this PR with
```
on:
    push:
```
And if it works, change it back to workflow dispatch

.github/workflows/docker-image-release.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          context: ./examples/docker/
+          file: ./examples/docker/Dockerfile
```
LucilleH, line 25: I wonder if you can simply run `devbox generate dockerfile` command here, instead of pushing a hardcoded Dockerfile (so that it does not get out of sync)

LucilleH (approved):

PR #1422, opened 2023-08-28
Fix bug when comparing profile item added by store path

ipince:
## Summary
When an package is added  by store path, we can no longer recover the original package from the profile item itself, because the only information we have is the nix store path. So, some comparisons where we were trying to find if a profile item matched a devbox package were broken.

Here we introduce a `Item.Matches(Package)` function to return true if the item was installed from the package. It attempts to hide the complexity of whether the item was installed by store path or not inside it, so callers don't have to worry about it. Modified the callsites to `ToPackage()` to use `Matches` instead, which simplified some logic.

Also added a `len == 0` check to calls to `nix profile remove`, since I noticed we were sometimes calling it with no arguments. No failures here, but just wasted time.

## How was it tested?
Manually tested by adding/removing packages and logging all the profile-related calls and item/package comparisons.


savil (approved):
nice!

Issue #1420, opened 2023-08-28
Programs installed with `devbox global` print an update message


gcurtis:
**Current Behavior (bug)**

```
$ go                   
Info: New devbox available: 0.5.10 -> v0.5.11. Please run `devbox version update`.
Go is a tool for managing Go source code.
$ zsh -c "echo hi"
Info: New devbox available: 0.5.10 -> v0.5.11. Please run `devbox version update`.
hi
```

**Expected Behavior (fix)**

Devbox wrappers never print anything.

mohsenari:
my guess is that the wrapper for these binaries run a devbox command and that command runs a version check. Or the devbox command we have in zshrc/bashrc runs a version check.

savil:
fixed

PR #1424, opened 2023-08-28
[bug fix] do not check version for devbox log, and for commands whose prefix is in the skip-list

savil:
## Summary

In the generated shellrc file, we invoke `devbox log` to track shell start times.
```
> git grep "devbox log"
internal/impl/shellrc.tmpl:devbox log shell-ready {{ .ShellStartTime }}
internal/impl/shellrc.tmpl:devbox log shell-interactive {{ .ShellStartTime }}
internal/impl/shellrc_fish.tmpl:devbox log shell-ready {{ .ShellStartTime }}
internal/impl/shellrc_fish.tmpl:devbox log shell-interactive {{ .ShellStartTime }}
```

We should skip checking if devbox version is up-to-date for this case.

Also, made a change to check if the commandPath matches the prefix of the commandSkipList items. This will #1420 

## How was it tested?

Did not test. It compiles.

### DIFF
--- a/internal/vercheck/vercheck.go
+++ b/internal/vercheck/vercheck.go
@@ -69,7 +69,11 @@ func CheckVersion(w io.Writer, commandPath string) {
 		return
 	}
 
-	if lo.Contains(commandSkipList, commandPath) {
+	hasSkipPrefix := lo.ContainsBy(
+		commandSkipList,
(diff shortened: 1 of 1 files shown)


mohsenari (approved):
thank you

PR #1435, opened 2023-08-29
Use `envsubst` in our NGINX plugin

Lagoja:
## Summary
NGINX does not support using environment variables in it's configuration file, which makes it difficult for our plugin to configure NGINX to run in a local devbox shell.

This update to the NGINX plugin installs envsubst, and then uses it to generate a valid `nginx.conf` from an `nginx.template` file. Developers can further customize their nginx.conf by adding environment variables to the `nginx.template` file. 

Envsubst will generate the nginx.conf whenever `devbox services up` or `devbox service start` is run. 

## How was it tested?

Using the nginx example:

1. Change the NGINX_WEB_PORT variable
2. Run `devbox services up`
3. NGINX should generate a valid nginx.conf in the `devbox.d/nginx` folder, and run on the NGINX_WEB_PORT

### DIFF
--- a/examples/servers/nginx/devbox.d/nginx/nginx.template
+++ b/examples/servers/nginx/devbox.d/nginx/nginx.template
@@ -1,10 +1,10 @@
 events {}
 http{
 server {
-         listen       $NGINX_WEB_PORT;
-         listen       [::]:$NGINX_WEB_PORT;
-         server_name  localhost;
-         root         ../../../devbox.d/web;
+         listen       $WEB_PORT;
+         listen       [::]:$WEB_PORT;
+         server_name  $WEB_SERVER_NAME;
+         root         $WEB_ROOT;
 
          error_log error.log error;
          access_log access.log;
--- a/examples/servers/nginx/devbox.d/nginx/nginx.template
+++ b/examples/servers/nginx/devbox.d/nginx/nginx.template
@@ -1,0 +1,20 @@
+events {}
+http{
+server {
+         listen       $NGINX_WEB_PORT;
+         listen       [::]:$NGINX_WEB_PORT;
+         server_name  localhost;
+         root         ../../../devbox.d/web;
+
+         error_log error.log error;
+         access_log access.log;
+         client_body_temp_path temp/client_body;
+         proxy_temp_path temp/proxy;
+         fastcgi_temp_path temp/fastcgi;
+         uwsgi_temp_path temp/uwsgi;
+         scgi_temp_path temp/scgi;
+
+         index index.html;
+         server_tokens off;
+    }
+}
--- a/examples/servers/nginx/devbox.d/nginx/nginx.template
+++ b/examples/servers/nginx/devbox.d/nginx/nginx.template
@@ -1,10 +1,10 @@
 events {}
 http{
 server {
-         listen       $WEB_PORT;
-         listen       [::]:$WEB_PORT;
-         server_name  $WEB_SERVER_NAME;
-         root         $WEB_ROOT;
+         listen       $NGINX_WEB_PORT;
+         listen       [::]:$NGINX_WEB_PORT;
+         server_name  $NGINX_WEB_SERVER_NAME;
+         root         $NGINX_WEB_ROOT;
 
          error_log error.log error;
          access_log access.log;
--- a/plugins/nginx.json
+++ b/plugins/nginx.json
@@ -1,7 +1,7 @@
 {
   "name": "nginx",
   "version": "0.0.3",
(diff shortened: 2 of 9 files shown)


plugins/nginx.json
```diff
@@ -1,15 +1,19 @@
 {
   "name": "nginx",
-  "version": "0.0.2",
-  "readme": "nginx can be configured with env variables\n\nTo customize:\n* Use $NGINX_CONFDIR to change the configuration directory\n* Use $NGINX_LOGDIR to change the log directory\n* Use $NGINX_PIDDIR to change the pid directory\n* Use $NGINX_RUNDIR to change the run directory\n* Use $NGINX_SITESDIR to change the sites directory\n* Use $NGINX_TMPDIR to change the tmp directory. Use $NGINX_USER to change the user\n* Use $NGINX_GROUP to customize.",
+  "version": "0.0.3",
+  "readme": "nginx can be configured with env variables\n\nTo customize:\n* Use $NGINX_CONFDIR to change the configuration directory\n* Use $NGINX_TMPDIR to change the tmp directory. Use $NGINX_USER to change the user\n* Use $NGINX_WEB_PORT to change the port NGINX runs on. \n Note: This plugin uses envsubst when running `devbox services` to generate the nginx.conf file from the nginx.template file. To customize the nginx.conf file, edit the nginx.template file.\n",
+  "__packages": ["envsubst"],
```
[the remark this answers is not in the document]
Lagoja: @mikeland73 do you know if this bypasses the search index? I wasn't able to add `@latest` when including the package

examples/servers/nginx/devbox.d/nginx/nginx.template
```diff
@@ -0,0 +1,20 @@
+events {}
+http{
+server {
+         listen       $NGINX_WEB_PORT;
+         listen       [::]:$NGINX_WEB_PORT;
+         server_name  localhost;
+         root         ../../../devbox.d/web;
```
jrseliga, line 7: Should `server_name` and `root` be configurable through environment variables too?

Lagoja: Adding variables for both of these

Lagoja: Added the following to the plugin: 

`WEB_PORT`
`WEB_SERVER_NAME`
`WEB_ROOT`

These are all set by the plugin, and can be overridden using the `.env` support or via the `env` section in the devbox.json. `devbox services up|start|restart` or will automatically regenerate the `nginx.conf` using: 

```
envsubst < $NGINX_CONFDIR/nginx.template > $NGINX_CONFDIR/nginx.conf
```

If for whatever reason you want to lock the `nginx.conf`, you can delete the `nginx.template` file.

LucilleH (approved):

plugins/nginx.json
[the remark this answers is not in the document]
mikeland73: I believe this will be installed as a legacy package. We could do `envsubst@latest` ? or pin the version?

plugins/nginx.json
```diff
@@ -1,15 +1,21 @@
...
   "env": {
-    "NGINX_CONFDIR": "{{ .DevboxDir }}/nginx.conf",
+    "NGINX_CONF": "{{ .DevboxDir }}/nginx.conf",
+    "NGINX_CONFDIR": "{{ .DevboxDir }}",
     "NGINX_PATH_PREFIX": "{{ .Virtenv }}",
-    "NGINX_TMPDIR": "{{ .Virtenv }}/temp"
+    "NGINX_TMPDIR": "{{ .Virtenv }}/temp",
+    "WEB_PORT": "8081",
+    "WEB_ROOT": "../../../devbox.d/web",
+    "WEB_SERVER_NAME": "localhost"
```
mikeland73, line 13: Should these be namespaced?

mikeland73 (commented):
Looks good. My only comment is whether the env vars should be namespaced with `NGINX`

Currently plugin environment variables are added to the entire environment. That means that if you want to run multiple custom servers we would probably want env variables to have different names. We have talked about isolating plugin environments but that's difficult to do and if we ever do that we would probably need to make it opt-in to keep backward-compatibility.

plugins/nginx.json
Lagoja: I think this would be relatively safe to not namespace them -- I guess the conflict would arise if you are running multiple servers (NGINX/Apache/Caddy) in the same shell, which seems a bit unlikely?

Lagoja: I'll namespace them for now until we decide on this

PR #1437, opened 2023-08-31
bug fix: FillNarInfoCache for versioned packages when adding package

savil:
## Summary

The call to `versionedPkg.ValidateExists()` would error because the NarInfoCache
was not pre-filled before we called `IsInBinaryCache` in it.

Now, we call FillNarInfoCache on the versioned packages as well. 

Did some minor refactoring so that we only do FillNarInfoCache for packages that
are to-be-added (thereby omitting existing packages) to be just a bit efficient.

## How was it tested?

```
export DEVBOX_FEATURE_REMOVE_NIXPKGS=1

devbox add vim
# BEFORE: this would add `vim` to devbox.json
# AFTER: this adds `vim@latest` to devbox.json
```

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -53,12 +53,28 @@ func (d *Devbox) Add(ctx context.Context, platforms, excludePlatforms []string, 
 	// to know the exact name to mark as allowed insecure later on.
 	addedPackageNames := []string{}
 	existingPackageNames := d.PackageNames()
+	newPackages := []*devpkg.Package{}
 	for _, pkg := range pkgs {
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #1437** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1437" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1437?utm_source=stack-comment).

ipince (approved):

Issue #1431, opened 2023-08-29
[Package Issue Report]: prometheus does not provide promtool


wrdls:
**If reporting an issue on an existing one, what is the name of the package?**

`prometheus`

**What changes are you requesting?**

Installing the `prometheus`  package should also provide the `promtool` binary.

This seems to be defined as a postinstall action (as far as my nix knowledge goes): https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/servers/monitoring/prometheus/default.nix#L119C19-L121

I tried Installing with `nix-env` as shown on the [nixos website](https://search.nixos.org/packages?channel=23.05&show=prometheus&from=0&size=50&sort=relevance&type=packages&query=promtool) to see if that would work and it does seem to give me the appropriate binaries:

```
$ nix-env -iA nixpkgs.prometheus
installing 'prometheus-2.38.0'
these 4 paths will be fetched (65.92 MiB download, 206.48 MiB unpacked):
  /nix/store/2ax1jq6jhqh826l62zz8sc1qrm8bsbx3-iana-etc-20220915
  /nix/store/fsxr40h41zb9s2g8rhz7b88fccz951za-mailcap-2.1.53
  /nix/store/msj01lbvcqwx62z7rz7bm5wwv5y0z622-prometheus-2.38.0
  /nix/store/yy7gb4g266l61kmwmc57q6jp7ls5yqvc-tzdata-2022e
copying path '/nix/store/2ax1jq6jhqh826l62zz8sc1qrm8bsbx3-iana-etc-20220915' from 'https://cache.nixos.org'...
copying path '/nix/store/fsxr40h41zb9s2g8rhz7b88fccz951za-mailcap-2.1.53' from 'https://cache.nixos.org'...
copying path '/nix/store/yy7gb4g266l61kmwmc57q6jp7ls5yqvc-tzdata-2022e' from 'https://cache.nixos.org'...
copying path '/nix/store/msj01lbvcqwx62z7rz7bm5wwv5y0z622-prometheus-2.38.0' from 'https://cache.nixos.org'...
building '/nix/store/mp31nc3pz9grpd7qc74bhwmbdcilzqd6-user-environment.drv'...

$ ls -l /nix/var/nix/profiles/per-user/wardloos/profile/bin/
total 462336
-r-xr-xr-x  1 root  nixbld  110703424 Jan  1  1970 prometheus
[...]

PR #1442, opened 2023-09-01
[rm nixpkgs] use path-from-hash-part command when updating lock file

savil:
## Summary

**Motivation**
With Remove Nixpkgs feature on, installing `curl@8.0` was failing with:
```
[3/7] curl@8.0
        don't know how to build these paths:
          /nix/store/0xyi7w3cm3g7gzwfpwqzx0n7xks2sdc6-curl-8.0.1
        asked 'https://cache.nixos.org/' for '/nix/store/0xyi7w3cm3g7gzwfpwqzx0n7xks2sdc6-curl-8.0.1' but got '/nix/store/0xyi7w3cm3g7gzwfpwqzx0n7xks2sdc6-curl-8.0.1-bin'
        error: path '/nix/store/0xyi7w3cm3g7gzwfpwqzx0n7xks2sdc6-curl-8.0.1' is required, but there is no substituter that can build it
```

**Explanation**
Thanks to @gcurtis.
curl’s first (and therefore default) output is `bin`, not `out`, so Nix is appending `-bin` to the path.

From `man nix-build`:
> If a derivation has multiple outputs, nix-build will build the default (first) output. You can also build all outputs:
$ nix-build '<nixpkgs>' --attr openssl.all
This will create a symlink for each output named result-outputname. The suffix is omitted if the output name is out. So if openssl has outputs out, bin and man, nix-build will create symlinks result, result-bin and result-man.

**Fix**
Observe:
```
> nix store path-from-hash-part --store https://cache.nixos.org/ bbkz21vmx16wl0h3xk36g7gxpbymnm6d --extra-experimental-features "nix-command flakes"
/nix/store/bbkz21vmx16wl0h3xk36g7gxpbymnm6d-curl-8.0.1-bin
```

So:
- During `devbox update`, we derive the store-path from the search API and write to `devbox.lock`
- Previously, we would manually construct store-path from `hash-name-version` from search API response
- Now, we invoke `nix store path-from-hash-part <hash>` and write the response store-path in the `devbox.lock`.
  - if this fails (it shouldn't!), we skip writing the store-path to the `devbox.lock`, and print a warning. Maybe we should track it in sentry?
  

## How was it tested?

```
> cat devbox.json
{
  "packages": [
    "curl@8.0"
  ]
}

> devbox update
Info: Updated system information for curl@8.0

Installing package: curl@8.0.

[1/1] curl@8.0
[1/1] curl@8.0: Success
Info: Running "nix flake update"

> cat devbox.lock
{
  "lockfile_version": "1",
  "packages": {
    "curl@8.0": {
      "last_modified": "2023-05-31T02:09:55Z",
      "resolved": "github:NixOS/nixpkgs/9cfaa8a1a00830d17487cb60a19bb86f96f09b27#curl",
      "source": "devbox-search",
      "version": "8.0.1",
      "systems": {
        "aarch64-darwin": {
          "store_path": "/nix/store/7qwsslwdrjyg8hs654nsqpq7lly6dhq5-curl-8.0.1-bin"
        },
        "aarch64-linux": {
          "store_path": "/nix/store/g9g1p4l9146kclnx4lv0hfbzhr9hk7n7-curl-8.0.1-bin"
        }
      }
    }
  }
}
```

### DIFF
--- a/internal/lock/resolve.go
+++ b/internal/lock/resolve.go
@@ -4,16 +4,22 @@
 package lock
 
 import (
+	"context"
 	"fmt"
+	"os"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/searcher"
+	"go.jetpack.io/devbox/internal/ux"
 	"golang.org/x/exp/maps"
+	"golang.org/x/sync/errgroup"
 )
 
 // FetchResolvedPackage fetches a resolution but does not write it to the lock
@@ -35,7 +41,10 @@ func (f *File) FetchResolvedPackage(pkg string) (*Package, error) {
 
 	sysInfos := map[string]*SystemInfo{}
 	if featureflag.RemoveNixpkgs.Enabled() {
-		sysInfos = buildLockSystemInfos(packageVersion)
+		sysInfos, err = buildLockSystemInfos(packageVersion)
+		if err != nil {
+			return nil, err
+		}
 	}
 	packageInfo, err := selectForSystem(packageVersion)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #1442** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1442" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1442?utm_source=stack-comment).

savil:
@gcurtis instead of running this logic in the CLI, I'm now thinking it may make sense for the search indexer to run it. WDYT?

gcurtis:
This is something we should add to the API, but feel free to keep it as a workaround until the API is updated.

I plan on adding a new `outputs` field to the JSON alongside the current `store_hash/name/version` fields. The new response will look like:

```json
{
  "store_hash": "00263n102bi1r6qihvdghxb3jx6n3k1d",
  "store_name": "libcxxabi",
  "store_version": "16.0.6",
  "outputs": [
    {
      "path": "/nix/store/00263n102bi1r6qihvdghxb3jx6n3k1d-libcxxabi-16.0.6",
      "name": "out",
      "hash": "00263n102bi1r6qihvdghxb3jx6n3k1d",
      "default": true
    },
    {
      "path": "/nix/store/04r1bxcai0bzb2z01n5r47j8r2387s2r-libcxxabi-16.0.6-dev",
      "name": "dev",
      "hash": "04r1bxcai0bzb2z01n5r47j8r2387s2r",
      "default": false
    }
  ]
}
```

The `store_hash` field will be kept for compatibility, but new code will use the `outputs` array instead. Would this new format get you what you're looking for?

savil:
@gcurtis That would be neat. A few questions:

1. I understand that `default` would refer to the first output, as per `nix build` docs. Does that _always_ necessarily imply its also the output cached by `cache.nixos.org`? Will you confirm that server-side, or do I need to keep a check similar to the one in this PR in the CLI?

2. For packages with multiple outputs (ref. #1431 where prometheus has `cli` output as well), your proposal would be really helpful. However, I'm guessing the `cache.nixos.org` will not have a prebuilt binary for the `cli` output, since its not the default output. Let me know if I'm incorrect. I'm thinking we may need the user to specify outputs they'd like. Something like:
```
devbox add prometheus --outputs default,cli
```
which are set as:
```
// devbox.json
{
    "prometheus: {
        "version": "latest",
        "outputs": [default, cli],
    },
}
```
And unfortunately then we'd need to install this `cli` using the nixpkgs-full path.
Does that sounds good, directionally at least?

savil:
Lets land this PR then so we can start dogfooding internally!

gcurtis (approved):

gcurtis:
> 1. I understand that `default` would refer to the first output, as per `nix build` docs. Does that _always_ necessarily imply its also the output cached by `cache.nixos.org`? Will you confirm that server-side, or do I need to keep a check similar to the one in this PR in the CLI?

Default isn't always just the first output. Some packages define multiple outputs to install by default. That's a good point about the cache check. I think the best approach would be to add a `cached` field to each output in the response.

> However, I'm guessing the `cache.nixos.org` will not have a prebuilt binary for the `cli` output, since its not the default output.

I had to look this one up to double check. It looks like all of the prometheus outputs are cached, including `cli`. I'm not sure what the best way is for installing non-default outputs. Off the top of my head, we could either:

- Do your suggestion - add an outputs field to `devbox.json`. I'm not a huge fan of adding yet another field though.
	- Alternatively, we just work it into the name like `prometheus-cli@latest` or `curl-lib@latest`.
- Install all outputs by default (`nixpkgs#prometheus.all`). Downside is this will probably really bloat installs.
- Special-case outputs to always install. In other words, always install `out`, `bin`, `man`, and `cli`.

PR #1448, opened 2023-09-01
[Github Action] cli-test: improve naming summary

savil:
## Summary

**Motivation**
The name summary for `cli-tests` in the Github PR page is:
```
test (false, ubuntu-latest, true, 2.12.0)
```
This isn't very helpful.

This PR:
- removes the `booleans` and changes their true/false states to explicit names
- renames `devbox-json-tests` to `project-tests` for brevity.

## How was it tested?

will observe the testscripts pass...

Now, it looks like:
```
cli-tests / test (not-main, ubuntu-latest, project-tests, 2.12.0)
```
which is much more understandable.

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -89,7 +89,7 @@ jobs:
           - is-main: "not-main"
             os: "${{ inputs.run-mac-tests && 'dummy' || 'macos-latest' }}"
           - is-main: "is-main"
-            run-project-tests: "project-tests"
+            run-project-tests: "project-tests-off"
           - run-project-tests: "project-tests"
             os: macos-latest
 
@@ -130,7 +130,7 @@ jobs:
         env:
           # For devbox.json tests, we default to non-debug mode since the debug output is less useful than for unit testscripts.
           # But we allow overriding via inputs.example-debug
-          DEVBOX_DEBUG: ${{ (matrix.run-project-tests == 'not-project-tests' || inputs.example-debug) && '1' || '0' }}
+          DEVBOX_DEBUG: ${{ (matrix.run-project-tests == 'project-tests-off' || inputs.example-debug) && '1' || '0' }}
           DEVBOX_RUN_PROJECT_TESTS: ${{ matrix.run-project-tests == 'project-tests' && '1' || '0' }}
           # Used in `go test -timeout` flag. Needs a value that time.ParseDuration can parse.
           DEVBOX_GOLANG_TEST_TIMEOUT: "${{ (github.ref == 'refs/heads/main' || inputs.run-mac-tests) && '35m' || '30m' }}"
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -19,8 +19,6 @@ on:
         type: boolean
   workflow_dispatch:
     inputs:
-      run-devbox-json-tests:
-        type: boolean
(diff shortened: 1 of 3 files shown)


LucilleH (approved):

.github/workflows/cli-tests.yaml
```diff
@@ -76,24 +74,25 @@ jobs:
...
         nix-version: ["2.12.0", "2.17.0"]
         exclude:
-          - is-main: false
+          - is-main: "not-main"
             os: "${{ inputs.run-mac-tests && 'dummy' || 'macos-latest' }}"
-          - is-main: true
-            run-devbox-json-tests: false
-          - run-devbox-json-tests: true
+          - is-main: "is-main"
+            run-project-tests: "project-tests"
```
ipince, line 92: shouldn't this be `project-tests-off`?

.github/workflows/cli-tests.yaml
```diff
@@ -131,8 +130,8 @@ jobs:
         env:
           # For devbox.json tests, we default to non-debug mode since the debug output is less useful than for unit testscripts.
           # But we allow overriding via inputs.example-debug
-          DEVBOX_DEBUG: ${{ (!matrix.run-devbox-json-tests || inputs.example-debug) && '1' || '0' }}
-          DEVBOX_RUN_DEVBOX_JSON_TESTS: ${{ matrix.run-devbox-json-tests }}
+          DEVBOX_DEBUG: ${{ (matrix.run-project-tests == 'not-project-tests' || inputs.example-debug) && '1' || '0' }}
```
ipince, line 133: shouldn't this be `project-tests-off`?

ipince (approved):
Approving to unblock, but see comments

.github/workflows/cli-tests.yaml
```diff
@@ -76,24 +74,25 @@ jobs:
...
-        run-devbox-json-tests: [true, false]
+        run-project-tests: ["project-tests", "project-tests-off"]
         # Run tests on:
         # 1. the oldest supported nix version (which is 2.9.0? But determinate-systems installer has 2.12.0)
         # 2. nix version 2.17.0 which introduces a new code path that minimizes nixpkgs downloads.
         # 3. latest nix version (currently, that is 2.17.0, so omitted)
         nix-version: ["2.12.0", "2.17.0"]
         exclude:
-          - is-main: false
+          - is-main: "not-main"
```
mikeland73, line 89: I agree that improving the labels is good, but lines like this make my head spin :(

mikeland73 (commented):
I totally agree the labels suck, but I really have trouble following our rules (Many of which I wrote myself - lol)

Alternative:

* Create two different workflows that both call a job in a shared workflow.
* The shared workflow takes inputs that decide what to run (what OS, and what scope of tests).


So the jobs in non-main can be named:
* Basic tests on mac
* All tests on linux

On main:
* All tests on (mac|linux)

(under the hood, each of these is just calling some shared jobs in another workflow which is never triggered directly).

This should really simplify names and also remove much of the exclude logic which is hard to grok.

.github/workflows/cli-tests.yaml
savil: good catch!

savil (commented):
@mikeland73 some refactoring like that seems called for. I'll circle back to doing that in a bit.

Will land this for now, so at least we get rid of boolean labels.

Issue #1444, opened 2023-09-01
[Bug]: Cannot build devbox dockerfile on arm64


eljohnson92:
**Current Behavior (bug)**
When trying to build a multi-platform image with devbox and nix there is a eBPF issue causing the dockerfile to fail to build. I have been able to reproduce this by running the base Devbox rootUser dockerfile in this repo. This does not seem to happen if I use the non-root Dockerfile
```bash
docker buildx build  --platform linux/amd64,linux/arm64  . -f internal/impl/generate/tmpl/DevboxImageDockerfileRootUser
```
The exact error can be seen here:
```
51.36 ~~> Setting up shell profiles for Fish with conf.d/nix.fish inside /etc/fish /usr/local/etc/fish /opt/homebrew/etc/fish /opt/local/etc/fish
51.36
51.42 ~~> Setting up the default profile
51.42
51.48 ---- sudo execution ------------------------------------------------------------
51.48 I am executing:
51.48
51.48     $ sudo HOME=/root /nix/store/9670sxa916xmv8n1kqs7cdvmnsrhrdjv-nix-2.17.0/bin/nix-env -i /nix/store/9670sxa916xmv8n1kqs7cdvmnsrhrdjv-nix-2.17.0
51.48
51.48 to install a bootstrapping Nix in to the default profile
51.48
52.00 installing 'nix-2.17.0'
52.25 error:
52.25        … while setting up the build environment
52.25
52.25        error: unable to load seccomp BPF program: Invalid argument
52.38
52.38 ---- oh no! --------------------------------------------------------------------
52.38 Oh no, something went wrong. If you can take all the output and open
52.42 an issue, we'd love to fix the problem so nobody else has this issue.
52.42
52.42 :(
52.42
52.43 We'd love to help if you need it.
52.43
52.43 You can open an issue at
52.43 https://github.com/NixOS/nix/issues/new?labels=installer&template=installer.md
52.43
52.43 Or get in touch with the community: https://nixos.org/community
52.43
```
**Expected Behavior (fix)**
Devbox Docker images should build cross platform

**Additional context**
I have currently tried building this on an amd64 Macbook and ubuntu VM, this might work if built on an arm64 macbook

savil:
@mohsenari IIRC you'd implemented the Dockerfile root user functionality. Any leads on this?

dimaqq:
I wonder if you managed to solve this for users on a Mac building this image locally?

eljohnson92:
[...]

PR #1455, opened 2023-09-07
Update docker-image-release.yml to support ARM builds for newer Macbooks

eljohnson92:
## Summary
Add arm64 builds to the non-root image build(currently the root image cannot be built because of #1444) 
## How was it tested?
I have tested this locally and confirmed the non-root image is built successfully with this change

### DIFF
--- a/.github/workflows/docker-image-release.yml
+++ b/.github/workflows/docker-image-release.yml
@@ -8,6 +8,8 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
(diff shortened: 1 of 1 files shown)


mohsenari (approved):
Thanks for looking into this.

mohsenari:
@eljohnson92 I'm going to merge this and publish a new docker image to docker hub. Once done I'll ping you here. Please test it and see if it works for you.

mohsenari:
@eljohnson92 I just published the [jetpackio/devbox](https://hub.docker.com/r/jetpackio/devbox/tags) image in dockerhub. It now shows support for both linux/amd64 and linux/arm64 support. Please give it a try and let me know if your issue is resolved.

PR #1473, opened 2023-09-12
[narInfoCache] Use sync.Map and simplify IsInBinaryCache

mikeland73:
## Summary

This change simplifies code, eliminates manual locks. It:

* uses sync.Map
* Changes `fillNarInfoCache` to be `fillNarInfoCacheIfNeeded`
* Simplifies `IsInBinaryCache`
* Simplifies goroutine in `FillNarInfoCache` because we no longer have to check for values that are already set and we don't have to handle error.

## How was it tested?

`devbox run build`


mikeland73:
Ug nevermind, concurrent reads/writes will actually cause a panic.

internal/devpkg/narinfo_cache.go
```diff
@@ -125,17 +102,21 @@ func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
...
 // assuming it is eligible for the binary cache. Callers are responsible
 // for checking isEligibleForBinaryCache before calling this function.
 //
 // NOTE: this must be concurrency safe.
-func (p *Package) fillNarInfoCache() error {
+func (p *Package) fillNarInfoCacheIfNeeded() (bool, error) {
+	// This check if fine to do without a lock, because we never remove/replace values
+	if status, alreadySet := isNarInfoInCache.status[p.Raw]; alreadySet {
+		return status, nil
+	}
```
savil, line 114: I don't think this is correct:

https://go.dev/doc/faq#atomic_maps
> Map access is unsafe only when updates are occurring. As long as all goroutines are only reading—looking up elements in the map, including iterating through it using a for range loop—and not changing the map by _assigning to elements_  or doing deletions, it is safe for them to access the map concurrently without synchronization.

https://go.dev/blog/maps
> [Maps are not safe for concurrent use](https://go.dev/doc/faq#atomic_maps): it’s not defined what happens when you read and write to them simultaneously. If you need to read from and write to a map from concurrently executing goroutines, the accesses must be mediated by some kind of synchronization mechanism. One common way to protect maps is with [sync.RWMutex](https://go.dev/pkg/sync/#RWMutex).

savil: It also makes sense to me that its not safe, because writing to a map can involve needing to allocate more memory and potentially move elements around...

internal/devpkg/narinfo_cache.go
```diff
@@ -103,39 +72,30 @@ func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
...
 // assuming it is eligible for the binary cache. Callers are responsible
 // for checking isEligibleForBinaryCache before calling this function.
 //
 // NOTE: this must be concurrency safe.
-func (p *Package) fillNarInfoCache() error {
+func (p *Package) fillNarInfoCacheIfNeeded() (bool, error) {
+	if status, alreadySet := isNarInfoInCache.Load(p.Raw); alreadySet {
+		return status.(bool), nil
+	}
+	isNarInfoInCache.Store(p.Raw, false) // store in case of failure
```
gcurtis, line 93: There could be a race here where multiple goroutines see `alreadySet == false` before reaching line 93. Or one goroutine might reach line 93 and then another sees `status.(bool) == false` on line 91.

gcurtis (commented):
If the cache is just a global variable, using a map of `sync.Once`s might simplify things.

(I haven't tested any of this)

```go
var cache sync.Map

func isInBinaryCache(key string) (bool, error) {
    type inCacheFunc func() (bool, error)
    f, ok := cache.Load(key)
    if !ok {
        f = inCacheFunc(sync.OnceValues(func() (bool, error) { /* check cache */ }))
        f, _ = cache.LoadOrStore(key, f)
    }
    return f.(inCacheFunc)()
```

internal/devpkg/narinfo_cache.go
mikeland73: Good call. I think our calling patterns make this unlikely but better to avoid if possible.

I think using `defer` might solve most race conditions. It's not immune to a success following a failure (but neither is the current code so it would not be a regression)

internal/devpkg/narinfo_cache.go
```diff
@@ -103,39 +72,34 @@ func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
...
 // NOTE: this must be concurrency safe.
-func (p *Package) fillNarInfoCache() error {
+func (p *Package) fillNarInfoCacheIfNeeded() (bool, error) {
+	if status, alreadySet := isNarInfoInCache.Load(p.Raw); alreadySet {
+		return status.(bool), nil
+	}
+	defer func() {
+		// If there's an error and nothing got cached, just store false.
+		_, _ = isNarInfoInCache.LoadOrStore(p.Raw, false)
+	}()
```
mikeland73, line 96: @gcurtis @savil using defer here is functionally identical to what was previously handled in line 119 except now it is handled 100% of the time. 

This in conjunction with `LoadOrStore` in line 125 guarantee that the first result is preserved forever (so a race condition cannot replace an existing value). This fixes an existing race condition where two calls to `fillNarInfoCacheIfNeeded` could override each other if they had different results.

This fix also removes the race condition where an early `false` causes another goroutine to return `false` early.

It partially* fixes multiple goroutines trying to fetch the same result at the same time. In that case only the first result is saved and other goroutines return the same result. Previously, each call to `fillNarInfoCacheIfNeeded` could return a different result.

* I say "partially" because ideally we don't make 2 http requests, but I think that's not a big deal so unless the fix is easy, I would avoid complicating the code.

savil: > This fix also removes the race condition where an early false causes another goroutine to return false early.

I don't quite understand this scenario. Care to elaborate?

> I say "partially" because ideally we don't make 2 http requests, but I think that's not a big deal so unless the fix is easy, I would avoid complicating the code.

Did you try @gcurtis 's suggestion of using `sync.Once`s? I think that would address this.

savil (approved):
@mikeland73 thanks for simplifying and hardening this code!

mikeland73:
>> This fix also removes the race condition where an early false causes another goroutine to return false early.

> I don't quite understand this scenario. Care to elaborate?

This was in reference to the first version of this PR (and @gcurtis's comment). In the first version if `fillNarInfoCacheIfNeeded` got called twice simultaneously, it's possbie the first one would initialize the status to false, and the second one would return false right away, even thought he ultimate value would be true. Using `defer` which does the cleanup at the end eliminates this case.

> Did you try @gcurtis 's suggestion of using sync.Onces? I think that would address this.

I think his solution is better because it guarantees only one call per package. I'll hold off on merging and see if I can get it to work.

internal/devpkg/narinfo_cache.go
```diff
@@ -103,39 +65,44 @@ func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
...
+// once and cache the result.
+func (p *Package) fetchNarInfoStatusOnce() (bool, error) {
+	type inCacheFunc func() (bool, error)
+	f, ok := narInfoStatusFnCache.Load(p.Raw)
+	if !ok {
+		f = inCacheFunc(sync.OnceValues(func() (bool, error) {
+			return p.fetchNarInfoStatus()
+		}))
+		f, _ = narInfoStatusFnCache.LoadOrStore(p.Raw, f)
+	}
```
mikeland73, line 92: We could replace this with a single LoadOrStore but that would create throwaway OnceValues function every time.

mikeland73:
@gcurtis implemented your suggestion. PTAL

internal/devpkg/narinfo_cache.go
```diff
@@ -103,39 +65,44 @@ func FillNarInfoCache(ctx context.Context, packages ...*Package) error {
...
+
+// fetchNarInfoStatusOnce is like fetchNarInfoStatus, but will only ever run
+// once and cache the result.
+func (p *Package) fetchNarInfoStatusOnce() (bool, error) {
+	type inCacheFunc func() (bool, error)
+	f, ok := narInfoStatusFnCache.Load(p.Raw)
+	if !ok {
+		f = inCacheFunc(sync.OnceValues(func() (bool, error) {
+			return p.fetchNarInfoStatus()
+		}))
```
gcurtis, line 90:
```suggestion
		f = inCacheFunc(sync.OnceValues(p.fetchNarInfoStatus))
```

gcurtis (approved):

PR #1474, opened 2023-09-13
[envsec] Initial envsec integration

mikeland73:
## Summary

This pre-mvp integration is mostly just to get something working. It's missing a lot of stuff. Specifically:

* envsec auto-install
* error handling (this requires new envsec `--json` output so we don't have to do clever parsing). 
* Performance. Currently `envsec.Env` gets called a few times and not cached.

One open question is how generic we want to make the integration. This implementation is pretty coupled. 

## How was it tested?

```bash
envsec set FOO=BAR
devbox run echo \$FOO
```

### DIFF
--- a/internal/integrations/envsec/envsec.go
+++ b/internal/integrations/envsec/envsec.go
@@ -1,0 +1,63 @@
+package envsec
+
+import (
+	"encoding/json"
+	"os/exec"
+
+	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
+	"go.jetpack.io/devbox/internal/cmdutil"
+)
+
+func Env(projectDir string) (map[string]string, error) {
+
+	if err := ensureEnvsecInstalled(); err != nil {
+		return nil, err
+	}
+
+	if err := ensureEnvsecInitialized(); err != nil {
+		return nil, err
+	}
+
+	return envsecList(projectDir)
+}
+
+func ensureEnvsecInstalled() error {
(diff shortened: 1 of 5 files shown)


internal/integrations/envsec/envsec.go
```diff
@@ -0,0 +1,63 @@
...
+	cmd := exec.Command("envsec", "init")
+	// TODO handle user not logged in
+	// envsec init is currently broken in that it exits with 0 even if the user is not logged in
+	return cmd.Run()
+}
+
+func envsecList(projectDir string) (map[string]string, error) {
+	cmd := exec.Command(
+		"envsec", "ls", "--show",
+		"--format", "json",
```
LucilleH, line 42: I'm not sure if we should do this. I worry that the secrets printed can be discovered through some proc file somewhere. 

Can we just use the envsec go library directly?

mikeland73: I was under impression we wanted to do CLI integration in order to avoid bloating devbox. This has the benefit of allowing other 3rd parties to integrate as well if we provide a standard interface.

If stderr and stdout are captured in a buffer, is there really any risk the outputs ends up anywhere else?

We can also go the RPC route, but that's more work and feels a bit overkill.

LucilleH: Looking deeper it might be ok based on [this line](https://cs.opensource.google/go/go/+/master:src/os/exec/exec.go;l=984). So I guess we are good

LucilleH (approved):

PR #1477, opened 2023-09-14
[Script Exit on Error] enable feature flag, and add stability advice to CICD docs

### LINKED ISSUE #1457: [Feature]: Stop scripts executions on error
Let's consider a simple case
```json
{
  "packages": [ "..."],
  "env": { "...": "..." },
  "shell": {
    "scripts": {
      "start": [
        "bash ./check_something.sh",
        "bash ./do_some_long_and_verbose_task.sh"
      ]
    }
  }
}

```
Right now if `check_something.sh` exit with error, script `do_some_long_task.sh` executed despite that.
If `do_some_long_task` generate a lot of debug info in stdout, developer even not have a chance to see error from first script in terminal

## Possible solution:
I expect that devbox check exit code of every script and stop execution on first exit with error.
In simple words - it would be cool if it will work like `./check_something.sh && ./do_some_long_task.sh` syntax.
(that can be used as workaround right now, but it less readable and not cross platform)

savil:
## Summary

Fixes #1457

This feature enables `set -e` in the generated shell files for the "scripts" and "init hooks" of Devbox.
This feature has been running internally for a long time to power our CICD tests for the projects in `examples/`,
and has been very useful in ensuring correctness.

This may expose bugs in user code. While this results in `devbox version update` possibly leading to 
code that used to seemingly work previously now needing some fixes, we feel this is acceptable because:

1. We no longer auto-update Devbox. Users are intentionally updating when they are ready.
2. For CICD, they can pin the devbox-version for additional Stability. This PR updates the CICD docs
to call this out.

## How was it tested?

testscripts should pass

### DIFF
--- a/docs/app/docs/continuous_integration/github_action.md
+++ b/docs/app/docs/continuous_integration/github_action.md
@@ -45,3 +45,5 @@ jobs:
 ## Configuring the Action
 
 See the [GitHub Marketplace page](https://github.com/marketplace/actions/devbox-installer) for the latest configuration settings and an example.
+
+For stability over new features and bug fixes, consider pinning `devbox-version`. Remember to update this pinned version when you update your local Devbox via `devbox version update`.
--- a/internal/boxcli/featureflag/feature.go
+++ b/internal/boxcli/featureflag/feature.go
@@ -19,6 +19,9 @@ type feature struct {
 
 var features = map[string]*feature{}
 
+// Prevent lint complaining about unused function
+//
+//nolint:unparam
 func disable(name string) *feature {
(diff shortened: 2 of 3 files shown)


savil:
* main
  * **PR #1477** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1477" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1477?utm_source=stack-comment).

Lagoja (approved):

PR #1484, opened 2023-09-15
[global] Fix global services

mikeland73:
## Summary

cobra doesn't chain `PersistentPreRunE`. So if a parent declares one and a child declares it as well, the parent function is ignored. `global` uses `PersistentPreRunE` to set the global project directory. The `services` command was running it to ensure nix is installed. The child replaced the parent, so we were never setting th

The fix here is a bit ugly but it solves them problem.

There's a UX issue I rediscovered while testing this. When adding global packages the environment is not automatically refreshed. This means services wont work until a new shell is opened or if the user does `eval "$(devbox global shellenv)"` manually.

## How was it tested?
```
devbox global add nginx
devbox global services up
```

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -19,8 +19,12 @@ func globalCmd() *cobra.Command {
 	globalCmd := &cobra.Command{}
 
 	*globalCmd = cobra.Command{
-		Use:                "global",
-		Short:              "Manage global devbox packages",
+		Use:   "global",
+		Short: "Manage global devbox packages",
+		// PersistentPreRunE is inherited only if children do not implement it
+		// (i.e. it's not chained). So this is fragile. Ideally we stop
+		// using PersistentPreRunE. For now a hack is to pass it down to commands
+		// that declare their own.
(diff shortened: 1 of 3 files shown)


internal/boxcli/global.go
```diff
@@ -19,8 +19,12 @@ func globalCmd() *cobra.Command {
 	globalCmd := &cobra.Command{}
 
 	*globalCmd = cobra.Command{
-		Use:                "global",
-		Short:              "Manage global devbox packages",
+		Use:   "global",
```
savil, line 22: Instead of copy-pasting, it'll be nice to use a common variable...
```
persistentPreRunE := setGlobalConfigForDelegatedCommands(globalCmd)

*globalCmd = cobra.Command{
 		Use:   "global",
 		PersistentPreRunE: persistentPreRunE,
 		 ...
 		 }
 		 
 		 ...
 addCommandAndHideConfigFlag(
 		 globalCmd,
 		servicesCmd(persistentPreRunE),
 ),
```

savil (approved):

PR #1488, opened 2023-09-18
[script-on-error] disable for fish shell, and add testscript unit-test

savil:
## Summary

`set -e` can fail in fish-shells (sometimes), so we disable script-on-error for them.

I also add a testscript unit-test for script-exit-on-error, in general.

## How was it tested?

I run fish-shell, and did a sanity test that the `go/hello-world` example passes `run_test`.

BEFORE:
with this devbox.json
```
> cat devbox.json
{
    "packages": [
        "go@1.19.8"
    ],
    "shell": {
        "init_hook": [
            "export \"GOROOT=$(go env GOROOT)\"",
            "echo \"foo\""
        ],
        "scripts": {
            "run_test": "go run main.go"
        }
    }
}
```

doing `devbox shell` would give:
```
> devbox shell

Installing package: go@1.19.8.

[1/1] go@1.19.8
[1/1] go@1.19.8: Success
Starting a devbox shell...
set: -e: option requires an argument

~/code/jetpack/devbox/examples/development/go/hello-world/.devbox/gen/scripts/.hooks.sh (line 1):
set -e
^
from sourcing file ~/code/jetpack/devbox/examples/development/go/hello-world/.devbox/gen/scripts/.hooks.sh
	called on line 130 of file /var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/devbox2997650637/config.fish
from sourcing file /var/folders/zv/r3sx92_94gq86_rq3yn1ky2h0000gn/T/devbox2997650637/config.fish

(Type 'help set' for related documentation)
foo
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
(devbox)
```

AFTER:
```
> devbox shell

Installing package: go@1.19.8.

[1/1] go@1.19.8
[1/1] go@1.19.8: Success
Starting a devbox shell...
foo
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
(devbox)>
```

### DIFF
--- a/internal/shellgen/scripts.go
+++ b/internal/shellgen/scripts.go
@@ -25,6 +25,7 @@ type devboxer interface {
 	Lockfile() *lock.File
 	AllInstallablePackages() ([]*devpkg.Package, error)
 	InstallablePackages() []*devpkg.Package
+	IsUserShellFish() (bool, error)
 	PluginManager() *plugin.Manager
 	ProjectDir() string
 }
@@ -97,7 +98,13 @@ func WriteScriptFile(d devboxer, name string, body string) (err error) {
 	}
 
 	if featureflag.ScriptExitOnError.Enabled() {
-		body = fmt.Sprintf("set -e\n\n%s", body)
+		isFish, err := d.IsUserShellFish()
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		if !isFish {
+			body = fmt.Sprintf("set -e\n\n%s", body)
+		}
 	}
 	_, err = script.WriteString(body)
 	return errors.WithStack(err)
--- a/internal/shellgen/scripts.go
+++ b/internal/shellgen/scripts.go
@@ -81,8 +81,8 @@ func WriteScriptsToFiles(devbox devboxer) error {
 	return nil
 }
 
-func WriteScriptFile(d devboxer, name string, body string) (err error) {
-	script, err := os.Create(ScriptPath(d.ProjectDir(), name))
+func WriteScriptFile(devbox devboxer, name string, body string) (err error) {
+	script, err := os.Create(ScriptPath(devbox.ProjectDir(), name))
 	if err != nil {
 		return errors.WithStack(err)
 	}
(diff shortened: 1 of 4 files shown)


savil:
* main
  * **PR #1488** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1488" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1488?utm_source=stack-comment).

mikeland73 (commented):
Wait, do we run scripts in the shell of choice? I thought we always used bash for portability. (Though looking at code it looks like we use `sh` or `/bin/sh`)

Fish in general is not a scripting language and more designed to be interactive only. (At least acording to [Mr. Ridiculous Fish](https://stackoverflow.com/users/1441328/ridiculous-fish) in [this answer](https://stackoverflow.com/questions/23147048/how-can-i-do-bash-set-e-set-u-and-set-x-in-fish)

Would this not break portability? At least with bash we know it will be installed. That's why plugins look for bash.

internal/shellgen/scripts.go
```diff
@@ -25,6 +25,7 @@ type devboxer interface {
 	Lockfile() *lock.File
 	AllInstallablePackages() ([]*devpkg.Package, error)
 	InstallablePackages() []*devpkg.Package
+	IsUserShellFish() (bool, error)
```
mikeland73, line 28: Not really introduced in this PR, but I think we should cut down on this interface pattern for pure data like this. (I admin to be much to blame for this)

There are some cases where the circular dependency is hard to avoid (for example if you need to call a function with side-effects that would create a cycle). If what you are passing in is pure data, cheap to compute (like in this case), then using a struct argument keeps code simpler and more linear.

mikeland73 (approved):

savil:
@mikeland73 good point. The init-hooks will run in `fish` shell when doing `devbox shell`. Regular Devbox scripts will _not_ run in fish shell, and will use `sh` (didn't verify, but IIRC you're correct).

I added a comment in code to clarify this, and also showed more output in the Test Plan.

internal/shellgen/scripts.go
savil: Yeah, I agree 💯 

In this particular package, there are more public APIs than I think are good. For example, `WriteScriptToFile` should maybe not be a public method. Having a tighter public interface, would enable a constructor which can take some of these arguments (as you suggest), minimizing the need for the interface.

gcurtis (approved):

PR #1491, opened 2023-09-18
[perf] Profile possibly slow functions

mikeland73:
## Summary

Profile a few functions. This shows any function that takes more than 1 ms. 

Q: does trace have this functionality? It looks like it does time events but we have to explicitly enable it using the `trace` flag? 

## How was it tested?

```bash
export DEVBOX_PRINT_EXEC_TIME=1

devbox add curl
...
Exec times over 1ms:
"syncPackagesToProfile" took 343.118292ms
"nixEnv" took 128.018333ms
"nixEnv" took 128.219083ms
"CreateWrappers" took 257.527167ms
"ensurePackagesAreInstalled" took 768.378084ms
"CreateWrappers" took 2.56ms
"devbox add curl" took 1.21434s
```

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -106,6 +106,7 @@ func Execute(ctx context.Context, args []string) int {
 }
 
 func Main() {
+	timer := debug.Timer(strings.Join(os.Args, " "))
 	ctx := context.Background()
 	if strings.HasSuffix(os.Args[0], "ssh") ||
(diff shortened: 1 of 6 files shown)


Lagoja (changes requested):
Tested this out locally and it seems to work as expected. Surprisingly few functions seem to go over 1ms though.

Lagoja (approved):

PR #1492, opened 2023-09-19
[auth] Use new auth package

mikeland73:
## Summary

Migrate to new auth package. The feature on devbox that used previous auth was not publicized and there's no data on S3 that we need to migrate.

## How was it tested?

```bash
devbox auth login
devbox auth whoami
```


loreto (approved):
LGTM

PR #1493, opened 2023-09-19
[script-exit-on-error] disable for init-hooks

savil:
## Summary

@Lagoja identified an issue with scripts-exit-on-error. We source init-hooks into the host shell. So, `set -e` will get set in the host shell. Any subsequent error will cause the shell to exit (error may be from the init-hook, or later in the shell). 

For now, this PR disables this feature entirely for init hooks. We'll revisit this later: https://github.com/jetpack-io/devbox/issues/1494

Also, this PR undoes the previous change to restrict this feature to fish-shells, since that only applied to init-hooks. Regular Devbox scripts always run in `sh`.

## How was it tested?

testscript unit-tests

did a sanity check that regular init_hooks work.

### DIFF
--- a/internal/shellgen/scripts.go
+++ b/internal/shellgen/scripts.go
@@ -25,7 +25,6 @@ type devboxer interface {
 	Lockfile() *lock.File
 	AllInstallablePackages() ([]*devpkg.Package, error)
 	InstallablePackages() []*devpkg.Package
-	IsUserShellFish() (bool, error)
 	PluginManager() *plugin.Manager
 	ProjectDir() string
 }
@@ -53,7 +52,7 @@ func WriteScriptsToFiles(devbox devboxer) error {
 	}
 	hooks := strings.Join(append(pluginHooks, devbox.Config().InitHook().String()), "\n\n")
 	// always write it, even if there are no hooks, because scripts will source it.
-	err = WriteScriptFile(devbox, HooksFilename, hooks)
+	err = writeHookFile(devbox, hooks)
 	if err != nil {
 		return errors.WithStack(err)
 	}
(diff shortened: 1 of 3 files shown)


savil:
* main
  * **PR #1493** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1493" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1493?utm_source=stack-comment).

internal/shellgen/scripts.go
```diff
@@ -82,39 +81,53 @@ func WriteScriptsToFiles(devbox devboxer) error {
...
 		return errors.WithStack(err)
 	}
-	defer func() {
-		cerr := script.Close()
-		if err == nil {
-			err = cerr
-		}
-	}()
-	err = script.Chmod(0755)
+	defer func() { _ = script.Close() }() // best effort: close file
```
mikeland73, line 89: Does `defer script.Close()` not work? (and add `nolint` if you want to ignore error)

internal/shellgen/scripts.go
```diff
@@ -82,39 +81,53 @@ func WriteScriptsToFiles(devbox devboxer) error {
...
-		isFish, err := devbox.IsUserShellFish()
-		if err != nil {
-			return errors.WithStack(err)
-		}
-		if !isFish {
-			body = fmt.Sprintf("set -e\n\n%s", body)
-		}
+		// NOTE: Devbox scripts will run using `sh` for consistency.
+		// However, we need to disable this for `fish` shell if/when we allow this for init_hooks,
+		// since init_hooks run in the host shell, and not `sh`.
```
mikeland73, line 105: I don't think we'll never be able to add this to init hooks unless we decide not to source.

mikeland73 (approved):

PR #1497, opened 2023-09-20
[envsec] Use runx to install envsec

mikeland73:
## Summary

Mostly for testing. Need to merge a few runx changes to finish this PR.

## How was it tested?

```
devbox run echo \$Foo
```


internal/integrations/envsec/envsec.go
```diff
@@ -20,10 +32,23 @@ func Env(projectDir string) (map[string]string, error) {
...
 func ensureEnvsecInstalled() error {
+	// In newer runx version this will return the paths
+	if err := runx.Install("jetpack-io/envsec"); err != nil {
+		return errors.Wrap(err, "failed to install envsec")
+	}
+
+	// temporary, mac only
+	home, _ := os.UserHomeDir()
+	path := filepath.Join(home, "Library/Caches/jetpack.io/pkgs/jetpack-io/envsec/v0.0.1/darwin/arm64/")
+	os.Setenv("PATH", path+string(os.PathListSeparator)+os.Getenv("PATH"))
```
LucilleH, line 50: Why do we need this? Does runx not handle macOS?

internal/integrations/envsec/envsec.go
```diff
@@ -3,15 +3,27 @@ package envsec
...
+	"go.jetpack.io/pkg/sandbox/runx"
 )
 
+var envCache map[string]string
+
 func Env(projectDir string) (map[string]string, error) {
 
+	defer debug.FunctionTimer().End()
+
+	if envCache != nil {
```
LucilleH, line 23: How do you validate/invalidate this cache

internal/integrations/envsec/envsec.go
mikeland73: This is for testing. The runx library will now return the path (for all os's)

internal/integrations/envsec/envsec.go
mikeland73: It never gets invalidated, it's lifetime is just the command. So if you do `devbox run` you only fetch it once.

Later on I think we can look into refreshing env vars for long running shells, but even then this cache is OK.

internal/integrations/envsec/envsec.go
```diff
@@ -20,10 +31,23 @@ func Env(projectDir string) (map[string]string, error) {
...
-	return envsecList(projectDir)
+	var err error
+	envCache, err = envsecList(projectDir)
+
+	return envCache, err
 }
 
 func ensureEnvsecInstalled() error {
+	// In newer runx version this will return the paths
+	paths, err := runx.Install("jetpack-io/envsec@v0.0.2")
```
LucilleH, line 42: Should we not pin this? Or is this necessary?

LucilleH (approved):

PR #1500, opened 2023-09-22
[plugins] Fix init hooks for custom plugins

mikeland73:
## Summary

Ensure custom plugin init hooks run. Previously they were only running for builtins.

## How was it tested?

`devbox run run_test`

### DIFF
--- a/internal/shellgen/scripts.go
+++ b/internal/shellgen/scripts.go
@@ -46,7 +46,10 @@ func WriteScriptsToFiles(devbox devboxer) error {
 
 	// Write all hooks to a file.
 	written := map[string]struct{}{} // set semantics; value is irrelevant
-	pluginHooks, err := plugin.InitHooks(devbox.InstallablePackages(), devbox.ProjectDir())
+	pluginHooks, err := devbox.PluginManager().InitHooks(
+		devbox.InstallablePackages(),
+		devbox.Config().Include,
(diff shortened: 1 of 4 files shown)


savil (approved):

internal/shellgen/scripts.go
```diff
@@ -46,7 +46,10 @@ func WriteScriptsToFiles(devbox devboxer) error {
 
 	// Write all hooks to a file.
 	written := map[string]struct{}{} // set semantics; value is irrelevant
-	pluginHooks, err := plugin.InitHooks(devbox.InstallablePackages(), devbox.ProjectDir())
+	pluginHooks, err := devbox.PluginManager().InitHooks(
+		devbox.InstallablePackages(),
+		devbox.Config().Include,
```
savil, line 51: nit: should this field be plural: `Includes`?

mikeland73: I think this ship has sailed, has been in the `devbox.json` for a while now.

FWIW, I think we chatted about this an decided to use as a verb instead of noun (similar to `import` instead of `imports`).  I don't feel super strongly, but it seems like it would be difficult to change (but I guess we could support both)

savil: oh yeah, I do remember chatting about it being a verb...

PR #1501, opened 2023-09-22
[update] Add sync flag which syncs all dependencies to same resolved version under current working dir

mikeland73:
## Summary

Adds new `devbox update --sync-lock` command. It syncs all lockfiles to the same version. 

I'm also thinking of adding a new `--sync-from [lockfile]` flag for @LucilleH's use case. Specifically that would allow a certain lockfile to take priority.

## How was it tested?

`devbox update --sync-lock` on this repo (see lockfile changes)


internal/lock/sync.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	// Step 2: Update the devbox.lock files
+	return filepath.Walk(".", func(path string, fi os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !fi.IsDir() && filepath.Base(path) == "devbox.lock" {
```
LucilleH, line 24: Is this going to walk all of my `node_modules` files, or are we excluding directories according to `.gitignore` somewhere

internal/lock/sync.go
```diff
@@ -0,0 +1,92 @@
...
+
+func latestPackages() (map[string]*Package, error) {
+	latestPackages := make(map[string]*Package)
+
+	err := filepath.Walk(".", func(path string, fi os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !fi.IsDir() && filepath.Base(path) == "devbox.lock" {
```
LucilleH, line 65: ditto

internal/lock/sync.go
```diff
@@ -0,0 +1,92 @@
...
+
+			for key, pkg := range lockFile.Packages {
+				if latestPkg, exists := latestPackages[key]; exists {
+					currentTime, err := time.Parse(time.RFC3339, pkg.LastModified)
+					if err != nil {
+						return err
+					}
+					latestTime, err := time.Parse(time.RFC3339, latestPkg.LastModified)
+					if err != nil {
+						return err
```
LucilleH, line 79: Does this error if one of the lockfile is super old and does not have `LastModified` in there? (or is that scenario non existent)

internal/lock/sync.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+		}
+		return nil
+	})
+}
+
+func latestPackages() (map[string]*Package, error) {
+	latestPackages := make(map[string]*Package)
+
+	err := filepath.Walk(".", func(path string, fi os.FileInfo, err error) error {
```
LucilleH, line 60: Is it possible to have a separate function that simply returns all the `devbox.json` file paths, so that we only walk it once? That way if a `.gitignore` doesn't catch a giant directory such as `node_modules`, at least the delay only happens once 🤷‍♀️

LucilleH (approved):

internal/lock/sync.go
mikeland73: That's a good call. Yeah I think using `.gitignore` is good.

internal/lock/sync.go
mikeland73: I'm not sure that ever happens, but we should handle that. I think a missing last modified should be considered old and replaced if possible.

internal/lock/sync.go
LucilleH: Want to add that the monorepo might have a top level `.gitignore` that handles all the sub directories, and individual `.gitignore` in each subdirectory to add more stuff to ignore

internal/lock/sync.go
mikeland73: will do

mikeland73:
@LucilleH took me a bit to figure out how to handle the multi gitignores but I think I figured it out.

This runs on devbox in 1s (which has 45 devbox.lock and similar number of gitignores) and runs in 300-600ms in our front-end repo with all dependencies installed.\

PTAL

internal/lock/sync.go
```diff
@@ -0,0 +1,147 @@
...
+	type filterFuncStackEntry struct {
+		filterFunc func(string) bool
+		path       string
+	}
+
+	var filterFuncStack []filterFuncStackEntry
+
+	var filterFunc = func(path string) bool {
+		// Start at the top of the stack which has most specific gitignore.
+		for i := len(filterFuncStack) - 1; i >= 0; i-- {
```
LucilleH, line 100: I'm trusting that this works as intended 😅

LucilleH (approved):

mikeland73:
@LucilleH it works, but the performance improvement doesn't really work. Turns out that testing (almost) every single file and directory against gitignore slows it down more than the savings of not going into every directory. 

Given the complexity I lean towards removing it.

mikeland73:
There's a chance that limiting it to only directories could yield gains, but given how fast it is right now I'm tempted to remove the gitignore code and ship it.

LucilleH:
@mikeland73 can you try against the frontend repo with `node_modules` populated with pnpm?

mikeland73:
> @mikeland73 sounds good. Just one more thing: can you try against the frontend repo with node_modules populated with pnpm? Just want to know how much time it takes without the gitignore

Looks like there's caching of some sort at the OS level. First run is 900ms and subsequent ones are ~350ms.

PR #1503, opened 2023-09-25
[sync-lock] Add sync-lock-reference flag

mikeland73:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1501

This adds a `--sync-lock-reference` flag that allows `--sync-lock` to prioritize dependencies in a lock file. This is useful when you want to enforce a repo to mimic a single lock file.

This is an alternative to introducing workspaces (e.g. yarn or golang workspaces). 

## How was it tested?

```
# modified examples/databases/mariadb/devbox.lock to have an older dependency
devbox update --sync-lock --sync-lock-reference examples/databases/mariadb/devbox.lock
# observed that all lock files now follow older dependency.
```

### DIFF
--- a/internal/boxcli/update.go
+++ b/internal/boxcli/update.go
@@ -13,8 +13,9 @@ import (
 )
 
 type updateCmdFlags struct {
-	config configFlags
-	sync   bool
+	config                configFlags
+	sync                  bool
+	ReferenceLockFilePath string
 }
 
 func updateCmd() *cobra.Command {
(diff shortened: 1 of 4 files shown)


LucilleH:
I'm not sure we need this if the search endpoint can always return cached packages. Thoughts?

mikeland73:
> I'm not sure we need this if the search endpoint can always return cached packages. Thoughts?

Agreed

mikeland73:
OTOH,  this could still be useful when adding a new project to a repo. e.g. you want to add a new project and use existing versions without forcing everything to change to the newer version.

gcurtis:
Could you go into more detail on when synchronizing to a specific lock file is useful? I've mostly run into cases where I want to:

1. Bulk edit a bunch of devbox.jsons to set a package version.
2. Run a bulk `devbox update` to lock all of those to the same constrained version.

It seems rare to want to synchronize to the non-latest nixpkgs commit for a package by pointing to a lock file.

mikeland73:
@gcurtis this is mostly because our frontend repo has a specific version of node that is pre-built. But if search is already only returning prebuilt by default, that would eliminate the need for this for the the most part.

I still think there's a use case for adding new dependencies that match what you have in the project, but this may be the wrong solution.

maybe:

```
devbox add <pkg> --resolve-using path/to/devbox.lock
devbox install --resolve-using path/to/devbox.lock
``` 

The reason being that I don't want to be forced to update all dependencies whenever I want to add a new project, but I do want all projects to resolve to the same hash.

PR #1507, opened 2023-09-27
[code-cleanup] Rename devbox.Writer to Stderr

mikeland73:
## Summary

This renames the devbox `Writer` field to the more accurate `Stderr`. It changes a few functions that used to print stuff and now they return strings instead.

Also renamed a few functions that where called `PrintX` to something more accurate because they don't actually print anything.

How to review:

* grep for `devopt.Opts` this will show all the places where we used to pass in stdout sometimes. Specifically @mohsenari please look at integrate.go. @ipince please look at telemetry.go. I think both of these were wrong.
* Please take a look at functions that used to print stuff and now return strings (i.e. Info and PrintGlobalList). I tried to keep formatting identical so nothing should change.

## How was it tested?

compiles

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -16,41 +16,36 @@ import (
 
 // Devbox provides an isolated development environment.
 type Devbox interface {
-	// Add adds Nix packages to the config so that they're available in the devbox
-	// environment. It validates that the Nix packages exist, and install them.
-	// Adding duplicate packages is a no-op.
 	Add(ctx context.Context, platforms, excludePlatforms []string, pkgs ...string) error
 	Config() *devconfig.Config
-	ProjectDir() string
-	// Generate creates the directory of Nix files and the Dockerfile that define
-	// the devbox environment.
 	EnvVars(ctx context.Context) ([]string, error)
-	Generate(ctx context.Context) error
-	GenerateDevcontainer(ctx context.Context, generateOpts devopt.GenerateOpts) error
-	GenerateDockerfile(ctx context.Context, generateOpts devopt.GenerateOpts) error
-	GenerateEnvrcFile(ctx context.Context, force bool, envFlags devopt.EnvFlags) error
 	Info(ctx context.Context, pkg string, markdown bool) (string, error)
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
 	NixEnv(ctx context.Context, includeHooks bool) (string, error)
 	PackageNames() []string
+	ProjectDir() string
 	Pull(ctx context.Context, opts devopt.PullboxOpts) error
 	Push(ctx context.Context, opts devopt.PullboxOpts) error
-	// Remove removes Nix packages from the config so that it no longer exists in
-	// the devbox environment.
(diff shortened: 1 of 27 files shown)


internal/impl/devbox.go
```diff
@@ -92,7 +89,7 @@ func Open(opts *devopt.Opts) (*Devbox, error) {
 		nix:                      &nix.Nix{},
 		projectDir:               projectDir,
 		pluginManager:            plugin.NewManager(),
-		writer:                   opts.Writer,
+		Stderr:                   opts.Stderr,
```
ipince, line 92: does `Stderr` need to be public?

ipince (approved):
LGTM, thanks!

devbox.go
```diff
@@ -16,41 +16,36 @@ import (
...
+	EnvVars(ctx context.Context) ([]string, error)
+	Info(ctx context.Context, pkg string, markdown bool) (string, error)
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	PrintEnv(ctx context.Context, includeHooks bool) (string, error)
-	PrintEnvVars(ctx context.Context) ([]string, error)
-	PrintGlobalList() error
+	NixEnv(ctx context.Context, includeHooks bool) (string, error)
+	PackageNames() []string
```
savil, line 27: Are these config PackageNames or installable PackageNames?

internal/boxcli/root.go
```diff
@@ -49,6 +52,8 @@ func RootCmd() *cobra.Command {
 		SilenceErrors: true,
 		SilenceUsage:  true,
 	}
+	command.SetOut(os.Stdout)
```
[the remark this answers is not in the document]
savil: maybe I am misunderstanding. 

I thought the idea behind `cmd.OutOrStdout()` was that tests could do `command.SetOut` and then inspect the output. 

If we're overriding here what the tests may set, then it won't work as desired.

savil (approved):

internal/boxcli/integrate.go
mohsenari, line 1: In this case either works, but stdErr is logically correct. 👍

mohsenari (approved):

internal/impl/devbox.go
mikeland73: good point. Will fix.

internal/boxcli/root.go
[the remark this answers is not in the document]
mikeland73: So `out` is not set by default in cobra. So `cmd.OutOrStdout` returns stdout because out is not set. But `cmd.Print` uses `OutOrStderr` (don't ask me why, I don't understand it) so it prints to stderr.

I'll fix this by removing `cmd.Print` and replacing with `fmt.Fprint(cmd.OutOrStdout()`

devbox.go
mikeland73: Just what is in config. This is a no-op (PrintGlobaList used to call PackageNames).

We can chat about whether this is correct or not (I think it's fine), but would rather not change it in this PR.

PR #1508, opened 2023-09-28
[shellenv] fix PATH nesting for devbox-project and devbox-global

savil:
## Summary

**Problem:**
High level problem description:
1. A devbox-global binary is used when executing a devbox script. 
   - For example, `devbox global add bash@5.1-p12` adds bin-wrappers for `sh` and `bash`. This `sh` wrapper is used to execute the devbox script.
2. The devbox-global binary (example, `sh`) needs to re-evaluate its own environment
3. As a result, the devbox-global binary amends the PATH, but in doing so it may drop the devbox script's PATH elements, leading to errors because the devbox-project's binaries cannot be found.

**Cause**
In our implementation of `shellenv`, we save an "OG Path" in an env-var. This is the PATH prior to entering `devbox shell`. We then calculate the `print-dev-env` PATH. Finally, we construct the PATH for the `shellenv` as `<print-dev-env paths>:<og-paths>`.

However, if another devbox-project has introduced additional path-elements in PATH, those will get dropped. Because when this shellenv is re-evaluated: it will do `PATH=<new nixEnvPath>:<og-path>`.

**Solution:**
Develop a `PATH_STACK` that can track which path parts come from which devbox-projects. 

So, for a devbox-project nested within a shell that is using devbox-global, the PATH_STACK looks like `{<devbox-project-nix-env-path>, <devbox-global-nix-env-path>, <devbox-init-path>}`
  - earlier parts get priority, just like PATH elements
  - devbox-init-path, is analogous to og-path, but not identical: this PATH_STACK applies to _all_ devbox projects within this shell. The og-path was specific to a devbox-project.

Now, if the global shellenv is re-evaluated, then the `<devbox-global-nix-env-path>` value is overidden, but the PATH_STACK structure is maintained.

Finally, we construct PATH for `shellenv` as `PATH=<devbox-project-nix-env-path>:<devbox-global-nix-env-path>:<devbox-init-path>`.

## How was it tested?

- [x] add unit test for `envpath.Stack`

TODO: fill in.

1. `devbox global add bash` doesn't lead to errors with `devbox run build`.
2. Inside a `devbox add vim && devbox shell`, then doing `devbox rm vim` doesn't give the nix store path i.e. #687 is still fixed.

3. Adhoc test:
```
mkdir -p foo-hello/bar-cowsay
cd foo-hello && devbox init && devbox add hello
cd bar-cowsay && devbox init && devbox add cowsay

cd ../..
eval $(devbox -c foo-hello shellenv)
hello
eval $(devbox -c foo-hello/bar-cowsay shellenv)
cowsay wow

# before, this would drop the bar-cowsay bin-wrappers and nix-profile from PATH
eval $(devbox -c foo-hello shellenv)
cowsay works
```
Previously, the last "cowsay works" would fail
https://gist.github.com/savil/2c6e5f86208ff95ae54e717c17c794d6


savil:
* main
  * **PR #1508** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1508" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1508?utm_source=stack-comment).

savil:
@mikeland73 have a look please. I think the main approach is in place. I needed to move to a struct to keep the logic encapsulated in both `nixEnv` and `envvars.go`'s ``IsEnvEnabled`.

Not 100% confident about all the places I added `false /*pathStackInPlace*/`. Need to audit them again.
I also need to make another pass over the comments in the `pathstack.go` file. Feel free to point out nits, or suggested changes.

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// AddToEnv adds the new nixEnvPath for the devbox-project identified by projectHash to the env.
+// It does so by modifying the following env-vars:
+// 1. nixEnvPath key
+// 2. PathStack
+// 3. PATH
+//
+// Returns the modified env map
+func (s *Stack) AddToEnv(
```
savil, line 63: I don't like this name. Maybe it should be `stack.PushAndUpdateEnv` to indicate that a new element is pushed on the stack, and the `env` is also modified.

internal/boxcli/shellenv.go
```diff
@@ -44,6 +45,11 @@ func shellEnvCmd() *cobra.Command {
 
 	command.Flags().BoolVar(
 		&flags.pure, "pure", false, "If this flag is specified, devbox creates an isolated environment inheriting almost no variables from the current environment. A few variables, in particular HOME, USER and DISPLAY, are retained.")
+	command.Flags().BoolVar(
+		&flags.pure, "path-stack-in-place", true,
+		"If true, Devbox will not give top priority to the PATH of this project's shellenv. "+
```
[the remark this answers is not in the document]
mikeland73: Why is it defaulting to true? Should be false. I'm confused by the name maybe

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+	InitPathEnv = "DEVBOX_INIT_PATH"
+)
+
+// Stack has the following design:
+// 1. The PathStack enables tracking which sub-paths in PATH come from which devbox-project
+// 2. What it stores: The PathStack is an ordered-list of nixEnvPathKeys
+// 3. Each nixEnvPathKey is set as an env-var with the value of the nixEnvPath for that devbox-project.
+// 4. The final PATH is reconstructed by concatenating the env-var values of each nixEnvPathKey env-var.
+// 5. The Stack is stored in its own env-var shared by all devbox-projects in this shell.
+type Stack struct {
```
mikeland73, line 23: dont export

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// keys holds the stack elements.
+	// Earlier (lower index number) keys get higher priority.
+	// This keeps the string representation of the Stack aligned with the PATH value.
+	keys []string
+}
+
+func NewStack(env map[string]string) *Stack {
```
mikeland73, line 31: Just call this `Stack`

internal/boxcli/shellenv.go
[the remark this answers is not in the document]
savil: oh yeah, I was struggling to recall the name we'd agreed on. I can change it.

devbox.go
```diff
@@ -20,10 +20,10 @@ type Devbox interface {
 	Config() *devconfig.Config
 	EnvVars(ctx context.Context) ([]string, error)
 	Info(ctx context.Context, pkg string, markdown bool) (string, error)
-	Install(ctx context.Context) error
+	Install(ctx context.Context, pathStackInPlace bool) error
```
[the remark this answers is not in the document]
savil: I can do that but I feel kinda mixed about it. 

We're not eliminating the boolean param, but sending it via a somewhat global state.

internal/boxcli/shellenv.go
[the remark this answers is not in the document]
savil: ugh, it should default to false.

The PR's state is a bit subpar b/c my fish shell has gotten into a bad state and I've been having problems running programs. So, sent this for prelim feedback.

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+// Has tests if the stack has the specified key. Refer to the Key function for constructing
+// the appropriate key for any devbox-project.
+func (s *Stack) Has(key string) bool {
+	for _, k := range s.keys {
+		if k == key {
+			return true
+		}
+	}
+	return false
+}
```
mikeland73, line 99: `lo.Contains`

devbox.go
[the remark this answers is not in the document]
savil: > We're not eliminating the boolean param, but sending it via a somewhat global state.

Yeah, but I see the utility in not complicating the core funciton interfaces for this edge-case handling.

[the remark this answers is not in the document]
mikeland73: Yeah agree devbox options are semi-global (not quite global as this [PR](https://github.com/jetpack-io/devbox/pull/1515) can attest) 

But it really makes everything else easier to read and the risk is minimal (specifically a single devbox object being used multiple times in a single command which is a pattern we don't use at all)

internal/boxcli/shellenv.go
```diff
@@ -44,6 +49,11 @@ func shellEnvCmd() *cobra.Command {
 
 	command.Flags().BoolVar(
 		&flags.pure, "pure", false, "If this flag is specified, devbox creates an isolated environment inheriting almost no variables from the current environment. A few variables, in particular HOME, USER and DISPLAY, are retained.")
+	command.Flags().BoolVar(
+		&flags.preservePathStack, "preserve-path-stack", false,
+		"If true, Devbox will not give top priority to the PATH of this project's shellenv. "+
+			"This project's place in the path stack will be unchanged.")
```
mikeland73, line 55: Seggested description:

"Preserves existing PATH order if this project's environment is already in PATH. Useful if you want to avoid overshadowing another devbox project that is already active"

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+	// Add this nixEnvPath to env
+	env[key] = nixEnvPath
+
+	// Common case: ensure this key is at the top of the stack
+	if !preservePathStack ||
+		// Case preservePathStack == true, usually from bin-wrapper or (in future) shell hook.
+		// Add this key only if absent from the stack
+		!lo.Contains(s.keys, key) {
+
+		s.keys = lo.Uniq(slices.Insert(s.keys, 0, key))
```
mikeland73, line 80: totally get why you chose `Insert` over append (so that we keep same order as needed by `JoinPathLists`) but I always think of golang slices as natural stacks where `append` can be used as a natural `push`

mikeland73 (approved):

internal/impl/envpath/pathstack.go
savil: yes! I totally get you. 

I first implemented `stack.keys` with later values having higher priority. Because append is a natural push, as you say. 

But then changed to have the highest-priority be the zeroth index, because it aligns the path-stack's golang and string representations with the expanded form $PATH variable. This makes debugging much easier, when we dump out these values in `fmt.Printf` or `echo $DEVBOX_PATH_STACK` or `echo $PATH`.

savil (commented):
@mikeland73 made updates as requested. Also, renamed `envpath.go` to `envpath/pathlists.go` and moved its unit test to this package.

I still need to write a unit test for `envpath.stack`.

I'm also thinking it may make sense to wait on landing this until Remove Nixpkgs and the Script Exit on Error are released. To minimize risk.

internal/boxcli/shellenv.go
savil: like it!

internal/impl/devbox.go
```diff
@@ -901,14 +896,16 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	})
 	debug.Log("PATH after filtering with buildInputs (%v) is: %s", buildInputs, nixEnvPath)
 
-	env["PATH"] = JoinPathLists(nixEnvPath, originalPath)
+	env = pathStack.PushAndUpdateEnv(env, d.projectDirHash(), nixEnvPath, d.preservePathStack)
+	debug.Log("New path stack is: %s", pathStack)
```
mikeland73, line 900: I thought I had written this comment previously, but github seems to have swallowed it. Here's a suggestion for an interface I find a bit easier to follow:

```golang
pathStack := envpath.Stack(env)
if !d.preservePathStack || !pathStack.Has(d.projectDirHash()) {
  pathStack.PushAndDedup(d.projectDirHash(), nixEnvPath)
}
env["PATH"] = pathStack.Path()
```

Benefits:

* Remove preserve logic from stack
* All code is concentrated together.
* Split out push from computing the path
* Change `Has` to take the hash instead of key. key can now be private

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+func Stack(env map[string]string) *stack {
+	stackEnv, ok := env[PathStackEnv]
+	if !ok {
+		// if path stack is empty, then push the current PATH, which is the
+		// external environment prior to any devbox-shellenv being applied to it.
+		stackEnv = InitPathEnv
+		env[InitPathEnv] = env["PATH"]
+	}
+	return &stack{
+		keys: strings.Split(stackEnv, ":"),
```
gcurtis, line 42: Not sure if this matters, but `keys` will be a slice with a single empty string if `stackEnv` is an empty string.

internal/impl/envvars.go
```diff
@@ -136,5 +138,11 @@ func markEnvsAsSetByDevbox(envs ...map[string]string) {
 // as a proxy for this. This allows us to differentiate between global and
 // individual project shells.
 func (d *Devbox) IsEnvEnabled() bool {
-	return os.Getenv(d.ogPathKey()) != ""
+	env := map[string]string{}
+	for _, keyVal := range os.Environ() {
+		parts := strings.Split(keyVal, "=")
```
gcurtis, line 143: Use strings.Cut, otherwise you'll lose the ends of variables that have an `=` in them. I think you could also export `envir.mapToPairs` and reuse it here.

internal/impl/envpath/pathstack.go
```diff
@@ -0,0 +1,99 @@
...
+// 2. PathStack
+// 3. PATH
+//
+// Returns the modified env map
+func (s *stack) PushAndUpdateEnv(
+	env map[string]string,
+	projectHash string,
+	nixEnvPath string,
+	preservePathStack bool,
+) map[string]string {
```
gcurtis, line 73: Maps are reference types, so you don't need to return a new one.

gcurtis (approved):
I know you guys will probably disagree, but the Go-ish way would be to make the zero-value do what the constructor is doing:

```go
type Stack struct {
    keys []string
}

// Remove func Stack()

func (s *Stack) Push(env map[string]string, ...) {
    if s.keys == nil {
        // do what Stack() was doing
    }

    // continue with PushAndUpdateEnv
}
```

Usage:

```go
var s envpath.Stack
s.Push(...)
```

internal/impl/devbox.go
savil: I like this.

Another benefit: 
- we move out handling env["PATH"]
- we can later have an `envpath.Path` struct that encapsulates further a lot of the complexity of this function, that then calls the `envpath.Stack` as shown here.

internal/impl/envpath/pathstack.go
savil: good point! I can check for stackEnv being an empty string

internal/impl/envvars.go
savil: oh i like that. Done!

internal/impl/devbox.go
savil: There is a problem with this. We need to update the env-vars for the nixEnvPathKey regardless of the if-condition. And I think that should be inside the stack, since its internal logic.

savil: So, a fix could be:
```
pathStack := envpath.Stack(env) 
if !d.preservePathStack || !pathStack.Has(d.projectDirHash()) { 
  pathStack.PushAndDedup(d.projectDirHash(), nixEnvPath) 
}
env[envpath.Key(d.projectDirHash())] = nixEnvPath // exposes internal logic of pathstack.
env["PATH"] = pathStack.Path()
```

but my concern would be that it breaks encapsulation and exposes internal logic.

mikeland73:
@gcurtis I don't think dislike [this pattern](https://github.com/jetpack-io/devbox/pull/1508#pullrequestreview-1651475847)  at all! I think it's cool. But it doesn't quite work because the stack needs to be initialized with `env` before calling `Push`

In that case the struct would contain an Env field.

One downside to not initializing is that now you need to have the initialization code in every public method.

gcurtis:
@mikeland73 it seems like the `Push` method could do that, no? I was thinking of it as:

1. Push the initial environment
2. Push the devbox environment
3. Push the nested devbox environment
4. Etc.

I don't think initializing the slice is too bad if the number of methods is small. It matches how types like `strings.Builder`, `bytes.Buffer`, `list.List`, etc. behave.

internal/impl/devbox.go
mikeland73: Good catch. Yeah, agree that it's not great to expose the internals. 

In that case maybe just put the stack creation and use next to each other?

savil: yup, did that.

mikeland73:
@gcurtis I think we're mixing pushing environment and pushing the path (we call the path `nixEnvPath` which is confusing).

What gets pushed on the stack is PATH fragments. But we also need to pass in the full environment because the stack itself is stored in env variables, as well as another env variable that stores the stack keys.

internal/impl/devbox.go
mikeland73: Actually we could still do:

```golang
pathStack := envpath.Stack(env) 
pathStack.PushAndUpdateEnv(d.projectDirHash(), nixEnvPath, d.preservePathStack)
env := pathStack.Env()
```

savil: I'm going to land as-is, so can focus on other work.

Changing to `.Env()` I have some concerns with:
1. We'll be making a copy of `env` internally in the constructor. Then applying that over any changes between the constructor and doing `pathStack.Env()` could potentially lead to issues, since it can lose information.

I prefer the existing design so that:
1. Internal state is restricted to the path-stack keys. Other state is directly stored in `env` that is passed in.
2. The `PATH` is updated externally. That enables us adding a separate data structure to abstract the PATH handling in this function (future work).

Either way, don't feel strongly and we can iterate on it.

PR #1511, opened 2023-09-28
[errors] Show better errors when user script errors out and fix poetry plugin

mikeland73:
* When a user script errors out, show the error code (e.g. `Error: exit status 1`) This helps in cases when stderr is empty (which the previous `--quiet` was causing)
* Improve poetry plugin by removing `--quiet` flag and adding more required into in readme.

Possible future improements:
* Better error when python is not installed
* Better error when pyproject is missing
* Allow custom pyproject locations

## How was it tested?

image

New implementation

image

### DIFF
--- a/internal/boxcli/midcobra/midcobra.go
+++ b/internal/boxcli/midcobra/midcobra.go
@@ -81,6 +81,7 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 		var exitErr *exec.ExitError
 		var userExecErr *usererr.ExitError
 		if errors.As(err, &userExecErr) {
+			ux.Ferror(ex.cmd.ErrOrStderr(), userExecErr.Error()+"\n")
 			return userExecErr.ExitCode()
 		}
 		if errors.As(err, &exitErr) {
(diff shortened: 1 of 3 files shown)


savil:
I don't understand the Test Plan. Why is `devbox run echo 5` complaining about python?

internal/boxcli/midcobra/midcobra.go
```diff
@@ -81,6 +81,7 @@ func (ex *midcobraExecutable) Execute(ctx context.Context, args []string) int {
 		var exitErr *exec.ExitError
 		var userExecErr *usererr.ExitError
 		if errors.As(err, &userExecErr) {
+			ux.Ferror(ex.cmd.ErrOrStderr(), userExecErr.Error()+"\n")
```
savil, line 84: Suggestion. Also print the Stderr since it may have other info.

We do similar in midcobra/debug.go: https://github.com/jetpack-io/devbox/blob/42d2f713fde840558faeb9d404b8f5e9e2f697dd/internal/boxcli/midcobra/debug.go#L67

savil (approved):

mikeland73:
> I don't understand the Test Plan. Why is devbox run echo 5 complaining about python?

Sorry, I should have been more specific. I added `poetry` and then ran the echo line. The error you see is poetry complaining python is not installed. The error is not visible because `--quiet` was removed. The next line (Error exit status 1) is now visible due to the new code that prints the ExitError

mikeland73:
@savil I changed the implementation a bit. Instead of printing error in `midcobra` which is wrong, we print it in the debug middleware.

I added code to ensure telemetry does not log user errors.

savil (approved):
coolio

PR #1513, opened 2023-09-28
all: (mostly) automated clean up

gcurtis:
- Unset any user-defined Go environment variables `devbox.json` to make builds more reproducible. Clean up some of the build scripts.
- Remove `exp/slices` and `exp/maps` now that they're in the Go 1.21 stdlib.
- Run `find . -name '*.go' -exec gofumpt -extra -w {} \+` to format Go files.
- Upgrade dependencies that don't have any backwards incompatible changes.
- Organize direct vs. indirect modules and run `go mod tidy`.


devbox.json
mikeland73, line 1: Add `gofumpt` here script here. Then we can add CICD that runs it and errors out if there are any changes.

mikeland73 (approved):

mikeland73:
Tests fails just look like timeouts. Probably the dependency changes caused longer build times.

devbox.json
gcurtis: Added a `devbox run fmt` script that runs in the lint job right before golangci-lint. If `CI` is set, it prints any unformatted files and exits with an error. When `CI` isn't set, it silently formats the files.

gcurtis:
Merging because the CI error was due to a GitHub API rate limit (presumably due to force-pushing debug commits).

PR #1515, opened 2023-09-29
[RFC][update] Add --all-projects flag

mikeland73:
## Summary

RFC for flag that updates all projects under a given directory (recursively) 

Also adds ability to sync specific packages via `--sync-lock` and moved sync code into `multi` package.

## How was it tested?

`devbox update --all-projects` in a few of our repos.


gcurtis (approved):
Makes sense to me.

PR #1520, opened 2023-10-02
[print-dev-env] improve cache hit rate in ensurePackagesAreInstalled

savil:
## Summary

In `ensurePackagesAreInstalled`, we force `computeNixEnv` to be run with `false` for using the `print-dev-env` cache.

We then run `createWrappers` > `nixBins` > `nixEnv`. This re-runs `computeNixEnv` with `false` for using the print-dev-env cache, because the lockfile changes in `ensurePackagesAreInstalled` were not committed.

## How was it tested?

Diagnosis done by running `devbox global --trace run -- echo "hello"`.

I'm trying to verify the fix.

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -234,16 +234,13 @@ func (d *Devbox) ensurePackagesAreInstalled(ctx context.Context, mode installMod
 	}
 
 	// Force print-dev-env cache to be recomputed.
+	fmt.Fprintln(d.stderr, "Re-computing the devbox environment.")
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #1520** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1520" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1520?utm_source=stack-comment).

savil:
The risk with this change is that if bin-wrappers fail to be generated, then the lockfile is out of sync.

PR #1521, opened 2023-10-02
[UX] print statement for recomputing the nixEnv

savil:
## Summary

This print-dev-env calculation may be a bit slow, so we should print a statement
for the waiting user.

I do this in `ensurePackagesAreInstalled`, instead of in `nix.PrintDevEnv` because otherwise we can print it 3x due to our cache misses.

Note, this PR is focussed on the least-risky change for the release to improve UX for improved wait times on switching to Remove Nixpkgs mode.

## How was it tested?

compiles


savil:
* main
  * **PR #1521** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1521" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1521?utm_source=stack-comment).

mikeland73 (approved):

PR #1522, opened 2023-10-03
[bin-wrappers] Reduce/improve when we create bin wrappers

mikeland73:
## Summary

This removes all `CreateWrappers` calls except one.  It also removes the interface used to prevent a circular dependency and passed in all arguments at the callsite. It no longer needs to recompute nix env.

It helps address performance issues where CreateWrappers was causing extra calls to ComputeNixEnv.

## How was it tested?

It builds and runs. This will need extensive testing.

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -134,7 +134,7 @@ func (d *Devbox) Add(ctx context.Context, platforms, excludePlatforms []string, 
 		}
 	}
 
-	return d.lockfile.Save()
+	return nil
 }
 
 // Remove removes the `pkgs` from the config (i.e. devbox.json) and nix profile
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -235,7 +235,7 @@ func (d *Devbox) ensurePackagesAreInstalled(ctx context.Context, mode installMod
 	// Ensure we clean out packages that are no longer needed.
 	d.lockfile.Tidy()
 
-	nixBins, err := d.nixBins(ctx, nixEnv)
+	nixBins, err := d.nixBins(nixEnv)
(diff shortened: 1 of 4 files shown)


internal/wrapnix/wrapper.go
```diff
@@ -134,12 +125,11 @@ func CreateDevboxSymlinkIfPossible() error {
 }
 
 type createWrapperArgs struct {
-	devboxer
```
savil, line 137: yes! thank god we removed this.

internal/impl/packages.go
```diff
@@ -183,11 +179,7 @@ func (d *Devbox) Remove(ctx context.Context, pkgs ...string) error {
 		return err
 	}
 
-	if err := d.saveCfg(); err != nil {
-		return err
-	}
-
-	return wrapnix.CreateWrappers(ctx, d)
+	return d.saveCfg()
```
savil, line 182: why don't we save the lockfile here, after the config has been changed? Similar to the end of AddPackages.

savil (approved):
why did we have all the additional calls?

mikeland73:
> why did we have all the additional calls? 

They predate the call inside `ensurePackagesAreInstalled`. Previously all our code paths were not centralized through `ensurePackagesAreInstalled` so each command had its own bin wrapper creation. But I think tying it to `ensurePackagesAreInstalled` is exactly what we need. I still want to test it thoroughly.

internal/impl/packages.go
mikeland73: It's actually not needed in either because `ensurePackagesAreInstalled` already does `lockfile.Tidy` 

will remove from `Add()`

PR #1523, opened 2023-10-03
[cleanup] remove unused nixEnv cache

savil:
## Summary

cleanup this unused nixEnv cache.

I don't think it makes sense to re-use it just now. Reasons:
- inside `computeNixEnv` the most time intensive work is `nix print-dev-env`. The
rest happens quite quickly.
- We already rely on the lockfile state to guide us on re-using the print-dev-env cache.

Therefore, the value of this nixEnv cache is pretty marginal, and by re-introducing
code to set the cache, we may introduce correctness errors. So I propose 
removing it.

## How was it tested?

compiles.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -916,16 +916,11 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	return env, d.addHashToEnv(env)
 }
 
-var nixEnvCache map[string]string
-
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #1523** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1523" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1523?utm_source=stack-comment).

PR #1524, opened 2023-10-03
[runx] Proof of concept for runx integration

mikeland73:
## Summary

Inspired by golangci-lint being broken in nixpkgs, I wanted to see if I could get a quick and dirty implementation so we can replace golangci-lint from nixpkgs with runx. The current PR implements this using `runx:<path>` syntax. Since we already use `github:<path>` for nix flakes, I can't use that without breaking backwards compatibility. We could overload it.

What this PR implements:

* Add and rm of runx, with and without versions.
* runx installed packages are added to PATH for run, shell, etc

Things that are not implemented by this PR:

* Search and validation. Currently trying to install a non-existing package will fail with a semi-cryptic error.
* lockfile support. 

## How was it tested?

image

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,7 +1,7 @@
 {
   "packages": [
-    "golangci-lint@1.52.2",
-    "go@latest"
+    "go@latest",
+    "runx:golangci/golangci-lint@latest"
   ],
   "env": {
     "GOENV": "off",
@@ -20,8 +20,8 @@
         "GOOS=linux GOARCH=arm64 go build -o dist/devbox-linux-arm64 ./cmd/devbox"
       ],
       "code": "code .",
-      "lint": "golangci-lint run",
       "fmt": "scripts/gofumpt.sh",
+      "lint": "golangci-lint run",
       "test": "go test -race -cover ./...",
       "update-examples": "devbox run build && go run testscripts/testrunner/updater/main.go"
     }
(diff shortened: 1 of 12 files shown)


devbox.json
```diff
@@ -1,7 +1,7 @@
 {
   "packages": [
-    "golangci-lint@1.52.2",
-    "go@latest"
+    "go@latest",
+    "runx:golangci/golangci-lint@latest"
```
savil, line 4: Would be preferable to include a source reference. In the future, runx may support github, gitlab, sourcehut, custom domain etc.

```
runx: github.com/golangci/golangci-lint@latest
```

savil (approved):
yes, lets do this and start using internally!

Before announcing this to Devbox users:

I have a significant concern about using `runx:` as the user-facing prefix in devbox.json. I think that'll confuse people: "I'm not running anything". IIRC @mohsenari was confused when you used this term in a meeting.

Finding an alternative would be good. One suggestion is `binary:`. Good discussion to have with the team and/or users.

mikeland73:
@savil should I merge this with `runx` or should we decide on the name first?

Another issue I need to fix before merging is the github rate limits on their `releases.json` endpoint. I need to make changes to runx so that `releases.json` gets called with github access token.

savil:
I think we can merge, and make any follow up changes prior to a public announcement. Incremental progress ftw!

PR #1525, opened 2023-10-03
[fish] unbreak init-hook

savil:
## Summary

I don't see any of these env-vars in my devbox shell, so didn't try unsetting
them in fish ;-)  the fish command is `set -e <name>` to "erase" the env-var.

## How was it tested?

could start devbox shell. 
before: it would break since `unset` is unrecognized.

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -9,7 +9,7 @@
   },
   "shell": {
     "init_hook": [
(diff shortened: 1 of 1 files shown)


savil:
* main
  * **PR #1525** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1525" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1525?utm_source=stack-comment).

PR #1530, opened 2023-10-04
[lockfile] When updating store-paths, we should update all fields

savil:
## Summary

When we update the store-paths in the lockfile, we should also update the other fields, including "resolved" for the nixpkgs commit hash, and the modified-time.

This way the exact same derivation is used whether installing the binary directly via store-path, or via relying on the derivation in the package's "resolved" field. This ensures parity for the following scenarios:
1. A team has a devbox project. Some team members are using nix >= 2.17, while others are using older nix.
2. The store-path binary is no longer in the nix binary cache, and the package is installed via the older way. This should alleviate @lagoja's concern that there may be a discrepancy between the store-path and the derivation source in "resolved" field of the package.

## How was it tested?

Did `devbox update` and saw that `go@latest` had new store-paths, and the "resolved" field was also updated.

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -3,21 +3,21 @@
   "packages": {
     "go@latest": {
       "last_modified": "2023-09-10T10:53:27Z",
-      "resolved": "github:NixOS/nixpkgs/78058d810644f5ed276804ce7ea9e82d92bee293#go_1_21",
+      "resolved": "github:NixOS/nixpkgs/46688f8eb5cd6f1298d873d4d2b9cf245e09e88e#go_1_21",
       "source": "devbox-search",
       "version": "1.21.1",
       "systems": {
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1530** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1530" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1530?utm_source=stack-comment).

mikeland73 (approved):
I think we should update the entire lock package. The new resolved package will have a different `last_modified` date as well (and possibly other keys in the future).

Pretty much the only thing that can't change is the version because if it did, we already replace the entire object in the `if existing.Version != resolved.Version {` check above.

PR #1534, opened 2023-10-06
[global] global shellenv should not recompute env

mikeland73:
## Summary

This causes the default behavior for `devbox global shellenv` to no longer recompute environment. It adds a new flag: 

`devbox global shellenv -r` which causes the recompute if needed (i.e. not cached)

Recomputing the environment entails two steps:

* Ensuring everything is installed
* Running print-dev-env

It also removed a bunch of `ensurePackagesAreInstalled` calls that are now handled by `nixEnv()`. Note that in all cases, we were calling `ensurePackagesAreInstalled(ensure)` followed by `nixEnv()`

## How was it tested?

* Modified devbox.json manually
* Opened new tab
* Saw message:

```bash
Info: Your devbox environment may be out of date. Please run

eval "$(devbox global shellenv -r)"
```

* Ran `eval "$(devbox global shellenv -r)"`
* Opened another tab and did not see message.

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/samber/lo"
 	"github.com/spf13/cobra"
 
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
@@ -72,7 +73,7 @@ func RootCmd() *cobra.Command {
 	command.AddCommand(servicesCmd())
 	command.AddCommand(setupCmd())
 	command.AddCommand(shellCmd())
-	command.AddCommand(shellEnvCmd())
+	command.AddCommand(shellEnvCmd(lo.ToPtr(false) /*dontRecomputeEnv*/))
 	command.AddCommand(updateCmd())
 	command.AddCommand(versionCmd())
 	// Preview commands
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -22,7 +22,7 @@ type shellEnvCmdFlags struct {
 	preservePathStack bool
 }
 
-func shellEnvCmd() *cobra.Command {
+func shellEnvCmd(recomputeEnvIfNeeded *bool) *cobra.Command {
 	flags := shellEnvCmdFlags{}
 	command := &cobra.Command{
 		Use:     "shellenv",
@@ -30,7 +30,7 @@ func shellEnvCmd() *cobra.Command {
 		Args:    cobra.ExactArgs(0),
 		PreRunE: ensureNixInstalled,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			s, err := shellEnvFunc(cmd, flags)
+			s, err := shellEnvFunc(cmd, flags, *recomputeEnvIfNeeded)
 			if err != nil {
 				return err
 			}
@@ -63,7 +63,11 @@ func shellEnvCmd() *cobra.Command {
 	return command
 }
 
-func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
+func shellEnvFunc(
+	cmd *cobra.Command,
+	flags shellEnvCmdFlags,
+	recomputeEnvIfNeeded bool,
+) (string, error) {
 	env, err := flags.Env(flags.config.path)
 	if err != nil {
 		return "", err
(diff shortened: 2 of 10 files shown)


internal/boxcli/root.go
```diff
@@ -72,7 +73,7 @@ func RootCmd() *cobra.Command {
 	command.AddCommand(servicesCmd())
 	command.AddCommand(setupCmd())
 	command.AddCommand(shellCmd())
-	command.AddCommand(shellEnvCmd())
+	command.AddCommand(shellEnvCmd(lo.ToPtr(false) /*dontRecomputeEnv*/))
```
savil, line 76: this reads as `dontRecomputeEnv` param is set to `false`. I think this is the opposite of what is happening?

```
lo.ToPtr(false) /*recomputeEnvIfNeeded*/)
```
OR
```
// don't recompute the env in shellenv:
command.AddCommand(shellEnvCmd(lo.ToPtr(false)))
```

internal/boxcli/shellenv.go
```diff
@@ -85,7 +89,10 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 		}
 	}
 
-	envStr, err := box.NixEnv(cmd.Context(), flags.runInitHook)
+	envStr, err := box.NixEnv(cmd.Context(), devopt.NixEnvOpts{
+		DontRecomputeEnvironment: !recomputeEnvIfNeeded,
```
savil, line 93: nbd, but I prefer the `recomputeEnvIfNeeded` as a param name. Can we reuse that within `NixEnvOpts` as well?

Alternatively, could it be `skipRecomputeEnvironment`?

internal/impl/devbox.go
```diff
@@ -299,11 +305,6 @@ func (d *Devbox) NixEnv(ctx context.Context, includeHooks bool) (string, error)
 func (d *Devbox) EnvVars(ctx context.Context) ([]string, error) {
 	ctx, task := trace.NewTask(ctx, "devboxEnvVars")
 	defer task.End()
-	// this only returns env variables for the shell environment excluding hooks
```
savil, line 302: hmm I think this comment was actually for `nixEnv`, right? If so, could we leave it in place?

internal/impl/devbox.go
```diff
@@ -913,6 +914,12 @@ func (d *Devbox) nixEnv(ctx context.Context) (map[string]string, error) {
 		usePrintDevEnvCache = true
 	}
 
+	if !usePrintDevEnvCache {
+		if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
+			return nil, err
+		}
```
savil, line 920: 🤔 I think this code will result in running `nix print-dev-env` two times in this scenario.

I believe we can set:
```
usePrintDevEnvCache = true
```
after `ensurePackagesAreInstalled`, because inside `ensurePackagesAreInstalled` we'll call `computeNixEnv(ctx, true)`.

Or, even better directly hardcode below:
`return d.computeNixEnv(ctx, true /*usePrintDevEnvCache*/)`

internal/impl/devbox.go
savil: Its explaining why Mohsen did `nixEnv` and not `NixEnv`.

internal/impl/devbox.go
```diff
@@ -167,9 +167,6 @@ func (d *Devbox) Shell(ctx context.Context) error {
 	ctx, task := trace.NewTask(ctx, "devboxShell")
 	defer task.End()
 
-	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
-		return err
-	}
```
savil, line 172: Should we move `Starting a devbox shell...` to after `nixEnv` so the printed output is as before:
first, nixEnv output about recomputing devbox environment, etc.
then, "Starting a devbox shell....`
then, any init-hook output or shellrc output

savil (approved):
This is great! Just see one of the comments below where I think the current code will run `nix print-dev-env` twice.

internal/impl/devbox.go
```diff
@@ -273,22 +266,35 @@ func (d *Devbox) ListScripts() []string {
...
+	var envs map[string]string
+	var err error
+
+	if opts.DontRecomputeEnvironment {
+		upToDate, _ := d.lockfile.IsUpToDateAndInstalled()
+		if !upToDate {
+			ux.Finfo(
+				d.stderr,
+				"Your devbox environment may be out of date. Please run \n\n"+
+					"eval \"$(devbox global shellenv -r)\"\n\n",
```
savil, line 282: oh two comments on this:
1. could we say: `devbox global shellenv --recompute`? I assume users will copy-paste it so not a problem to be a bit more verbose, and does explain better what it'll do.

2. For fish shells, we need to print a different output:
```
devbox global shellenv -r | source
```

internal/boxcli/shellenv.go
mikeland73: Yeah I struggled with this as well. Here's why I went with this:

The issue with `recomputeEnvIfNeeded` is that we would need to default it to `true`. I prefer if the default is true and only pass in a param to make it false.

The reason that flag is the other way around is because `devbox global shellenv --no-recompute=false` would be super awkward because of the double negative.

PR #1535, opened 2023-10-06
[perf] Replace only-path-without-wrappers with sed

mikeland73:
## Summary

This replaces our existing `only-path-without-wrappers` wrapper implementation to use `sed`. 

Context:

We use `only-path-without-wrappers` to remove bin wrappers in the environment of our binaries. This is particularly useful when a binary calls another binary. See https://github.com/jetpack-io/devbox/pull/1160 for more context. Unfortunately this adds a perf hit to every single binary call. In my informal experiments this cost was ~40ms. For example:

Previously, `time go version` was taking ~60ms even if environment is fully up to date.
Now, `time go version` takes ~20ms.

@Lagoja I'm getting an error in the elixer example (since that was the motivation for the original fix) but it looks like nix elixer on Sonoma is broken: https://elixirforum.com/t/bus-error-after-upgrading-to-sonoma-beta/56354/33 (Our CICD runners will test on Ubuntu and older macOS, Monterrey I believe)

## How was it tested?

* Tested time to call `go version` before and after
* Inspected PATH of shell to ensure `.wrappers` was still there. 
* Copy pasted `sed` code from one of the bin wrappers

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -94,13 +94,14 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 }
 
 func shellEnvOnlyPathWithoutWrappersCmd() *cobra.Command {
+	// Deprecated: will be removed after devbox 0.7.0
+	// Don't add deprecated field to avoid printing anything to stdout
 	command := &cobra.Command{
-		Use:        "only-path-without-wrappers",
-		Deprecated: "This command is deprecated and will be removed after devbox 0.7.0",
-		Hidden:     true,
-		Short:      "[internal] Print shell command that exports the system $PATH without the bin-wrappers paths.",
-		Args:       cobra.ExactArgs(0),
-		PreRunE:    ensureNixInstalled,
+		Use:     "only-path-without-wrappers",
+		Hidden:  true,
+		Short:   "[internal] Print shell command that exports the system $PATH without the bin-wrappers paths.",
+		Args:    cobra.ExactArgs(0),
+		PreRunE: ensureNixInstalled,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			s := shellEnvOnlyPathWithoutWrappersFunc()
 			fmt.Fprintln(cmd.OutOrStdout(), s)
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -95,11 +95,12 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 
 func shellEnvOnlyPathWithoutWrappersCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:     "only-path-without-wrappers",
-		Hidden:  true,
-		Short:   "[internal] Print shell command that exports the system $PATH without the bin-wrappers paths.",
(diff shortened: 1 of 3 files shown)


internal/boxcli/shellenv.go
```diff
@@ -95,11 +95,12 @@ func shellEnvFunc(cmd *cobra.Command, flags shellEnvCmdFlags) (string, error) {
 
 func shellEnvOnlyPathWithoutWrappersCmd() *cobra.Command {
 	command := &cobra.Command{
-		Use:     "only-path-without-wrappers",
-		Hidden:  true,
-		Short:   "[internal] Print shell command that exports the system $PATH without the bin-wrappers paths.",
-		Args:    cobra.ExactArgs(0),
-		PreRunE: ensureNixInstalled,
+		Use:        "only-path-without-wrappers",
+		Deprecated: "This command is deprecated and will be removed after devbox 0.7.0",
```
savil, line 99: lets confirm this prints to Stderr and not Stdout, otherwise existing bin-wrappers may not work.

That said, I suppose we will re-generate the bin-wrappers due to the Devbox version change when this is released, so may be not a concern either way.

internal/wrapnix/wrapper.sh.tmpl
```diff
@@ -25,12 +25,13 @@ eval "$(DO_NOT_TRACK=1 devbox shellenv --preserve-path-stack -c {{ .ProjectDir }
 fi
 
 {{/*
-We call only-path-without-wrappers so that we do not invoke other bin-wrappers from
+Remove wrapper bin path from PATH so that we don't call more bin-wrappers from
 this bin-wrapper. Instead, we directly invoke the binary from the nix store, which
 should be in PATH.
 
-DO_NOT_TRACK=1 can be removed once we optimize segment to queue events.
+This is implemented in sed for efficiency. sed is POSIX so we assume it's available.
```
savil, line 32: oh interesting. Availability was one of the concerns earlier.

savil (approved):
nice!

internal/wrapnix/wrapper.sh.tmpl
mikeland73: @savil if it's a concern we can implement in pure bash. Here's an implementation that only uses bash builtins:

```bash
IFS=: read -ra dirs <<< "$PATH"  # Split PATH into an array using colon as a delimiter

new_PATH=""
for dir in "${dirs[@]}"; do
    if [[ "$dir" != "/path/to/remove" ]]; then
        if [[ -z "$new_PATH" ]]; then
            new_PATH="$dir"
        else
            new_PATH="$new_PATH:$dir"
        fi
    fi
done

export PATH="$new_PATH"
```

internal/boxcli/shellenv.go
mikeland73: Good call. It's actually `OutOrStdErr` which is currently Stderr, but we can't guarantee that. Will remove.

internal/wrapnix/wrapper.sh.tmpl
savil: oh not too bad. That's fairly understandable except for:
```
IFS=: read -ra dirs <<< "$PATH" 
```

:D

savil: but now that makes sense too...

PR #1538, opened 2023-10-06
[minor cleanup] use flake dir instead of flake-file-path in nix print-dev-env

savil:
## Summary

This has been bugging me for many months. Previously, with DEVBOX_DEBUG=1 and doing
some command that evaluated `nix print-dev-env`, we'd see output like:
```
path '/Users/savil/code/jetpack/devbox-projects/lagoja-global/.devbox/gen/flake/flake.nix' does not contain a 'flake.nix', searching up
```

This is because `nix print-dev-env` expects the path to the flake directory but not including the flake.nix itself.

## How was it tested?

changed devbox.json to trigger a re-evaluation of `nix print-dev-env`, and then
`DEVBOX_DEBUG=1 devbox install` and didn't see such output

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -773,7 +773,7 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	maps.Copy(originalEnv, env)
 
 	vaf, err := d.nix.PrintDevEnv(ctx, &nix.PrintDevEnvArgs{
-		FlakesFilePath:       d.nixFlakesFilePath(),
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1538** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1538" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1538?utm_source=stack-comment).

gcurtis (approved):

PR #1539, opened 2023-10-06
devconfig: allow comments and trailing commas

gcurtis:
This allows comments and trailing comments in devbox.json as supported by hujson/JWCC.

```json5
// Top-level comment about the config.
{
	"packages": {
		"go": "1.20",
		
		// Needed to decompress stuff on macOS
		"xz": {
			"version": "latest",
			"platforms": ["aarch64-darwin"],
		},
	},
	"shell": {
		"scripts": {
			/*
			A longer block comment about some build script.
			*/
			"build": "go install ./..",
		},
	},
}
```

As a part of this change, the way the `add`/`rm` commands edit and marshal devbox.json has changed significantly. We can no longer overwrite the previous JSON, since that would lose the user's comments and formatting. There's a new `configAST` type to handle surgical edits to the JSON so that those things aren't lost.

Unmarshalling the config is still done the same way. The docs in `ast.go` have more details.


mikeland73 (commented):
I think this is really interesting, but the complexity is a big concern.

```golang
//   - Marshalling is done by calling configAST.root.Pack to encode the AST as
//     hujson/JWCC. Therefore, any changes to a Config struct will NOT
//     automatically be marshaled back to JSON. Support for modifying a part of
//     the JSON must be explicitly implemented in configAST.
//   - Validation with the AST is complex, so it doesn't do any. It will happily
//     append duplicate object keys and panic on invalid types. The higher-level
//     Config type is responsible for tracking state and making valid edits to
//     the AST.
```

Does this mean every new field we want to add to devbox.json needs custom handling in `configAST`?

If yes, that would be a really big tradeoff in terms of amount of work and maintainability.

gcurtis:
@mikeland73

> Does this mean every new field we want to add to devbox.json needs custom handling in `configAST`?

Only for fields that are edited by the CLI. Fields that are only read by the CLI and edited by the user don't need to go in the AST. In other words, the CLI can't just set a struct field to a new value:

```go
if cfg.SomeField == "foo" {
    cfg.SomeField = "new"
}
b := cfg.Bytes() // SomeField will still be "foo"
```

There's a lot of logic that would be simpler with consistent typing. Much of the code comes from converting packages between an array of strings, a map of strings, and a map of objects while preserving comments/formatting/member order. Other than that, most of it is just boilerplate that's creating or setting JSON values.

I think this feature is worthwhile. We've had a lot of users ask for it, and it's still much simpler than tyson.

internal/devconfig/config_test.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+
+	outBytes := out.Bytes()
+	if _, err := hujson.Parse(outBytes); err != nil {
+		t.Fatalf("loaded default config JSON is invalid: %v\n%s", err, outBytes)
+	}
+	if string(inBytes) != string(outBytes) {
+		t.Errorf("got different JSON after load/save/load:\ninput:\n%s\noutput:\n%s", inBytes, outBytes)
+	}
+}
```
savil, line 614: Do we need sanity checking tests for fields in `shell`?

savil (commented):
Reviewed everything except the `ast.go`. So far, this is looking great to me!

internal/devconfig/init.go
```diff
@@ -33,6 +35,30 @@ func Init(dir string, writer io.Writer) (created bool, err error) {
...
+	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0o644)
+	if errors.Is(err, os.ErrExist) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		if err != nil {
+			os.Remove(file.Name())
```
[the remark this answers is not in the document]
gcurtis: This is only for initializing a new `devbox.json`. The file will only be created if it doesn't already exist (because of `os.O_EXCL`). If something goes wrong I think it makes sense to delete it instead of leaving a partially written `devbox.json`.

internal/devconfig/packages.go
```diff
@@ -1,3 +1,4 @@
+//nolint:dupl // add/exclude platforms needs refactoring
```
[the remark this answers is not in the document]
gcurtis: The linter doesn't like that there's a lot of duplicated lines in the `AddPlatforms` and `ExcludePlatforms` methods. I just didn't want to refactor it as part of this PR to minimize the number of changes.

internal/devconfig/packages.go
```diff
@@ -49,27 +37,33 @@ func (pkgs *Packages) VersionedNames() []string {
...
 // Get returns the package with the given versionedName
 func (pkgs *Packages) Get(versionedName string) (*Package, bool) {
-	name, version := parseVersionedName(versionedName)
-	for _, pkg := range pkgs.Collection {
-		if pkg.name == name && pkg.Version == version {
-			return &pkg, true
-		}
+	name, _ := parseVersionedName(versionedName)
+	i := pkgs.index(name)
+	if i == -1 {
```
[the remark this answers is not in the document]
gcurtis: I didn't think that was possible so I omitted it (it's definitely impossible with the object). I assumed it would complain that the two packages conflict. I can add it back.

internal/devconfig/packages.go
```diff
@@ -81,77 +75,85 @@ func (pkgs *Packages) AddPlatforms(writer io.Writer, versionedname string, platf
...
 	}
-	return errors.Errorf("package %s not found", versionedname)
+	if len(pkg.Platforms) > oldLen {
+		pkgs.ast.appendPlatforms(pkg.name, "platforms", pkg.Platforms[oldLen:])
+	}
+
+	ux.Finfo(writer,
+		"Added platform %s to package %s\n", strings.Join(platforms, ", "),
+		pkg.VersionedName(),
+	)
```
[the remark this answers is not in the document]
gcurtis: I wasn't sure because then we wouldn't print anything if the platforms already existed. I can update if that sounds okay to you.

internal/devconfig/packages_test.go
```diff
@@ -172,42 +159,31 @@ func TestJsonifyConfigPackages(t *testing.T) {
...
-			// the json pretty-printed.
-			expected := &bytes.Buffer{}
-			if err := json.Indent(expected, []byte(testCase.jsonConfig), "", cuecfg.Indent); err != nil {
-				t.Errorf("unexpected error: %v", err)
+				t.Fatal(err)
 			}
-			if string(marshalled) != expected.String() {
-				t.Errorf("expected: %v, got: %v", testCase.jsonConfig, string(marshalled))
+			if string(got) != testCase.jsonConfig {
+				t.Errorf("expected: %v, got: %v", testCase.jsonConfig, string(got))
```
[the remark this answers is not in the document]
gcurtis: It should be checking both. It unmarshals from `testCase.jsonConfig` and then marshals back with `got, err := hujson.Minimize(config.Bytes())` on line 170. It doesn't do `cuecfg.Marshal` anymore because the new code isn't using `cuecfg`.

internal/devconfig/ast.go
```diff
@@ -0,0 +1,282 @@
...
+	// Add a packages field to the root config object and initialize it with
+	// an empty object.
+	rootObject.Members = append(rootObject.Members, hujson.ObjectMember{
+		Name: hujson.Value{
+			Value:       hujson.String("packages"),
+			BeforeExtra: []byte{'\n'},
+		},
+		Value: hujson.Value{Value: &hujson.Object{}},
+	})
+	c.root.Format()
```
savil, line 94: what does Format do?

internal/devconfig/ast.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// appendPackage appends a package to the packages field.
+func (c *configAST) appendPackage(name, version string) {
+	pkgs := c.packagesField(false)
+	switch val := pkgs.Value.Value.(type) {
+	case *hujson.Object:
+		c.appendPackageMember(val, name, version)
+	case *hujson.Array:
+		c.appendPackageElement(val, joinNameVersion(name, version))
```
savil, line 105: hmm gotta admit the distinction between `member` and `element` feels fairly obscure to me. Could we find a more explicit name, such as ... `appendPackageToMap` and `appendPackageToArray`?

internal/devconfig/ast.go
```diff
@@ -0,0 +1,282 @@
...
+func (c *configAST) appendPackageMember(pkgs *hujson.Object, name, version string) {
+	i := c.memberIndex(pkgs, name)
+	if i != -1 {
+		return
+	}
+
+	// Add a new member to the packages object with the package name and
+	// version.
+	pkgs.Members = append(pkgs.Members, hujson.ObjectMember{
+		Name:  hujson.Value{Value: hujson.String(name), BeforeExtra: []byte{'\n'}},
```
savil, line 126: what does the "Extra" represent in BeforeExtra? Trying to understand why it is used.

internal/devconfig/ast.go
```diff
@@ -0,0 +1,282 @@
...
+		return
+	}
+
+	// We need to ensure that the package value is a full object
+	// (not a version string) before we can add a platform.
+	c.convertVersionToObject(&pkgs.Members[i].Value)
+
+	pkgObject := pkgs.Members[i].Value.Value.(*hujson.Object)
+	var arr *hujson.Array
+	if i := c.memberIndex(pkgObject, fieldName); i == -1 {
```
savil, line 193: could benefit from a helper function to be used as:
`if !c.hasMember(pkgObject, fieldName) {`

savil (approved):
This lgtm! I think I can live with some of the increased overhead of needing to some golang code for fields we have the CLI code write to. Do check in with @mikeland73 if his concerns are blocking...

That said, I also advocate for us shipping a change that migrates all users to the map-style packages when they do `devbox update`. That will help reduce the size of code in `ast.go`, and the test-cases, and general complexity. Note, we should plan this a bit so that docs can also be appropriately updated at the right time.

mikeland73:
> I think this feature is worthwhile. We've had a lot of users ask for it, and it's still much simpler than tyson.

FWIW, I don't see this competing with the tyson implementation in https://github.com/jetpack-io/devbox/pull/1555
That PR is tightly scoped so it does not affect users that don't use tyson. It is also very small in terms of code and does not import tyson as a golang dependency at all.

Functionally they are different as well. The tyson PR is mostly to augment `devbox.json`, not replace it.

My concern with this PR is that it replaces a straightforward JSON marshalling with a custom marshaller/unmarshaller. This means we have to maintain a custom marshaller/unmarshaller forever. The cost-benefit of a decision like that needs to be very clear because this would add new overhead to config development and this would be a pretty painful change to revert.

mikeland73 (changes requested):
I'm not a hard no on this PR but would like to chat about it before merging. I also want @loreto's opinion.

gcurtis:
> FWIW, I don't see this competing with the tyson implementation in https://github.com/jetpack-io/devbox/pull/1555

I don’t see it as competing with any PRs either. I’m just saying that out of the discussed potential changes to devbox configuration, this is about as minimal as it gets.

I’d really like to find a way to allow comments in devbox.json. I’m open to ideas, but I don’t see a simpler way to do that that’s also backwards compatible.

loreto:
> I'm not a hard no on this PR but would like to chat about it before merging. I also want @loreto's opinion.

Are you guys asking me to comment on the feature (allowing comments and trailing commas), or on the implementation?

I'm a yes on the feature itself, but I can also take a look at the implementation if that's what you want me to take a look at.

mikeland73 (approved):
Concept Ack. Nice work!

internal/devconfig/ast.go
gcurtis: Generally applies indentation and makes things look nice. If the input has any commas or comments (making it JWCC and not JSON), it also adds trailing commas and aligns fields.

internal/devconfig/ast.go
gcurtis: The extra fields track the whitespace, commas, and comments before/after/inside JSON values. So in ` /* comment */    "foo"` the foo string would have ` /* comment */    ` in its BeforeExtra field.

internal/devconfig/ast.go
gcurtis: We need the index a few lines below to get the key or value.

internal/devconfig/packages.go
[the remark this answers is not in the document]
gcurtis: Updated it to use the name and version everywhere.

internal/devconfig/ast.go
gcurtis: Renamed.

gcurtis:
@weilbith you can add `// comment` and `/* comment */` style comments anywhere in devbox.json as long as you're using  a newer Devbox release.

If your editor allows you to override the syntax, you can tell it to use `jsonc` or `json5`. I know in VSCode you can do this by adding a file association to settings.json. For example:

```
"files.associations": {
  "devbox.json": "jsonc"
}
```

weilbith:
Quick question. Not only regards of comments, readability (?) but also multi line scripts/init_hooks, would YAML be very easy to be supported? I know only very basic Go, but shouldn't most serialization libraries make that a no-brainer? Sorry, just curious. 🙈

PR #1542, opened 2023-10-08
[runx] Add runx validate, lockfile resolve

mikeland73:
## Summary

A few major changes to runx:

* Updates newest runx version to allow for github token, validation, etc. 
* lockfile can now resolve runx packages. Exact versions are saved in lockfile and `install installs deterministic versions according to lockfile information.
* devpkg can now validate runx packages

## How was it tested?

```bash
devbox add runx:golangci/golangci-lint@latest
devbox add runx:golangci/golangci-lint@v1.54.1
devbox add runx:golangci/golangci-lint
devbox rm runx:golangci/golangci-lint
```

Inspected devbox.json and devbox.lock in between every command.


savil:
> Note: While implementing changes I discovered a few API issues in runx that can be improved. Specifically:

this seems to have gotten dropped

internal/devpkg/validation.go
```diff
@@ -1,16 +1,17 @@
...
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/nix"
 )
 
-func (p *Package) ValidateExists() (bool, error) {
+func (p *Package) ValidateExists(ctx context.Context) (bool, error) {
 	if p.IsRunX() {
-		// TODO implement runx validation
-		return true, nil
+		_, err := p.lockfile.Resolve(p.Raw)
```
savil, line 13: I notice that `ctx` is unused, but I think you were trying to add it to the `Resolve` function signature but gave up b/c there are way too many callsites?

If so, no problem. We can complete that TODO in some follow up.

savil (approved):
nice!

mikeland73:
> Note: While implementing changes I discovered a few API issues in runx that can be improved. Specifically:

> this seems to have gotten dropped

I was gonna write something but then decided against it.

internal/devpkg/validation.go
mikeland73: Yeah, it's a bunch. I wanted to prevent this PR from growing more.

PR #1546, opened 2023-10-09
[perf] cache nix.searchSystem

savil:
## Summary

From observing times for adding and removing packages, I notice that a lot of time is spent in `syncPackagesToProfile` and most of that is from `nix.Search`, which is a fairly slow operation. However, from observing the results, it seems that these should be constant. 

For example, here are some search results keyed by the queries:
https://gist.github.com/savil/584c4d9bd6c468aaa8c289265287a23a
Correct me if wrong: all these seem to be constants that won't change, because the queries we run have the nixpkgs-commit-hash in them.

The general [`nix search` command](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-search.html) seems to be slower since the query it accepts is an `installable regex`, which can match a large number of concrete queries. Furthermore, it needs to accept queries like`nixpkgs#vim` and figure out which nixpkgs commit hash this would resolve to. In our case, our queries are much more specific being of the form `nixpkgs/<commit>#attribute`

## How was it tested?

This effect is more pronounced for packages that are not using store-paths. This will affect all users who are still on nix < 2.17, which is a fairly large percentage for now.

I have a devbox global of about 43 packages. For the times below, I ran `devbox global add vim` (where vim is already in my /nix/store), and did `export DEVBOX_FEATURE_REMOVE_NIXPKGS=0`

BEFORE:
syncPackagesToProfile takes about 1 min 30+ seconds

AFTER:
syncPackagesToProfile takes 500ms

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -284,8 +284,25 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	return p.normalizedPackageAttributePathCache, nil
 }
 
+// search attempts to find the package in the local nixpkgs. It may be an expensive
+// call, if it is has not been cached.
+func (p *Package) search(query string) (map[string]*nix.Info, error) {
+	if p.IsDevboxPackage() {
+		if strings.HasPrefix(query, "runx:") {
+			// TODO implement runx search
+			return map[string]*nix.Info{}, nil
+		}
+		// This will be slow if its the first time on the user's machine that this
+		// query is running. Otherwise, it will be cached and fast.
+		return nix.SearchNixpkgsAttribute(query)
+	}
+
+	// fallback to the slow but generalized nix.Search
+	return nix.Search(query)
+}
+
 // normalizePackageAttributePath calls nix search to find the normalized attribute
-// path. It is an expensive call (~100ms).
+// path. It may be an expensive call (~100ms).
 func (p *Package) normalizePackageAttributePath() (string, error) {
 	var query string
 	if p.IsDevboxPackage() {
@@ -302,9 +319,9 @@ func (p *Package) normalizePackageAttributePath() (string, error) {
 		query = p.String()
 	}
 
-	// We prefer search over just trying to parse the URL because search will
+	// We prefer search over just trying to parse the package's "URL" because search will
 	// guarantee that the package exists for the current system.
-	infos, err := nix.Search(query)
+	infos, err := p.search(query)
 	if err != nil {
 		return "", err
 	}
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -270,7 +270,7 @@ func (p *Package) FullPackageAttributePath() (string, error) {
 }
 
 // NormalizedPackageAttributePath returns an attribute path normalized by nix
-// lookupNixInfo. This is useful for comparing different attribute paths that may
+// search. This is useful for comparing different attribute paths that may
 // point to the same package. Note, it may be an expensive call.
 func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	if p.normalizedPackageAttributePathCache != "" {
@@ -284,7 +284,7 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	return p.normalizedPackageAttributePathCache, nil
 }
 
-// normalizePackageAttributePath calls nix lookupNixInfo to find the normalized attribute
+// normalizePackageAttributePath calls nix search to find the normalized attribute
 // path. It may be an expensive call (~100ms).
 func (p *Package) normalizePackageAttributePath() (string, error) {
 	var query string
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -270,7 +270,7 @@ func (p *Package) FullPackageAttributePath() (string, error) {
 }
 
 // NormalizedPackageAttributePath returns an attribute path normalized by nix
-// search. This is useful for comparing different attribute paths that may
+// lookupNixInfo. This is useful for comparing different attribute paths that may
 // point to the same package. Note, it may be an expensive call.
 func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	if p.normalizedPackageAttributePathCache != "" {
@@ -284,24 +284,7 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	return p.normalizedPackageAttributePathCache, nil
 }
 
-// search attempts to find the package in the local nixpkgs. It may be an expensive
-// call, if it is has not been cached.
-func (p *Package) search(query string) (map[string]*nix.Info, error) {
-	if p.IsDevboxPackage() {
-		if strings.HasPrefix(query, "runx:") {
-			// TODO implement runx search
-			return map[string]*nix.Info{}, nil
-		}
-		// This will be slow if its the first time on the user's machine that this
-		// query is running. Otherwise, it will be cached and fast.
-		return nix.SearchNixpkgsAttribute(query)
-	}
-
-	// fallback to the slow but generalized nix.Search
-	return nix.Search(query)
-}
-
-// normalizePackageAttributePath calls nix search to find the normalized attribute
+// normalizePackageAttributePath calls nix lookupNixInfo to find the normalized attribute
 // path. It may be an expensive call (~100ms).
 func (p *Package) normalizePackageAttributePath() (string, error) {
 	var query string
@@ -319,11 +302,24 @@ func (p *Package) normalizePackageAttributePath() (string, error) {
 		query = p.String()
 	}
 
-	// We prefer search over just trying to parse the package's "URL" because search will
-	// guarantee that the package exists for the current system.
-	infos, err := p.search(query)
-	if err != nil {
-		return "", err
+	// We prefer nix.Search over just trying to parse the package's "URL" because
+	// nix.Search will guarantee that the package exists for the current system.
+	var infos map[string]*nix.Info
+	var err error
+	if p.IsDevboxPackage() && !p.IsRunX() {
+		// Perf optimization: For queries of the form nixpkgs/<commit>#foo, we can
+		// use a nix.Search cache.
+		//
+		// This will be slow if its the first time on the user's machine that this
+		// query is running. Otherwise, it will be cached and fast.
+		if infos, err = nix.SearchNixpkgsAttribute(query); err != nil {
+			return "", err
+		}
+	} else {
+		// fallback to the slow but generalized nix.Search
+		if infos, err = nix.Search(query); err != nil {
+			return "", err
+		}
 	}
 
 	if len(infos) == 1 {
--- a/internal/nix/search.go
+++ b/internal/nix/search.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/debug"
@@ -33,11 +32,7 @@ func (i *Info) String() string {
 }
 
 func Search(url string) (map[string]*Info, error) {
-	if strings.HasPrefix(url, "runx:") {
-		// TODO implement runx search
-		return map[string]*Info{}, nil
-	}
-	return searchSystemUsingCache(url, "")
+	return searchSystem(url, "" /* system */)
 }
 
 func parseSearchResults(data []byte) map[string]*Info {
@@ -110,21 +105,27 @@ func searchSystem(url, system string) (map[string]*Info, error) {
 	return parseSearchResults(out), nil
 }
 
+// searchSystemCache is a machine-wide cache of search results. It is shared by all
+// Devbox projects on the current machine. It is stored in the XDG cache directory.
 type searchSystemCache struct {
 	QueryToInfo map[string]map[string]*Info `json:"query_to_info"`
 }
 
 const (
+	// searchSystemCacheSubDir is a sub-directory of the XDG cache directory
 	searchSystemCacheSubDir   = "devbox/nix"
 	searchSystemCacheFileName = "search-system-cache.json"
 )
 
 var cache = searchSystemCache{}
 
-func searchSystemUsingCache(url, system string) (map[string]*Info, error) {
-	if system != "" {
-		return searchSystem(url, system)
-	}
+// SearchNixpkgsAttribute is a wrapper around searchSystem that caches results.
+// NOTE: we should be very conservative in where we use this function. `nix search`
+// accepts generalized `installable regex` as arguments but is slow. For certain
+// queries of the form `nixpkgs/<commit-hash>#attribute`, we can know for sure that
+// once `nix search` returns a valid result, it will always be the very same result.
+// Hence we can cache it locally and answer future queries fast, by not calling `nix search`.
+func SearchNixpkgsAttribute(query string) (map[string]*Info, error) {
 
 	if cache.QueryToInfo == nil {
 		contents, err := readSearchSystemCacheFile()
@@ -134,16 +135,16 @@ func searchSystemUsingCache(url, system string) (map[string]*Info, error) {
 		cache.QueryToInfo = contents
 	}
 
-	if result := cache.QueryToInfo[url]; result != nil {
+	if result := cache.QueryToInfo[query]; result != nil {
 		return result, nil
 	}
 
-	info, err := searchSystem(url, system)
+	info, err := searchSystem(query, "" /*system*/)
 	if err != nil {
 		return nil, err
 	}
 
-	cache.QueryToInfo[url] = info
+	cache.QueryToInfo[query] = info
 	if err := writeSearchSystemCacheFile(cache.QueryToInfo); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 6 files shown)


savil:
* main
  * **PR #1546** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1546" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1546?utm_source=stack-comment).

internal/nix/search.go
```diff
@@ -34,7 +37,7 @@ func Search(url string) (map[string]*Info, error) {
 		// TODO implement runx search
 		return map[string]*Info{}, nil
 	}
-	return searchSystem(url, "")
+	return searchSystemUsingCache(url, "")
```
savil, line 40: I think this is not the best. 

Instead, I'll do:
- revert this callsite to `searchSystem`. 
- rename `searchSystemUsingCache` to `SearchNixpkgAttribute`
- In [normalizePackageAttributePath()](https://github.com/jetpack-io/devbox/blob/b2a883596325a29b207756d325df2afbfe7b8cf8/internal/devpkg/package.go#L289-L303), I'll have `DevboxPackages` call `SearchNixpkgsAttribute`, while non-Devbox Packags continue calling `nix.Search`.

savil: done

internal/devpkg/package.go
```diff
@@ -284,8 +284,25 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
 	return p.normalizedPackageAttributePathCache, nil
 }
 
+// search attempts to find the package in the local nixpkgs. It may be an expensive
+// call, if it is has not been cached.
+func (p *Package) search(query string) (map[string]*nix.Info, error) {
```
mikeland73, line 289: This API is really weird. Having the package as a receiver and also requiring a query. I would suggest a different API, but I'm not sure we need this function at all. See comments below

internal/devpkg/package.go
```diff
@@ -284,8 +284,25 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
...
+	if p.IsDevboxPackage() {
+		if strings.HasPrefix(query, "runx:") {
```
mikeland73, line 291: Does this code path hit? `normalizePackageAttributePath` should never be called on a runx.

Also, there's a `package.IsRunX` function

internal/devpkg/package.go
```diff
@@ -284,8 +284,25 @@ func (p *Package) NormalizedPackageAttributePath() (string, error) {
...
+			// TODO implement runx search
+			return map[string]*nix.Info{}, nil
+		}
+		// This will be slow if its the first time on the user's machine that this
+		// query is running. Otherwise, it will be cached and fast.
+		return nix.SearchNixpkgsAttribute(query)
+	}
+
+	// fallback to the slow but generalized nix.Search
+	return nix.Search(query)
```
mikeland73, line 301: Why the distinction between devbox packages and flakes? Can't you cache the flakes as well?

internal/nix/search.go
```diff
@@ -19,22 +21,18 @@ var (
...
 
 func (i *Info) String() string {
 	return fmt.Sprintf("%s-%s", i.PName, i.Version)
 }
 
 func Search(url string) (map[string]*Info, error) {
-	if strings.HasPrefix(url, "runx:") {
-		// TODO implement runx search
-		return map[string]*Info{}, nil
-	}
```
mikeland73, line 36: Yeah, this is wrong, but I'm not sure we should fix it in this PR?

internal/nix/search.go
```diff
@@ -106,3 +104,85 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+
+var cache = searchSystemCache{}
+
+// SearchNixpkgsAttribute is a wrapper around searchSystem that caches results.
+// NOTE: we should be very conservative in where we use this function. `nix search`
+// accepts generalized `installable regex` as arguments but is slow. For certain
+// queries of the form `nixpkgs/<commit-hash>#attribute`, we can know for sure that
+// once `nix search` returns a valid result, it will always be the very same result.
+// Hence we can cache it locally and answer future queries fast, by not calling `nix search`.
+func SearchNixpkgsAttribute(query string) (map[string]*Info, error) {
```
mikeland73, line 128: Why not just `CachedSearch`

internal/nix/search.go
```diff
@@ -106,3 +104,85 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+
+	cache.QueryToInfo[query] = info
+	if err := writeSearchSystemCacheFile(cache.QueryToInfo); err != nil {
+		return nil, err
+	}
+
+	return info, nil
+}
+
+func readSearchSystemCacheFile() (map[string]map[string]*Info, error) {
```
mikeland73, line 154: Thoughts on using something like https://github.com/jetpack-io/opensource/blob/main/envsec/internal/filecache/filecache.go#L2-L2 ? 

(it's in internal but we can pull it out)

Caching per result is a tiny bit more inefficient but has significant benefits:

* No race conditions
* Per result expiration
* Easy to inspect file system

Using that library this could look like

```golang
var searchCache := filecache.New("jetpack.io")
func CachedSearch(query string) (map[string]*Info, error) {
  if result, err := searchCache.Get(query); if err == nil {
    // unrmarshall and return
  }
  info, err := searchSystem(query, "" /*system*/)
  if err != nil {
    return nil, err
  }
  marshalledInfo, err := json.Marshall(info)
  if err != nil {
    return nil, err
  }
  if err = searchCache.set(query, marshalledInfo, time.Hour * 24 * 30); err != nil {
    return nil, err
  }
  return info, nil
}
```

mikeland73 (changes requested):

internal/devpkg/package.go
mikeland73: It's weird because you are searching for the query (which is determined by the package), but tangentially using the package information to determine how the search will happen as well.

It's a bit like

```golang
mysql := devpkg.New("mysql")
mysql.Search("mysql")
```

* Search should not have a receiver OR should not have a query param.

I'm not sure renaming it addresses the concern.

An alternative:

```golang
func (p *Package) lookupNixInfo() (map[string]*nix.Info, error) {
        var query string
	if p.IsDevboxPackage() {
		if p.isVersioned() {
			entry, err := p.lockfile.Resolve(p.Raw)
			if err != nil {
				return "", err
			}
			query = entry.Resolved
		} else {
			query = p.lockfile.LegacyNixpkgsPath(p.String())
		}
	} else {
		query = p.String()
	}
        return nix.Search(query)
}
```

internal/nix/search.go
mikeland73: Another benefit of reusing code is to not have a custom cache solution and instead reusing code we already have. This would be the main reason.

> The "per result expiration" doesn't apply here. We should only cache for queries that always resolve to the same result.

Disagree, expiration is always good. Especially for performance only cache. If you make expiration 1 month, you know that if you ever change the code, the entire cache will turnover in a month. We also can't rely on nix search returning the same thing forever.

> Again, I don't think this should be CachedSearch for reasons explained above.

I'm not sure I understand the hesitation of naming it `CachedSearch`. It precisely what it is doing. Your above comment says:

> I wanted to keep it specific to search queries I have high confidence should always resolve to the very same result. The nix search api accepts installable regex queries which this doesn't. Hence I chose a very specific name to indicate what scenarios this works with.

But naming it `SearchNixpkgsAttribute` doesn't tell me at all that the result is cached. 

I'm not sure it matters if it doesn't support regex (does it not?) as long as we document it no?

mikeland73:
@savil I just realized the reason this is super slow for you it's because it's downloading `nixpkgs`. We are searching for the "resolved" packages. Previously nixpkgs always existed because you could not install the package otherwise. But now that we allow installing packages without nixpkgs, this function will needlessly download nixpkgs.

I still think caching is a good idea, but now I think how we do profile comparisons for packages with store path is broken. We should not be running `search` on a reference that includes nixpkgs if we don't intend on installing the package using `nixpkgs`

Just to give you an example, if our lockfile includes store paths, then simply comparing the store path on the nix profile and the store path in the lockfile is sufficient. 

@gcurtis, for more context:

 we use search to find a normalized way to compare packages in devbox.json with packages in the nix profile. 

Nix has a set of rules of how it determines the default output. To avoid re-implementing those rules we rely on search. If two flake references with outputs are searched we can compare the results and that way we know we are using same rules as nix.

For example:

`github:F1bonacc1/process-compose#process-compose`
`github:F1bonacc1/process-compose`

Are both the same.

Similarly, nixpkgs can be referred to using `github`, `nixpkgs` or as a devbox package. This function allows us to compare all of these correctly.

Using `eval` could work, but in practice can be slower (~600ms) than search. I think the solution is:

* We need to ensure we never download nixpkgs for a package that was installed via store path
* Add caching to avoid slowish calls to `nix search`

savil:
> @savil I just realized the reason this is super slow for you it's because it's downloading nixpkgs. We are searching for the "resolved" packages. Previously nixpkgs always existed because you could not install the package otherwise. But now that we allow installing packages without nixpkgs, this function will needlessly download nixpkgs.

> I still think caching is a good idea, but now I think how we do profile comparisons for packages with store path is broken. We should not be running search on a reference that includes nixpkgs if we don't intend on installing the package using nixpkgs

> Just to give you an example, if our lockfile includes store paths, then simply comparing the store path on the nix profile and the store path in the lockfile is sufficient.


Ah, no, this is not accurate. In my testing, I had disabled remove-nixpkgs, removed `.devbox`, and installed all packages in the devbox.json. All timings are after doing that setup and then doing `devbox add`.

For remove-nixpkgs, @ipince and I added various higher-level functions (like `NixProfileItem.Matches`) so packages with store-paths don't call this function. Otherwise, you'd be witnessing "Downloading nixpkgs" a lot more!

gcurtis:
Directly-installed packages slowing things down makes sense. Even if it's not downloading a new nixpkgs commit, just evaluating the package for the first time can take a while.

Taking a step back, why are we comparing packages in the profile to the ones in devbox.json? I'm a little unsure about adding caching, because that's what Nix is supposed to do. We should just have the devbox-generated flake refer to the packages and save them to a file.

Here's a small example of a flake that could be generated to track Devbox dependencies:

```nix
{
  description = "A Devbox environment";

  outputs = { self, nixpkgs }: {
    packages.aarch64-darwin.default = derivation {
      name = "devbox-env";
      system = "aarch64-darwin";
      builder = "/bin/bash";
      args = let
        script = builtins.toFile "builder.sh" ''
          echo "$hello
          $go" > $out
        '';
        in [ "${script}" ];

      # Note the packages don't need to be nixpkgs attributes. They could also
      # be direct store paths or `fetchClosure` calls.
      hello = nixpkgs.legacyPackages.aarch64-darwin.hello;
      go = nixpkgs.legacyPackages.aarch64-darwin.go;
    };
  };
}
```

If you run `nix path-info` it spits out all the references:

```json5
// $ nix path-info --json | jq
[
  {
    "deriver": "/nix/store/kpl2nb2b3xwkyw0ma4d4qnxvsz3zllgh-devbox-env.drv",
    "narHash": "sha256-KzYFyXkgG8kWpdHNABX/zu6qchMP7TWylosl8/zirOk=",
    "narSize": 224,
    "path": "/nix/store/c9046f6p7dg0v248cvrfhlqbj5bsy544-devbox-env",
    "references": [
      "/nix/store/c4n3i1nfhvcim30s4ij502cw84yl4vdn-hello-2.12.1",
      "/nix/store/dpnvxm37r1z9vk77mdc3h68b95kxvsji-go-1.20.8"
    ],
    "registrationTime": 1697048631,
    "ultimate": true,
    "valid": true
  }
]
```

internal/devpkg/package.go
savil: We cannot use `nix.Search` caching for flakes, local flakes at least. For now, I'd lean towards being conservative and just support queries of the form `nixpkgs/commit#attribute`

To illustrate the problem, consider the `examples/flakes/php`. This has a package ` "path:my-php-flake#php"`

The `query` for this package is of the form: `<absolute path on laptop>/devbox/examples/flakes/php/my-php-flake#php`

Originally, that flake exported:
```
php = pkgs.php.withExtensions ({ enabled, all }: enabled ++ (with all; [ ds ]));
```
and the `infos` returned from `nix.Search` is ` map[packages.x86_64-darwin.php:php-with-extensions-8.1.19]`

However, inside the package's flake.nix, I can make this edit:
```
 php = pkgs.cowsay;
```
and now the `infos` is `map[packages.x86_64-darwin.php:cowsay-3.7.0]` for the same `query`.

savil (commented):
@mikeland73 switched to using `filecache`. I've left the API as `nix.SearchNixpkgsAttribute` 
- This is a specialized api for queries of the form `nixpkgs/<commit>#attribute`
- I want it distinct from `nix.Search` which is the more general API that supports all queries (`nixpkgs/<commit>#attribute`, all other flake references, etc.)
- open to alternate names, but not to having "cache" in it since that doesn't tell users when to use it, and users shouldn't need to worry about it being cached or not.

@gcurtis I hear your concerns. However, I'm a bit reluctant to change the machinery of how we're currently implementing all of with `devpkg` and nix-profile. That would be a more invasive change.

To your concerns:
> I'm a little unsure about adding caching, because that's what Nix is supposed to do. 

`nix search` is a very general API. This makes it hard for nix to internally optimize it. In our case, we are adding caching for a strict subset of it that we _know_ is always guaranteed to return the same result: `nixpkgs/<commit>#attribute`. 

> We should just have the devbox-generated flake refer to the packages and save them to a file.

The problem with this is that we can't query it while incrementally adding or removing packages quickly

savil:
ping

internal/nix/search.go
```diff
@@ -106,3 +111,78 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+// read as: filecache.NeedsUpdate(err)
+// TODO savil: this should be implemented in the filecache package
+func filecacheNeedsUpdate(err error) bool {
+	return errors.Is(err, filecache.NotFound) || errors.Is(err, filecache.Expired)
+}
+
+// cacheKey sanitizes the search query to be a valid unix filename.
+// This cache key is used as the filename to store the cache value, and having a
+// representation of the query is important for debuggability.
+func cacheKey(input string) string {
```
gcurtis, line 173: Do we handle collisions for queries that have sanitized characters? Using the infamous `emacsPackages.@` package as an example, the input would get cached under the key `emacsPackages.`. Does that mean `devbox add emacsPackages.` would install that package if it was in the cache?

internal/nix/search.go
gcurtis: Could we throw in a regex to verify that the input query is `^github:NixOS/nixpkgs/[0-9a-f]{40}#`? Just to make sure we don't end up caching a ref like `github:NixOS/nixpkgs/nixpkgs-unstable#go_1_19`.

gcurtis (approved):

internal/nix/search.go
savil: good call, done

internal/nix/search.go
savil: Ooh good catch. I added that `@` to the "replace disallowed chars with underscores"

I tried searching the other non-alphanumeric-characters on my standard American keyboard, and no packages showed up so I think its okay (practically speaking). 

Theoretically speaking this scheme is not collision-resistant since two packages with `foo-<disallowed-char-number-one>` and `foo-<disallowed-char-number-two>` would have a collision in the sanitized name. Same for the name-length-truncation we do.

internal/nix/search.go
```diff
@@ -106,3 +111,88 @@ func searchSystem(url, system string) (map[string]*Info, error) {
 	}
 	return parseSearchResults(out), nil
 }
+
+type CachedSearchResult struct {
+	Results map[string]*Info `json:"results"`
+	// Query is added easier to grep for debuggability
+	Query string `json:"query"`
+}
```
mikeland73, line 119: rm

internal/nix/search.go
```diff
@@ -106,3 +111,88 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+
+var allowableQuery = regexp.MustCompile("^github:NixOS/nixpkgs/[0-9a-f]{40}#[^#]+$")
+
+func isAllowableQuery(query string) bool {
```
mikeland73, line 123: nit this function is probably not needed?

internal/nix/search.go
```diff
@@ -106,3 +111,88 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+	const oneYear = 12 * 30 * 24 * time.Hour
+	if err := cache.Set(key, marshalled, oneYear); err != nil {
+		return nil, err
+	}
+
+	return infos, nil
+}
+
+// read as: filecache.NeedsUpdate(err)
+// TODO savil: this should be implemented in the filecache package
```
mikeland73, line 175: agree!

internal/nix/search.go
```diff
@@ -106,3 +111,88 @@ func searchSystem(url, system string) (map[string]*Info, error) {
...
+func filecacheNeedsUpdate(err error) bool {
```
mikeland73, line 176: fwiw I would just name this `isNotFoundOrExpired`

Maybe in filecache, using new `errors.Join` we can return an error that is both NotFound and also something like `CacheMiss`

so `return errors.Join(NotFound, CacheMiss)`

and then here we just check if it is `CacheMiss`

internal/devpkg/package.go
```diff
@@ -302,11 +302,24 @@ func (p *Package) normalizePackageAttributePath() (string, error) {
...
-	// We prefer search over just trying to parse the URL because search will
-	// guarantee that the package exists for the current system.
-	infos, err := nix.Search(query)
-	if err != nil {
-		return "", err
+	// We prefer nix.Search over just trying to parse the package's "URL" because
+	// nix.Search will guarantee that the package exists for the current system.
+	var infos map[string]*nix.Info
+	var err error
+	if p.IsDevboxPackage() && !p.IsRunX() {
```
mikeland73, line 309: Nice! this feels really understandable to me.

mikeland73 (approved):
Looks great! no blocking comments.

internal/nix/search.go
savil: yeah, that sgtm

savil: named it as: `filecacheIsCacheMiss`

PR #1548, opened 2023-10-10
add helpful env vars during shell + run

Lagoja:
## Summary

Adds a few helpful Env Vars to the shell/run/services environment. These should help in situations where a developer needs to refer to files within their project directory, but doesn't want to hardcode the path into their configuration

Since this just adds a few variables to the shell environment, it should be a fairly low risk change

## How was it tested?

Locally

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -764,9 +764,6 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	// Append variables from current env if --pure is not passed
 	currentEnv := os.Environ()
 	env, err := d.parseEnvAndExcludeSpecialCases(currentEnv)
-	env["DEVBOX_PROJECT_ROOT"] = d.projectDir
-	env["DEVBOX_CONFIG_DIR"] = d.projectDir + "/devbox.d"
-	env["DEVBOX_PACKAGES_DIR"] = d.projectDir + "/.devbox/virtenv/.wrappers"
 
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


internal/impl/devbox.go
```diff
@@ -764,6 +764,10 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	// Append variables from current env if --pure is not passed
 	currentEnv := os.Environ()
 	env, err := d.parseEnvAndExcludeSpecialCases(currentEnv)
+	env["DEVBOX_PROJECT_ROOT"] = d.projectDir
+	env["DEVBOX_CONFIG_DIR"] = d.projectDir + "/devbox.d"
+	env["DEVBOX_PACKAGES_DIR"] = d.projectDir + "/.devbox/virtenv/.wrappers"
+
```
savil, line 770: A couple of concerns:
1. We should insert this _after_ the `err != nil` check below.
2. We make a copy of the env on line 786 into `originalEnv`. So we likely want to add it some place after that.

My suggestion is add it before line 864 `configEnv, err := d.configEnvs(ctx, env)` so it can be used to derive other env-vars etc. in the devbox.json config

Lagoja: Good suggestion, updated!

savil (approved):

mikeland73:
One concern I have with this is that in these point to the _last_ devbox project you shellenv'd into. It's not a deal breaker, but just a concern.

mikeland73 (approved):

Lagoja:
> One concern I have with this is that in these point to the _last_ devbox project you shellenv'd into. It's not a deal breaker, but just a concern.

I'm not sure I follow the concern, is there an issue that could cause? Since you can't have nested devbox shells, that should be ok?

PR #1549, opened 2023-10-10
[cicd] Unify lint, add caching

mikeland73:
## Summary

* unified gofumpt and lint
* Added go caching to tests
* Added pre-building devbox binary to be shared between tests

## How was it tested?
CICD

### DIFF
--- a/.github/workflows/cli-tests.yaml
+++ b/.github/workflows/cli-tests.yaml
@@ -208,7 +208,10 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: devbox-${{ runner.os }}
-      - run: chmod +x ./devbox
+      - name: Add devbox to path
+        run: |
+          chmod +x ./devbox
+          sudo mv ./devbox /usr/local/bin/
       - name: Install nix and devbox packages
         run: |
           export NIX_INSTALLER_NO_CHANNEL_ADD=1
@@ -237,7 +240,10 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: devbox-${{ runner.os }}
-      - run: chmod +x ./devbox
+      - name: Add devbox to path
+        run: |
+          chmod +x ./devbox
+          sudo mv ./devbox /usr/local/bin/
       - name: Install Nix
         uses: DeterminateSystems/nix-installer-action@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/cli-tests.yaml
```diff
@@ -85,15 +111,10 @@ jobs:
...
 
       # Use main devbox for now to ensure it supports runx
-      - run:  go run ./cmd/devbox run fmt
-      
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.7.0
-        with:
-          args: "--out-${NO_FUTURE}format colored-line-number --timeout=10m"
-          skip-cache: true
+      - run:  go run ./cmd/devbox run lint
```
gcurtis, line 114: Use the built artifact here too?

.github/workflows/cli-tests.yaml
```diff
@@ -171,17 +197,18 @@ jobs:
...
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v4
+      - name: Download devbox
+        uses: actions/download-artifact@v3
         with:
-          go-version-file: ./go.mod
-      - name: Build devbox
-        run: go install ./cmd/devbox
+          name: devbox-${{ runner.os }}
+      - run: chmod +x ./devbox
```
gcurtis: I think it also needs to be added to the PATH.

.github/workflows/cli-tests.yaml
```diff
@@ -46,6 +46,32 @@ env:
   DEVBOX_DEBUG: 1
 
 jobs:
+  build-devbox:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
```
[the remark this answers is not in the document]
mikeland73: That would be a bit slower no? In this case they can run in parallel?

.github/workflows/cli-tests.yaml
mikeland73: I changed this, but I think longer term we can use the released devbox instead for speed since this is one of the first tests we want to see complete.

mikeland73:
Comments addressed. @gcurtis PTAL

gcurtis (approved):

PR #1551, opened 2023-10-10
[perf UX] add spinner to print-dev-env output

savil:


savil:
* main
  * **PR #1551** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1551" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1551?utm_source=stack-comment).

internal/impl/devbox.go
```diff
@@ -781,11 +783,22 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	originalEnv := make(map[string]string, len(env))
 	maps.Copy(originalEnv, env)
 
+	var spinny *spinner.Spinner
+	if !usePrintDevEnvCache {
+		spinny = spinner.New(spinner.CharSets[11], 100*time.Millisecond, spinner.WithWriter(d.stderr))
+		spinny.FinalMSG = "✓ Got env-vars from nix.\n"
+		spinny.Suffix = " Getting env-vars from nix...\n"
```
mikeland73, line 790: Computing new devbox environment...
...
Devbox environment computed

?

mikeland73 (approved):

internal/impl/devbox.go
savil: Yeah, we could do that. I was trying to indicate to the user that this is nix being slow  (which it is, in this instance) 😃

PR #1553, opened 2023-10-12
[poetry plugin] ensure initHook runs with /bin/sh

### LINKED ISSUE #1550: [Bug]: Incomplete fish support
**Current Behavior (bug)**

Devbox makes a good attempt at supporting fish, but it's not complete enough to work. For example, it generates a fish config in `/tmp` but then in the config it tries to `source` a traditional shell script:

```fish
# in /tmp/devbox614133565/config.fish

# Source the hooks file, which contains the project's init hooks and plugin hooks.
source /home/user/some/path/.devbox/gen/scripts/.hooks.sh
```

This is kind of like... a bash script trying to source a python script. You can see where this falls down when you try to use the `poetry` package. Devbox shows this error in fish after running `devbox shell`:

```
Starting a devbox shell...
~/some/path/.devbox/gen/scripts/.hooks.sh (line 1): $(...) is not supported. In fish, please use '(command)'.
poetry env use $(command -v python) --no-interaction
               ^
from sourcing file ~/Code/Terminus/.devbox/gen/scripts/.hooks.sh
	called on line 162 of file /tmp/devbox614133565/config.fish
from sourcing file /tmp/devbox614133565/config.fish
source: Error while reading file “/home/phil/Code/Terminus/.devbox/gen/scripts/.hooks.sh”
```

This also makes the shell environment incomplete at best, because the fish config is not fully executed to the end.

**Expected Behavior (fix)**

While fish is only partially supported, I would expect that the hooks are not sourced, with a warning printed to screen saying that init hooks aren't supported for fish yet.

**Additional context**

```
Version:     0.6.0
Platform:    linux_amd64
Commit:      fdbd72f230b3520dde60d72c57d9460922ade656
Commit Time: 2023-10-03T18:04:08Z
Go Version:  go1.21.1
Launcher:
```

```json
{
  "packages": [
    "python@3.8",
    "poetry@latest"
  ],
  "shell": {
    "init_hook": [
      "poetry install"
    ],
    "scripts": {
    }
  }
}
```

savil:
## Summary

A user is saying that the poetry plugin's init-hook was failing for them with Fish 
shell. While the init-hook's command does work with modern fish shells, it fails
for older versions of fish. 

In this case, the user is on Ubuntu whose package manager seems to use an older
version of fish.

Regardless, we can ensure the plugin's init-hooks work for consistently if we
ensure that they are written as posix scripts executed via /bin/sh. This PR
seeks to do that.

Fixes #1550

## How was it tested?

```
cd examples/development/poetry/poetry_demo
devbox run run_test
```

### DIFF
--- a/examples/development/python/poetry/poetry-demo/devbox.lock
+++ b/examples/development/python/poetry/poetry-demo/devbox.lock
@@ -3,7 +3,7 @@
   "packages": {
     "poetry@1.4": {
       "last_modified": "2023-05-19T19:44:39Z",
-      "plugin_version": "0.0.1",
+      "plugin_version": "0.0.2",
       "resolved": "github:NixOS/nixpkgs/4a22f6f0a4b4354778f786425babce9a56f6b5d8#poetry",
       "source": "devbox-search",
       "version": "1.4.2",
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -766,7 +766,6 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	// Append variables from current env if --pure is not passed
 	currentEnv := os.Environ()
 	env, err := d.parseEnvAndExcludeSpecialCases(currentEnv)
-
 	if err != nil {
 		return nil, err
 	}
--- a/plugins/poetry.json
+++ b/plugins/poetry.json
@@ -1,15 +1,18 @@
 {
     "name": "poetry",
-    "version": "0.0.2",
+    "version": "0.0.3",
     "readme": "This plugin automatically configures poetry to use the version of python installed in your Devbox shell, instead of the Python version that it is bundled with. Your pyproject.toml must be in the same directory as your devbox.json.",
     "env": {
         "POETRY_VIRTUALENVS_IN_PROJECT": "true",
         "POETRY_VIRTUALENVS_CREATE": "true",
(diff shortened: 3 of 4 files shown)


savil:
* main
  * **PR #1553** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1553" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1553?utm_source=stack-comment).

mikeland73 (approved):
Should all init hooks run in bash? I'm not sure we can claim devbox.json is portable if they don't.

savil:
@mikeland73 currently, we state that the init-hooks run in the shell they are imported into. For `devbox shell`, this is the user's shell. For devbox scripts, this is `sh`. 

I think changing init-hooks to always run in `sh` makes sense for portability. OTOH it would be a breaking change since some users may have written init-hooks that work just fine in their shells, but are not portable for `sh`.  So, we need some plan for introducing this change gradually. 

Perhaps this could be introduced as part of a move of `init-hooks` out of the `"shell"` field in `devbox.json`. If its an opt-in migration process, or a new devbox.json field, then we can insist that the new `init-hooks` field run in `sh`. Can discuss more IRL.

savil:
Ugh, this won't solve the problem. I misunderstood how sourcing works in shells. Will send an update to fix this for older fish shells.

PR #1566, opened 2023-10-17
[sed] Fix bin wrappers when sed is installed with devbox

mikeland73:
## Summary

TSIA

## How was it tested?

Using this config: https://discord.com/channels/903306922852245526/1009933892385517619/1163763578399105064

Pre-change, could not shell in, got bash nesting error.
Post-change, could shell in.

### DIFF
--- a/internal/wrapnix/wrapper.go
+++ b/internal/wrapnix/wrapper.go
@@ -48,6 +48,7 @@ func CreateWrappers(ctx context.Context, args CreateWrappersArgs) error {
 	_ = os.MkdirAll(destPath, 0o755)
 
 	bashPath := cmdutil.GetPathOrDefault("bash", "/bin/bash")
+	sedPath := cmdutil.GetPathOrDefault("sed", "sed")
 
 	if err := CreateDevboxSymlinkIfPossible(); err != nil {
 		return err
@@ -58,6 +59,7 @@ func CreateWrappers(ctx context.Context, args CreateWrappersArgs) error {
 			WrapperBinPath:     destPath,
 			CreateWrappersArgs: args,
 			BashPath:           bashPath,
+			SedPath:            sedPath,
 			Command:            bin,
 			DevboxSymlinkDir:   devboxSymlinkDir,
 			destPath:           filepath.Join(destPath, filepath.Base(bin)),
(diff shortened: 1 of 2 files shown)


Lagoja:
Interesting, so this issue specifically happens if the user installs `sed` via one of the `coreutils` packages?

mikeland73:
> Interesting, so this issue specifically happens if the user installs sed via one of the coreutils packages?

yep, it creates infinite recursion. coreutils also installs `echo` but that doesn't seem to lead to the infinite loop (I guess bash defaults to the built-in). I could pre-emptively add the same logic for `echo` but it doesn't seem to be an issue.

internal/wrapnix/wrapper.go
```diff
@@ -48,6 +48,7 @@ func CreateWrappers(ctx context.Context, args CreateWrappersArgs) error {
 	_ = os.MkdirAll(destPath, 0o755)
 
 	bashPath := cmdutil.GetPathOrDefault("bash", "/bin/bash")
+	sedPath := cmdutil.GetPathOrDefault("sed", "sed")
```
gcurtis, line 51: Won't the default value trigger the bug again? Should the default be `/usr/bin/sed` to mirror `/bin/bash` above?

mikeland73: Yeah if `sed` is not in path this will trigger same issue. Though if `sed` is not in path the wrapper would have failed anyway. `sed` is POSIX so I kinda assume it exists.

There's an additional problem that affects `sed` and `bash`. If you are using global and a project devbox and both install sed and/or bash, you could end up cycling between both wrappers recursivelly. For example:

* Add coreutils to project (bin wrappers use `/usr/bin/sed`)
* Add coreutils to global while shelled in to project (bin wrappers use `sed` in project)
* Make some change to project, now sed in bin wrapper points to global sed
* Cycle 

Another edge case:

* if 2 devbox processes are creating wrappers at the same time, theres a chance the second one will point to the bin wrapper incorrectly.

These are pretty edge cases, but would suck. I think a better solution is to just special case `sed` and `bash` and never bin wrap them. Thoughts?

mikeland73: Follow up https://github.com/jetpack-io/devbox/pull/1571

PR #1567, opened 2023-10-17
[poetry plugin] enable pyproject-dir to be configurable

### LINKED ISSUE #1563: [Bug]: Repo with poetry submodule failing since 0.7
**Current Behavior (bug)**
I have a project with 2 submodules, one is a python project for a rest api backend and the other a pnpm project for the frontend SPA.

I created the project with following structure:

```
+ myproject
  + devbox.json
  +- service
  |  + pyproject.toml
  |  + ...
  +- frontend
     +- package.json
     +- ...
```

My devbox.json:

```
{
  "packages": [
    "nodejs@latest",
    "poetry@latest",
    "python3@latest",
    "nodePackages.pnpm@latest",
    "redis@latest"
  ],
  "shell": {
    "init_hook": [
    ],
    "scripts": {
      "agent": [
        "cd service && poetry run uvicorn service:app --reload"
      ],
      "frontend": [
        "pnpm --dir frontend dev"
      ],
      "install": [
        "pnpm --dir frontend install",
        "cd service",
        "poetry install"
      ],
      "test": [
        "cd service && poetry run pytest -vv"
      ]
    }
  }
}
```

This was working fine in 0.5.13 but when I updated to 0.7.0 I started getting this error and can't use any of defined scripts:

```
Poetry could not find a pyproject.toml file in /Users/user/myproject or its parents
```

Can you provide a fix? Or alternate project structure, configuration, or workaround we should use when having a mono repo (a repo with multiple sub "environments") like this example?

Regards

savil:
## Summary

We can enable the poetry project to exist in a sub-directory of the devbox-project by allowing a user to set `DEVBOX_PYPROJECT_DIR`.

Note, this only works if there is a _single_ poetry project in the devbox-project. For multiple poetry projects in a monorepo, I think users may need to define distinct devbox-projects.

Fixes #1563 

## How was it tested?

In the added example, ran `devbox run run_test`. Previously, this would fail.

Also verified the default directory is the project dir, by removing the DEVBOX_PYPROJECT_DIR env-var, and seeing this error message:
```
examples/development/python/poetry/poetry-pyproject-subdir/service> devbox run run_test

Poetry could not find a pyproject.toml file in /Users/savil/code/jetpack/devbox/examples/development/python/poetry/poetry-pyproject-subdir or its parents
Error: error running script "run_test" in Devbox: exit status 1
```


savil:
* main
  * **PR #1567** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1567?utm_source=stack-comment).

mikeland73 (approved):
I think this is OK, but does it solve the bug in question? I think the user wanted to use this in a multi pyproject repo. 

I feel like a better solution is to only run `poetry env use` if there's a pyproject file 

(or maybe a combination of both)

savil:
> I think this is OK, but does it solve the bug in question? I think the user wanted to use this in a multi pyproject repo.
 
Yes, it solves it. The user has a single poetry `service` folder.

savil:
@mikeland73 I'm not a fan of the conditional check, because then the poetry plugin is less robust. It'll use the python-from-devbox-json in some circumstances and not in others, which could lead to more subtle bugs.

PR #1569, opened 2023-10-18
[telemetry] fix segfault

savil:
## Summary

`newSentryException` can error [here](https://github.com/jetpack-io/devbox/blob/main/internal/telemetry/sentry.go#L55) due to `err` being `nil`.

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -113,8 +113,8 @@ func commandEvent(meta Metadata) (id string, msg *segment.Track) {
 }
 
 // Error reports an error to the telemetry server.
-func Error(err error, meta Metadata) {
-	if !started || err == nil {
+func Error(errToLog error, meta Metadata) {
+	if !started || errToLog == nil {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


savil:
* main
  * **PR #1569** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1569?utm_source=stack-comment).

internal/telemetry/telemetry.go
```diff
@@ -113,8 +113,8 @@ func commandEvent(meta Metadata) (id string, msg *segment.Track) {
 }
 
 // Error reports an error to the telemetry server.
-func Error(err error, meta Metadata) {
-	if !started || err == nil {
+func Error(errToLog error, meta Metadata) {
```
gcurtis, line 116: Could we leave this one as `err` just to keep the package API clean?

gcurtis (approved):
Could you also add a test for `Error` for this? It would just need to set `segmentBufferDir` and `sentryBufferDir` to a temp dir so that they don't actually get sent to Segment/Sentry. I should've added some basic tests back when this was created.

savil:
@gcurtis good call. I added a very very basic test that would fail without this PR's code changes.

Issue #1565, opened 2023-10-17
[Bug]: NGINX `envsubst` regression in `0.7.0`


jrseliga:
## Current Behavior (bug)
After upgrading to `0.7.0`  the variables in the file passed `--env-file` option of `services up` are not replaced by the `envsubst` command

## Expected Behavior (fix)
NGINX plugin `envsubst` respects `--env-file`

### Given
`.env.devbox`
```
NGINX_WEB_PORT=8004
```

`nginx.template`
```
# ..
server {
         # ..
         listen       $NGINX_WEB_PORT;
         # ..
}
# ..
```

### Unexpected outcome on `0.7.0`
`DEVBOX_USE_VERSION=0.7.0 devbox services up --env-file .env.devbox`

Results in (unexpected):
```
# ..
server {
         # ..
         listen       8081;
         # ..
}
# ..
```
### Expected outcome on `0.6.0`
`DEVBOX_USE_VERSION=0.6.0 devbox services up --env-file .env.devbox`

Results in (expected):
```
# ..
server {
         # ..
         listen       8004;
         # ..
}
# ..
```

Lagoja:
Verified that I can repro on my machine. Looking into this now

PR #1570, opened 2023-10-18
[services] Fix issue where env var is ignored

mikeland73:
## Summary

Fixes https://github.com/jetpack-io/devbox/issues/1565

The bugs:
* env vars specified via `--env-file` (and possibly via other methods) are ignored in nginx config.
* When inside a devbox shell or a direnv context, starting services would ignore env vars.

Intermediate causes:
* `envsubst` binary was a bin wrapped binary. This was resetting the env vars. Fix: Never use bin wrappers when inside a `devbox run` (This was previously only scoped to using the CLI command explicitly, but now we do it any time we use RunScript function)
* Previously we would only run services in new shell if we were not already in a shell or direnv. But this causes us to ignore environment variables that are set after the environment is created due to bin wrappers. Fix: always run services in new shell.

Root cause:
* I'm not 100% sure what caused this. It's possible the [PATH stack PR](https://github.com/jetpack-io/devbox/commit/c11031a034bf4b84aa902be64ab91f59388dffb1) did, but I don't fully understand the mechanism. I'm still confident the changes in this PR are good and simplify uncertainty about how services are called.

Sorry in advance for all the boolean params. We should move to structs, but I didn't want to make those changes here.

## How was it tested?

* Copied setup from https://github.com/jetpack-io/devbox/issues/1565. 
* Tried changing port and restarting services.
* Tested outside of devbox shell
* Tested inside of devbox shell
* Tested with direnv enabled
* Tested with direnv disabled

### DIFF
--- a/examples/stacks/lapp-stack/devbox.json
+++ b/examples/stacks/lapp-stack/devbox.json
@@ -7,7 +7,8 @@
     "apache@2.4"
   ],
   "env": {
-    "PGPORT": "5432"
+    "PGPORT": "5432",
+    "PGHOST": "/tmp/devbox/lapp"
   },
   "shell": {
     "scripts": {
@@ -19,7 +20,6 @@
       "init_db": "initdb",
       "run_test": [
         "mkdir -p /tmp/devbox/lapp",
-        "export PGHOST=/tmp/devbox/lapp",
         "initdb",
         "devbox services start",
         "echo 'sleep 1 second for the postgres server to initialize.' && sleep 1",
--- a/examples/stacks/lepp-stack/devbox.json
+++ b/examples/stacks/lepp-stack/devbox.json
@@ -9,7 +9,8 @@
   "env": {
     "NGINX_WEB_PORT": "8089",
     "NGINX_WEB_ROOT": "../../../my_app",
-    "PGPORT": "5433"
+    "PGPORT": "5433",
+    "PGHOST": "/tmp/devbox/lepp"
   },
   "shell": {
     "scripts": {
@@ -22,6 +23,7 @@
       "run_test": [
         "mkdir -p /tmp/devbox/lepp",
         "export PGHOST=/tmp/devbox/lepp",
+        "rm -rf .devbox/virtenv/postgresql/data",
         "initdb",
         "devbox services up -b",
         "echo 'sleep 2 second for the postgres server to initialize.' && sleep 2",
--- a/examples/stacks/lepp-stack/devbox.json
+++ b/examples/stacks/lepp-stack/devbox.json
@@ -22,7 +22,6 @@
       "init_db": "initdb",
       "run_test": [
         "mkdir -p /tmp/devbox/lepp",
-        "export PGHOST=/tmp/devbox/lepp",
         "rm -rf .devbox/virtenv/postgresql/data",
         "initdb",
         "devbox services up -b",
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -48,7 +48,13 @@ func servicesCmd(persistentPreRunE ...cobraFunc) *cobra.Command {
 	serviceStopFlags := serviceStopFlags{}
 	servicesCommand := &cobra.Command{
 		Use:   "services",
-		Short: "Interact with devbox services",
+		Short: "Interact with devbox services.",
+		Long: "Interact with devbox services. Services start in a new shell. " +
+			"Plugin services use envrinment variables specified by plugin unless " +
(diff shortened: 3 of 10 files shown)


examples/stacks/lapp-stack/devbox.json
```diff
@@ -19,7 +20,6 @@
       "init_db": "initdb",
       "run_test": [
         "mkdir -p /tmp/devbox/lapp",
-        "export PGHOST=/tmp/devbox/lapp",
```
savil, line 22: TODO savil: come back to this change. Refer to slack convo b/w mike and john

examples/stacks/lepp-stack/devbox.json
```diff
@@ -22,6 +23,7 @@
       "run_test": [
         "mkdir -p /tmp/devbox/lepp",
         "export PGHOST=/tmp/devbox/lepp",
+        "rm -rf .devbox/virtenv/postgresql/data",
```
savil, line 26: why is this needed?

examples/stacks/lepp-stack/devbox.json
```diff
@@ -9,7 +9,8 @@
   "env": {
     "NGINX_WEB_PORT": "8089",
     "NGINX_WEB_ROOT": "../../../my_app",
-    "PGPORT": "5433"
+    "PGPORT": "5433",
+    "PGHOST": "/tmp/devbox/lepp"
```
savil, line 13: do we need this if we're exporting PGHOST in run_test?

internal/boxcli/services.go
```diff
@@ -47,7 +48,13 @@ func servicesCmd(persistentPreRunE ...cobraFunc) *cobra.Command {
 	serviceStopFlags := serviceStopFlags{}
 	servicesCommand := &cobra.Command{
 		Use:   "services",
-		Short: "Interact with devbox services",
+		Short: "Interact with devbox services.",
+		Long: "Interact with devbox services. Services start in a new shell. " +
+			"Plugin services use envrinment variables specified by plugin unless " +
```
savil, line 53: typo: envrinment

internal/impl/devbox.go
```diff
@@ -526,15 +538,22 @@ func (d *Devbox) Services() (services.Services, error) {
 	return result, nil
 }
 
-func (d *Devbox) StartServices(ctx context.Context, serviceNames ...string) error {
-	if !d.IsEnvEnabled() {
```
savil, line 530: 🤔 I did re-implement this function in the PathStack PR. I wonder if that update caused this bug.

internal/impl/envpath/pathlists.go
```diff
@@ -35,3 +36,22 @@ func JoinPathLists(pathLists ...string) string {
 	}
 	return strings.Join(cleaned, string(filepath.ListSeparator))
 }
+
+func RemoveFromPath(path, pathToRemove string) string {
+	paths := strings.Split(path, string(os.PathListSeparator))
```
savil, line 41: nit: can use `filepath.SplitList`

internal/impl/envpath/pathlists.go
```diff
@@ -35,3 +36,22 @@ func JoinPathLists(pathLists ...string) string {
...
+
+	// Iterate through the paths and add them to the newPaths slice if they are not equal to pathToRemove
+	for _, p := range paths {
+		if p != pathToRemove {
+			newPaths = append(newPaths, p)
+		}
+	}
+
+	// Join the modified paths using ":" as the delimiter
+	newPath := strings.Join(newPaths, string(os.PathListSeparator))
```
savil, line 54: nit: can use `filepath.Join`

savil: line 37 too :)

savil (commented):
Made a quick pass. No major concerns, except for the `lepp-stack` changes, which I don't follow.

I didn't quite understand how it all fits together, so need to make another review pass.
I need to pause for 🐕 walk and dinner. Will re-review soon...

examples/stacks/lepp-stack/devbox.json
mikeland73: This just makes testing easier by making `run_test` more idempotent.

internal/impl/envpath/pathlists.go
mikeland73: `filepath.Join` is for directory paths! not for path lists!

e.g. /path/to/my/file vs path1:path2:path3

savil: lol

savil (approved):

PR #1571, opened 2023-10-18
[wrappers] Always use system bash and sed in bin wrappers

mikeland73:
## Summary

This feels like the safest solution to https://github.com/jetpack-io/devbox/pull/1566#discussion_r1364318788

It also avoids creating bin wrappers for `bash` and `sed` which reduces risk and improves performance.

## How was it tested?

Used setup from https://github.com/jetpack-io/devbox/pull/1566

```
devbox shell
devbox add bash
hash -r
which bash # ensure I get profile bash instead of wrapped bash
```

### DIFF
--- a/internal/boxcli/root.go
+++ b/internal/boxcli/root.go
@@ -17,6 +17,7 @@ import (
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/midcobra"
 	"go.jetpack.io/devbox/internal/cloud/openssh/sshshim"
+	"go.jetpack.io/devbox/internal/cmdutil"
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/telemetry"
 	"go.jetpack.io/devbox/internal/vercheck"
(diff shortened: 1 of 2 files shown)


gcurtis:
Does this affect the user's devbox environment or just the wrappers? I think it would be surprising for users that explicitly want the coreutils version of sed. The macOS sed and the coreutils sed have some big differences in how they handle flags.

What if we do the following:

1. When devbox starts, check to see if `DEVBOX_SYSTEM_BASH` and `DEVBOX_SYSTEM_SED` are set. 2. If they aren't, find them in the PATH (before we modify it at all) and then set the env vars to their absolute paths.
3. When generating the wrappers, use those env var values. If for some reason the env vars are missing (for example, because the user unset them), assume `/bin/bash` and `/usr/bin/sed`.

mikeland73:
@gcurtis sure I think that's a reasonable solution, but to be honest there's absolutely no benefit to generating wrappers for bash and sed. Why not avoid it?

mikeland73:
FWIW, I'm pretty sure coreutils sed/bash works fine for our use case. But agree it's probably not great that the user can change which binaries we use. On the other hand if `sed` is not fully portable, we are not sure what binaries we're using either.

gcurtis (approved):
I might be misunderstanding. I thought that if we didn't wrap bash or sed then the user's devbox environment would end up using their OS's version of those program. If ~~it's~~ the user's devbox shell is still using the nix-installed bash and sed, then this looks good to me.

mikeland73:
@gcurtis no, if we remove the wrapper, then we use binaries in the devbox nix profile.

PR #1573, opened 2023-10-19
Update process-compose.yml in rails stack

dallasgoldswain:
Corrected it to reflect rails and not django

## Summary
Changed `django` to `rails`

## How was it tested?
No test needed, comment update


Lagoja:
🤦

Lagoja (approved):

PR #1574, opened 2023-10-19
devconfig,shellgen: option to patch ELF binaries with newer glibc

gcurtis:
Add an optional `patch_glibc` boolean field to packages in devbox.json. When true, devbox will patch any ELF binaries in a package to link against the latest `nixpkgs#glibc` at runtime. This is a workaround for dynamic linking issues that can arise when using older packages.

For example, an old Python interpreter that's linked against glibc 2.35 might run a script that loads a native library that requires glibc 2.37. Because the linker only loads a library into a process once (regardless of version), linking will fail when the script references a 2.37 symbol.

The following devbox.json will reproduce a glibc version bug when the `virtenv` and `crash` scripts are run on x86-64-linux (aarch64 will not work). Setting the `patch_glibc` field to true will fix it:

```json
{
  "packages": {
    "binutils": "latest",
    "libpqxx": "latest",
    "libxcrypt": "latest",
    "libz": "latest",
    "python37Packages.pip": "latest",
    "python": {
      "version": "3.7",
      "patch_glibc": false
    }
  },
  "shell": {
    "scripts": {
      "crash": "python3 -c 'import psycopg2'",
      "virtenv": [
        "rm -rf \"$VENV_DIR\"",
        "echo \"python3 path: $(readlink -f $(command -v python3))\"",
        "python3 -m venv \"$VENV_DIR\"",
        ". \"$VENV_DIR/bin/activate\"",
        "echo \"pip3 path: $(readlink -f $(command -v pip3))\"",
        "pip3 install psycopg2==2.9.5"
      ]
    }
  }
}
```

This field is intended to be a "last resort" for when a package cannot be updated to a newer version. Upgrading to a newer version of Python in the example above is preferable.


internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -29,25 +37,59 @@
...
+
+	{{ if .PatchGlibc -}}
+        patchGlibc = pkg: derivation {
+          name = "devbox-patched-glibc";
+          system = "{{.System}}";
+
+          # Set these attributes so that glibc-patch.bash has access to their
+          # paths via environment variables of the same name.
+          inherit pkg;
+          glibc = nixpkgs.legacyPackages."{{.System}}".glibc;
```
savil, line 62: I suggest using `stdenv.cc.libc_lib`.  I don't know if `glibc` nix package is always reliably updated. For example, glibc 2.36 versions are missing https://www.nixhub.io/packages/glibc

internal/shellgen/tmpl/glibc-patch.bash
```diff
@@ -0,0 +1,99 @@
...
+cp -R "$pkg" "$out"
+
+# Find the new linker that we'll patch into all of the package's executables as
+# the interpreter.
+interp="$(find "$glibc/lib" -type f -maxdepth 1 -executable -name 'ld-linux-*.so*' | head -n1)"
+readonly interp
+
+patch() {
+	declare -r binary="$1" # ELF binary to patch
+	declare -r type="$2"   # "exe" or "lib"
```
savil, line 38: IIUC we only need this for `exe` files, since the linker-loader will cache the glibc-lib paths even when resolving NEEDED for any `lib` that are encountered later.

Was there a specific reason to include `lib` as well, or more for consistency?

internal/shellgen/tmpl/glibc-patch.bash
```diff
@@ -0,0 +1,99 @@
...
+	fi
+
+	perm=$(stat -c "%a" "$binary")
+	old_rpath="$(patchelf --print-rpath "$binary")"
+	new_rpath="$glibc/lib${old_rpath:+:$old_rpath}"
+
+	echo "running patchelf file=\"$binary\" type=\"$type\" rpath=\"$new_rpath\" perm=\"$perm\""
+	chmod u+w "$binary"
+	patchelf --set-rpath "$new_rpath" \
+	         --add-needed libBrokenLocale.so.1 \
```
savil, line 52: should `ld-linux-x86-64.so.2` (or equivalent for aarch64) also be in this list? Is there never a chance it is used as a library by some executable?

internal/shellgen/tmpl/glibc-patch.bash
```diff
@@ -0,0 +1,99 @@
...
+	declare -r type="$2"   # "exe" or "lib"
+
+	declare set_interpreter
+	if [[ "$type" == "exe" ]]; then
+		set_interpreter=true
+	fi
+
+	perm=$(stat -c "%a" "$binary")
+	old_rpath="$(patchelf --print-rpath "$binary")"
+	new_rpath="$glibc/lib${old_rpath:+:$old_rpath}"
```
savil, line 47: TIL about this conditional parameter expansion syntax!

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -29,25 +37,59 @@
...
+
+          overlays = [
+            {{- range $flake.Packages }}
+            {{- if .PatchGlibc }}
+            (final: prev: {
+              {{.PackageAttributePath}} = prev.{{.PackageAttributePath}};
+            })
+            {{- end }}
+            {{- end }}
+          ];
```
savil, line 49: Can you comment on the tradeoffs between using an overlay versus a flake as done in https://github.com/savil/devbox-issue-1276? What led you down this path?

savil (commented):
Nice to get to review this!

I'd like to better understand why the python plugin wasn't working. In  https://github.com/savil/devbox-issue-1276, I hadn't needed to reset the venv-dir.

internal/shellgen/generate.go
```diff
@@ -178,8 +214,9 @@ func makeFlakeFile(d devboxer, outPath string, plan *flakePlan) error {
 		return errors.WithStack(err)
 	}
 
-	// add the flake.nix file to git
-	cmd = exec.Command("git", "-C", flakeDir, "add", "flake.nix")
+	// Any files that flake.nix needs at build time must be in git.
+	// Otherwise, Nix won't copy it into the flake's build environment.
+	cmd = exec.Command("git", "-C", flakeDir, "add", ".")
```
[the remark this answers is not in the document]
gcurtis: I don't think so. Aren't we using the flake.lock right now anyway?

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -3,9 +3,14 @@
 
    inputs = {
      nixpkgs.url = "{{ .NixpkgsInfo.URL }}";
+
      {{- range .FlakeInputs }}
      {{.Name}}.url = "{{.URLWithCaching}}";
      {{- end }}
+
+     {{- if .PatchGlibc }}
+     nixpkgs-unstable.url = "nixpkgs";
```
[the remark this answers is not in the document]
gcurtis: It should be locked by the flake.lock file. I'm not as familiar with the updating logic, but I think it should only change when `devbox update` is run.

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
gcurtis: I specified `glibc` because the patch would break if the stdenv libc was different (for example, it could be musl).

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
gcurtis: So this overlay is actually wrong and could explain why virtualenvs are finding the wrong Python interpreter. I meant to set it to the patched version of the package.

From how I understand it, the overlay overrides the package for any other packages that depend on it. So if a python37Packages.package is installed, it should use the overridden python from the overlay. I'll fix this and see if it addresses the virtual env stuff.

internal/shellgen/tmpl/glibc-patch.bash
gcurtis: Good point. I forgot that we decided to skip the libraries. I'll skip those and re-test.

internal/shellgen/tmpl/glibc-patch.bash
gcurtis: I don't think it's necessary because the interpreter should just be using itself. If this turns out to be wrong I can add it back.

gcurtis:
@savil PTAL. I fixed the overlay, removed the lib patching, and added a system check.

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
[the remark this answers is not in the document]
savil: So, generally speaking, `devbox update` doesn't affect the `flake.lock` directly. It does affect how the `flake.nix`'s inputs are defined, and in that manner will affect the `flake.lock` file.

I think this could be problematic. My understanding is:
1. In general, in a flake, `nixpkgs` will resolve to the current `nix channel` for the user. 
2. The `flake.lock` will be generated from that. For users on a team, this could pin `nixpkgs` to different commit-hashes in the generated flake.lock, assuming they are even on the same nix channel.

Let me think some more (have to go to a meeting, jsut sharing my prelim. thoughts)

savil (commented):
Generally lgtm, assuming overlays is how we wanna go. I think the `nixpkgs` reference in the generated flake.nix is problematic due to huirting reproducibility. Will circle back shortlky....

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
mikeland73, line 12: Make this a hard coded reference.

mikeland73: As discussed:

avoid creating `nixpkgs-unstable` and just use `nixpkgs` from above. Just set `nixpkgs` to be the hard coded commit you want in go code.

mikeland73 (approved):
Recapping offline discussion:

This PR works because bin wrappers point to the nix store directly (and skip the profile). We are wanting to get rid of bin wrappers soon, so we don't want to rely on this functionality.

For a fix in a follow up:

* Move glibc patching to a flake (one flake per package that requires patching)
* Revert top level flake so it no longer has awareness of patching (we decided for now to not do the overlay)
* You can modify `AllInstallablePackages` (similar to how `ProcessPluginPackages` works) to replace the original package with the local flake. This will ensure the package gets installed to profile and also included in the top level flake. You should not need any custom code to handle the patching flake inside the top level flake.

internal/shellgen/flake_input.go
```diff
@@ -47,28 +47,33 @@ func (f *flakeInput) PkgImportName() string {
...
-		return attributePath
-	})
-	if err != nil {
-		return nil, err
+type buildInput struct {
+	AttrPath   string
+	PatchGlibc bool
+}
+
+func (f *flakeInput) BuildInputs() ([]buildInput, error) {
```
mikeland73, line 55: You should be able to revert this as well once you do the individual flake approach since `PatchGlibc` is not needed in the top level flake.

gcurtis:
@savil @mikeland73 please take another look.

- Patching now happens in a separate `.devbox/gen/flake/glibc-patch` flake.
- Any packages that need to be patched have their flake input overridden to refer to the glibc-patch flake.
- There's some duplication between the top-level flake template and the glibc flake template to handle cached packages and initializing each nixpkgs input.
- The virtual environment bug still happens, which I think should go away once wrappers are removed.

internal/shellgen/generate.go
```diff
@@ -50,6 +51,15 @@ func GenerateForPrintEnv(ctx context.Context, devbox devboxer) error {
 		return errors.WithStack(err)
 	}
 
+	if plan.needsGlibcPatch() {
+		patch, err := newGlibcPatchFlake(devbox.Config().NixPkgsCommitHash(), plan.Packages)
```
savil, line 55: Lets ship this. 

To flesh this out further, we should think more about how to enable configuring the `devbox.Config().NixPkgsCommitHash()` parameter. The field it uses is one we've been hoping to deprecate. Its not great to use a value that was there for a completely different reason or is used in a myriad of different use-cases, since users will not know what _other_ side-effects may arise if they change the commit-hash's value.

One suggestion is:
```
// devbox.json
{
    "meta-packages": {
           "patchGlibc": <glibc-installable> or `glibc@<version>`
     }
}
```
This would let users update it via `devbox update` as well.

internal/shellgen/generate.go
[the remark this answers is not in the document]
savil: could we revert this change, unless there's a real reason to make it? I'd rather we track the minimal set of files that we need to. That makes it easier to reason about the system.

internal/shellgen/tmpl/glibc-patch.nix.tmpl
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  outputs = args@{ self, nixpkgs-glibc {{- range $name, $_ := .Inputs -}}, {{ $name }} {{- end }} }:
+    {{ with .Outputs -}}
+    let
+      # Initialize each nixpkgs input into a new attribute set with the
+      # schema "pkgs.<input>.<system>.<package>".
+      #
+      # Example: pkgs.nixpkgs-80c24e.x86_64-linux.python37
+      pkgs = builtins.mapAttrs (name: flake:
```
savil, line 19: where is this `name` used? i.e. in the Example provided above, where is the `nixpkgs-80c24e` being set?

internal/shellgen/tmpl/glibc-patch.nix.tmpl
```diff
@@ -0,0 +1,64 @@
...
+        if builtins.hasAttr "legacyPackages" flake then
+          {
+            {{- range $system, $_ := .Packages }}
+            {{ $system }} = (import flake {
+              system = "{{ $system }}";
+              config.allowUnfree = true;
+              config.allowInsecurePredicate = pkg: true;
```
savil, line 26: would be nice to have this respect the `boolean` value from the `devpkg.Package.AllowInsecure()`

internal/shellgen/tmpl/glibc-patch.nix.tmpl
```diff
@@ -0,0 +1,64 @@
...
+          {{- end }}
+        };
+        {{- end }}
+      };
+
+      formatter = {
+        {{- range $system, $_ := .Packages }}
+        {{ $system }} = nixpkgs-glibc.legacyPackages.{{ $system }}.nixpkgs-fmt;
+        {{- end }}
+      };
```
savil, line 61: whats the formatter for? where is it used?

savil (approved):
Excited to ship a fix for glibc dynamic linking issues!

Could you add an example devbox-project under `examples/` that exercises the patch glibc? That way, we'll ensure good CICD coverage and also have a ready example to debug issues if someone reports something related to patch-glibc.

internal/shellgen/flake_input.go
```diff
@@ -54,6 +57,9 @@ func (f *flakeInput) BuildInputs() ([]string, error) {
 		if attributePathErr != nil {
 			err = attributePathErr
 		}
+		if pkg.PatchGlibc {
+			return strings.Replace(attributePath, "legacyPackages", "packages", 1)
```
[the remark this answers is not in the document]
gcurtis: Sure. It's because the attribute path from nixpkgs is something like "legacyPackages.x86_64-linux.hello` but when it comes from the flake it's `packages.x86_64-linux.hello` (following the standard flake output schema).

internal/shellgen/flake_plan.go
```diff
@@ -65,3 +68,91 @@ func newFlakePlan(ctx context.Context, devbox devboxer) (*flakePlan, error) {
...
+	if err != nil {
+		return err
+	}
+	// Remove the legacyPackages.<system> prefix.
+	outputName := strings.SplitN(attrPath, ".", 3)[2]
+
+	if g.Outputs.Packages == nil {
+		g.Outputs.Packages = map[string]map[string]string{nix.System(): {}}
+	}
+	if cached, _ := pkg.IsInBinaryCache(); cached {
```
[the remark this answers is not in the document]
gcurtis: Can it return `cached == true` and an error?

internal/shellgen/generate.go
gcurtis: This was something that Mike and I discussed as an interim solution. I agree that it's not great how we're using it as a catch-all nixpkgs hash for random things.

internal/shellgen/generate.go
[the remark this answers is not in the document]
gcurtis: I'll update it to specifically add the new flake file (which we do need to add or the flake build fails).

internal/shellgen/tmpl/flake.nix.tmpl
```diff
@@ -45,7 +45,7 @@
           buildInputs = with pkgs; [
             {{- range $_, $flake := .FlakeInputs }}
             {{- range $flake.BuildInputs }}
-            {{.}}
+            {{.AttrPath}}
```
gcurtis, line 48: Forgot to revert this in the old flake template too.

internal/shellgen/tmpl/glibc-patch.nix.tmpl
gcurtis: It's not used anywhere, but it's a required argument to `mapAttrs`.

This is just mapping the existing names from the flake inputs (set by the template) to a new attribute set. The new set has the configured nixpkgs. So `nixpkgs-80c24e → pkgs.nixpkgs-80c24e` where `pkgs.nixpkgs-80c24e = (import nixpkgs-80c24e { ... })`. 

This way we don't have to try and guess if an input resolves to nixpkgs in Go. Instead we can check at eval-time with Nix.

internal/shellgen/tmpl/glibc-patch.nix.tmpl
gcurtis: It makes `nix fmt` work. I threw it in because it's convenient when debugging and manually editing the flake.

internal/shellgen/flake_input.go
```diff
@@ -77,48 +83,65 @@ func (f *flakeInput) BuildInputs() ([]string, error) {
 // i.e. have a commit hash and always resolve to the same package/version.
 // Note: inputs returned by this function include plugin packages. (php only for now)
 // It's not entirely clear we always want to add plugin packages to the top level
-func flakeInputs(ctx context.Context, packages []*devpkg.Package) []*flakeInput {
+func flakeInputs(ctx context.Context, packages []*devpkg.Package) []flakeInput {
```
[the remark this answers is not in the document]
gcurtis: Slices should generally contain values unless there's a specific reason they're needed. A slice of pointers allocates each element separately on the heap and adds a layer of indirection. A slice of values is a single chunk of memory which is nicer for GC and locality. We weren't using any of the pointer semantics (other than keying the lookup while building the slice) so I figured I'd switch it.

I moved the "get-or-append" logic into its own method to try and address your feedback about line 136.

gcurtis:
After rebasing main the patch was no longer working. I attempted to get the patched flake into the nix profile, but kept running into bugs with the getting the various normalized/full/resolved/etc attribute paths right. The other problem is that right now we install to the profile before generating the flake. This breaks because the patched flake doesn't exist to install to the profile yet. I don't want to flip that logic right now because it's a much bigger change.

For now I'm just going to hack this by not removing the store path from the PATH. I really want to clean up the devpkg and flake generation code so that it's easier to reason about and revisit this.

PR #1581, opened 2023-10-23
devpkg: better flake references and installable parsing

gcurtis:
In anticipation of adding support for non-default package outputs, I wanted to clean up some of the flakeref parsing in `devpkg`. This will make it easier to handle Nix's `flakeref#attrpath^outputs` installable syntax. This PR is broken up into two commits: one that handles flakeref parsing and one that handles installable parsing.

The `ParseFlakeRef` function is able to parse the more common flakerefs that devbox currently supports ("path", "flake", "indirect", etc.). It doesn't support mercurial, gitlab, and sourcehuthut for now. The `ParseFlakeInstallable` function parses Nix flake installables, which are a superset of the flake reference syntax. It allows an attribute path and/or outputs to be specified using the `flake#attrpath^outputs` syntax. The tests were validated using the nix expression `builtins.parseFlakeRef "ref"` and the `nix flake metadata <ref>` CLI.

The goal isn't to perfectly replicate Nix's parsing of flake references (which itself is sometimes inconsistent), but to make it good enough to handle flakerefs in devbox.json.

### DIFF
--- a/internal/devpkg/flakeref.go
+++ b/internal/devpkg/flakeref.go
@@ -1,0 +1,261 @@
+package devpkg
+
+import (
+	"net/url"
+	"strings"
+
+	"go.jetpack.io/devbox/internal/redact"
+)
+
+// FlakeRef is a parsed Nix flake reference. A flake reference is as subset of
+// the Nix CLI "installable" syntax. Installables may specify an attribute path
+// and derivation outputs with a flake reference using the '#' and '^' characters.
+// For example, the string "nixpkgs" and "./flake" are valid flake references,
+// but "nixpkgs#hello" and "./flake#app^bin,dev" are not.
+//
+// See the [Nix manual] for details on flake references.
+//
+// [Nix manual]: https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake
+type FlakeRef struct {
+	// Type is the type of flake reference. Some valid types are "indirect",
+	// "path", "file", "git", "tarball", and "github".
+	Type string `json:"type,omitempty"`
+
+	// ID is the flake's identifier when Type is "indirect". A common
+	// example is nixpkgs.
+	ID string `json:"id,omitempty"`
+
+	// Path is the path to the flake directory when Type is "path".
+	Path string `json:"path,omitempty"`
+
+	// Owner and repo are the flake repository owner and name when Type is
+	// "github".
+	Owner string `json:"owner,omitempty"`
+	Repo  string `json:"repo,omitempty"`
+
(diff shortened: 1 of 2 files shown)


internal/devpkg/flakeref.go
```diff
@@ -0,0 +1,356 @@
+package devpkg
+
+import (
+	"net/url"
+	"strings"
+
+	"go.jetpack.io/devbox/internal/redact"
+)
+
+// FlakeRef is a parsed Nix flake reference. A flake reference is as subset of
```
savil, line 10: `...is a subset of...`

internal/devpkg/flakeref.go
```diff
@@ -0,0 +1,356 @@
...
+	// "github".
+	Owner string `json:"owner,omitempty"`
+	Repo  string `json:"repo,omitempty"`
+
+	// Rev and ref are the git revision (commit hash) and ref
+	// (branch or tag) when Type is "github" or "git".
+	Rev string `json:"rev,omitempty"`
+	Ref string `json:"ref,omitempty"`
+
+	// Dir is non-empty when the directory containinig the flake.nix file is
```
savil, line 41: typo: containing

internal/devpkg/flakeref.go
```diff
@@ -0,0 +1,356 @@
...
+		return parsed, nil
+	}
+	parsed, _, err := parseFlakeURLRef(ref)
+	return parsed, err
+}
+
+func parseFlakeURLRef(ref string) (parsed FlakeRef, fragment string, err error) {
+	// A good way to test how Nix parses a flake reference is to run:
+	//
+	// 	nix eval --json --expr 'builtins.parseFlakeRef "ref"' | jq
```
savil, line 101: good pro-tip 👍🏾

savil (approved):
This is great! What is the plan to integrate it with the rest of the code?

> I wanted to clean up some of the flakeref parsing in devpkg. This will make it easier to handle Nix's flakeref#attrpath^outputs installable syntax. This PR is broken up into two commits: one that handles flakeref parsing and one that handles installable parsing.

Is the expectation that users would directly run `devbox add flakeref#attrpath^outputs`? If so, I worry that is a complicated string for our users. We can find a simpler way.

On the other hand, if we're saying the user's commands will get translated by Devbox CLI code into being of the form `flakeref#attrpath^outputs` and then this makes sense to me.

This product discussion is probably better in a different forum than this PR. Happy to chat elsewhere more :)

gcurtis:
It's the second one. For handling the flake-like syntax supported in devbox.json (which users have to add manually).

PR #1584, opened 2023-10-24
[shell] Remove bin wrappers 🎉 

mikeland73:
## Summary

This removes bin wrappers. It also does the following:

* Creates `refresh` and `refresh-global` aliases.
* If environment is out of date, it prompts user to run the alias.
* If alias is missing, it shows the ugly raw command.
* If user has direnv active, we don't show message (because we refresh automatically)
* We add aliases in two ways: 1) we add as part of `shellrc` for shell. 2) we add as part of shellenv for global, run, etc

Caveats:

* calling `refresh` causes environment to refresh including environment variables (plugin and devbox.json)
* If user does not call refresh, it is possible some packages that require setup hooks will not work, but the binary is available and in PATH.

## How was it tested?

On zsh, fish shells:

* I installed `hello` globally, verified with `which hello`
* Installed `hello` in shell, got refresh message. Did `which` again and verified still global
* Did `refresh` and `which` again and verified correct.

On global zsh:

* Added `hello` and got `refresh-global` message.
* Ran refresh and verified env was up to date.

On direnv:

* Installed `hello` and verified environment was up to date.


internal/impl/shell.go
```diff
@@ -52,6 +52,7 @@ var ErrNoRecognizableShellFound = errors.New("SHELL in undefined, and couldn't f
 // DevboxShell configures a user's shell to run in Devbox. Its zero value is a
 // fallback shell that launches a regular Nix shell.
 type DevboxShell struct {
+	devbox          *Devbox
```
mikeland73, line 55: this is hacky, I'm sorry. I ran out of steam at the end of this PR.

mikeland73:
@savil @gcurtis PTAL

internal/impl/refresh.go
```diff
@@ -0,0 +1,75 @@
...
+		devboxCmd = "global shellenv --preserve-path-stack -r"
+	}
+	if os.Getenv("SHELL") == "fish" {
+		return fmt.Sprintf(`eval (devbox %s  | string collect)`, devboxCmd)
+	}
+	return fmt.Sprintf(`eval "$(devbox %s)" && hash -r`, devboxCmd)
+}
+
+func (d *Devbox) refreshAlias() string {
+	if os.Getenv("SHELL") == "fish" {
```
gcurtis, line 55: `SHELL` can be a path, so check the base name.

internal/impl/refresh.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return "refresh"
+}
+
+func (d *Devbox) refreshCmd() string {
+	devboxCmd := fmt.Sprintf("shellenv --preserve-path-stack -c %s", d.projectDir)
+	if d.isGlobal() {
+		devboxCmd = "global shellenv --preserve-path-stack -r"
+	}
+	if os.Getenv("SHELL") == "fish" {
```
gcurtis, line 48: Same here about `SHELL` being a path.

internal/impl/refresh.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (d *Devbox) refreshAliasName() string {
+	if d.isGlobal() {
+		return "refresh-global"
+	}
+	return "refresh"
+}
+
+func (d *Devbox) refreshCmd() string {
+	devboxCmd := fmt.Sprintf("shellenv --preserve-path-stack -c %s", d.projectDir)
```
gcurtis, line 44: Quotes around project directory.

gcurtis (approved):

internal/impl/devbox.go
```diff
@@ -888,7 +873,7 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 	// Add helpful env vars for a Devbox project
 	env["DEVBOX_PROJECT_ROOT"] = d.projectDir
 	env["DEVBOX_CONFIG_DIR"] = d.projectDir + "/devbox.d"
-	env["DEVBOX_PACKAGES_DIR"] = d.projectDir + "/.devbox/virtenv/.wrappers"
+	env["DEVBOX_PACKAGES_DIR"] = d.projectDir + "/" + nix.ProfilePath
```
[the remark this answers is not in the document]
savil: suggest checking IRL with @Lagoja

savil (approved):
great!

internal/impl/refresh.go
```diff
@@ -0,0 +1,75 @@
...
+
+// In some cases (e.g. 2 projects somehow active at the same time),
+// refresh might not match. This is a tiny edge case, so no need to make UX
+// great, we just print out the entire command.
+func (d *Devbox) refreshAliasOrCommand() string {
+	if !d.isRefreshAliasSet() {
+		return d.refreshCmd()
+	}
+	return d.refreshAliasName()
+}
```
savil, line 34: will this occur if devbox global is enabled in the user's shellrc, and they are now using a local devbox project?

savil: If not, could you explain why? I was struggling to figure it out

mikeland73: We use different refresh commands `refresh` vs `refresh-global` so no, this will only happen if we ever somehow activate 2 non-global devboxes at the same time (e.g. manually running shellenv might do it). Very edge case. But if we don't do this, we'll tell the user to use `refresh` but that won't work.

PR #1589, opened 2023-10-25
devconfig: use two spaces instead of tab

gcurtis:
The default hujson formatting uses tabs, which causes a bunch of whitespace changes for existing configs. Most JSON formatters/LSP servers also tend to default to 2 spaces. Add a simple replace all to change all tabs to "  ", which should be safe since raw tab characters are not allowed in valid JSON values.

### DIFF
--- a/internal/devconfig/config.go
+++ b/internal/devconfig/config.go
@@ -4,6 +4,7 @@
 package devconfig
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


mikeland73 (approved):

LucilleH:
TY!

PR #1597, opened 2023-10-31
devpkg: use struct field values in FlakeRef.String

gcurtis:
Improve FlakeRef.String so that it reassembles the flakeref using the current field values instead of returning the original parsed string that it came from. This makes it easier to modify parsed flakerefs or to create new ones from scratch.

The strings are normalized so that when two flakerefs are equal, their strings will also be equal. The docs for FlakeRef.String go into more detail on what the normalized form looks like.

Being able to parse/modify/encode flake references is part of the work to support multiple package outputs. Having a single (well-tested type) for handling flakes should help manage the complexity as another syntax for outputs is added.

### DIFF
--- a/internal/devpkg/flakeref.go
+++ b/internal/devpkg/flakeref.go
@@ -2,6 +2,7 @@ package devpkg
 
 import (
 	"net/url"
+	"path"
 	"strings"
 
 	"go.jetpack.io/devbox/internal/redact"
@@ -52,9 +53,6 @@ type FlakeRef struct {
 	// or "git". Note that the URL is not the same as the raw unparsed
 	// flakeref.
 	URL string `json:"url,omitempty"`
-
-	// raw stores the original unparsed flakeref string.
-	raw string
 }
 
 // ParseFlakeRef parses a raw flake reference. Nix supports a variety of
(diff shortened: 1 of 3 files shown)


internal/devpkg/flakeref.go
```diff
@@ -231,9 +228,111 @@ func parseGitHubFlakeRef(refURL *url.URL, parsed *FlakeRef) error {
...
+		if err != nil {
+			// This should be rare and only happen if the caller
+			// messed with the parsed URL.
+			return ""
+		}
+		url.RawQuery = buildQueryString("ref", f.Ref, "rev", f.Rev, "dir", f.Dir)
+		return url.String()
+	case "github":
+		if f.Owner == "" || f.Repo == "" {
+			return ""
```
savil, line 279: should this be an error?

internal/devpkg/flakeref.go
```diff
@@ -260,6 +359,33 @@ func isArchive(path string) bool {
 		strings.HasSuffix(path, ".zip")
 }
 
+// buildOpaquePath escapes and joins path elements for a URL flakeref. The
+// resulting path is cleaned according to url.JoinPath.
+func buildOpaquePath(elem ...string) string {
```
savil, line 364: The Opaque word confused me. How about `buildEscapedPath` or `buildURLSafePath`?

savil (approved):
awesome!

internal/devpkg/flakeref.go
gcurtis: Ideally, but I want FlakeRef to be a `fmt.Stringer` so this function is unable to return an error.

internal/devpkg/flakeref.go
gcurtis: Sounds good. I'm going to rename it in another PR just so I don't have to rebase/merge all the other PRs.

PR #1600, opened 2023-10-31
devpkg: use struct field values in FlakeInstallable.String

gcurtis:
Improve `FlakeInstallable.String` so that it reassembles the installable string using the current field values instead of returning the original parsed string that it came from. This mirrors the other commit that updates `FlakeRef.String`. Similar to `FlakeRef.String`, `FlakeInstallable.String` normalizes the resulting string so that if two installables are equal, then their strings will always be equal.


savil (approved):
excellent-o

PR #1617, opened 2023-11-10
Added json schema for 0.8 version

mohsenari:
## Summary
Title says it.

I put the file in `.schema/` directory so that in future if we change the schema we can align it with version. Meaning, we can add new schema files with version numbers and enforce the schema based on the version of devbox a user has installed. Putting it in root of this repository would make the root too crowded especially if we add more schema files.

I'm going to host our schema in this repository and subsequently send a PR to https://github.com/SchemaStore/schemastore/ with a link to the schema file in our repo to register our repo in https://schemastore.org

afterwards, I can update the vscode extension to reference the schemastore.org link and enforce our defined schema on files that are named `devbox.json`

## How was it tested?
validated schema and examples of devbox.json in https://www.jsonschemavalidator.net/


.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+            },
+            "required": [
+                "init_hook"
+            ]
+        }
+    },
+    "required": [
+        "packages",
+        "shell"
+    ]
```
mohsenari, line 87: I put these fields as required, If they're wrong or missing something lmk.

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "https://github.com/jetpack-io/devbox",
+    "title": "Devbox json definition",
+    "description": "Defines fields and acceptable values of devbox.json",
+    "type": "object",
+    "properties": {
+        "packages": {
+            "description": "A specification of packages to be installed via devbox.",
```
gcurtis, line 9:
```suggestion
            "description": "Collection of packages to install",
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                        "type": "string"
+                    }
+                },
+                {
+                    "type": "object",
+                    "description": "Name of each package in {\"name\": \"version\"} format.",
+                    "patternProperties": {
+                        ".*": {
+                            "type": "string",
+                            "description": "Version number of the package in this object."
```
gcurtis, line 24:
```suggestion
                            "description": "Version of the package to install."
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                        }
+                    }
+                },
+                {
+                    "type": "object",
+                    "description": "Name of each package in {\"name\": {\"version\": \"1.2.3\"}} format.",
+                    "patternProperties": {
+                        ".*": {
+                            "type": "object",
+                            "description": "Version number of the specified package in {\"version\": \"1.2.3\"} formmat.",
```
gcurtis, line 34: This object can contain a few more fields. I think they are:

- version
- platforms - only install the package on these Nix systems (e.g., `aarch64-darwin`, `x86_64-linux`, etc.)
- excluded_platforms - don't install the package on these Nix systems.
- glibc_patch - apply the glibc-patch feature to this package's binaries.

gcurtis, line 34:
```suggestion
                            "description": "Version number of the specified package in {\"version\": \"1.2.3\"} format.",
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                                    "description": "Version of the package"
+                                }
+                            }
+                        }
+                    }
+                }
+            ]
+        },
+        "env": {
+            "description": "List of additional environment variables to be set in devbox shell. Values can contain $PATH and $PWD.",
```
gcurtis, line 47:
```suggestion
            "description": "List of additional environment variables to be set in the Devbox environment. Values containing $PATH or $PWD will be expanded. No other variable expansion or command substitution will occur.",
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+            "type": "object"
+        },
+        "shell": {
+            "description": "Definitions of scripts and actions to take when in devbox shell.",
+            "type": "object",
+            "properties": {
+                "init_hook": {
```
gcurtis, line 54: This can also be a single string.

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                    "type": "array",
+                    "items": {
+                        "description": "list of shell commands/scripts to run right after devbox shell starts.",
+                        "type": "string"
+                    }
+                },
+                "scripts": {
+                    "description": "List of commads/scripts definitions to run with `devbox run <script_name>` command.",
```
gcurtis, line 62:
```suggestion
                    "description": "List of command/script definitions to run with `devbox run <script_name>`.",
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+            "type": "object"
+        },
+        "shell": {
+            "description": "Definitions of scripts and actions to take when in devbox shell.",
+            "type": "object",
+            "properties": {
+                "init_hook": {
+                    "type": "array",
+                    "items": {
+                        "description": "list of shell commands/scripts to run right after devbox shell starts.",
```
gcurtis, line 57:
```suggestion
                        "description": "List of shell commands/scripts to run right after devbox shell starts.",
```

.schema/0.8.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                    "patternProperties": {
+                        ".*": {
+                            "description": "Alias name for the script.",
+                            "type": [
+                                "array",
+                                "string"
+                            ],
+                            "items": {
+                                "type": "string",
+                                "description": "command/script to run as part of a list of commands defined here."
```
gcurtis, line 73:
```suggestion
                                "description": "The script's shell commands."
```

.schema/0.8.schema.json
gcurtis, line 87: Are these required? I thought an empty devbox.json object was valid.

gcurtis (approved):
Awesome! I've been wanting this for a while. I left some wording suggestions, but feel free to take them or leave them as you see fit.

Question about the schema versioning - could we have a single schema.json and just link to specific commits in GitHub? For example, https://github.com/jetpack-io/devbox/blob/0.8.1/.schema/0.8.schema.json would return the schema as of devbox 0.8.1.

mohsenari:
> Awesome! I've been wanting this for a while. I left some wording suggestions, but feel free to take them or leave them as you see fit.
> 
> Question about the schema versioning - could we have a single schema.json and just link to specific commits in GitHub? For example, https://github.com/jetpack-io/devbox/blob/0.8.1/.schema/0.8.schema.json would return the schema as of devbox 0.8.1.

I think that also works

> Awesome! I've been wanting this for a while. I left some wording suggestions, but feel free to take them or leave them as you see fit.
> 
> Question about the schema versioning - could we have a single schema.json and just link to specific commits in GitHub? For example, https://github.com/jetpack-io/devbox/blob/0.8.1/.schema/0.8.schema.json would return the schema as of devbox 0.8.1.

The link we provide has to point to the raw file content. If we can have the link to that within multiple versions then we can have a single file, otherwise we gotta keep a separate file per version

.schema/0.8.schema.json
mohsenari: @gcurtis can you send me an example of devbox.json with these parameters included? 
Also a list of all possible values for platforms/excluded_platforms.
Side note: If I remember correctly, platforms and excluded_platforms can not co-exist right?

.schema/0.8.schema.json
mohsenari: I actually don't know. I'll test it out and if they weren't I'll remove.

gcurtis:
@mohsenari it should be possible to get the raw content. For example, https://raw.githubusercontent.com/jetpack-io/devbox/ab2e606cc3c71b9f84d92d1d22441b6b400bf959/.schema/0.8.schema.json

.schema/0.8.schema.json
mohsenari: Seems like empty devbox shell is accepted. I'll remove the required fields

.schema/devbox.schema.json
```diff
@@ -0,0 +1,87 @@
...
+                        ".*": {
+                            "type": "object",
+                            "description": "Version number of the specified package in {\"version\": \"1.2.3\"} format.",
+                            "properties": {
+                                "version": {
+                                    "type": "string",
+                                    "description": "Version of the package"
+                                },
+                                "platform": {},
+                                "excluded_platform": {},
```
savil, line 41: I think these should be `platforms` and `excluded_platforms` (note the `s` for plural)

PR #1618, opened 2023-11-10
[easy][auth] Update auth package

mikeland73:
## Summary

Use newest version of auth which fetches access token.

## How was it tested?

```
devbox auth login
devbox auth whoami
```


loreto (approved):
LGTM

mohsenari (approved):

PR #1624, opened 2023-11-13
add fields, restrict additional props

Lagoja:
## Summary
Adds missing fields to the json schema
Restricts additional properties -- this ensures that developers can't put `init_hook` in the wrong location

## How was it tested?
Manually using the root `devbox.json`


mohsenari (approved):

mohsenari:
thanks! let's ship it

Issue #1612, opened 2023-11-08
[Bug]: devbox keeps trying to install mysql80 but fails when the services are already running


calvinbaart:
**Current Behavior (bug)**
When we start the services using devbox services up and then in another shell do devbox shell we get this error:

```
➜  test2 devbox shell

Installing package: path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80.

[1/1] path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80
Error: error installing package path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80: error searching for pkg path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80: exit status 1

Error: There was an internal error. Run with DEVBOX_DEBUG=1 for a detailed error message, and consider reporting it at https://github.com/jetpack-io/devbox/issues
```

When running with DEVBOX_DEBUG=1 we get this log:

```
➜  test2 DEVBOX_DEBUG=1 devbox shell
2023/11/08 09:50:29 findProjectDir: path is
2023/11/08 09:50:29 finding devbox.json in dir: /Users/calvinbaart/dev/test2
2023/11/08 09:50:29 findProjectDir: path is
2023/11/08 09:50:29 finding devbox.json in dir: /Users/calvinbaart/dev/test2
2023/11/08 09:50:29 findProjectDir: path is
2023/11/08 09:50:29 finding devbox.json in dir: /Users/calvinbaart/dev/test2
2023/11/08 09:50:29 Creating files for package "mysql80@8.0.34" create files
2023/11/08 09:50:29 Creating file "/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80/setup_db.sh" from contentPath: "mysql/setup_db.sh"
2023/11/08 09:50:29 Creating file "/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80/process-compose.yaml" from contentPath: "mysql/process-compose.yaml"
2023/11/08 09:50:29 Creating file "/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80/flake.nix" from contentPath: "mysql/flake.nix"

Installing package: path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80.

[1/1] path:/Users/calvinbaart/dev/test2/.devbox/virtenv/mysql80
[...]

PR #1625, opened 2023-11-14
[ensure] Don't recompute environment if up to date

mikeland73:
## Summary

This bug was introduced in https://github.com/jetpack-io/devbox/pull/1584 and has at least 2 effects:

* Perf regression (because we recompute environment all the time)
* Plugins that create flakes may break if they create additional files in the same directory. (Another PR will address this issue in more depth)

## How was it tested?

```bash
devbox add mysql80
devbox services up -b
devbox services stop
```

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -228,12 +228,14 @@ func (d *Devbox) ensurePackagesAreInstalled(ctx context.Context, mode installMod
 	if err != nil {
 		return err
 	}
-	// if mode is ensure and we are up to date, then we can skip the rest
-	if mode == ensure && upToDate {
-		return nil
-	}
 
-	fmt.Fprintln(d.stderr, "Ensuring packages are installed.")
+	if mode == ensure {
+		// if mode is ensure and we are up to date, then we can skip the rest
+		if upToDate {
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


internal/impl/packages.go
```diff
@@ -230,9 +230,11 @@ func (d *Devbox) ensurePackagesAreInstalled(ctx context.Context, mode installMod
 	}
 	// if mode is ensure and we are up to date, then we can skip the rest
 	if mode == ensure && upToDate {
-		fmt.Fprintln(d.stderr, "Ensuring packages are installed.")
+		return nil
 	}
 
+	fmt.Fprintln(d.stderr, "Ensuring packages are installed.")
```
savil, line 236: Shouldn't this be:
```

 	if mode == ensure { 
 	   // if mode is ensure and we are up to date, then we can skip the rest
 	   if upToDate {
 		return nil
           }
 	}
```
?

savil (approved):
Oof, can't believe I overlooked this in the code review. My bad too!

internal/impl/packages.go
mikeland73: yep! will change

jrseliga:
@mikeland73 does this resolve https://github.com/jetpack-io/devbox/issues/1612?

PR #1626, opened 2023-11-14
[plugins] flakes should be in own dir

mikeland73:
## Summary

**What this fixes:**
When plugins create flake in a directory shared with other files, `nix search` can break. (We use `nix search` to determine output path). For example in the mysql plugin the flake would live side-by-side with the `run` directory which has a sock file. `nix search` would recurse in that directory and try to search for nix expressions and break because the file had an unexpected format. 

**How it was surfaced:**
This issue was previously hidden because we don't call `nix search` if a package has already been installed. The bug fixed in https://github.com/jetpack-io/devbox/pull/1625 was causing us to needlessly call `nix search` on stuff that was already 

**Solution:**
The partial solution in this PR is to always create the flake in its own dir. This is a bit fragile, e.g. someone can forget if writing their own plugin. Longer term, we need a better way to extract the output path from a custom flake. (I think `nix flake show` could do it without causing the issues surfaced by search)

## How was it tested?

CICD tests all plugins with exampels.

### DIFF
--- a/plugins/haskell.json
+++ b/plugins/haskell.json
@@ -1,6 +1,6 @@
 {
   "name": "haskell",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "readme": "Haskell plugin",
   "__packages": [
     "path:{{ .Virtenv }}/flake"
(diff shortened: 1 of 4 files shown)


savil (approved):

Lagoja:
This makes sense, should I move the flakes in the Github plugins to a subfolder as well?

Lagoja (approved):

PR #1636, opened 2023-11-17
[direnv] Fix direnv fish support

mikeland73:
## Summary

Adding alias to `direnv` environment is causing problems in fish and doesn't work anyway (because direnv doesn't support aliases)

## How was it tested?

* For both zsh and fish
* Started shell
* activated direnv hook
* cd to devbox directory
* confirmed environment was actice
* installed package and tested it

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -16,10 +16,11 @@ import (
 type shellEnvCmdFlags struct {
 	envFlag
 	config            configFlags
-	runInitHook       bool
 	install           bool
-	pure              bool
+	noRefreshAlias    bool
 	preservePathStack bool
+	pure              bool
+	runInitHook       bool
 }
 
 func shellEnvCmd(recomputeEnvIfNeeded *bool) *cobra.Command {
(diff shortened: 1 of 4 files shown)


savil (approved):
This is good, but just to clarify, we _suspect_ the issue is fish but don't have conclusive evidence. For example, direnv works fine on my fish shell.

mikeland73:
@savil I can repro when installing fish on devbox global. direnv doesn't work, but doing `SHELL=fish devbox shell` works fine (including the refresh alias).

I think it happens because `.envrc` is evaluated with bash (even for fish users) (See this [comment](https://github.com/direnv/direnv/issues/109#issuecomment-38694428))

So our issue is:

* When user is using fish, we change shellenv syntax to use fish. This is correct for shellenv but wrong for `.envrc`.

That said, I'm not 100% confident, so I'm not ready to change `shellenv` inside `.envrc` to use bash always. Instead, I'm removing the alias which is the only fish specific part of the shellenv output. 

Longer term we should make sure shellenv for direnv is only environment exports.

PR #1637, opened 2023-11-17
Update `packages` object

Lagoja:
## Summary

Updates the schema to better reflect the packages object -- packages can be either: 

* A list of Strings
* An object, where each key contains either a version string or an object

## How was it tested?

Used the following package list: 

```json
  "packages": {
    "go":                          "latest",
    "runx:golangci/golangci-lint": "latest",
    "runx:mvdan/gofumpt":          "latest",
    "glibcLocales": {
      "version":            "latest",
      "excluded_platforms": ["aarch64-darwin"]
    }
  },
```

### DIFF
--- a/.schema/devbox.schema.json
+++ b/.schema/devbox.schema.json
@@ -33,9 +33,24 @@
                                     "type": "string",
                                     "description": "Version of the package"
                                 },
-                                "platform": {},
-                                "excluded_platform": {},
-                                "glibc_patch": {}
+                                "platforms": {
+                                    "type": "array",
+                                    "description": "Names of platforms to install the package on. This package will be skipped for any platforms not on this list",
+                                    "items": {
+                                        "enum": ["i686-linux", "aarch64-linux", "aarch64-darwin", "x86_64-darwin", "x86_64-linux", "armv7l-linux"]
+                                    }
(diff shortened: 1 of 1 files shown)


.schema/devbox.schema.json
```diff
@@ -45,6 +37,12 @@
                                 "excluded_platform": {},
                                 "glibc_patch": {}
                             }
+                        },
```
savil, line 40: i think lines 36 and 37 should be `platforms` and `excluded_platforms` with `s` for plural, and their type should be array-of-strings

Lagoja: Gotcha, will update. Is there a list of valid entries for the platform? I could make this an enum so that the autocomplete is more useful.

savil: oh yeah, good call, can be an enum. Here ya go:
```
❯ devbox add hello --platform=foo
Info: Adding package "hello@latest" to devbox.json

Error: Unsupported platform: foo. Valid platforms are: [aarch64-darwin aarch64-linux i686-linux x86_64-darwin x86_64-linux armv7l-linux]
```

savil (approved):
nice!

Issue #1640, opened 2023-11-20
[Package Issue Report]: python312Packages.pip missing


mdomke:
The [documentation](https://www.jetpack.io/devbox/docs/devbox_examples/languages/python/#installing-packages-with-pip) of devbox indicates that Python with `pip` should be setup by adapting the snippet with the actual version number of Python 
```
{
    "packages": [
        "python312",
        "python312Packages.pip"
    ],
    "shell": {
        "init_hook": ". $VENV_DIR/bin/activate"
    }
}
```
but the `python312Packages.pip` package does not exist. In fact, no `python312Packages.xxx` [exists at all](https://www.nixhub.io/search?q=python312). Is there a reason why these packages are missing for Python 3.12?

Lagoja:
@mdomke thanks for reporting this. It's not clear to me why those packages do not exist for python312 -- it looks like they were never added or surfaced in the core nixpkgs repo. 

I think a better way to work around this is to use the `pip` binary that is bundled with `python@3.12`. Our current plugin only creates a virtual environment if you explicitly add a `pip` package, but we could make it work with the `python` package as well. I can link a PR once I finish testing this

Lagoja:
Merged now! Will work on getting a release out with this in the next week.

savil:
@Lagoja close task?

PR #1645, opened 2023-11-21
[Plugin] Create a virtual environment with the `python` plugin

Lagoja:
## Summary

The Nix Python packages come bundled with `pip`, so we can go ahead and create a virtual environment whenever a user installs a `python` package, even if they do not also install `pip`. 

This plugin reuses the `venvShellHook` and `VENV_DIR` environment variable from the `pip` plugin to create a virtual environment when the user starts their shell (assuming one has not been created). The user can activate the virtual environment by running `source $VENV_DIR/bin/activate`

A few notes

* This should not cause issues with virtual environments that have been created with `pythonPackages.pip`, meaning behavior should remain the same for existing python projects
* Devbox will use `VENV_DIR` variable and virtual environment for the package the appears last in the `packages` list.
* Going forward, we should probably recommend developers use the `pip` bundled with the python package. 

## How was it tested?

Using the modified python/pip example in this PR

### DIFF
--- a/docs/app/docs/devbox_examples/languages/python.md
+++ b/docs/app/docs/devbox_examples/languages/python.md
@@ -27,14 +27,27 @@ This will install Python 3.10 in your shell. You can find other versions of Pyth
 
 [pip](https://pip.pypa.io/en/stable/) is the standard package manager for Python. Since it installs python packages globally, we strongly recommend using a virtual environment.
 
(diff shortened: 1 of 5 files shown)


mikeland73 (approved):

Lagoja:
This issue helps address #1640, since the user no longer needs to install pip explicitly to activate the plugin + install packages.

PR #1650, opened 2023-11-24
fix: allow github plugins with slash in rev name

jay-aye-see-kay:
## Summary

You can't currently reference a plugin on a rev if that rev has a `/` character in it. This seems like a bug, here's a simple fix that I don't believe will cause any other issues (but I'm new to the codebase, so could be wrong).

## How was it tested?

1. Unit tests in this PR
2. Manual verification:

**verify current build breaks:**
```sh
devbox run build
./dist/devbox run echo "success"
# works fine

# add plugin with branch with slash -> `"include": ["github:jetpack-io/devbox-plugins/jl/add-readmes?dir=mongodb"]`
./dist/devbox run echo "success"
# Error: invalid github plugin url "jetpack-io/devbox-plugins/jl/add-readmes?dir=mongodb". Must be of the form org/repo/[revision]
```

**checkout this branch and verify it works again:**
```sh
git reset --hard
git checkout jdr/allow-github-plugins-with-slash-in-rev-name

# add plugin with branch with slash -> `"include": ["github:jetpack-io/devbox-plugins/jl/add-readmes?dir=mongodb"]`
./dist/devbox run echo "success"
# works fine
```

### DIFF
--- a/internal/plugin/github.go
+++ b/internal/plugin/github.go
@@ -28,9 +28,9 @@ func newGithubPlugin(rawURL string) (*githubPlugin, error) {
 		return nil, err
 	}
 
-	parts := strings.Split(pluginURL.Path, "/")
+	parts := strings.SplitN(pluginURL.Path, "/", 3)
 
-	if len(parts) < 2 || len(parts) > 3 {
+	if len(parts) < 2 {
 		return nil, usererr.New(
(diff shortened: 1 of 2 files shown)


mikeland73 (approved):
Nice!

Yeah, this was a bug. 

Thanks for contributing and adding unit tests.

PR #1659, opened 2023-12-02
Create a json schema for plugins

Lagoja:
## Summary

Adds a JSON schema for creating plugins. Hopefully, this should make authoring plugins a bit easier

## How was it tested?

Tested against the `devbox plugins` repo

### DIFF
--- a/.schema/devbox-plugin.schema.json
+++ b/.schema/devbox-plugin.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft-04/schema#",
-  "$id": "https://github.com/jetpack-io/devbox",
+  "$id": "https://github.com/jetpack-io/devbox/plugins",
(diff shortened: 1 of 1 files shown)


.schema/devbox-plugin.schema.json
```diff
@@ -0,0 +1,56 @@
+{
+  "$schema": "https://json-schema.org/draft-04/schema#",
+  "$id": "https://github.com/jetpack-io/devbox",
```
mohsenari, line 3: This should be "https://github.com/jetpack-io/devbox-plugin" to differentiate between devbox schema and devbox plugin schema

mohsenari (approved):

.schema/devbox-plugin.schema.json
Lagoja: Do you have ideas on how we should change it? I'd like to keep them in the same repo

PR #1662, opened 2023-12-04
[update] Fix update

mikeland73:
## Summary

nix 2.19 changed how `nix flake update` [works](https://nixos.org/manual/nix/stable/release-notes/rl-2.19).

This fixes it. 

## How was it tested?

- [x] Run update on nix < 2.19 and 2.19
- [x] Add update unit test

### DIFF
--- a/testscripts/update/update.test.txt
+++ b/testscripts/update/update.test.txt
@@ -11,7 +11,7 @@ exec devbox update
   ]
 }
 
--- devbox.lock -- 
+-- devbox.lock --
 {
   "lockfile_version": "1",
   "packages": {


.github/workflows/cli-tests.yaml
```diff
@@ -242,7 +242,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        nix-version: [2.15.1, 2.16.1, 2.17.0, 2.18.0]
+        nix-version: [2.15.1, 2.16.1, 2.17.0, 2.18.0, 2.19.2]
```
savil, line 245: can we drop 2.16.1 and 2.18?

mikeland73:
> if Devbox works on nix 2.19, we can also update the hardcoded 2.18 version that we install during the default nix install...

We're on 2.17. I'm gonna bump to 2.18 in another PR

.github/workflows/cli-tests.yaml
Lagoja: I don't think there's a reason to drop older versions -- It's hard enough to upgrade Nix, I don't want to break early users

PR #1669, opened 2023-12-06
[glibc] Add patch-glibc patch and fix when package in binary cache

mikeland73:
## Summary

* Adds new `--patch-glibc` flag 
* Fixes glibc feature (I think) when package is in binary cache. We were previouwly always installing the cached version instead of the patched version.

## How was it tested?

* Flag was tested using `devbox add python --patch-glibc`
* I'm not 100% sure how to test the patch itself, but I did verify we are creating the patch flake

### DIFF
--- a/internal/devconfig/packages.go
+++ b/internal/devconfig/packages.go
@@ -177,7 +177,20 @@ func (pkgs *Packages) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-func (pkgs *Packages) DisablePlugin(versionedName string, v bool) error {
+func (pkgs *Packages) SetPatchGLibc(versionedName string, v bool) error {
+	name, version := parseVersionedName(versionedName)
+	i := pkgs.index(name, version)
+	if i == -1 {
+		return errors.Errorf("package %s not found", versionedName)
+	}
+	if pkgs.Collection[i].PatchGlibc != v {
+		pkgs.Collection[i].PatchGlibc = v
+		pkgs.ast.setPackageBool(name, "patch_glibc", v)
+	}
+	return nil
+}
+
+func (pkgs *Packages) SetDisablePlugin(versionedName string, v bool) error {
 	name, version := parseVersionedName(versionedName)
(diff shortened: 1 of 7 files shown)


internal/devconfig/packages.go
```diff
@@ -177,7 +177,20 @@ func (pkgs *Packages) UnmarshalJSON(data []byte) error {
...
+	i := pkgs.index(name, version)
+	if i == -1 {
+		return errors.Errorf("package %s not found", versionedName)
+	}
+	if pkgs.Collection[i].PatchGlibc != v {
+		pkgs.Collection[i].PatchGlibc = v
+		pkgs.ast.setPackageBool(name, "patch_glibc", v)
+	}
+	return nil
+}
```
savil, line 191: just to confirm, if the packages is a `[]string` in json, then this will convert it to a `map[string] (string | object)`?

savil (approved):
Great! I have a couple of projects I can test this locally with.

savil:
> I'm not 100% sure how to test the patch itself, but I did verify we are creating the patch flake

wait, we should only be generating the flake on linux (since this issue affects linux, and not macos)

internal/devpkg/narinfo_cache.go
```diff
@@ -25,6 +25,10 @@ const BinaryCache = "https://cache.nixos.org"
 // ALERT: Callers in a perf-sensitive code path should call FillNarInfoCache
 // before calling this function.
 func (p *Package) IsInBinaryCache() (bool, error) {
+	// Patched glibc packages are not in the binary cache.
+	if p.PatchGlibc {
```
savil, line 29: this should be `PatchGlibc && linux`

or better still the PatchGlibc boolean should only be set if it is specified in the devbox.json _and_ the current OS is linux

mikeland73:
> wait, we should only be generating the flake on linux (since this issue affects linux, and not macos)

Only the linux version is patched but the flake still gets created. I'll fix

Issue #1657, opened 2023-11-30
[Bug]: User should be able to activate the shell even if no internet is available


carlotm:
**Current Behavior (bug)**

If I am not connected to the internet, I got:

```
carloratm@localhost $ devbox shell
→ Downloading version 0.4.9...
✘ Command failed (exit code 6): curl --fail --silent --location --output /tmp/tmp.fsp4QClTWG/devbox_0.4.9_linux_amd64.tar.gz https://github.com/jetpack-io/devbox/releases/download/0.4.9/devbox_0.4.9_linux_amd64.tar.gz
```

**Expected Behavior (fix)**

I should be able to activate the shell even if I am not connected to the internet

Lagoja:
Hey @carlotm,

Which version of devbox are you using, and how did you install it?

carlotm:
Dear @Lagoja ,
I just updated to `0.8.3` (it was `0.8.2` when I first encountered this issue)

What I get when not connected to the internet is:

```
~/w/oha[main]
> devbox shell
Ensuring packages are installed.
Error: Head "https://cache.nixos.org/frsl3935sgp53hdgmmknhbl33d4fb7nn.narinfo": dial tcp: lookup cache.nixos.org on 10.6.26.1:53: dial udp 10.6.26.1:53: connect: network is unreachable
```

I think it would be great if a user could activate a shell when there's no internet.
As the packages are already downloaded, I'd love to be able to work on the project (while commuting for example :))

Cheers!

carlotm:
@mikeland73 thanks a lot!

PR #1673, opened 2023-12-06
[offline] Allow devbox to work offline if packages are already fetched

mikeland73:
## Summary

Very basic fix. Doesn't handle all cases, but fixes run/shell.

Fixes https://github.com/jetpack-io/devbox/issues/1657

## How was it tested?

Turned off my wifi, modified `.local.lock` file to bust cache. Was able to run/shell.

### DIFF
--- a/internal/impl/devbox.go
+++ b/internal/impl/devbox.go
@@ -953,8 +953,10 @@ func (d *Devbox) ensurePackagesAreInstalledAndComputeEnv(
 
 	// When ensurePackagesAreInstalled is called with ensure=true, it always
 	// returns early if the lockfile is up to date. So we don't need to check here
-	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
+	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil && !strings.Contains(err.Error(), "no such host") {
 		return nil, err
+	} else if err != nil {
+		ux.Fwarning(d.stderr, "Error connecting to the internet. Will attempt to use cached environment.\n")
 	}
 
 	// Since ensurePackagesAreInstalled calls computeNixEnv when not up do date,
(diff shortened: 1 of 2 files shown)


savil (approved):

internal/impl/devbox.go
```diff
@@ -953,8 +953,10 @@ func (d *Devbox) ensurePackagesAreInstalledAndComputeEnv(
 
 	// When ensurePackagesAreInstalled is called with ensure=true, it always
 	// returns early if the lockfile is up to date. So we don't need to check here
-	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil {
+	if err := d.ensurePackagesAreInstalled(ctx, ensure); err != nil && !strings.Contains(err.Error(), "no such host") {
 		return nil, err
+	} else if err != nil {
+		ux.Fwarning(d.stderr, "Error connecting to the internet. Will attempt to use cached environment.\n")
```
savil, line 959: I guess there is some danger that `ensurePackagesAreInstalled` has done some work already (like nix profile install) but then internet fails and we fallback to using a stale `print-dev-env` cache value

savil: (i think its okay; better than blocking)

Lagoja: What kind of things would be stale?

savil: like it may have some package that has been removed, or an older version of a package, etc.

mikeland73: Usually, this means one of 2 things:

* Something in devbox.json changed
* devbox version changed (In this case we bust the cache because the .devbox dir structure might change)

The cached environment is nix only. So if you add a script or env var to your `devbox.json` we will still pick it up. If you add a package we will not.

So one edge case where this doesn't completely work is if you add a local flake. There's no internet needed to install it, but it won't work anyway. This is harder to fix given our current code structure and feels pretty edge case so I wouldn't worry.

PR #1675, opened 2023-12-07
[Renaming] ComputeNixEnv -> ComputeDevboxEnv; ensurePackagesAreInstalled -> ensureProjectStateIsCurrent

savil:
## Summary

These names were getting rather stale. Hopefully, the new ones represent
them more accurately.

## How was it tested?

compiles and cicd tests should pass


savil:
* main
  * **PR #1674** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1674?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1675** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1675?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1675?utm_source=stack-comment).

mikeland73 (commented):
> There's a bunch of other functions you will need to change if you want to keep consistency:

Good call.

* NixEnv -> DevboxEnvExports
* NixEnvOpts -> DevboxEnvExportsOpts
* ensurePackagesAreInstalledAndComputeEnv -> computeCurrentDevboxEnv
* NixEnvPath (maybe?) -> yep, `devboxEnvPath`

@mikeland73 PTAL

savil:
Also, treating `current` as a shorter synonym for `upToDate` (shorter in terms of number of words) but maybe I should change to that?

savil:
@mikeland73 PTAL

savil:
ping

devbox.go
```diff
@@ -23,7 +23,7 @@ type Devbox interface {
 	Install(ctx context.Context) error
 	IsEnvEnabled() bool
 	ListScripts() []string
-	NixEnv(ctx context.Context, opts devopt.NixEnvOpts) (string, error)
+	DevboxEnvExports(ctx context.Context, opts devopt.DevboxEnvExports) (string, error)
```
mikeland73, line 26: `EnvExports` ? So if the struct name is devbox:

`devbox.EnvExports` ?

Also a bit repetitive but usually our option structs end in `Opt`. We should leave for consistency. (so `EnvExportOpts`)

internal/impl/devbox.go
```diff
@@ -267,7 +267,10 @@ func (d *Devbox) ListScripts() []string {
 	return keys
 }
 
-func (d *Devbox) NixEnv(ctx context.Context, opts devopt.NixEnvOpts) (string, error) {
+// DevboxEnvExports returns a string of the env-vars that would need to be applied
+// to define a Devbox environment. The string is of the form `export KEY=VALUE` for each
+// env-var that needs to be applied.
+func (d *Devbox) DevboxEnvExports(ctx context.Context, opts devopt.DevboxEnvExports) (string, error) {
 	ctx, task := trace.NewTask(ctx, "devboxNixEnv")
```
mikeland73, line 274: make sure trace tasks and profiling names are changed too.

internal/impl/devbox.go
```diff
@@ -288,9 +291,9 @@ func (d *Devbox) NixEnv(ctx context.Context, opts devopt.NixEnvOpts) (string, er
 			)
 		}
 
-		envs, err = d.computeNixEnv(ctx, true /*usePrintDevEnvCache*/)
+		envs, err = d.computeDevboxEnv(ctx, true /*usePrintDevEnvCache*/)
 	} else {
-		envs, err = d.ensurePackagesAreInstalledAndComputeEnv(ctx)
+		envs, err = d.computeCurrentDevboxEnv(ctx)
```
mikeland73, line 296: This is confusing. A reader can't tell what the difference between these two function calls is.

What about `ensureProjectStateIsCurrentAndComputeEnv`

internal/impl/devbox.go
```diff
@@ -889,15 +891,15 @@ func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (m
 
 	markEnvsAsSetByDevbox(pluginEnv, configEnv)
 
-	nixEnvPath := env["PATH"]
-	debug.Log("PATH after plugins and config is: %s", nixEnvPath)
+	devboxEnvPath := env["PATH"]
```
mikeland73, line 894: I'm not sure if I totally agree with all the changes in this function. Many of these are nix variables (e.g. nix path)

internal/impl/packages.go
```diff
@@ -236,19 +236,20 @@ const (
...
+//  4. The Devbox environment is re-computed, if necessary, and cached
+//  5. Lockfile is synced
 //
 // The `mode` is used for:
 // 1. Skipping certain operations that may not apply.
 // 2. User messaging to explain what operations are happening, because this function may take time to execute.
-// TODO: Rename method since it does more than just ensure packages are installed.
-func (d *Devbox) ensurePackagesAreInstalled(ctx context.Context, mode installMode) error {
-	defer trace.StartRegion(ctx, "ensurePackages").End()
+func (d *Devbox) ensureProjectStateIsCurrent(ctx context.Context, mode installMode) error {
```
mikeland73, line 251: Could be shortened to `ensureStateIsCurrent`

internal/impl/devbox.go
```diff
@@ -771,11 +774,10 @@ func (d *Devbox) StartProcessManager(
 // programs.
 //
 // Note that the shellrc.tmpl template (which sources this environment) does
-// some additional processing. The computeNixEnv environment won't necessarily
+// some additional processing. The computeDevboxEnv environment won't necessarily
 // represent the final "devbox run" or "devbox shell" environments.
-// TODO: Rename to computeDevboxEnv?
-func (d *Devbox) computeNixEnv(ctx context.Context, usePrintDevEnvCache bool) (map[string]string, error) {
-	defer trace.StartRegion(ctx, "computeNixEnv").End()
+func (d *Devbox) computeDevboxEnv(ctx context.Context, usePrintDevEnvCache bool) (map[string]string, error) {
```
mikeland73, line 779: Could be `computeEnv`. The struct name is `Devbox`

mikeland73 (approved):
Some comments. Approving to unblock.

Only blocker for me:

`computeDevboxEnv` vs `computeCurrentDevboxEnv`

This distinction means nothing to me. `computeCurrentDevboxEnv` should be something like `ensureProjectStateIsCurrentAndComputeEnv`. If you want to shorten it could also be `ensureStateIsCurrentAndComputeEnv`

I think when a function is complex, giving it a longer name for clarity is better.

I also think in many cases we can drop `Devbox` from variables and function names. So `computeEnv` or `computeEnvironment`. At the end of the day, the struct is named devbox so it's fairly clear what the subject is.

devbox.go
savil: Yeah, lets call it `EnvExports`. Good call.

Yeah, I drop the `Opt` because `devopt` already specifies it. For consistency, I can drop it from the other places as well? Happy to do a follow up.

internal/impl/devbox.go
savil: I don't like this one. I'll explain elsewhere.

internal/impl/devbox.go
savil: The idea is that we are computing the `devboxEnvPath`. Even though the initial `env["PATH"]` is taken from `nix`, in this function we are transforming this variable to be the "PATH of the devbox environment".

internal/impl/packages.go
savil: I was a tad concerned we'd have some other state (now, or in the future) that is not handled here and so wanted to qualify `State`. But don't feel strongly. Happy to shorten it.

internal/impl/devbox.go
savil: can just make it `envPath` or `path` as well

savil:
> Some comments. Approving to unblock.
> 
> Only blocker for me:
> 
> `computeDevboxEnv` vs `computeCurrentDevboxEnv`
> 
> This distinction means nothing to me. `computeCurrentDevboxEnv` should be something like `ensureProjectStateIsCurrentAndComputeEnv`. If you want to shorten it could also be `ensureStateIsCurrentAndComputeEnv`
> 
> I think when a function is complex, giving it a longer name for clarity is better.
> 
> I also think in many cases we can drop `Devbox` from variables and function names. So `computeEnv` or `computeEnvironment`. At the end of the day, the struct is named devbox so it's fairly clear what the subject is.

The distinction is: `computeEnvThatWillBeUpToDate` versus `computeEnvThatMightBeStale`.

I don't see the value in being so explicit TBH, because then we'll end up with super long sentences for functions.

`ensureStateIsCurrentAndComputeEnv` is not a good name because it focuses on the implementation and not the actual meaning. The goal is to get the environment that we can guarantee is up-to-date.

If you're not sure the first time you encounter the name, you can inspect the docblock of the function to understand what it does, and in your mind you'll realize "Ah, the _current_ is referring to _ensuring the environment is current or up to date_". And henceforth it'll be clear.

internal/impl/devbox.go
mikeland73: You can leave as is, not a blocker.

mikeland73:
> The distinction is: computeEnvThatWillBeUpToDate versus computeEnvThatMightBeStale.

This is not quite why `ensurePackagesAreInstalledAndComputeEnv` was introduced (see below)

> ensureStateIsCurrentAndComputeEnv is not a good name because it focuses on the implementation and not the actual meaning goal. The goal is to get the environment that we can guarantee is up-to-date.

My issues with `computeCurrentDevboxEnv` are:

* It's very similar to `computeDevboxEnv`. 
* `current` is not a strong enough descriptor of what it is doing. `current` means "the current thing" and it also means "up to date". The former is probably more common. 
* The function is not only computing the up to date environment, it is making the internal state be up to date. It changes the hidden files and installs nix packages. I'm also not sure I agree that making it up to date is not a goal (I would argue it definitely is). This function was introduced because there were 3-4 places where we repeated the exact same steps, but the goal is to do both, 1) bring the state up to date 2) compute the env. 
* My view with naming is that when meaning in doubt, be explicit. 

Does that make sense?

devbox.go
savil: Upon further inspection, I am renaming to `EnvExportOpts` and adding this naming guideline in the devopts package:
```
// Naming Convention:
// - suffix Opts for structs corresponding to a Devbox api function
// - omit suffix Opts for other structs that are composed into an Opts struct
```

savil:
> > The distinction is: computeEnvThatWillBeUpToDate versus computeEnvThatMightBeStale.
> 
> This is not quite why `ensurePackagesAreInstalledAndComputeEnv` was introduced (see below)
> 
> > ensureStateIsCurrentAndComputeEnv is not a good name because it focuses on the implementation and not the actual meaning goal. The goal is to get the environment that we can guarantee is up-to-date.
> 
> My issues with `computeCurrentDevboxEnv` are:
> 
> * It's very similar to `computeDevboxEnv`.
> * `current` is not a strong enough descriptor of what it is doing. `current` means "the current thing" and it also means "up to date". The former is probably more common.

I'd already indicated that I wasn't sure if `current` was best. I've switched to `upToDate`.

> * The function is not only computing the up to date environment, it is making the internal state be up to date. It changes the hidden files and installs nix packages. I'm also not sure I agree that making it up to date is not a goal (I would argue it definitely is). This function was introduced because there were 3-4 places where we repeated the exact same steps, but the goal is to do both, 1) bring the state up to date 2) compute the env.
> * My view with naming is that when meaning in doubt, be explicit.
> 
> Does that make sense?

Fair enough. And its not such a big deal to me so I've changed it for now. 

But I have added a TODO to find a better name.

I consider it a mild code-smell to introduce a function when the callers can just directly invoke the core-primitives themselves (in this case, `ensureStateIsUpToDate` and `computeEnv`). By adding this function, we do get the small convenience of avoiding two function calls, but then add the cognitive overhead of needing to remember that "oh, we also have this small convenience function". 

And then, over time, as has happened here with the "no such host" check, we introduce behavior changes as well. So, it is no longer exactly composing the two functions. And then this begs the question, what is the "core concept of this function" and when should one invoke it versus directly invoking the two functions it is composing?

PR #1678, opened 2023-12-08
[envsec] Add (hidden) envsec init command

mikeland73:
## Summary

TSIA

## How was it tested?

`devbox envsec init`

### DIFF
--- a/internal/boxcli/envsec.go
+++ b/internal/boxcli/envsec.go
@@ -1,0 +1,58 @@
+// Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package boxcli
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"go.jetpack.io/devbox"
+	"go.jetpack.io/devbox/internal/impl/devopt"
+	"go.jetpack.io/devbox/internal/integrations/envsec"
+)
+
+type envsecInitCmdFlags struct {
+	config configFlags
+}
+
+func envsecCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "envsec",
+		Short: "envsec commands",
+	}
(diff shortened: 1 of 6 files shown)


internal/boxcli/envsec.go
```diff
@@ -0,0 +1,58 @@
...
+	})
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	box.Config().SetStringField("EnvFrom", "envsec")
+	if err := box.Config().SaveTo(box.ProjectDir()); err != nil {
+		return errors.WithStack(err)
+	}
+
+	return envsec.EnsureInitialized(cmd.Context(), box.ProjectDir())
```
savil, line 57: can we swap the order of envsec.EnsureInitialized and modifying the devbox.json?

i.e. I feel we should only modify the devbox.json once we know envsec has been properly initialized.

internal/devconfig/env.go
```diff
@@ -18,7 +18,8 @@ func (c *Config) ComputedEnv(
 	if c.IsEnvsecEnabled() {
 		env, err = envsec.Env(ctx, projectDir)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error reading secrets from envsec: %s\n\n", err)
+			ux.Ferror(os.Stderr, "Error reading secrets from envsec: %s\n\n", err)
```
savil, line 21: Given that we are continuing on, should this be a `ux.Fwarning` instead?

savil: i.e. something like:

```
ux.Fwarning(os.Stderr, "Computing Devbox environment without envsec secrets. There was an error from envsec: %s\n\n", err)
```

savil (approved):

internal/devconfig/env.go
mikeland73: I'll can switch to warning. My motivation is that it's a fairly big error (not really a warning) so I wanted to make sure it's clear. But generally we don't use errors for commands that end in success.

Issue #1680, opened 2023-12-10
[Bug]: PHP xdebug extension is not being loaded


skyosev:
**Current Behavior (bug)**

After installing `php82Extensions.xdebug@latest`, PHP does not load it, rendering xdebug unusable.

Tried to add `zend_extension=xdebug.so` in `php.ini`, but this produces the following error when executing `php -v` in shell:

_Warning: Failed loading Zend extension 'xdebug.so' (tried: /nix/store/jzq5y2f3ivhyapw69rr3vzwqr391psaw-php-8.2.13/lib/php/extensions/xdebug.so (/nix/store/jzq5y2f3ivhyapw69rr3vzwqr391psaw-php-8.2.13/lib/php/extensions/xdebug.so: cannot open shared object file: No such file or directory), /nix/store/jzq5y2f3ivhyapw69rr3vzwqr391psaw-php-8.2.13/lib/php/extensions/xdebug.so.so (/nix/store/jzq5y2f3ivhyapw69rr3vzwqr391psaw-php-8.2.13/lib/php/extensions/xdebug.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0_

**Expected Behavior (fix)**

The expected  `php -v` output should be:

```
PHP 8.2.12 (cli) (built: Oct 26 2023 17:33:49) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.12, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.12, Copyright (c), by Zend Technologies
    with Xdebug v3.2.1, Copyright (c) 2002-2023, by Derick Rethans
```

**Additional context**

`devbox.json`

```
{
  "packages": [
    "php@8.2",
    "php82Extensions.xdebug@latest",
  ]
}
```

PR #1687, opened 2023-12-20
[plugins] Ensure plugin flakes get upgraded on every install

mikeland73:
## Summary

Run `nix profile upgrade` on any virtenv (plugin) flakes after any package is installed. This is a bit blunt, but works. (ideally we determine which plugin is being touched and we only upgrade that flake, but that's a bit tricky)

Fixes https://github.com/jetpack-io/devbox/issues/1680

## How was it tested?

```bash
devbox add php
devbox add php82Extensions.xdebug
devbox run php -m | grep xdebug
```

### DIFF
--- a/internal/impl/packages.go
+++ b/internal/impl/packages.go
@@ -392,7 +392,7 @@ func (d *Devbox) syncPackagesToProfile(ctx context.Context, mode installMode) er
 				return err
 			}
 			pending = append(pending, pkg)
-		} else if f, _ := pkg.FlakeInstallable(); d.pluginManager.PathIsInVirtenv(f.Ref.Path) {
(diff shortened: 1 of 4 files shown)


internal/impl/packages.go
```diff
@@ -390,6 +392,10 @@ func (d *Devbox) syncPackagesToProfile(ctx context.Context, mode installMode) er
 				return err
 			}
 			pending = append(pending, pkg)
+		} else if f, _ := pkg.FlakeInstallable(); d.pluginManager.PathIsInVirtenv(f.Ref.Path) {
```
savil, line 395: lets not ignore the error here? `f, _ := pkg.FlakeInstallable();`

savil (approved):

PR #1692, opened 2023-12-22
Derive nix profile from the generated flake

savil:
## Summary

This PR seeks to unify our `nix print-dev-env` and `nix profile` steps. Other benefits:
1. It will enable us to remove some hacks (example, this one [introduced for glibc-patch](https://github.com/jetpack-io/devbox/blob/d66f6ea5388adb3b90008c64a1b9b9abdc4550cd/internal/devbox/devbox.go#L938-L944))
2. Re-enable some testscripts that were disabled (example, this [one for php extension removal](https://github.com/jetpack-io/devbox/blob/main/testscripts/languages/php.test.txt#L4-L15))

Previously, these would be two separate processes: we'd manually do `nix profile install`.
Now, we get the `buildInputPaths` from `nix print-dev-env` and install those, so the nix profile is _derived_ from the flake.

Some highlights of code movements/changes:
1. Adds `devbox.syncNixProfile`. Initially copied from #1588, and then modified quite a bit.
2. Move `devpkg.storePathParts` to `nix.StorePathParts` so it can be called from `syncFlakeToProfile`
3. Needed to modify `nixprofile.ProfileInstall` to work on `Installables` rather than package-names, so it can install `/nix/store/<hash>-<pkg>` buildPaths from the flake.
4. Moved `validatePackages` from `nixprofile.ProfileInstall` to `ensureStateIsUpToDate`, since we must run it prior to `computeEnv` that evaluates `nix print-dev-env`.
5. Added `devbox.installNixPackagesToStore` so that packages are locally installed in nix store when adding or removing or updating, and not in devbox-environment.

**Behavior change:**
This PR leads to more `computeEnv` calls that do `nix print-dev-env` in our Examples tests. The reason is that `devbox add/rm/update` will always invalidate the `local.lock` cache, leading to subsequent operations like `devbox run/shell/shellenv` to invoke `computeEnv`. Do note that this is _more correct_ than before where the `nix print-dev-env` cache would be stale for `devbox add/rm/update` when not in a devbox environment.

It led to some Examples testscripts failing for nix 2.17.0 due to errors like https://gist.github.com/savil/03ae7f4e55c0796a60bdf0cfae588fac. It appears this was an [issue with `nix` that was fixed](https://github.com/NixOS/nix/issues/6253), affecting nix operations that happen in parallel, which affects our cicd tests.

The issue was fixed in August 2023, so it doesn't fail for nix 2.19.  It also doesn't fail for nix 2.12, so perhaps it was an issue introduced in between.

For now, I've dropped 2.17 from the nix versions that we run the cli-tests on. I think this is okay, since we still test with nix 2.12 and 2.19. 

## How was it tested?

1. Automated tests should pass.
2. Did `devbox add` and `devbox rm` with `hello` and `cowsay` and also with the process-compose flake.
3. `devbox add glibcLocales` prints the correct error message and suggestion to use `--exclude-platform`. And then verified that `devbox add glibcLocales --exclude-platform x86_64-darwin` works as expected.

is there some edge case I may be missing?

### DIFF
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -2,7 +2,6 @@ package devbox
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -12,25 +11,17 @@ import (
 
 // syncFlakeToProfile ensures the buildInputs from the flake's devShell are
 // installed in the nix profile.
-func (d *Devbox) syncFlakeToProfile(ctx context.Context) error {
+// buildInputs is a space-separated list of store paths from the nix print-dev-env output's buildInputs.
+func (d *Devbox) syncFlakeToProfile(ctx context.Context, buildInputs string) error {
 	profilePath, err := d.profilePath()
 	if err != nil {
 		return err
 	}
 
-	// Get the build inputs (i.e. store paths) from the generated flake's devShell.
-	buildInputPaths, err := nix.Eval(
-		ctx,
-		d.stderr,
-		d.flakeDir()+"#devShells."+nix.System()+".default.buildInputs",
-		"--json",
-	)
-	if err != nil {
-		return fmt.Errorf("nix eval devShells: %v", err)
-	}
-	storePaths := []string{}
-	if err := json.Unmarshal(buildInputPaths, &storePaths); err != nil {
-		return fmt.Errorf("unmarshal store paths: %s: %v", buildInputPaths, err)
+	// Get the build inputs (i.e. store paths) from the generated flake's print-dev-env output
+	wantStorePaths := []string{}
+	if buildInputs != "" { // if buildInputs is empty, then we don't want wantStorePaths to be an array with a single "" entry
+		wantStorePaths = strings.Split(buildInputs, " ")
 	}
 
 	// Get the store-paths of the packages currently installed in the nix profile
@@ -38,13 +29,13 @@ func (d *Devbox) syncFlakeToProfile(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("nix profile list: %v", err)
 	}
-	got := make([]string, 0, len(items))
+	gotStorePaths := make([]string, 0, len(items))
 	for _, item := range items {
-		got = append(got, item.StorePaths()...)
+		gotStorePaths = append(gotStorePaths, item.StorePaths()...)
 	}
 
 	// Diff the store paths and install/remove packages as needed
-	add, remove := diffStorePaths(got, storePaths)
+	add, remove := diffStorePaths(gotStorePaths, wantStorePaths)
 	if len(remove) > 0 {
 		packagesToRemove := make([]string, 0, len(remove))
 		for _, p := range remove {
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -1,0 +1,88 @@
+package devbox
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"slices"
+
+	"go.jetpack.io/devbox/internal/nix"
+)
+
+func syncFlakeToProfile(ctx context.Context, flakePath, profilePath string) error {
+	cmd := exec.CommandContext(ctx, "nix", "eval", "--json", flakePath+"#devShells."+nix.System()+".default.buildInputs")
+	b, err := cmd.Output()
+	if err != nil {
+		return fmt.Errorf("nix eval devShells: %v", err)
+	}
+	storePaths := []string{}
+	if err := json.Unmarshal(b, &storePaths); err != nil {
+		return fmt.Errorf("unmarshal store paths: %s: %v", b, err)
+	}
+
+	listCmd := exec.CommandContext(ctx, "nix", "profile", "list", "--json", "--profile", profilePath)
+	b, err = listCmd.Output()
+	if err != nil {
+		return err
+	}
+	var profile struct {
+		Elements []struct {
+			StorePaths []string
+		}
+	}
+	if err := json.Unmarshal(b, &profile); err != nil {
+		return fmt.Errorf("unmarshal profile: %v", err)
+	}
+	got := make([]string, 0, len(profile.Elements))
+	for _, e := range profile.Elements {
+		got = append(got, e.StorePaths...)
+	}
+
+	add, remove := diffStorePaths(got, storePaths)
+	if len(remove) > 0 {
+		removeCmd := exec.CommandContext(ctx, "nix", "profile", "remove", "--profile", profilePath)
+		removeCmd.Args = append(removeCmd.Args, remove...)
+		if err := removeCmd.Run(); err != nil {
+			return err
+		}
+	}
+	if len(add) > 0 {
+		addCmd := exec.CommandContext(ctx, "nix", "profile", "install", "--profile", profilePath)
+		addCmd.Args = append(addCmd.Args, add...)
+		if err := addCmd.Run(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func diffStorePaths(got, want []string) (add, remove []string) {
+	slices.Sort(got)
+	slices.Sort(want)
+
+	var g, w int
+	for {
+		if g >= len(got) {
+			add = append(add, want[w:]...)
+			break
+		}
+		if w >= len(want) {
+			remove = append(remove, got[g:]...)
+			break
+		}
+
+		switch {
+		case got[g] == want[w]:
+			g++
+			w++
+		case got[g] < want[w]:
+			remove = append(remove, got[g])
+			g++
+		case got[g] > want[w]:
+			add = append(add, want[w])
+			w++
+		}
+	}
+	return add, remove
+}
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/nix/nixprofile"
+	"go.jetpack.io/devbox/internal/ux"
 )
 
 // syncNixProfile ensures the nix profile has the packages specified in wantStorePaths.
@@ -28,7 +30,7 @@ func (d *Devbox) syncNixProfile(ctx context.Context, wantStorePaths []string) er
 	}
 
 	// Diff the store paths and install/remove packages as needed
-	add, remove := diffStorePaths(gotStorePaths, wantStorePaths)
+	remove, add := lo.Difference(gotStorePaths, wantStorePaths)
 	if len(remove) > 0 {
 		packagesToRemove := make([]string, 0, len(remove))
 		for _, p := range remove {
@@ -36,9 +38,9 @@ func (d *Devbox) syncNixProfile(ctx context.Context, wantStorePaths []string) er
 			packagesToRemove = append(packagesToRemove, fmt.Sprintf("%s@%s", storePath.Name, storePath.Version))
 		}
 		if len(packagesToRemove) == 1 {
-			fmt.Fprintf(d.stderr, "Removing %s\n", strings.Join(packagesToRemove, ", "))
+			ux.Finfo(d.stderr, "Removing %s\n", strings.Join(packagesToRemove, ", "))
 		} else {
-			fmt.Fprintf(d.stderr, "Removing packages: %s\n", strings.Join(packagesToRemove, ", "))
+			ux.Finfo(d.stderr, "Removing packages: %s\n", strings.Join(packagesToRemove, ", "))
 		}
 
 		if err := nix.ProfileRemove(profilePath, remove...); err != nil {
@@ -69,28 +71,4 @@ func (d *Devbox) syncNixProfile(ctx context.Context, wantStorePaths []string) er
 		}
 	}
 	return nil
-}
-
-func diffStorePaths(got, want []string) (add, remove []string) {
-	gotSet := map[string]bool{}
-	for _, g := range got {
-		gotSet[g] = true
-	}
-	wantSet := map[string]bool{}
-	for _, w := range want {
-		wantSet[w] = true
-	}
-
-	for _, g := range got {
-		if _, ok := wantSet[g]; !ok {
-			remove = append(remove, g)
-		}
-	}
-
-	for _, w := range want {
-		if _, ok := gotSet[w]; !ok {
-			add = append(add, w)
-		}
-	}
-	return add, remove
 }
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -9,19 +9,12 @@ import (
 	"go.jetpack.io/devbox/internal/nix/nixprofile"
 )
 
-// syncFlakeToProfile ensures the buildInputs from the flake's devShell are
-// installed in the nix profile.
-// buildInputs is a space-separated list of store paths from the nix print-dev-env output's buildInputs.
-func (d *Devbox) syncFlakeToProfile(ctx context.Context, buildInputs string) error {
+// syncNixProfile ensures the nix profile has the packages specified in wantStorePaths.
+// It also removes any packages from the nix profile that are not in wantStorePaths.
+func (d *Devbox) syncNixProfile(ctx context.Context, wantStorePaths []string) error {
 	profilePath, err := d.profilePath()
 	if err != nil {
 		return err
-	}
-
-	// Get the build inputs (i.e. store paths) from the generated flake's print-dev-env output
-	wantStorePaths := []string{}
-	if buildInputs != "" { // if buildInputs is empty, then we don't want wantStorePaths to be an array with a single "" entry
-		wantStorePaths = strings.Split(buildInputs, " ")
 	}
 
 	// Get the store-paths of the packages currently installed in the nix profile
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"slices"
 	"strings"
 
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/nix/nixprofile"
 )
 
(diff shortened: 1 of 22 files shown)


savil:
* `main`
  * **PR #1692** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1692?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1692?utm_source=stack-comment).

savil:
fails with:
```
--- FAIL: TestScripts (0.05s)
    --- FAIL: TestScripts/php.test (319.64s)
        testscript.go:534: > exec devbox run php index.php
            [stderr]
            Error: error running script "php" in Devbox: Command: /nix/var/nix/profiles/default/bin/nix print-dev-env $WORK/.devbox/gen/flake --extra-experimental-features ca-derivations --option experimental-features nix-command flakes fetch-closure --json: exit status 1

            program not built with -cover
            [exit status 1]
            FAIL: languages/php.test.txt:1: unexpected command failure

    --- FAIL: TestScripts/unfree.test (353.82s)
        testscript.go:534: # Test ensures that we can add and remove "unfree" nix packages (1.499s)
            # we could test with slack and/or vscode. Using slack since it is lighter. (352.309s)
            > exec devbox add slack
            [stdout]
            Running /nix/var/nix/profiles/default/bin/nix eval --json $WORK/.devbox/gen/flake#devShells.x86_64-darwin.default.buildInputs --extra-experimental-features ca-derivations --option experimental-features nix-command flakes fetch-closure
            [stderr]
            Ensuring nixpkgs registry is downloaded.
            Downloaded 'github:NixOS/nixpkgs/fd04bea4cbf76f86f244b9e2549fca066db8ddff' to '/nix/store/pfixw35k0jwhz6mrhd6jw0vlppjjkamg-source' (hash 'sha256-imn+/Rj+bqagOSm7GmRDbrqkxtc7QnjY3Cu85gv46BU=').
            Ensuring nixpkgs registry is downloaded: Success
            Info: Adding package "slack@latest" to devbox.json
            program not built with -cover
            > stderr 'Installing package: slack.'
            FAIL: packages/unfree.test.txt:7: no match for `Installing package: slack.` found in stderr

FAIL
coverage: [no statements]
FAIL	go.jetpack.io/devbox/testscripts	438.322s
FAIL
Error: error running script "test" in Devbox: exit status 1
```

savil:
Made some progress. 

The `unfree` testscript needed a update for which output to look for. So fixed.

The php-testscript was failing due to `nix profile install` adding multiple buildInputs for the same package (in this case, `composer` is added by both the php plugin and as a devbox package). Made a change to reuse our current approach of setting `--priority` in the `nix profile install`.

So, this is very much usable. However, I feel there are some UX fixes to be made:
1. We'll need to prefetch `nixpkgs` for when we are going to download them. This allows us to show a nicer message than when the `flake.nix` downloads them as a side-effect of being evaluated.
2. `devbox rm <pkg>` needs UX improvement to print some nice messages. The current messages have gotten dropped.
3. Adding a package to the nix profile currently displays the `/nix/store` path as in:
```
[1/1] /nix/store/c9mw21jbrpprfd8vh735zvangxmh7j5a-cowsay-3.7.0
[1/1] /nix/store/c9mw21jbrpprfd8vh735zvangxmh7j5a-cowsay-3.7.0: Success
```
We should parse the package name and version from this path, and display just that.

But this seems pretty doable and should simplify our logic quite a bit.

savil:
Currently, running into this CICD error:

```
       2024/01/03 01:37:52 running command: /nix/var/nix/profiles/default/bin/nix profile install --profile $WORK/.devbox/nix/profile/default --impure --priority 8 /nix/store/cyqwmgbir56zsn600z0sgnvdribnfgpj-process-compose-0.40.2 --extra-experimental-features ca-derivations --option experimental-features nix-command flakes fetch-closure
        don't know how to build these paths:
          /nix/store/cyqwmgbir56zsn600z0sgnvdribnfgpj-process-compose-0.40.2
        error: path '/nix/store/cyqwmgbir56zsn600z0sgnvdribnfgpj-process-compose-0.40.2' is required, but there is no substituter that can build it
```

internal/devbox/packages.go
```diff
@@ -536,3 +387,37 @@ func (d *Devbox) InstallRunXPackages(ctx context.Context) error {
...
+		inCache, err := pkg.IsInBinaryCache()
+		if err != nil {
+			return err
+		}
+
+		if !inCache && nix.IsGithubNixpkgsURL(pkg.URLForFlakeInput()) {
+			if err := nix.EnsureNixpkgsPrefetched(d.stderr, pkg.HashFromNixPkgsURL()); err != nil {
+				return err
+			}
+			if exists, err := pkg.ValidateInstallsOnSystem(); err != nil {
```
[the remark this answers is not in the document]
savil: I think one approach is:
1. fetch existing profile-items from nix profile
2. diff with existing packages in the config to find packages that need to be added to the profile
3. run the validation check on just those.

savil:
converting to Draft while I fix the perf issues with package validation.

internal/devbox/packages.go
```diff
@@ -290,10 +293,21 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 		return err
 	}
 
-	// Use the printDevEnvCache if we are adding or removing or updating any package,
-	// AND we are not in the shellenv-enabled environment of the current devbox-project.
-	usePrintDevEnvCache := mode != ensure && !d.IsEnvEnabled()
-	if _, err := d.computeEnv(ctx, usePrintDevEnvCache); err != nil {
```
mikeland73, line 296: This undos an optimization which is to skip recomputing envs for add/rm when not in a devbox environment. 

I think the reason we introduced that is because sometimes this triggered some slow stuff like downloading nixpkgs for some other package. I'm not entirely sure this is fixable, just a tradeoff.

internal/devbox/nixprofile.go
```diff
@@ -0,0 +1,96 @@
...
+				Writer:      d.stderr,
+			}); err != nil {
+				return fmt.Errorf("error installing package %s: %w", addPath, err)
+			}
+		}
+	}
+	return nil
+}
+
+func diffStorePaths(got, want []string) (add, remove []string) {
```
mikeland73, line 74: I think `lo.Difference` does this

internal/nix/profiles.go
```diff
@@ -29,47 +30,58 @@ func ProfileList(writer io.Writer, profilePath string, useJSON bool) (string, er
...
 		// Using an arbitrary priority to avoid conflicts with other packages.
 		// Note that this is not really the priority we care about, since we
 		// use the flake.nix to specify the priority.
-		"--priority", nextPriority(profilePath),
-		installable,
+		"--priority", nextPriority(args.ProfilePath),
 	)
+	if args.Offline {
+		cmd.Args = append(cmd.Args, "--offline")
+	}
```
mikeland73, line 63: I'm not totally familiar with the implications of this.

internal/nix/nixprofile/upgrade.go
```diff
@@ -4,25 +4,27 @@
...
 
 	"go.jetpack.io/devbox/internal/devpkg"
 	"go.jetpack.io/devbox/internal/lock"
 	"go.jetpack.io/devbox/internal/nix"
 )
 
-func ProfileUpgrade(ProfileDir string, pkg *devpkg.Package, lock *lock.File) error {
+func ProfileUpgrade(profileDir string, pkg *devpkg.Package, lock *lock.File) error {
 	idx, err := ProfileListIndex(
+		context.TODO(),
```
mikeland73, line 17: :(

mikeland73:
I lean yes on this, but just want to flag the risk on a potential performance regression on add/remove. 

Imagine removing a package and if you are missing nixpkgs for something or if a package was evicted from the nix store, you have to wait for the full sync. Not what you were expecting.

savil:
@mikeland73 yeah, that is a trade off for sure. I tried to acknowledge that in the PR summary (see Behavior Change section)

gcurtis (commented):
I think this is a good step toward making the Devbox environment more stable. There's a bunch of logic we can still offload to Nix to simplify and avoid perf issues (particularly with the `validatePackagesToBeInstalled` logic). With the POC I was trying to break down Devbox environment startup into the following phases:

1. Cache check - check the Devbox cache hash; skip to the last step if it's a hit.
2. Resolve packages - resolve the Devbox packages to installables using the lockfile and the search service.
3. Generate flake - create `.devbox/gen/flake/flake.nix` using the list of installables.
4. Build flake - build `.devbox/gen/flake/flake.nix` whose output is the shell environment; skip to the last step if nothing needed to be built.
5. Sync profile - sync the profile's store paths with the flake's `buildInputs` store paths.
6. Shell - start the shell environment.

Step 1 we already do.

Step 2 I think we also already do this. It should be fast unless we need to hit the search service to resolve a new package.

Step 3 could be cleaned up, but I think improving the other steps are more of a win. Specifically, right now nixpkgs is re-evaluated for the main flake + plugin flakes + the glibc flake.

Step 4 would be the biggest win. Make everything get installed by running `nix build` in `.devbox/gen/flake` (no separate `nix profile install` commands). Nix will only build/download packages that aren't already in the store, making it instantaneous if all packages are already installed. Otherwise it scales linearly with the number of new packages. Some implementation notes:

- Wrap each package in the flake with `builtins.tryEval` and `builtins.trace` to print a nicer error message (or a structured error message that Devbox can easily parse) if a package cannot be installed.
- `nix build` prints nothing if everything is already up-to-date so we can skip profile syncing.
- For pragmatism, we would also do things like add runx packages to the PATH outside of Nix. But ideally we'd push as much into the flake as we can.

Step 5 do what we're doing here with diffing the store paths. This is reasonably fast because the flake is already built and all the packages are already installed. In the worst case, we only need to run a single `nix profile remove <store-paths>` and a single `nix profile install <store-paths>` command to sync the profile.

The reason why I think this should all be fast is because `nix develop` is fast. If you run something like `nix develop .#devShells.aarch64-darwin.default` in `.devbox/gen/flake` it installs packages faster than Devbox. If you run it a second time, the shell starts instantly (not quite as fast as a cached Devbox env) because Nix saw that everything was up-to-date. If you add a package to the flake and run it a third time, you only see the one new package get installed.

gcurtis (approved):
LGTM if performance is still good.

internal/devbox/packages.go
```diff
@@ -536,3 +394,75 @@ func (d *Devbox) InstallRunXPackages(ctx context.Context) error {
 	}
 	return nil
 }
+
+// validatePackagesToBeInstalled will ensure that packages are available to be installed
+// in the user's current system.
+func (d *Devbox) validatePackagesToBeInstalled(ctx context.Context) error {
```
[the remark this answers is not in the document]
mikeland73: I think the issue is that the error message is bad. Maybe in that case we need to catch this error and show better error message

savil:
Perf fixes:
1. For `devbox add/rm` outside devbox environment, we'll skip the `computeEnv` and `syncNixProfile` steps.
2. We'll augment `validatePackagesToBeInstalled` to actually do the install by doing `nix build <package.installable>` so that `devbox add` can successfully install the package.

savil:
@mikeland73 @gcurtis PTAL. I think i've addressed the perf concern.

For the scenario we were concerned about (`devbox add/rm/update` not in a devbox environment) I now skip `computeEnv` and `syncNixProfile` entirely, and install into nix store via `nix build`. We then do the rest of the operations and delete the `local.lock` file (simplest way to "mark it dirty"), so that the next time we do `ensureStateIsUpToDate(mode=ensure)` it will recompute the env and syncNixProfile.

This PR is getting rather large. Let me know if you'd like me to split it up.

internal/devbox/packages.go
```diff
@@ -248,10 +248,20 @@ const (
...
+func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) error { //nolint:revive
 	defer trace.StartRegion(ctx, "devboxEnsureStateIsUpToDate").End()
 	defer debug.FunctionTimer().End()
 
+	if mode != ensure && !d.IsEnvEnabled() {
+		// if mode is install/uninstall/update and we are not in a devbox environment,
+		// then we skip some operations below for speed.
+		// Remove the local.lock file so that we re-compute the project state when
+		// we are in the devbox environment again.
+		defer func() { _ = d.lockfile.RemoveLocal() }()
```
savil, line 262: Actually, this should get moved to the else clause below (where we skip computeEnv and syncNixProfile)

internal/devbox/packages.go
```diff
@@ -248,10 +248,20 @@ const (
 // The `mode` is used for:
 // 1. Skipping certain operations that may not apply.
 // 2. User messaging to explain what operations are happening, because this function may take time to execute.
-func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) error {
+//
+// linter:revive is turned off due to complaining about function complexity
+func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) error { //nolint:revive
```
mikeland73, line 253: I think the logic an flow of this function can be improved (and maybe that can help split PRs) to make this lower risk. Proposal:

* Split lockfile.Save() into `lockfile.Save()` and `lock.UpdateAndSaveStateHash()`. This should actually live in different packages, but baby steps. The first function saves the lockfile, the other only deals with the hash file.
* define `mode == ensure || d.IsEnvEnabled()` as `recomputeLocalState`
* if `!recomputeLocalState` then we only add and build packages, or remove.
* if `recomputeLocalState` we `syncFlake`.
* if `recomputeLocalState` we call `lock.UpdateAndSaveStateHash()` 
* we always save lockfile at the end. 

In terms of splitting PRs we could split out the lockfile functions and do that first. Followed by most of the rest

internal/devbox/packages.go
```diff
@@ -248,10 +248,20 @@ const (
 // The `mode` is used for:
 // 1. Skipping certain operations that may not apply.
 // 2. User messaging to explain what operations are happening, because this function may take time to execute.
-func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) error {
+//
+// linter:revive is turned off due to complaining about function complexity
```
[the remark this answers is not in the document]
savil: Ah yeah, some context:
1. I wanted to see if the code would pass all cicd tests, and didn't want to do big refactoring before getting that information.
2. I like this linter too, but its a bit vague about how its rules work. How does it decide what is complex? How do i refactor my code?
3. I then didn't invest time in removing this, since I was awaiting feedback on the general approach from you and Greg.

internal/devbox/packages.go
```diff
@@ -290,11 +303,34 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
...
+		}
+		if err := d.syncNixProfile(ctx, buildInputs); err != nil {
+			return err
+		}
+
+	} else if mode == install || mode == update {
+		// Else: if we are not in a devbox environment, and we are installing or updating
+		// then we must ensure the new nix packages are in the nix store. This way, the
+		// next time we enter a devbox environment, we will have the packages available locally.
+		if err := d.installNixPackagesToStore(ctx); err != nil {
```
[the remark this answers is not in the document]
savil: Since we may be installing plugins, I thought it should come after the plugin manager create, and remove plugin invalid symlinks.

Will see if I can combine validatePackagesToBeInstalled with `nix build` and move that below the pluginmanager setup functions.

internal/nix/nixprofile/item.go
```diff
@@ -83,3 +83,7 @@ func (i *NixProfileListItem) String() string {
 		i.nixStorePaths,
 	)
 }
+
+func (i *NixProfileListItem) StorePaths() []string {
+	return i.nixStorePaths
+}
```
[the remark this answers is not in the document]
savil: since everything in this struct is private and controlled, will prefer leaving it this way

internal/nix/nixprofile/profile.go
```diff
@@ -189,57 +194,34 @@ func parseNixProfileListItem(line string) (*NixProfileListItem, error) {
 	}, nil
 }
 
+// TODO drop the Profile prefix from this name.
 type ProfileInstallArgs struct {
 	CustomStepMessage string
-	Lockfile          *lock.File
-	Package           string
+	Installable       string
+	Offline           bool
```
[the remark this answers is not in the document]
savil: `nix build` has its own `--offline` flag.

We do need the `--offline` flag for the reasons I've explained elsewhere

internal/nix/profiles.go
savil: > A ['substituter'](https://nixos.org/manual/nix/unstable/glossary.html#gloss-substituter) in Nix terminology is an upstream server that can serve prebuilt binaries for a given derivation (it can 'substitute' the binary instead of having to build from source).

yes, we are okay disabling them because the `nix print-dev-env` has already evaluated the derivations for the build inputs. And so the binaries are already present on the user's machine.

internal/devbox/packages.go
```diff
@@ -267,17 +277,20 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 		fmt.Fprintln(d.stderr, "Ensuring packages are installed.")
 	}
 
+	// Validate packages. Must be run up-front and definitely prior to computeEnv
+	// and syncNixProfile below that will evaluate the flake and may give
+	// inscrutable errors if the package is uninstallable.
```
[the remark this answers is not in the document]
savil: will see if that may work

internal/devbox/packages.go
savil: the lockfile.save function is:
```
func (f *File) Save() error {
	if err := cuecfg.WriteFile(lockFilePath(f.devboxProject), f); err != nil {
		return err
	}
	return updateLocal(f.devboxProject)
}
```
splitting this is a pretty small change. Here I'll do it in this comment itself:
```
func (f *File) Save() error {
	return cuecfg.WriteFile(lockFilePath(f.devboxProject), f)
}
func (f *File) UpdateAndSaveStateHash() error {
	return updateLocal(f.devboxProject)
}
```
It feels a bit unnecessary to make a separate PR just for this...

savil: I do like the overall breakdown of the logic you present. Let me refactor into that.

savil:
re-written in #1724 and the PR stack below that.

PR #1693, opened 2023-12-22
[auth] Improve whoami

mikeland73:
## Summary

TSIA

### DIFF
--- a/internal/integrations/envsec/envsec.go
+++ b/internal/integrations/envsec/envsec.go
@@ -63,8 +63,7 @@ func EnsureInstalled(ctx context.Context) (string, error) {
 }
 
 func ensureInitialized(projectDir string) error {
-	envsec := DefaultEnvsec(os.Stderr)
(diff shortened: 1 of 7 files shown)


internal/integrations/envsec/envsec.go
```diff
@@ -119,15 +120,14 @@ func handleError(stderr *bytes.Buffer, err error) error {
 	return errors.WithStack(err)
 }
 
-func DefaultEnvsec(stderr io.Writer, workingDir string) *envsec.Envsec {
+func DefaultEnvsec(stderr io.Writer) *envsec.Envsec {
```
savil, line 123: Given that most uses of this function are setting `WorkingDir`, I feel we should preserve the parameter. Otherwise, its very easy to overlook or forget it needs to be set.

Can we pass in an empty-value for the "WhoAmI" use-case?

savil (approved):

PR #1700, opened 2024-01-04
[nix] ensure nix version is >= 2.12

savil:
## Summary

Ensures that the nix used is >= 2.12 which is the minimum version we support.
Our CICD tests run this version, and I believe so does @mohsenari.

I considered renaming the function to `ensureSupportedNixInstalled` but
decided against it. The current name is good enough, given the function's scope
and location in the package.

## How was it tested?

ran `devbox shell` which results in this function being called.
Toggled the condition to trigger the error to verify it prints as we want.

### DIFF
--- a/internal/boxcli/setup.go
+++ b/internal/boxcli/setup.go
@@ -4,13 +4,15 @@
 package boxcli
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/samber/lo"
 	"github.com/spf13/cobra"
-
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/nix"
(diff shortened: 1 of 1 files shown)


savil:
* `main`
  * **PR #1700** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1700?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1700?utm_source=stack-comment).

mohsenari:
Thanks for considering and remembering my nix install version 😄 
I upgraded to 2.18 a few weeks ago

mohsenari (approved):

PR #1703, opened 2024-01-05
[envsec] Ensure build environment is the same

mikeland73:
## Summary

This is combined with https://github.com/jetpack-io/opensource/pull/251 to ensure devbox and envsec environments are always the same.

Otherwise, when using dev devbox, it will install a prod envsec and environments won't match which causes issues because dev environment uses a different authentication issuer so keys are not compatible.

### DIFF
--- a/internal/devbox/util.go
+++ b/internal/devbox/util.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/build"
 	"go.jetpack.io/devbox/internal/integrations/envsec"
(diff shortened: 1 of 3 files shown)


internal/devbox/util.go
```diff
@@ -40,16 +41,20 @@ func (d *Devbox) addDevboxUtilityPackage(ctx context.Context, pkg string) error
 // process-compose, etc
 func (d *Devbox) addUtilitiesToPath(
 	ctx context.Context,
-	path string,
-) (string, error) {
+	env map[string]string,
```
savil, line 44: update the function comment to match the new function name?

savil (approved):

PR #1713, opened 2024-01-10
[secrets] Implement ls, set, rm secrets

mikeland73:
## Summary

Implements `secrets list`, `secrets set`, and `secrets rm`.

Removed dependency on envsec binary.

Renamed `envsec` commands to `secrets`

TODO:  Improve envsec interface to avoid having to use embedded struct. 

## How was it tested?

```
devbox secrets list
devbox secrets set foo=bar
devbox secrets set foo=barprod --environment prod
devbox run echo $\foo
devbox ls --environment prod
```


internal/boxcli/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+		Aliases: []string{"envsec"},
+		Short:   "Interact with devbox secrets.",
+	}
+	cmd.AddCommand(secretsInitCmd(flags))
+	cmd.AddCommand(secretsListCmd(flags))
+	cmd.AddCommand(secretsRemoveCmd(flags))
+	cmd.AddCommand(secretsSetCmd(flags))
+	cmd.Hidden = true
+
+	flags.config.registerPersistent(cmd)
```
savil, line 40: need to update the `envsec` wording to `secrets` in:
https://github.com/jetpack-io/devbox/blob/11bb6f05a4562d16ee572e54a11c13a6c9943346/internal/boxcli/config.go#L21-L30

internal/boxcli/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+	show   bool
+	format string
+}
+
+func secretsCmd() *cobra.Command {
+	flags := &secretsFlags{}
+	cmd := &cobra.Command{
+		Use:     "secrets",
+		Aliases: []string{"envsec"},
+		Short:   "Interact with devbox secrets.",
```
savil, line 32: Instead of `...devbox secrets` should it be `Interact with secrets in jetpack cloud` ?

internal/boxcli/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+	)
+
+	return cmd
+}
+
+func secretsSetCmd(flags *secretsFlags) *cobra.Command {
+	return &cobra.Command{
+		Use:   "set <NAME1>=<value1> [<NAME2>=<value2>]...",
+		Short: "Securely store one or more environment variables",
+		Long:  "Securely store one or more environment variables. To test contents of a file as a secret use set=@<file>",
```
savil, line 71: can you test the file-upload scenario?

internal/boxcli/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+				vars, cmd.OutOrStdout(), flags.show, flags.format)
+		},
+	}
+
+	cmd.Flags().BoolVarP(
+		&flags.show,
+		"show",
+		"s",
+		false,
+		"Display the value of each environment variable (secrets included)",
```
savil, line 177: It feels odd to mention `environment variable` when the command is `devbox secrets`.

Can this be: `Display the value of the secret [default = false]`?

internal/boxcli/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+	)
+	cmd.Flags().StringVarP(
+		&flags.format,
+		"format",
+		"f",
+		"table",
+		"Display the key values in key=value format. Must be one of: table | dotenv | json",
```
savil, line 184: `"Display the key values of each secret in the specified format, one of: table | dotenv | json. [default = table]"` ?

internal/devbox/devbox.go
```diff
@@ -1109,9 +1105,34 @@ func (d *Devbox) configEnvs(
...
+				"Error reading secrets from jetpack cloud: %s\n\n",
+				err,
+			)
+		} else {
+			for _, secret := range cloudSecrets[envID] {
+				env[secret.Name] = secret.Value
+			}
+		}
+	} else if d.cfg.EnvFrom != "" {
+		return nil, usererr.New("unknown from_env value: %s", d.cfg.EnvFrom)
```
savil, line 1132: `usererr.New("unknown from_env value: %s. Supported value is: envsec.", d.cfg.EnvFrom, )`

internal/devbox/devbox.go
```diff
@@ -1266,12 +1287,19 @@ func (d *Devbox) RunXPaths(ctx context.Context) (string, error) {
 	return runxBinPath, nil
 }
 
+func (d *Devbox) Environment() string {
+	return d.environment
+}
```
savil, line 1292: does this need to be made visible externally? Seems like its used in `internal/devbox/secrets.go` which can use `d.environment`

savil (approved):
neat!

internal/boxcli/secrets.go
mikeland73: works!

PR #1717, opened 2024-01-10
set the ownership of /code to devbox user

agate:
## Summary
When set to non `RootUser` for generating Dockerfile. We need to make sure `/code` owned by the devbox user for allowing the script to create new files / dirs in `/code` dir.

## How was it tested?
Locally, Manually


mohsenari (approved):

agate:
> Did you run into an issue that made you create this PR? I want to know the use case where this change is needed.
@mohsenari
When you create a dockerfile with this template without using the root user. The `/code` will still be under `root`'s ownership. So you will not be able to add new files into it. Let's say I want to create a tmp dir and adding some files during the run time. Hope this make sense to you

PR #1720, opened 2024-01-11
[secrets] Add secrets download and upload

mikeland73:
## Summary

Adds

```
devbox secrets download
devbox secrets upload
```

Also, updated envsec dependency that has a simpler API.

## How was it tested?

```
devbox secrets download foo.json
# modified foo.json
devbox secrets upload foo.json
```

### DIFF
--- a/internal/boxcli/secrets.go
+++ b/internal/boxcli/secrets.go
@@ -236,10 +236,10 @@ func secretsInitFunc(
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	secrets, err := box.Secrets(ctx)
-	if err != nil {
-		return errors.WithStack(err)
-	}
+
+	// devbox.Secrets() by default assumes project is initialized (and shows
+	// error if not). So we use UninitializedSecrets() here instead.
+	secrets := box.UninitializedSecrets(ctx)
 
 	if _, err := secrets.ProjectConfig(); err == nil &&
 		box.Config().EnvFrom != "jetpack-cloud" {
--- a/internal/boxcli/secrets.go
+++ b/internal/boxcli/secrets.go
@@ -16,7 +16,7 @@ type secretsFlags struct {
 	config configFlags
 }
 
-func (f *secretsFlags) genSecrets(cmd *cobra.Command) (*envsec.Envsec, error) {
+func (f *secretsFlags) envsec(cmd *cobra.Command) (*envsec.Envsec, error) {
 	box, err := devbox.Open(&devopt.Opts{
 		Dir:         f.config.path,
 		Environment: f.config.environment,
@@ -99,7 +99,7 @@ func secretsSetCmd(flags *secretsFlags) *cobra.Command {
 			return envsec.ValidateSetArgs(args)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			secrets, err := flags.genSecrets(cmd)
+			secrets, err := flags.envsec(cmd)
 			if err != nil {
 				return errors.WithStack(err)
 			}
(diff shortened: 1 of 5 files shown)


internal/boxcli/secrets.go
```diff
@@ -16,6 +16,19 @@ type secretsFlags struct {
 	config configFlags
 }
 
+func (f *secretsFlags) genSecrets(cmd *cobra.Command) (*envsec.Envsec, error) {
```
savil, line 19: `genSecrets`: are we _generating_ them? :)

savil: Thoughts on `Envsec()` to indicate the struct being returned?

internal/boxcli/secrets.go
```diff
@@ -188,6 +166,63 @@ func secretsListCmd(commonFlags *secretsFlags) *cobra.Command {
 	return cmd
 }
 
+func secretsDownloadCmd(commonFlags *secretsFlags) *cobra.Command {
+	flags := secretsDownloadFlags{}
+	command := &cobra.Command{
+		Use:   "download <file1>",
+		Short: "Download environment variables into the specified file",
+		Long:  "Download environment variables stored into the specified file (most commonly a .env file). The format of the file is one NAME=VALUE per line.",
```
savil, line 174: could we add something about `unless --format is specified`?

internal/boxcli/secrets.go
```diff
@@ -188,6 +166,63 @@ func secretsListCmd(commonFlags *secretsFlags) *cobra.Command {
...
+
+	return command
+}
+
+func secretsUploadCmd(commonFlags *secretsFlags) *cobra.Command {
+	flags := &secretsUploadFlags{}
+	command := &cobra.Command{
+		Use:   "upload <file1> [<fileN>]...",
+		Short: "Upload variables defined in a .env file",
+		Long: "Upload variables defined in one or more .env files. The files " +
```
savil, line 202: again, need to modify it to mention `--format` or `json`

internal/devbox/secrets.go
```diff
@@ -9,41 +9,36 @@ import (
...
 		Auth: envsec.AuthConfig{
 			ClientID: envvar.Get("ENVSEC_CLIENT_ID", build.ClientID()),
 			Issuer:   envvar.Get("ENVSEC_ISSUER", build.Issuer()),
 		},
 		IsDev:      build.IsDev,
 		Stderr:     d.stderr,
+		Store:      &jetstore.JetpackAPIStore{},
 		WorkingDir: d.ProjectDir(),
+		EnvID: envsec.EnvID{
+			EnvName: d.environment,
```
savil, line 24: is it possible for `d.environment` to not be set? If so, could we add a validation check in this function?

internal/devbox/secrets.go
```diff
@@ -9,41 +9,36 @@ import (
...
+	if err != nil {
 		return nil, err
 	}
-	return &secrets{
-		Envsec:  envsecInstance,
-		EnvName: d.environment,
-	}, nil
-}
+	envsecInstance.EnvID.ProjectID = project.ProjectID.String()
+	envsecInstance.EnvID.OrgID = project.OrgID.String()
```
savil, line 37: its okay for now, but I feel the API could be made a bit safer by initializing `EnvID` as a complete struct in one place...

having it split between UninitializedSecrets and here feels like it could lead to bugs over time (once someone less experienced with this code refactors it, etc.)

savil (approved):

internal/devbox/secrets.go
mikeland73: it's already validated in `devbox.Open`

internal/devbox/secrets.go
mikeland73: yeah, this weirdness is to fix a bug that `secrets init` would not work because it depended on the project already existing. Agree it's a bit awkward API. 

2 possible fixes:

* We can always set the full EnvID object. Re-writing is fine.
* The `UninitializedSecrets` doesn't really need EnvID set. It just felt weird not to set it.

PR #1723, opened 2024-01-12
[Refactor ensureStateIsUpToDate] split into two modes for when to recompute state

savil:
## Summary

Refactor the implementation of `ensureStateIsUpToDate` as per the suggestion in this comment: https://github.com/jetpack-io/devbox/pull/1692#discussion_r1449453688

1. Split `lockfile.Save` into `lockfile.Save` and `lockfile.UpdateAndSaveStateHash`
2. For now, regardless of `recomputeState` (boolean variable), we always call `recomputeState` (the function). In the next PR, this will change.

NOTE:
In `pluginManager.create`, I only do  `lockfile.Save` and not `lockfile.UpdateAndSaveLocal` (as before), because:
When I do `devbox add php81Extensions.ds` then the `lockfile.Save` in `pluginManager.Create` here is resulting in the `local.lock` being up-to-date so that the next `devbox run` is thinking everything is up to date and early returning in `ensureStateIsUpToDate`.

## How was it tested?

```
devbox shell
devbox add hello
devbox rm hello
```

tests should pass.

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -255,13 +255,14 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 	defer trace.StartRegion(ctx, "devboxEnsureStateIsUpToDate").End()
 	defer debug.FunctionTimer().End()
 
-	// if mode is install or uninstall, then we need to update the nix-profile
-	// and lockfile, so we must continue below.
 	upToDate, err := d.lockfile.IsUpToDateAndInstalled()
 	if err != nil {
 		return err
 	}
 
+	// if mode is install or uninstall, then we need to compute some state
+	// like updating the flake or installing packages locally, so must continue
+	// below
 	if mode == ensure {
 		// if mode is ensure and we are up to date, then we can skip the rest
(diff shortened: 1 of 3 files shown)


savil:
* `main`
  * **PR #1723** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1723?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1723?utm_source=stack-comment).

gcurtis (approved):

mikeland73:
I'm not sure removing lock.Save() from plugin manager is great. It modifies the lock file so I would expect it to save it?

> When I do devbox add php81Extensions.ds then the lockfile.Save in pluginManager.Create here is resulting in the local.lock being up-to-date so that the next devbox run is thinking everything is up to date and early returning in ensureStateIsUpToDate.

Why is the case? 

If it only saves the lockfile but does not update the local cache state, would the local state not be out of date?

PR #1724, opened 2024-01-12
Derive nix profile from flake: attempt 2

savil:
## Summary

This PR is a re-write of #1692. See motivation and discussion there.

**Motivation**
This PR seeks to unify our nix print-dev-env and nix profile steps. Other benefits:
- It will enable us to remove some hacks (example, this one [introduced for glibc-patch](https://github.com/jetpack-io/devbox/blob/d66f6ea5388adb3b90008c64a1b9b9abdc4550cd/internal/devbox/devbox.go#L938-L944))
- Re-enable some testscripts that were disabled (example, this [one for php extension removal](https://github.com/jetpack-io/devbox/blob/main/testscripts/languages/php.test.txt#L4-L15))

Previously, these would be two separate processes: we'd manually do nix profile install.
Now, we get the `buildInputPaths` from `nix print-dev-env` and install those, so the nix profile is derived from the flake.

**Implementation notes**
1. In `ensureStateIsUpToDate`, I had to modify the logic in #1724: call `d.installPackages` that creates plugin files, installs runx and nix packages. Also, move all lockfile operations to the end for _all_ modes, instead of just for `recomputeState == true`.
2. The `nix build` operation now happens with the `[1/3] go@1.21: Success` output that previously would happen with `nix profile install`.
3. Instead of `nixprofiles.NixProfileInstall`, we first install via `nix build` to ensure the package is installed in the local nix store. We then call `devbox.syncNixProfile` to get the `buildInputs` from `devbox.computeEnv` and `nix profile install $buildInputs`.

## How was it tested?

- CICD tests pass
- Did basic sanity checking with `devbox add hello cowsay` and `devbox rm hello cowsay`. 
- Used the Devbox binary for development for a while.

To verify that we can install packages in `/nix/store` during `devbox add`, and then use that during offline in `devbox shell`:
1. Do `devbox add python@3.11.6`
2. Do `devbox shell` and exit, or just `devbox install`.
3. Turn Wifi off, and then open `devbox shell` and verify that the newly installed version of `python` is used.
BEFORE: we could skip step 2, because we directly installed packages in the nix profile
AFTER: step 2 is required because we re-evaluate the flake's environment and derive nix profile from it.

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -450,18 +450,33 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context) error {
 			fmt.Fprintf(d.stderr, "%s: ", stepMsg)
 			color.New(color.FgRed).Fprintf(d.stderr, "Fail\n")
 
-			platform := nix.System()
-			return usererr.WithUserMessage(
-				err,
-				"package %s cannot be installed on your platform %s.\n"+
-					"If you know this package is incompatible with %[2]s, then "+
-					"you could run `devbox add %[1]s --exclude-platform %[2]s` and re-try.\n"+
-					"If you think this package should be compatible with %[2]s, then "+
-					"it's possible this particular version is not available yet from the nix registry. "+
-					"You could try `devbox add` with a different version for this package.\n",
-				pkg.Raw,
-				platform,
-			)
+			// Check if the user is installing a package that cannot be installed on their platform.
+			// For example, glibcLocales on MacOS will give the following error:
+			// flake output attribute 'legacyPackages.x86_64-darwin.glibcLocales' is not a derivation or path
+			// This is because glibcLocales is only available on Linux.
+			// The user should try `devbox add` again with `--exclude-platform`
+			errMessage := strings.TrimSpace(err.Error())
+			fmt.Printf("errMessage is %s\n", errMessage)
+			maybePackageSystemCompatibilityError := strings.Contains(errMessage, "error: flake output attribute") &&
+				strings.Contains(errMessage, "is not a derivation or path")
+
+			if maybePackageSystemCompatibilityError {
+				platform := nix.System()
+				return usererr.WithUserMessage(
+					err,
+					"package %s cannot be installed on your platform %s.\n"+
+						"If you know this package is incompatible with %[2]s, then "+
+						"you could run `devbox add %[1]s --exclude-platform %[2]s` and re-try.\n"+
+						"If you think this package should be compatible with %[2]s, then "+
+						"it's possible this particular version is not available yet from the nix registry. "+
+						"You could try `devbox add` with a different version for this package.\n\n"+
+						"Underlying Error from nix is:",
+					pkg.Raw,
+					platform,
+				)
+			}
+
+			return usererr.WithUserMessage(err, "error installing package %s", pkg.Raw)
 		}
 
 		fmt.Fprintf(d.stderr, "%s: ", stepMsg)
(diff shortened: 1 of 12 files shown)


savil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1724?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **PR #1723** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1724** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1724?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1724?utm_source=stack-comment).

savil:
@mikeland73 @gcurtis I think this is 95% there. The `insecure` example was failing which I think I've now addressed 🤞🏾 

PTAL and let me know what you think.

internal/devbox/packages.go
```diff
@@ -571,3 +419,93 @@ func (d *Devbox) InstallRunXPackages(ctx context.Context) error {
...
+		fmt.Fprintf(d.stderr, stepMsg+"\n")
+
+		// --no-link to avoid generating the result objects
+		err = nix.Build(ctx, []string{"--no-link"}, installable)
+		if err != nil {
+			fmt.Fprintf(d.stderr, "%s: ", stepMsg)
+			color.New(color.FgRed).Fprintf(d.stderr, "Fail\n")
+
+			platform := nix.System()
+			return usererr.WithUserMessage(
```
savil, line 454: I can tighten this condition a bit, but its hard to make it as good as before.

Previously, we could check `strings.Contains(string(out), "is not available on the requested hostPlatform") {`

but now the error looks like:
```
source: nix build exit code: 1, 
output: error: 
flake output attribute 'legacyPackages.x86_64-darwin.glibcLocales' is not a derivation or path
, err: exit status 1
```

savil: With update, it looks like this now:

Screenshot 2024-01-16 at 10 12 33 AM

gcurtis (approved):

internal/nix/nixprofile/item.go
```diff
@@ -83,3 +83,7 @@ func (i *NixProfileListItem) String() string {
 		i.nixStorePaths,
 	)
 }
+
+func (i *NixProfileListItem) StorePaths() []string {
+	return i.nixStorePaths
+}
```
mikeland73, line 89: Export instead?

internal/devbox/packages.go
```diff
@@ -271,15 +272,15 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 		fmt.Fprintln(d.stderr, "Ensuring packages are installed.")
 	}
 
-	recomputeState := mode == ensure || d.IsEnvEnabled()
-	if recomputeState {
-		if err := d.recomputeState(ctx, mode); err != nil {
+	if mode == install || mode == update || mode == ensure {
```
mikeland73, line 275: `!= remove`?

mikeland73 (approved):

PR #1735, opened 2024-01-19
[gen-readme] Improve gen readme

mikeland73:
## Summary

Improve automatically generated readme. Highlights:

* Adds "getting started" section.
* documents how to install devbox, start shell, run script
* Adds more context around scripts, init hook.
* Adds link to devbox repo.
* Adds more spacing between script details

What it looks like: https://github.com/jetpack-io/devbox/blob/56e7427a8f3da266d2f3d2b4a5028a7c7a017fc3/devbox.md

## How was it tested?

`devbox gen readme devbox.md`

### DIFF
--- a/internal/devbox/docgen/readme.tmpl
+++ b/internal/devbox/docgen/readme.tmpl
@@ -7,11 +7,29 @@
 {{- if .Description }}
 {{ .Description }}
 {{ end }}
+## Getting Started
+This project uses [devbox](https://github.com/jetpack-io/devbox) to manage its development environment.
+
+Install devbox:
+```sh
+curl -fsSL https://get.jetpack.io/devbox | bash
+```
+
+Start the devbox shell:
+```sh 
+devbox shell
+```
+
+Run a script in the devbox environment:
+```sh
+devbox run <script>
+```
 
 {{- if .Scripts }}
 ## Scripts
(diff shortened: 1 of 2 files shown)


internal/devbox/docgen/readme.tmpl
```diff
@@ -7,11 +7,29 @@
...
+
+Install devbox:
+```sh
+curl -fsSL https://get.jetpack.io/devbox | bash
+```
+
+Start the devbox shell:
+```sh 
+devbox shell
+```
```
savil, line 21: WDYT about nudging people towards using direnv as well? It tends to be significantly faster than starting a `devbox shell` (since we don't start a new shell process)

savil (approved):

internal/devbox/docgen/readme.tmpl
mikeland73: Feels weird to me since I personally don't use it (I tend to use `devbox run` a lot which is faster than shell).

Feels a bit prescriptive no? i.e. devbox is absolutely needed to develop on this repo, but direnv is more a matter of opinion? 

Anyway, a follow up to this could be to allow people to specify their own readme templates so they can modify as needed.

savil: its just sooo much faster

PR #1741, opened 2024-01-22
[fish] Fix init hooks for fish

mikeland73:
## Summary

Fixes issue introduced in https://github.com/jetpack-io/devbox/pull/1709

`devbox run` uses `sh` instead of native shell. So init hooks must always be `sh` compatible for `run`. On the other hand, `devbox shell` uses native shell so init hooks must be `fish` if the native shell is fish.

This is ugly, but it's how it has always worked. 

https://github.com/jetpack-io/devbox/pull/1709 introduced recursion protection that would be `fish` or `sh` depending on native shell. This worked fine for `devbox shell` but would break `devbox run`. 

This change fixes that by creating 2 hooks files, one fish and one sh and sourcing the correct one in shellrc, while still always using the `sh` for run.

cc: @Lagoja 

## How was it tested?

```
devbox run echo 5
devbox shell
SHELL=fish devbox run echo 5
SHELL=fish devbox shell
```

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -204,7 +204,6 @@ func (d *Devbox) Shell(ctx context.Context) error {
 	}
 
 	opts := []ShellOption{
-		WithHooksFilePath(shellgen.ScriptPath(d.ProjectDir(), shellgen.HooksFilename)),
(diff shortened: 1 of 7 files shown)


Lagoja (commented):
Confirmed that this fixes the issue upon testing

Lagoja (approved):

PR #1749, opened 2024-01-25
[devbox add] --allow-insecure should handle multiple, user-specified packages

### LINKED ISSUE #1337: [Bug]: flake package is insecure error when using --allow-insecure
**Current Behavior (bug)**
if I have a flake.nix that uses an insecure package, it shows an error. 

flake.nix
```nix
{
  description =
    "This flake outputs a modified version of Yarn that uses NodeJS 20";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    flake-utils.url = "github:numtide/flake-utils";
  };

  outputs = { self, nixpkgs, flake-utils }:
  # Use the flake-utils lib to easily create a multi-system flake
  flake-utils.lib.eachDefaultSystem (system:
    let
      # You can define overlays as functions using the example below
      # This overlay will modify yarn to use nodejs-16_x
      overlay = (final: prev: rec {
        nodejs = prev.nodejs_14;
        yarn = prev.yarn.override { nodejs = nodejs; };
      });

      #
      pkgs =
        import nixpkgs {
          inherit system;
          # Add your overlays to the list below. Note that they will be applied in order
          overlays = [ overlay ];
        };

    in rec {
      # For our outputs, we'll return the modified Yarn package from our overridden nixpkgs.
      packages = {
        yarn = pkgs.yarn;
        nodejs = pkgs.nodejs;
      };

    }
  );
}

```

add the flake with devbox:
```
devbox add "path:overlay#nodejs" --allow-insecure
# or 
devbox add "path:overlay#yarn" --allow-insecure
```

it shows an error: 
```
Error: Package Package ‘nodejs-14.21.3’ is insecure. 

To override use `devbox add <pkg> --allow-insecure`
```


**Expected Behavior (fix)**
The flake with an insecure package is installed when using --allow-insecure

**Additional context**

```
devbox version                                                                                                                                                                                                                                                                                                                                                                       (latestnhost x!?)
0.5.7
```

```
{
  "packages": [
  ],
  "shell": {
    "init_hook": "yarn"
  },
  "nixpkgs": {
  }
}

```

savil:
## Summary

Changes
1. Modify `devbox add`'s `--allow-insecure` flag to take a `[]string` instead of `bool`.
2. Save `allow_insecure` field in the package in the `devbox.json` config
3. Set `NIXPKGS_ALLOW_INSECURE=1` in the `nix build` during `devbox add`
4. Set the packages specified in `allow_insecure` in the generated flake's `permittedInsecurePackages`.

TODO:
- [x] Handle existing projects with `allow_insecure: true` in their `devbox.lock` (#1754)

**Downside:**
If a user erroneously specifies the wrong package name or names in `--allow-insecure=<packages>`, then a `devbox add  <pkg> --allow-insecure=<packages>` outside the `devbox shell` environment will quietly work. However, the user will experience an error when running `devbox shell` (or similar) next time, and need to fix the erroneous `--allow-insecure` value.
 - Fix: we could fix this by doing "recomputeState" when installing-with-insecure in "ensureStateIsUpToDate", but I haven't to avoid adding complexity for this unlikely scenario. I call it unlikely because the error message (see below) specifically tells the user what values to use with `--allow-insecure=<packages>`

Fixes #1337 

## How was it tested?

`devbox add python@2.7` gives the following error now:
```
 devbox add python@2.7
Info: Adding package "python@2.7" to devbox.json
[1/1] python2
[1/1] python2: Fail

Error: Nix: Package  ‘python-2.7.18.7’ is insecure.

Known vulnerabilities:
Python 2.7 has reached its end of life after 2020-01-01. See https://www.python.org/doc/sunset-python-2/.

To override, use `devbox add <pkg> --allow-insecure=python-2.7.18.7`
```

Doing `devbox add python@2.7 --allow-insecure=python-2.7.18.1` works with the devbox.json now having:
```
+    "python": {
+      "version":        "2.7",
+      "allow_insecure": ["python-2.7.18.7"],
+    },
```

Also for the package reported in #1337:

Screenshot 2024-01-25 at 5 43 43 PM

### DIFF
--- a/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
+++ b/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
@@ -25,8 +25,10 @@
           config.allowUnfree = true;
           config.permittedInsecurePackages = [
             {{- range $flake.Packages }}
-            {{- if .AllowInsecure }}
-            "{{ .StoreName }}"
+            {{- if .HasAllowInsecure }}
+            {{- range .AllowInsecure }}
+            "{{ . }}"
+            {{- end }}
             {{- end }}
             {{- end }}
           ];
--- a/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
+++ b/internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
@@ -25,10 +25,8 @@
           config.allowUnfree = true;
           config.permittedInsecurePackages = [
             {{- range $flake.Packages }}
-            {{- if .HasAllowInsecure }}
             {{- range .AllowInsecure }}
             "{{ . }}"
-            {{- end }}
             {{- end }}
             {{- end }}
           ];
--- a/internal/boxcli/add.go
+++ b/internal/boxcli/add.go
@@ -19,7 +19,7 @@ const toSearchForPackages = "To search for packages, use the `devbox search` com
 
 type addCmdFlags struct {
 	config           configFlags
-	allowInsecure    bool
+	allowInsecure    []string
 	disablePlugin    bool
 	platforms        []string
 	excludePlatforms []string
@@ -53,8 +53,8 @@ func addCmd() *cobra.Command {
 	}
 
 	flags.config.register(command)
-	command.Flags().BoolVar(
-		&flags.allowInsecure, "allow-insecure", false,
+	command.Flags().StringSliceVar(
+		&flags.allowInsecure, "allow-insecure", []string{},
 		"allow adding packages marked as insecure.")
 	command.Flags().BoolVar(
 		&flags.disablePlugin, "disable-plugin", false,
--- a/internal/devbox/devopt/devboxopts.go
+++ b/internal/devbox/devopt/devboxopts.go
@@ -43,7 +43,7 @@ type Credentials struct {
 }
 
 type AddOpts struct {
-	AllowInsecure    bool
+	AllowInsecure    []string
 	Platforms        []string
 	ExcludePlatforms []string
 	DisablePlugin    bool
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -137,24 +137,9 @@ func (d *Devbox) setPackageOptions(pkgs []string, opts devopt.AddOpts) error {
 			d.stderr, pkg, opts.Outputs); err != nil {
 			return err
 		}
-	}
-
-	// Resolving here ensures we allow insecure before running ensureStateIsUpToDate
-	// which will call print-dev-env. Resolving does not save the lockfile, we
-	// save at the end when everything has succeeded.
-	if opts.AllowInsecure {
-		nix.AllowInsecurePackages()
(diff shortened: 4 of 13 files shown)


savil:
* `main`
  * **PR #1749** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1749?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1749?utm_source=stack-comment).

mikeland73:
Currently reviewing, but a quick comment about the json change. cc: @Lagoja 

Instead of the current json change, thoughts on:

```json
"python": {
  "version":        "2.7",
  "allow_insecure" true,
},
```

```json
"python": {
  "version":        "2.7",
  "dependencies": {
    "openssl": {
      "allow_insecure": true,
    }
  }
},
```

Motivation:

* Same format for top level package and dependencies
* allows for future input overrides (and any other configuration we want to apply to dependent packages). 


The only downside is that this format doesn't make a distinction between direct dependencies and sub-depedencies, but I think that's way too complicated.

savil:
@mikeland73 🤔 I'm a little concerned that it'll be introducing quite some complexity in this PR's implementation: especially around needing to distinguish the top-level package from dependency packages, and their properties. Can we cross that bridge when we need to later?

internal/shellgen/tmpl/flake_remove_nixpkgs.nix.tmpl
```diff
@@ -25,8 +25,10 @@
           config.allowUnfree = true;
           config.permittedInsecurePackages = [
             {{- range $flake.Packages }}
-            {{- if .AllowInsecure }}
-            "{{ .StoreName }}"
+            {{- if .HasAllowInsecure }}
+            {{- range .AllowInsecure }}
+            "{{ . }}"
+            {{- end }}
```
gcurtis, line 31: I think you can remove the if check here.

```suggestion
        {{- range .AllowInsecure }}
        "{{ . }}"
        {{- end }}
```

gcurtis (approved):
I'm liking the comprehensive tests.

savil:
@mikeland73 landed the code to close the week out, but happy to discuss more and make any followup fixes

mikeland73:
@savil I would run it by @Lagoja .

JSON changes are semi-1 way doors. This is OK but just wanted to make sure you are aware of:

* As currently implemented, this changes the previous behavior. Specifically, if the package is updated and the version changes, allow insecure version will no longer march. My suggested implementation keeps the same behavior. This is probably a bit safer, but wanted to call out that we're changing how it works.
* If we do add input overrides, we may want to add allow insecure to the individual packages themselves (in order to standardize fields between top level packages and dependencies). This may leave this new field as a bit of an awkward outlier (because it's a field that applies to both top level and also dependencies). Not enough to be a blocker, but it does feel  a bit weird if we end up in that state.

PR #1763, opened 2024-01-30
Update PostgreSQL example for cross-platform

Lagoja:
## Summary

Current PostgreSQL example doesn't work on MacOS because it attempts to install glibcLocales. This updates the example and documentation to only install glibcLocales on Linux

## How was it tested?

CI/CD, run example locally

### DIFF
--- a/docs/app/docs/devbox_examples/databases/postgres.md
+++ b/docs/app/docs/devbox_examples/databases/postgres.md
@@ -9,7 +9,7 @@ PostgreSQL can be automatically configured by Devbox via the built-in Postgres P
 
 ## Adding Postgres to your Shell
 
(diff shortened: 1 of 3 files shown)


docs/app/docs/devbox_examples/databases/postgres.md
```diff
@@ -9,13 +9,18 @@ PostgreSQL can be automatically configured by Devbox via the built-in Postgres P
 
 ## Adding Postgres to your Shell
 
-`devbox add postgresql glibcLocales`, or in your `devbox.json`, add
+You can install the PostgreSQL server and client by running`devbox add postgresql`. In some Linux distributions, you may also need to add `glibcLocales`, which can be added using `devbox add glibcLocales -p x86_64-linux,aarch64-linux`.
```
savil, line 12: nit: slight preference to use `--platform` instead of shorthand `-p` since this is documentation. Makes it more self-explanatory.

savil (approved):

PR #1765, opened 2024-01-30
[local-state] Local state and fish improvements

mikeland73:
## Summary

A few fish and local state improvements:

* [fish] Don't add recursion protection if init hook is empty or default. Recursion protection is shell-specific so it may cause issues in some edge cases (e.g. if someone uses a `shellenv --init-hooks` output on a different shell). 
* [fish] Incorporate fish into local state hash. This means that if a user changes shell, we consider the local state stale. This is needed because we generate different files (e.g. shellrc) depending on the shell.
* [local-state] Rename a bunch of stuff in local.go to be more accureately named ( `stateHash` instead of `localLock`)
* [local-state] Decoupled `stateHash` from lock file. These are different and should not depend on each other. They are still in the same package, but we can change that in the future.
* Removed the devbox interface from `statehash`. As @gcurtis warned us many moons ago, interfaces like this make code harder to follow.

## How was it tested?

Tested empty/default init hooks not having recursion protection:

```bash
devbox init
devbox run echo foo
cat .devbox/gen/scripts/.hooks.sh
```

Tested local state with fish

```bash
devbox run echo 5
cat .devbox/local.lock
SHELL=fish devbox  run echo 5
cat .devbox/local.lock
```


internal/lock/statehash.go
```diff
@@ -0,0 +1,116 @@
...
+func readStateHashFile(projectDir string) (*stateHashFile, error) {
+	lockFile := &stateHashFile{}
+	err := cuecfg.ParseFile(stateHashFilePath(projectDir), lockFile)
+	if errors.Is(err, fs.ErrNotExist) {
+		return lockFile, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	return lockFile, nil
```
savil, line 71: its confusing to call these lockfiles in the local variables. Why not `stateHashFile` and similar?

internal/lock/statehash.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	lockfileHash, err := getLockfileHash(args.ProjectDir)
+	if err != nil {
+		return nil, err
+	}
+
+	newLock := &stateHashFile{
```
savil, line 90: `newLock` -> `newHashFile`

internal/lock/statehash.go
```diff
@@ -0,0 +1,116 @@
...
+		LockFileHash:           lockfileHash,
+		NixPrintDevEnvHash:     printDevEnvCacheHash,
+		NixProfileManifestHash: nixHash,
+	}
+
+	return newLock, nil
+}
+
+func stateHashFilePath(projectDir string) string {
+	return filepath.Join(projectDir, ".devbox", "local.lock")
```
savil, line 103: I find the `.lock` pretty confusing usage here. Lockfiles in my mind represent a need to ensure that some particular state can be reproduced as done with `package.lock` or `devbox.lock`.

This feels to be more of a local-state-cache-representation or `.devbox/state_cache_key.json`, or simply `.devbox/cache_key.json`

WDYT?

internal/shellgen/scripts.go
```diff
@@ -110,6 +110,11 @@ func writeInitHookFile(devbox devboxer, body, tmpl, filename string) (err error)
 	}
 	defer script.Close() // best effort: close file
 
+	if body == devconfig.DefaultInitHook || strings.TrimSpace(body) == "" {
```
savil, line 113: Mind adding a comment to the effect of:
```
// skip adding init-hook recursion guard for the 
// default hook or any empty hook
```

As is, not obvious why the template below is being skipped.

Will also be a good idea to comment somewhere in the template that it will be skipped for default or empty hooks, in case we later introduce some other logic for init-hooks in the template files.

savil (approved):
thanks!

internal/lock/statehash.go
mikeland73: will change

internal/lock/statehash.go
mikeland73: for sure. I was not bold enough to change the name, but there's no downside really.

internal/lock/statehash.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2023 Jetpack Technologies Inc and contributors. All rights reserved.
```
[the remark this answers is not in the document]
mikeland73: github shows the file as deleted, but I split the PR into 2 commits. If you take a look https://github.com/jetpack-io/devbox/pull/1765/commits/a7160e394d78d3f5c53e54bbc233097c99eb8541 it contains the changes to the file.

Issue #1744, opened 2024-01-23
[Bug]: syntax error in hooks.sh when using fish


jasononeil:
**Current Behavior (bug)**
Since updating to 0.8.6 this morning, I'm now getting:

```
/Users/jason/workspace/local-ops/.devbox/gen/scripts/.hooks.sh:15: syntax error: unexpected end of file
```

I'm using Fish as my shell.

This is consistent across all my projects, and has been replicated by my co-worker @jay-aye-see-kay's mac also, though he had to delete his `.devbox` directory in the project to start seeing the error.

This prevents any commands from starting, for example:

```
> devbox run setup
/Users/jason/workspace/cultureamp/std-demographics-next/.devbox/gen/scripts/.hooks.sh: line 15: syntax error: unexpected end of file
Error: error running script "setup" in Devbox: exit status 2
```

**Expected Behavior (fix)**

There shouldn't be a syntax error when running devbox commands on a fish shell.

Notably, @jay-aye-see-kay pointed out the file is a `.sh` file with no shebang, but contains fish syntax.

If I run `./devbox/gen/scripts/.hooks.sh` I get the error, but if I source it with `. ./devbox/gen/scripts/.hooks.sh` it works as expected.

**Additional context**

Please include the output of `devbox version -v`

```
Version:     0.8.6
Platform:    darwin_amd64
Commit:      af58be9b50c1a3d19bc21485bcbb77eb8ed8076a
Commit Time: 2024-01-18T18:36:03Z
Go Version:  go1.21.5
Launcher:    0.2.1
```

A copy of your `devbox.json` file:

```
{
  "packages": ["flake:nixpkgs#nodejs_18", "flake:nixpkgs#nodejs_18.pkgs.pnpm"],
  "env": {
    "STD_DEMOGRAPHICS_PORT": "4267"
  },
  "shell": {
    "scripts": {
      "setup": "pnpm install"
    }
  }
}
```

Fish version: 3.6.1

jay-aye-see-kay:
Just saw version `0.8.7` was release with a fix. That's working for me thank you! 🙇

jasononeil:
Yep this was fixed in https://github.com/jetpack-io/devbox/pull/1741

Thanks!

phanirithvij:
I still have it wth 0.8.7, I actually only got this issue after updating to 0.8.7 yesterday.

```shell
```

ssoriche:
[...]

PR #1769, opened 2024-02-01
[urgent] Disable recursion guard

mikeland73:
## Summary

This disables recursion guard. That means that init hooks are always written as is (fish and non fish will look the same)

Fixes https://github.com/jetpack-io/devbox/issues/1744

## How was it tested?

Inspected hook scripts.


savil (approved):

mikeland73:
closing this in favor of https://github.com/jetpack-io/devbox/pull/1771

PR #1775, opened 2024-02-02
[telemetry] Fix release in telemetry

mikeland73:
## Summary

I think this is causing issues to be miss categorized.

If you go to Sentry none of our issues are actually getting associated to the correct release that has all the commits.

I would not find any documentation that says that releases needs to be `app@version`  

## How was it tested?

Not sure how to test this.


gcurtis (approved):

gcurtis:
I think the recommendation came from here - https://docs.sentry.io/product/cli/releases/#creating-releases. It might've changed though.

loreto:
We were following: https://docs.sentry.io/platforms/go/configuration/releases/#setting-a-release which still shows `app@version` ...

mikeland73:
@loreto @gcurtis I see. I think you guys are right.

I think in that case our issue might be when creating the release in github action instead.

On sentry the release that has the commits says it's not semver
image


While the release created by our issues is semver:

image

mikeland73:
Should we be doing 

`version_prefix: "devbox@"` in our `getsentry/action-release@v1` ?  I'm not sure.

loreto:
Maybe we need to make the github release match the semver release convention. I think it means:
1. Changing the GHA to create `devbox@{version}`
2. No longer using commit hashes, `version` needs to come from a tag, and we would only create the release in sentry when we tag a new release.

OR

We switch the code to also use commits instead of the semver notation to match GHA.

gcurtis:
Yeah, maybe try setting `version_prefix: "devbox@"` to make it match. I'm not sure why Sentry's docs recommend the @ naming convention but then the action doesn't follow that.

mikeland73:
I'm gonna change this to be an action change instead.

PR #1778, opened 2024-02-03
Revert "nix: improve redacted nix error (#1776)"

savil:


savil:
* `main`
  * **PR #1778** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1778?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1778?utm_source=stack-comment).

gcurtis (approved):

internal/nix/nix.go
```diff
@@ -77,22 +76,15 @@ func (*Nix) PrintDevEnv(ctx context.Context, args *PrintDevEnvArgs) (*PrintDevEn
...
 			return nil, insecureErr
 		} else if err != nil {
-			safeArgs := make([]string, 0, len(cmd.Args))
-			for _, a := range cmd.Args {
-				if a == args.FlakeDir {
-					a = "<redacted path>"
-				}
-				safeArgs = append(safeArgs, a)
-			}
-			return nil, redact.Errorf("nix command: %s", redact.Safe(safeArgs))
```
gcurtis, line 87: I fumbled this with the original PR. I was going to try and sanitize the individual args, but then decided not to and ended up with a weird in between implementation. The line was supposed to just be:

```go
redact.Errorf("nix command: %s", redact.Safe(cmd))
```

and not have the `safeArgs` stuff. 

Redacted error messages are not normally redacted and retain their normal message unless they're sent to Sentry. But here I modified the actual slice of args, so it ended up always redacted.

If this isn't blocking anything, I can just submit a fix PR later.

mikeland73 (approved):

internal/nix/nix.go
savil: I think I see. I need to learn about our redact logic more.

Assuming it's okay, I'm going to land this, just so that the Devbox release is unblocked. Since it's a small PR, perhaps you could re-submit with the better fix and if it works out timing-wise then it could be included in the release as well!

gcurtis:
@savil fixed in #1782.

PR #1782, opened 2024-02-05
internal/nix: fix unredacted error message

gcurtis:
Don't pass a slice of already-redacted arguments to `redact.Errorf` so that they're preserved in the regular error message.

Tested by forcing an error and seeing the full message:

```
$ go run ./cmd/devbox shell
Error: nix print-dev-env --json "path:/Users/gcurtis/src/devbox/.devbox/gen/flake": exit status 1

```

Gets redacted in Sentry as:

```
nix print-dev-env --json "path:<redacted string>": <redacted *exec.ExitError>
```


savil (approved):
Tested and verified that the error motivated #1778 is properly handled as well. Thanks!

savil:
Rebased to incorporate the CICD fix of #1786

PR #1789, opened 2024-02-05
internal/devpkg: move flake code into new nix/flake package

gcurtis:
Move `devpkg.FlakeRef` and `devpkg.FlakeInstallable` into a standalone `go.jetpack.io/devbox/nix/flake` package. This allows two things:

- The backend can import and use these types because they're no longer internal.
- The search client can use them without creating an import cycle.

Note that the types were renamed, which messed up git's ability to detect this is a file move. The summary of changes is:

- `FlakeRef` → `flake.Ref`
- `ParseFlakeRef` → `flake.ParseRef`
- `FlakeInstallable` → `flake.Installable`
- `ParseFlakeInstallable` → `flake.ParseInstallable`
- `FlakeTypeIndirect`, `FlakeTypePath`, ... → `flake.TypeIndirect`, `flake.TypePath`, ...


savil (approved):

internal/devpkg/package.go
cklim2024, line 1: internal/devpkg/package.go

PR #1790, opened 2024-02-05
internal/searcher,lock: use /v2/resolve endpoint

gcurtis:
Add a `DEVBOX_FEATURE_RESOLVE_V2` feature flag that enables the new `/v2/resolve` endpoint when resolving packages. The new endpoint unlocks a couple of improvements:

- We no longer need to call `nix store path-from-hash-part` (which incurs a network request) for every `fetchClosure` package. The full store path is returned directly by the API.
- The resolved package response includes store paths for all known outputs, allowing us to support `fetchClosure` for non-default outputs as well.
- The new endpoint returns a flake installable rather than a nixpkgs commit hash and attribute path. This gives more flexibility, allowing the API to potentially return packages from sources other than nixpkgs.

### DIFF
--- a/internal/searcher/model.go
+++ b/internal/searcher/model.go
@@ -3,6 +3,12 @@
 
 package searcher
 
+import (
+	"time"
+
+	"go.jetpack.io/devbox/nix/flake"
+)
+
 type SearchResults struct {
 	NumResults int       `json:"num_results"`
 	Packages   []Package `json:"packages,omitempty"`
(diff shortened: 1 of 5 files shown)


internal/searcher/model.go
```diff
@@ -35,3 +41,56 @@ type PackageInfo struct {
 	Version      string   `json:"version"`
 	Summary      string   `json:"summary"`
 }
+
+// ResolveResponse is a response from the /v2/resolve endpoint.
+type ResolveResponse struct {
```
mikeland73, line 46: Would be nice to share these with search service if possible.

mikeland73 (approved):
Would be nice to add an additional test run to matrix with this flag enabled that way we can start seeing tests (and perf implication if there is any) right away.

gcurtis:
> Would be nice to add an additional test run to matrix with this flag enabled that way we can start seeing tests (and perf implication if there is any) right away.

Agreed. I can do that once the backend change is deployed.

PR #1794, opened 2024-02-07
[gen-readme] Allow custom templates

mikeland73:
## Summary

This adds:

Flag that saves default template
```
devbox generate readme [path] --save-template
```

Flag that uses custom template
```
devbox generate readme [path] --template <custom-template>
```

It also:

* defaults readme [path] to `README.md`
* defaults template [path] to `readme.tmpl`
* If there is `readme.tmpl` in root, then `devbox generate readme` will use it automatically.
* Fixed missing env vars from default template

## How was it tested?

```
devbox generate readme --save-template # generate template
vi readme.tmpl # modified template
devbox generate readme devbox.md # tested creating new readme
```

### DIFF
--- a/devbox.md
+++ b/devbox.md
@@ -36,6 +36,13 @@ Scripts are custom commands that can be run using this project's environment. Th
 * [tidy](#devbox-run-tidy)
 * [update-examples](#devbox-run-update-examples)
 
+## Environment
+
+```sh
+GOENV="off"
+PATH="$PATH:$PWD/dist"
+```
+
 ## Shell Init Hook
 The Shell Init Hook is a script that runs whenever the devbox environment is instantiated. It runs 
 on `devbox shell` and on `devbox run`.
--- a/internal/boxcli/generate.go
+++ b/internal/boxcli/generate.go
@@ -22,6 +22,12 @@ type generateCmdFlags struct {
 	rootUser          bool
 }
 
+type GenerateReadmeCmdFlags struct {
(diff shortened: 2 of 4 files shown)


mikeland73:
@savil gentle ping

savil:
@mikeland73 sorry for the delay in reviewing this. I'm a little hesitant with this design. I don't love having _golang_ templates be the expected UX for our users.

I'm curious if there's a markdown template where one can more natively separate the auto-generated parts from the user-written parts of the markdown file...

mikeland73:
@savil we already use golang templates for plugins, and generally go applications expose go templating (e.g. Helm being a fairly prominent example) so I don't think it's unexpected or bad UX. 

I think if we try to use markdown for this we will end up inventing our own templating. I would much rather re-use golang templates.

I also considered  allowing injecting the readme into a specific portion of a manually created file, but that is very limiting (so limiting, that we would probably not use it ourselves). But as built, this would be useful for us right away in Axiom.

savil (approved):
Nice!

Allowed discusssed offline that we may want a follow up PR that can preserve the README content outside the `gen-readme` comments:
```
<!-- gen-readme start - generated by https://github.com/jetpack-io/devbox/ -->
...
<!-- gen-readme end -->
```

So, it may be nice to move those comments outside the template and into the file-content we _always_ generate.

This would enable users to also custom-modify the README's markdown directly, without needing to deal with the golang template.

PR #1797, opened 2024-02-08
ci: run example tests on macOS for releases

gcurtis:
We were accidentally excluding the example tests on macOS. Also add some descriptions to the workflow inputs to make the schema validate with the YAML language server.


savil (approved):
thanks, and good catch!

PR #1801, opened 2024-02-09
examples/stacks/drupal: fix for local

gcurtis:
Fix the drupal example so that it runs locally. It's still broken in CI due to some filesystem permissions issues that I don't have time to debug. The example had a few errors, some probably due to the example drifting too far from the most recent Devbox version.

- Add `curl` to `devbox.json` so that `run_test` works.
- Increase the sleep duration (2 seconds wasn't long enough on my M1).
- Symlink devbox.d/web -> ../web so that nginx serves the drupal site instead of a static index.html.
- Update the curl command to hit localhost:8081, which is what nginx is listening on.
- Set `MYSQL_UNIX_PORT` via the `env` field in `devbox.json`. I think otherwise the plugin overrides it back to its default value.
- Update some composer dependencies to address dependabot security issues.


gcurtis:
I put back the skip for the test in CI. It runs locally, but I don't have time to figure out the filesystem permissions issues that are going on in CI.

savil (approved):

PR #1803, opened 2024-02-09
docs: update dependencies

gcurtis:
This addresses some dependabot security issues and fixes a few syntax errors.


Lagoja (approved):

gcurtis:
@LucilleH is the 404 on the preview deployment legit? Other pages all show up when I search for them, but I'm not sure why the initial load is a not found.

LucilleH:
@gcurtis it's 404ing because the docs are hosted under `/devbox/docs` subpath. So if you go to the preview url and add `/devbox/docs` you'll see the index page

PR #1806, opened 2024-02-09
[CICD] Add slack message when cli release is complete

mikeland73:
## Summary

TSIA


Lagoja:
Does this only notify when the release completes successfully? I need a message on failure + cancellation as well, since that's the issue that causes me to miss releases

mikeland73:
@Lagoja now it does. It reports test failures/cancelations and also reports final release status (assuming tests did not fail or were canceled)

loreto (approved):
LGTM

PR #1809, opened 2024-02-12
ci: increase mac test timeout to 1 hour

gcurtis:
This is to temporarily unblock CI for a release. The example tests were recently re-enabled for macOS and are taking a long time to run.


savil (approved):
need to edit line 145 too

PR #1814, opened 2024-02-14
[Outputs] Add store-paths of package outputs to devbox.lock

savil:


savil:
* `main`
  * **PR #1814** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1814?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1814?utm_source=stack-comment).

internal/devbox/packages.go
```diff
@@ -107,6 +107,10 @@ func (d *Devbox) Add(ctx context.Context, pkgsNames []string, opts devopt.AddOpt
 		return err
 	}
 
+	if err := d.lockfile.Add(addedPackageNames...); err != nil {
+		return err
+	}
```
mikeland73, line 112: Main issue with this is that this saves the lock file right away. This will leave it in a bad state if something fails later on. It's also not 100% correct (see `newFlakePlan` calls `Add`)

A few alternatives:

* Instead of doing lockfile.Add just do `lockfile.UpdateSystemsFormat()`  in the 3 places you need it (add/remove/update). This is simplest.
* In lockfile.Save() only update the format is stuff is dirty (i.e if reloading the lockfile doesn't match what is in memory). (though it may be a bit complex to tell if dirty). This is cleanest solution because it doesn't require a new bool we may forget to add.
* Remove `Save` from `Add` (this is probably OK, but it feels risky). I don't love this one.

FWIW, currently `newFlakePlan` calls `lockfile.Add` on includes every time the state needs to be recomputed, so I think any project with includes will automatically update.

internal/lock/lockfile.go
```diff
@@ -30,6 +30,10 @@ type File struct {
 
 	// Packages is keyed by "canonicalName@version"
 	Packages map[string]*Package `json:"packages"`
+
+	// legacySystemsFormat aids in managing the legacy and/or modern format of
+	// the Package.Systems in this lockfile
+	legacySystemsFormat bool `json:"-"`
```
mikeland73, line 36: `json:"-"` is redundant. Unexported fields are never marshaled.

internal/lock/lockfile.go
```diff
@@ -46,6 +50,26 @@ func GetFile(project devboxProject) (*File, error) {
...
+			if sysInfo.StorePath != "" && len(sysInfo.Outputs) == 0 {
+				lockFile.Packages[pkgName].Systems[sys].Outputs = []*Output{
+					{
+						Default: true,
+						Name:    "out",
+						Path:    sysInfo.StorePath,
+					},
+				}
+				// Track that we read from the legacy format
+				lockFile.legacySystemsFormat = true
```
mikeland73, line 68: Can't we use the existence of `StorePath` to tell this? To generalize this boolean (or even get rid of it) we could:

The no boolean solution looks like:

* If lockfile contains store paths it is legacy.
* If lockfile is dirty (i.e. loaded version doesn't match file) we modernize.  (small perf hit for re-reading file, but we should rarely write to lock file and when we do, it's a slower operation anyway)

Version keeping a bool:

* If lockfile contains store paths it is legacy.
* Add `upgradeLockfile` bool that gets triggered manually during add/remove/update.

internal/lock/package.go
```diff
@@ -43,9 +64,50 @@ func (p *Package) IsAllowInsecure() bool {
...
+// Useful for debugging when we print the struct
+func (o *Output) String() string {
+	return fmt.Sprintf("%v", *o)
+}
+
+func (o *Output) Equals(other *Output) bool {
+	if o == nil || other == nil {
+		return o == other
+	}
+	return o.Name == other.Name && o.Path == other.Path && o.Default == other.Default
```
mikeland73, line 112: This is normal equality in golang. You can do `*o == *other`

internal/lock/package.go
```diff
@@ -22,13 +26,30 @@ type Package struct {
 }
 
 type SystemInfo struct {
-	// StorePath is the input-addressed path for the nix package in /nix/store
-	// It is the cache key in the Binary Cache Store (cache.nixos.org)
-	// It is of the form /nix/store/<hash>-<name>-<version>
-	// <name> may be different from the canonicalName so we store the full store path.
+	Outputs []*Output `json:"outputs,omitempty"`
```
mikeland73, line 29: Why pointers? If you make them structs you don't need the `Equals` function below you can just compare

`a == b`

internal/lock/package.go
```diff
@@ -43,9 +64,50 @@ func (p *Package) IsAllowInsecure() bool {
...
+	if len(i.Outputs) != len(other.Outputs) {
+		return false
+	}
+	for i, o := range i.Outputs {
+		if !o.Equals(other.Outputs[i]) {
+			return false
+		}
+	}
+	return isEqual
+}
```
mikeland73, line 101: `slices.Compare == 0` (assuming you don't use pointers, see comment above)

mikeland73 (changes requested):
Added some comments with alternative approach. There's a minor bug because of previous use of `Add`

I think the cleanest solution that would not require boolean:

* If lockfile contains store paths it is legacy.
* If lockfile is dirty (i.e. loaded version doesn't match file) we modernize. (small perf hit for re-reading file, but we should rarely write to lock file and when we do, it's a slower operation anyway)

But if you run into issues implementing this, I think using a boolean is fine.

internal/devbox/packages.go
savil: ugh, I thought I had checked for existing callers, but I think I missed this one...good catch

savil:
> If lockfile contains store paths it is legacy.

This doesn't work in the following scenario:
1. user reads legacy `devbox.lock` so lockfile has non-empty StorePath
2. user does `devbox remove`
3. During `lockfile.Save`, the lockfile will have the legacy StorePath so we'll save it in legacy format
BUT since the user did an explicit action that updates devbox.lock, we want to actually also write it in modern format.

> If lockfile is dirty (i.e. loaded version doesn't match file) we modernize. (small perf hit for re-reading file, but we should rarely write to lock file and when we do, it's a slower operation anyway)

yay! this works :)
Updated code.

internal/lock/package.go
```diff
@@ -43,9 +65,29 @@ func (p *Package) IsAllowInsecure() bool {
...
+	return i.Outputs[0].Path
+}
+
 func (i *SystemInfo) Equals(other *SystemInfo) bool {
 	if i == nil || other == nil {
 		return i == other
 	}
-	return *i == *other
+
+	return slices.Equal(i.Outputs, other.Outputs)
```
gcurtis, line 92: Does this need to take `i.StorePath` into account?

gcurtis (approved):
I think using "out" for legacy data is okay. I'm pretty sure the original `store_path` JSON field mapped to `.outPath` attribute of the package.

internal/lock/lockfile.go
```diff
@@ -92,12 +110,24 @@ func (f *File) Resolve(pkg string) (*Package, error) {
 	return f.Packages[pkg], nil
 }
 
-func (f *File) ForceResolve(pkg string) (*Package, error) {
-	delete(f.Packages, pkg)
-	return f.Resolve(pkg)
-}
-
 func (f *File) Save() error {
+	isDirty, err := f.isDirty()
```
mikeland73, line 114: sweet! forgot we had this.

internal/lock/package.go
mikeland73: I think this is OK? at least looking at the callsite https://github.com/jetpack-io/devbox/blob/main/internal/devbox/update.go#L142-L146

Only outputs matter moving forward and storePath is just kept for legacy reasons (to avoid updating everyone's lockfile)

mikeland73 (approved):

PR #1816, opened 2024-02-16
Bump process compose to version 0.85.0

przemek-pokrywka:
## Summary
It would be nice for the users of Devbox to be able to use new process-compose features.

## How was it tested?
No tests were done, counting on CI and hoping that the new process compose does not break its API.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/cloudflare/ahocorasick v0.0.0-20210425175752-730270c3e184
 	github.com/creekorful/mvnparser v1.5.0
 	github.com/denisbrodbeck/machineid v1.0.1
-	github.com/f1bonacc1/process-compose v0.43.1
+	github.com/f1bonacc1/process-compose v0.85.0
 	github.com/fatih/color v1.16.0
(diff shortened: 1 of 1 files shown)


Lagoja:
I think unfortunately this won't be enough to increment the version. Changing the go.mod only bumps the library that we use in Jetpack to check some process-compose types. We would need to actually increment the version of process compose that gets installed in the `utils` profile.

gcurtis:
We've put off doing this because there are some backwards-incompatible changes in the Go module that need to be addressed. I don't think it's anything too crazy, but it just requires some testing to make sure everything still works with the new process compose binary.

Lagoja:
I think the backwards incompatible changes are minimal, the main thing I've been struggling with is how to upgrade developers already on 0.43.1, since the Nix Profile UX is a little rough.

Lagoja:
https://github.com/jetpack-io/devbox/pull/1820 < Here's a PR that upgrades process-compose to 0.85.0

Lagoja:
The upgrade PR is now merged, I'll cut a pre-release shortly for testing

Issue #1772, opened 2024-02-01
Ruby 2.7 throws the error `GLIBC_2.38' not found


jamesgecko:
### What happened?

When running Rails with Ruby 2.7, I get the following stack trace:

```
bundler: failed to load command: rails (/home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/bin/rails)
LoadError: /nix/store/jxjy8r4k8pf7v0hasi37qgva9va81kj2-glibc-2.37-8/lib/libc.so.6: version `GLIBC_2.38' not found (required by /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/date-3.3.4/lib/date_core.so) - /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/date-3.3.4/lib/date_core.so
  /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/date-3.3.4/lib/date.rb:4:in `require'
  /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/date-3.3.4/lib/date.rb:4:in `<top (required)>'
  /nix/store/vs52hm8jbwgx19plicg02dzz2vmvbyy5-ruby-2.7.8/lib/ruby/2.7.0/time.rb:3:in `require'
  /nix/store/vs52hm8jbwgx19plicg02dzz2vmvbyy5-ruby-2.7.8/lib/ruby/2.7.0/time.rb:3:in `<top (required)>'
  /nix/store/vs52hm8jbwgx19plicg02dzz2vmvbyy5-ruby-2.7.8/lib/ruby/2.7.0/open-uri.rb:4:in `require'
  /nix/store/vs52hm8jbwgx19plicg02dzz2vmvbyy5-ruby-2.7.8/lib/ruby/2.7.0/open-uri.rb:4:in `<top (required)>'
  /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/railties-6.1.7.6/lib/rails/generators/app_base.rb:6:in `require'
  /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/railties-6.1.7.6/lib/rails/generators/app_base.rb:6:in `<top (required)>'
  /home/james/src/rails-crash/.devbox/virtenv/ruby_2_7/gems/railties-6.1.7.6/lib/rails/generators/rails/app/app_generator.rb:3:in `require'
[...]

PR #1818, opened 2024-02-16
shellgen: update store paths when patching glibc

gcurtis:
This addresses an issue where scripts (and other text files) could still refer to the unpatched version of a program after enabling `patch_glibc`. This is one of the reasons why Ruby would still segfault in issue #1772. For example, if you look at `bundler` (installed by default with `ruby`) it's a wrapper script with the shebang:

	#!/nix/store/vs52hm8jbwgx19plicg02dzz2vmvbyy5-ruby-2.7.8/bin/ruby

This means that commands like `bundler exec rails` would still run using the unpatched version of ruby. I suspect this was also happening with Python or other packages that rely on wrapper scripts.

The fix updates glibc-patch.bash to grep for the old store path and replace it using sed. It relies on ripgrep's ability to ignore binary files so that we only modify text files.

### DIFF
--- a/internal/shellgen/tmpl/glibc-patch.bash
+++ b/internal/shellgen/tmpl/glibc-patch.bash
@@ -7,26 +7,29 @@ declare -r glibc # new glibc that we're patching in
 declare -r out   # nix output path that will contain the patched package
 
 # Paths to this script's dependencies set by nix.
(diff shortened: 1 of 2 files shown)


savil (approved):
nice!

PR #1820, opened 2024-02-17
Update Process Compose to 0.85.0

Lagoja:
## Summary

Devbox will now install process-compose 0.85.0 if an older version (or no version) is installed on the user's system

If an older version of process compose was installed in the utility profile, Devbox will remove that older version first. This PR provides a general way to remove packages (either from Flakes or Nixpkgs) from the utils profile

This PR also fixes a bug where using `devbox services up` with the `--config` flag would cause process compose to fail.

This PR doesn't add any new 0.85.0 features, though that could be a follow up. 

## How was it tested?

* Install an older version in the utils profile, check that this upgrades
* Remove process-compose from the utlis profile, check that 0.85.0 is installed
* This should be pre-released first, for more extensive testing

### DIFF
--- a/internal/devbox/util.go
+++ b/internal/devbox/util.go
@@ -5,13 +5,15 @@ package devbox
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/devpkg"
-	"go.jetpack.io/devbox/internal/nix/nixprofile"
+	"go.jetpack.io/devbox/internal/nix"
 
 	"go.jetpack.io/devbox/internal/xdg"
 )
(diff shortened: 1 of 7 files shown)


internal/devbox/util.go
```diff
@@ -32,14 +35,66 @@ func (d *Devbox) addDevboxUtilityPackage(ctx context.Context, pkgName string) er
 		return err
 	}
 
-	return nixprofile.ProfileInstall(ctx, &nixprofile.ProfileInstallArgs{
```
mikeland73, line 35: If `nixprofile.ProfileInstall` and `nixprofile.ProfileInstallArgs` are no longer used you should remove.

internal/devbox/util.go
```diff
@@ -32,14 +35,66 @@ func (d *Devbox) addDevboxUtilityPackage(ctx context.Context, pkgName string) er
...
+		return err
+	}
+	profileString, err := nix.ProfileList(d.stderr, utilityProfilePath, true)
+	if err != nil {
+		return err
+	}
+
+	if err = json.Unmarshal([]byte(profileString), &utilProfile); err != nil {
+		return err
+	}
```
mikeland73, line 64: You should use `nixprofile.ProfileListItems` which handles legacy nix outputs. It may also simplify finding the index.

Edit: actually, you can use `ProfileListNameOrIndex` to find the index.

internal/nix/profiles.go
```diff
@@ -142,6 +142,16 @@ func readManifest(profilePath string) (manifest, error) {
 
 const DefaultPriority = 5
 
+type NixProfile struct {
```
mikeland73, line 145: Probably not needed

mikeland73 (changes requested):

internal/services/client.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/f1bonacc1/process-compose/src/types"
 )
 
-type processStates = types.ProcessStates
+type processStates = types.ProcessesState
```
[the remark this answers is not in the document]
Lagoja: Should be. Everything still seems to compile and work as expected.

internal/devbox/util.go
Lagoja: I was a little confused about having both nixprofile and nix.profile. Is nix.profile the one we should keep using? Can we completely remove nixprofile?

internal/devbox/util.go
Lagoja: I'm not really sure why we have nixprofile and nix.profile in the source code, but both of those methods you mentioned seem to require that we have a lockfile, which does not exist in the case of the utility profile (since we're manipulating the profile directly, instead of using Devbox). 

I think we'd need to either refactor those methods to not require a profile, or refactor the utility profile to use Devbox.

Lagoja: Went with using ProfileListItems to get the list, and then doing a quick match based on unlockedReference == installable. This should work for utility packages that we install from a Flake reference and not from a store path, which is currently the case for all utility packages.

internal/devbox/util.go
```diff
@@ -32,14 +34,49 @@ func (d *Devbox) addDevboxUtilityPackage(ctx context.Context, pkgName string) er
...
+}
+
+func findUtilPackage(installable string, profile []*nixprofile.NixProfileListItem) int {
+	for i := range profile {
+		if profile[i].MatchesInstallable(installable) {
+			return i
+		}
+	}
+	return -1
+}
```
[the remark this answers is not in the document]
Lagoja: Thanks, that's much better

PR #1824, opened 2024-02-21
[devbox.json] Add and implement imports field

mikeland73:
## Summary

Adds and implements `imports` field to devbox.json. Imports follow flake ref specs (but are limited to only path and github for now).

For example:

```json
"imports": [
  "./foobar",
  "path:./foobar2",
  "github:jetpack-io/devbox?dir=somedir",
]
```

Fields from imported configs are added to the environment. The merge algorithm is as follows:

* Imports are evaluated in order first. Base config is evaluated last.
* Slices are concatenated (init hooks, etc)
* maps are merged with later evaluation replacing previous one on conflict.
* Packages are deduped by name and concatenated. (i.e. multiple versions not allowed)

Caveats:

* This can be slow. When using remote imports, every devbox command does 1 or more HTTP requests to load the imports.
* central lock file. All dependencies are stored there. 

TODOs:

* Handle relative paths inside of imported configs
* Cycle detection

## How was it tested?

* Tested importing local devbox.json
* Tested importing github  devbox.json

### DIFF
--- a/nix/flake/reflike.go
+++ b/nix/flake/reflike.go
@@ -1,0 +1,76 @@
+package flake
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/samber/lo"
+	"go.jetpack.io/devbox/internal/boxcli/usererr"
+)
+
+// RefLike is like a flake ref, but in some ways more general. It can be used
+// to reference other types of files, e.g. devbox.json.
+type RefLike struct {
+	Ref
+	filename string
+}
+
+func ParseRefLike(s, filename string) (RefLike, error) {
+	r, e := ParseRef(s)
+	return RefLike{r, filename}, e
+}
+
+func (r RefLike) Fetch() ([]byte, error) {
+	switch r.Type {
+	case TypePath:
+		return os.ReadFile(r.withFilename(r.Path))
+	case TypeGitHub:
+		return r.fetchGithub()
+	default:
+		return nil, fmt.Errorf("unsupported ref type %q", r.Type)
+	}
+}
+
+// TODO: This is almost copy paste of plugins/github.go. We should refactor this
+func (r RefLike) fetchGithub() ([]byte, error) {
+	// Github redirects "master" to "main" in new repos. They don't do the reverse
+	// so setting master here is better.
+	contentURL, err := url.JoinPath(
+		"https://raw.githubusercontent.com/",
+		r.Owner,
+		r.Repo,
+		lo.Ternary(r.Rev == "", "master", r.Rev),
+		r.withFilename(r.Dir),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := http.Get(contentURL)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		return nil, usererr.New(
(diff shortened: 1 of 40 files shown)


nix/flake/reflike.go
gcurtis, line 1: I think there's a way we can leverage Nix's ability to fetch flake refs. I'll add it to the review comment.

gcurtis (commented):
Instead of adding our own download logic, we could use `nix flake prefetch --json <flake-ref>` to fetch the directory containing the devbox.json. Contrary to the name, it can download non-flake sources:

> This command downloads the source tree denoted by flake reference flake-url. Note that this does not need to be a flake (i.e. it does not have to contain a flake.nix file).

I think this would have a few advantages:

- All flake references work automatically.
- Nix handles caching for us by downloading to the Nix store.
- We get a bunch of info back that would be useful for locking the included config (if we want to eventually implement that).

Using the devbox repo as an example:

```
$ nix flake prefetch --json github:jetpack-io/devbox | jq
{
  "hash": "sha256-7i6hmy1EOAeNW6TXZKoXsBeuRksx3QDrqzatOGXqR1g=",
  "locked": {
    "lastModified": 1708479263,
    "narHash": "sha256-7i6hmy1EOAeNW6TXZKoXsBeuRksx3QDrqzatOGXqR1g=",
    "owner": "jetpack-io",
    "repo": "devbox",
    "rev": "b0d6032376ee21bd9d428358a3046984770e5fd2",
    "type": "github"
  },
  "original": {
    "owner": "jetpack-io",
    "repo": "devbox",
    "type": "github"
  },
  "storePath": "/nix/store/psjz43j55y9qx2sbpr035qywmrj42031-source"
}
$ cat /nix/store/psjz43j55y9qx2sbpr035qywmrj42031-source/devbox.json
{
  "name":        "devbox",
  "description": "Instant, easy, and predictable development environments",
...
```

We can also prefetch local paths, but they have to be in the URL form (`path:./foo` not `./foo`). The `flake.Ref.String()` method can already handle converting those though.

gcurtis:
It can download individual files if you transform the `github` flake reference to a `file` one. For example, `nix flake prefetch --json file+https://raw.githubusercontent.com/jetpack-io/devbox/main/devbox.json`.

That said, I think that downloading the full tree has some advantages.

- You can have a devbox.json that references files within its own repo. For example, something like axiom's Go linting and tidy scripts.
- You can have common plugin configs, such as an `nginx.conf`.
- It aligns with how flake references work, so we don't need to add another syntax.

We don't need to worry about caching if we rely on Nix doing it. After the initial download, `nix flake prefetch` returns instantly and skips the download. We could always make it even faster by incorporating it into the Devbox cache, but it's fast enough that we don't need to think about it for the first implementation.

nix/flake/reflike.go
gcurtis: If you end up leaving in the Devbox fetching logic, could the reflike code be moved to another package? Just so we can keep nix/flake specific to nix.

mikeland73: > If you end up leaving in the Devbox fetching logic, could the reflike code be moved to another package? Just so we can keep nix/flake specific to nix

Yep, for sure. I was thinking of this, but didn't end up doing it. I can do as part of this PR.

savil (changes requested):
putting back on your queue, since I think we're no longer doing this. Please re-request review if needed.

mikeland73:
Closing in favor of https://github.com/jetpack-io/devbox/pull/1850

PR #1825, opened 2024-02-21
When SIGINT is detected by `devbox run` send SIGTERM to the pid of the script that is running

### LINKED ISSUE #1815: Support passing `SIGINT` to scripts that are running via `devbox run`

Scripts run via `devbox run` cannot be cancelled with `SIGINT`. 

Handle the SIGINT signal and pass it to the script being execute by `devbox run`.

Not running scripts via devbox.  That is not desirable as the scripts defined in `devbox.json` provide a great way to catalog functionality.

kadaan:
## Summary

When SIGINT is detected by ``devbox run` send SIGTERM to the pid of the script that is running.  Fixes #1815 

## How was it tested?

I run a script before this change and `CTRL-C` could not cancel it.  With this change it can be canceled.


kadaan:
Is there a good way to add test for these cases into the repository? If there is, I will gladly do it.

mikeland73 (changes requested):

kadaan:
I made an initial stab of doing it this way and cancelling didn't work.  I will try to get back to this in a bit.

PR #1826, opened 2024-02-21
[UX] Show add (nix build) stdout

mikeland73:
## Summary

I mostly copied over what we have from `nix.ProfileInstall`. I thinks this is OK? It will pipe all output and errors to stderr so user can see progress and if anything breaks it will show up there as well.

## How was it tested?

Tried installing something slow:
image


savil (approved):
Thanks! I was just looking into this myself as well :)

Issue #1823, opened 2024-02-21
Hide banner message from python plugin


kadaan:
### What problem are you trying to solve?

The python plugin logs a banner message no matter what.. https://github.com/jetpack-io/devbox/blob/7f1633f6834f8772d31b06588c637b8f12cbe7c7/plugins/pip/venvShellHook.sh#L8

If you use the shell init_hook to enter the virtual env it would be nice to suppress this message.  This is especially important when scripts are execute via `devbox run ...`, because the message ends up just being cruft.

### What solution would you like?

Create a environment variable set automatically by devbox: `DEVBOX_ENTRYPOINT=run:<script_name>` or `DEVBOX_ENTRYPOINT=shell`.
Then the python plugin could not print the banner if `DEVBOX_ENTRYPOINT=run:*`.  This would solve the run case.

Optionally, change the python plugin to not log the banner if a specific environment variable is present (ie. `DEVBOX_PLYTHON_PLUGIN_HIDE_BANNER`

### Alternatives you've considered

Writing a wrapper around devbox would be another way to solve this, but that seems like more maintenance in the long run.

gcurtis:
I agree this is an annoying message. I'm not sure about adding a `DEVBOX_ENTRYPOINT` environment variable, but we could add a more general version of `DEVBOX_PLYTHON_PLUGIN_HIDE_BANNER`.

A few ideas:

- Only show the message once. We could achieve this by saving a list in the `.devbox` directory of messages we've already shown.
- Detect a `DEVBOX_NO_HINTS` environment variable (similar to brew's `HOMEBREW_NO_ENV_HINTS`) that suppresses plugin hints globally.
- Now that `devbox.json` allows comments, we could add the message as a comment above packages that trigger plugins. This might be annoying if it messes up strict JSON syntax highlighting in editors though.

@Lagoja @mikeland73 any thoughts on this?

kadaan:
Mind explaining the issue with allowing the plunging to know the entrypoint to them being called?  Seems like a useful and generic way of solving the problem that could enable other scenarios.

Lagoja:
@gcurtis For this message in particular, maybe we move it into the plugin README and out of the script itself. The README would show when you install Python, but not on subsequent runs?

kadaan:
[...]

PR #1830, opened 2024-02-22
Enabled plugins to detect devbox entrypoint and alter their behavior

kadaan:
## Summary

Extract getSubcommand from telemetry and use it to set the new DEVBOX_ENTRYPOINT environment variable in midcobra.  Use the new env var in venvShellHook.sh to stop printing the banner when running via devbox run ...

Fixes: #1823

## How was it tested?

Manually run a devbox with python package via both `devbox shell` and `devbox run`.  When `devbox shell` is used, the banner is printed.  When `devbox run` is used, the banner is not printed.


kadaan:
Another example of a command that should not print the python banner:
```
❯ devbox services ls
You can activate the virtual environment by running 'source $VENV_DIR/bin/activate'
No services currently running. Run `devbox services up` to start them:
```

kadaan:
@Lagoja @gcurtis @mikeland73 Can you offer some feedback on this?  I'm game to change this if this is not how you want to approach it, but the banner messages are increasingly annoying.

kadaan:
@Lagoja Any feedback?

kadaan:
Any reason this was closed?

loreto:
Hi @kadaan, I was just automatically closing PRs older than 4 months (since most of them are abandoned) without making any judgement (or reading) any particular PR.

For this one specifically, are you still hoping to check it in and do you still need feedback from John? (if yes, I can ping him to take a look and we can reopen it)

PR #1831, opened 2024-02-22
[plugins] Unify plugin config and devbox config

mikeland73:
## Summary

Changes plugins to use devbox config file struct. Additional fields (e.g. `create_files`) are added to plugin struct, but not devbox config.

A few other changes:

* readme is deprecated in favor of description. Backward compatible code added.
* `__packages` is renamed to `packages`. Structure can now match devbox.json  (which means it now allows options like allow_insecure, etc). This field has only ever been public and used for builtins.

## How was it tested?

* builds
* CICD


mikeland73:
Test failure is unrelated.

internal/devconfig/file.go
```diff
@@ -33,10 +33,11 @@ const (
 	tsonFormat
 )
 
-// configFile defines a devbox environment as JSON.
-type configFile struct {
+// ConfigFile defines a devbox environment as JSON.
+type ConfigFile struct {
 	Name        string `json:"name,omitempty"`
 	Description string `json:"description,omitempty"`
+	Version     string `json:"version"`
```
gcurtis, line 40: Can we make this a separate change for now? It would be nice if `devbox.json` didn't require a version field.

internal/plugin/plugin.go
```diff
@@ -281,3 +276,14 @@ func (m *Manager) shouldCreateFile(
...
+
+func (c *config) Description() string {
+	if c == nil {
+		return ""
+	}
+	return lo.Ternary(
+		c.ConfigFile.Description != "",
+		c.ConfigFile.Description,
+		c.DeprecatedDescription,
+	)
```
gcurtis, line 288:
```suggestion
	return cmp.Or(c.ConfigFile.Description, c.DeprecatedDescription)
```

gcurtis (approved):

savil (approved):

PR #1832, opened 2024-02-22
skip lapp for now

Lagoja:
## Summary

LAPP examples test is consistently failing in CI with the following: 

```
dropdb: error: connection to server on socket "/tmp/devbox/lapp/.s.PGSQL.5432" failed: No such file or directory
Is the server running locally and accepting connections on that socket?
```

I'm unable to consistently reproduce this failure in either devbox.sh, or locally on my Mac. Until we can resolve, I want to unblock CI

### DIFF
--- a/testscripts/testrunner/examplesrunner.go
+++ b/testscripts/testrunner/examplesrunner.go
@@ -75,6 +75,15 @@ func RunDevboxTestscripts(t *testing.T, dir string) {
 			return nil
 		}
 
+		if strings.Contains(path, "lapp") {
+			if envir.IsCI() {
+				// lapp is mysteriously failing with an "unable to find socket" error.
+				// This has not been reproduced locally, so disabling until we can resolve it.
+				t.Logf("skipping lapp, config at: %s\n", path)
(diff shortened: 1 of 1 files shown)


savil (approved):

savil:
@Lagoja should we disable `lepp` as well? @mikeland73 is also hitting the same error sporadically but for lepp as well

Lagoja:
Are the two tests conflicting somehow? I've consistently gotten LAPP to work but not LEPP, but disabling one didn't trigger a failure in the other.

mikeland73:
It happened once on both, but usually it's lapp that fails.

I'm trying out changing the port to see if that works: https://github.com/jetpack-io/devbox/pull/1831/commits/b7eec160d09a787d84db12ff44446d26c36df9be

mikeland73:
We may not need to disable. https://github.com/jetpack-io/devbox/pull/1835 looks like it fixes it.

Lagoja:
Closing for now since we have a lead on a workaround

PR #1835, opened 2024-02-23
[process-compose] Improve port selection

mikeland73:
## Summary

This helps fix the lapp and lepp stack test issues. It also provides a more reliable way to provide an unused port that does not have race condition issues or the limit of 10 ports.

## How was it tested?
devbox services start

### DIFF
--- a/internal/services/manager.go
+++ b/internal/services/manager.go
@@ -25,8 +25,6 @@ import (
 
 const (
 	processComposeLogfile = ".devbox/compose.log"
-	startingPort          = 8260
-	maxPortTries          = 10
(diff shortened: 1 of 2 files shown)


Lagoja:
I'm not entirely sure what it does, but it looks like it worked in the GH Action. 

One question I had -- Do we want it to potentially grab _any_ open port? Or would we rather have it randomly select a port within a given range? That would make it easier for developers to avoid potential port conflicts between their service and process-compose.

Lagoja (approved):

mikeland73:
@Lagoja I disallowed a few common ports. PTAL

PR #1837, opened 2024-02-23
[Outputs] Update generated flake to handle store paths for each output

savil:
## Summary

Goal:
This PR ensures we add store-paths for package outputs in the generated flake. This means users can skip the nixpkgs download saving 30-60 seconds, for packages with non-default outputs.

Along the way, it also improves how we handle multiple default outputs of packages. Well, the follow up PR #1842 achieves that.

Implementation:
- Needed to add an `outputs` struct in `devpkg`. It was needed to abstract the handling of user-selected outputs versus default outputs.
- The logic in `devpkg/narinfo_cache.go` to determine if the store-path is cached needed updating to be output-specific. 
- Needed to add output-specific versions of `package.Installable()` and `package.InputAddressedPath`

For ease of reviewing, I suggest toggling "Hide Whitespace":
Screenshot 2024-02-26 at 12 23 59 PM

## How was it tested?

`devbox add prometheus --outputs cli,out`

The testscripts have some flakes that are added, which continue to work.

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -77,11 +76,7 @@ type Package struct {
 	Raw string
 
 	// Outputs is a list of outputs to build from the package's derivation.
-	// If empty, the default output is used.
-	//
-	// Note the distinction: these are user-selected outputs, whereas the lockfile has outputs
-	// whose store-paths are available outputs with store-paths.
-	Outputs []string
+	outputs *outputs
 
 	// PatchGlibc applies a function to the package's derivation that
 	// patches any ELF binaries to use the latest version of nixpkgs#glibc.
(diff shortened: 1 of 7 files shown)


savil:
* **#1837** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1837?utm_source=stack-comment).

internal/devpkg/package.go
```diff
@@ -79,9 +75,12 @@ type Package struct {
 	//    example: github:nixos/nixpkgs/5233fd2ba76a3accb5aaa999c00509a11fd0793c#hello
 	Raw string
 
-	// Outputs is a list of outputs to build from the package's derivation.
+	// Outputss is a list of outputs to build from the package's derivation.
```
gcurtis, line 78:
```suggestion
	// Outputs is a list of outputs to build from the package's derivation.
```

gcurtis (approved):
Nice! I know the package and lockfile code has become a bit of a maze, which makes changes like this tricky. From what I can tell, this makes sense though.

I have some cleanup ideas (outside the scope of this PR) that I want to jot down in a separate issue. The way we install and lock packages has changed so much, that I think there's a lot we can do to simplify it. Hopefully we can make it easier to change this part of the code.

PR #1839, opened 2024-02-24
[cicd] split out project and non-projec tests

mikeland73:
## Summary

This splits out CICD tests into 2 distinct non-overlapping set of tests:

* Fast tests (golang)
* Slow tests (Examples and testscripts)

This helps avoid a bit of duplicate work and paralelizes a bit more. I'm hoping it makes tests a bit faster (so far in anecdotal testing in CICD, it does seem to shave a minute or two from the slowest test).

Other improvements:

* Use devbox to run tests
* Re-enable insecure example test

Update: looks like a couple of minute improvement so far:

image

## How was it tested?

CICD


devbox.json
```diff
@@ -36,6 +36,7 @@
       "lint":            "golangci-lint run --timeout 5m && scripts/gofumpt.sh",
       "fmt":             "scripts/gofumpt.sh",
       "test":            "go test -race -cover ./...",
+      "test-projects-only": "DEVBOX_RUN_PROJECT_TESTS=1 go test -v -timeout ${DEVBOX_GOLANG_TEST_TIMEOUT:-30m} ./... -run \"TestExamples|TestScriptsWithProjects\"",
```
gcurtis, line 39: Are all of the testscript tests under ./testscripts? Could this be `./testscripts/...` instead of the `-run` flag?

gcurtis (approved):

devbox.json
mikeland73: Good call, I think so. I'll confirm.

mikeland73: They are not quite the same (and maybe we should split the out in the future).

testscripts dir include 3 types of tests:

* manual test scripts 
* "project" test scripts. These are examples that live under testsciprts (Just 1, shellenv/node) 
* examples (everything in `examples` dir that has a `run_test` script

(Only 2 and 3 respect the `DEVBOX_RUN_PROJECT_TESTS` env var)

The reason they are combined is that we use testscript library to run them. 

In terms of speed, examples is slowest (usually taking 10 min on linux and 30+ on mac), manual test scripts take 3-4 minutes, and "project" test script is close to zero.

Since the goal of this PR is to avoid double work, I'm gonna keep `-run` as is (otherwise we end up doing the manual test scripts twice). 

I think medium term we need to decide a better way to run these. They are incredibly valuable, but they make releases really painful (sometimes we have to wait 40-60 min just to get a failure). This PR is a quick hack, but longer term could:

* Cache the nix store so all examples are fast.
* Run 1 uncached run every day wit cronjob for testing uncached scenarios.

We probably can't fit the cache in GHA cache.

PR #1841, opened 2024-02-24
[examples] Update all examples

mikeland73:
## Summary

Updating examples.

* Did `devbox update --all-projects` in examples dir. (This broke in one of the projects, need to investigate. Update: broke do to missing `excluded_platforms`)
* Updated a few devbox.json to ensure we're using similar versions (to speed up testing)

Hoping this speeds up CICD a bit. Update: surprisingly no CICD improvement (Saw small improvements in newer nix versions, small regression in nix 2.12). Still think this is worth it.

@Lagoja notes: 

* I can't get the ory-cli flake example to work.
* I think we can update a few examples where the plugin has made a package obsolete (e.g. composer for php, pip for python)

## How to review

This Pr only changes a single prod file (the change template dir). Everything else is from examples

image

## How was it tested?

CICD


Lagoja:
@mikeland73 Fixed Ory. What are the changes you'd propose for python + php? They seem ok from what I can tell?

examples/data_science/tensorflow/devbox.json
```diff
@@ -1,8 +1,8 @@
 {
   "packages": [
-    "python310",
     "python310Packages.pip",
-    "stdenv.cc.cc.lib"
+    "stdenv.cc.cc.lib",
+    "python310@latest"
```
savil, line 5: we should make a pass (not blocking) to update all such examples to be of the modern form: `python@3.10`

savil (approved):

examples/data_science/tensorflow/devbox.json
mikeland73: Agreed. This pr was mostly automated stuff (which makes it large and hard to review) so I was hoping to keep manual changes to minimum. I can do that in follow up.

PR #1844, opened 2024-02-26
[cicd] Run mac tests on cron only

mikeland73:
## Summary

* Don't run mac tests on release
* Don't run mac tests on main
* Only run mac tests on request or nightly

## How was it tested?

CICD

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -26,10 +26,6 @@ permissions:
 jobs:
   tests:
     uses: ./.github/workflows/cli-tests.yaml
-    with:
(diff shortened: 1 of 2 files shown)


gcurtis (approved):

mikeland73:
> It would be great to be able to further toggle Mac tests as:
testscripts tests
projects tests


We currently don't split out these tests at all anywhere. we use testcripts infra to run examples, txt testscripts, and testscript examples.

Did you mean testscripts vs non-tescripts (golang stuff)? (in this yaml we refer to them as `project-tests-only` vs `project-tests-off`, we may want to come up with better names)

by default we run both in parallel.

PR #1845, opened 2024-02-26
[plugins] Use reflikes for plugins

mikeland73:
## Summary

This changes plugins to use new `reflike` struct. It brings it more in line with how flake refs work. For now, only some types are supported.

## How was it tested?

CICD


nix/flake/flakeref.go
```diff
@@ -396,7 +397,13 @@ func splitPathOrOpaque(u *url.URL) ([]string, error) {
 	upath = path.Clean(upath)
 
 	var err error
-	split := strings.Split(upath, "/")
+	var split []string
+	if allowSlashesIfRef {
+		split = strings.SplitN(upath, "/", 3)
```
savil, line 402: I don't understand what scenario is being dealt with here. What does `allowSlashesifRef` do, and why does it only care about 3 slashes, while the `else` can deal with an unlimited number of slashes?

nix/flake/flakeref.go
```diff
@@ -206,7 +206,7 @@ func parseGitHubRef(refURL *url.URL, parsed *Ref) error {
 	// github:<owner>/<repo>(/<rev-or-ref>)?(\?<params>)?
 
 	parsed.Type = TypeGitHub
-	split, err := splitPathOrOpaque(refURL)
+	split, err := splitPathOrOpaque(refURL, true)
```
savil, line 209: nit: comment bool param

nix/flake/flakeref.go
```diff
@@ -129,7 +129,7 @@ func parseURLRef(ref string) (parsed Ref, fragment string, err error) {
 		// [flake:]<flake-id>(/<rev-or-ref>(/rev)?)?
 
 		parsed.Type = TypeIndirect
-		split, err := splitPathOrOpaque(refURL)
+		split, err := splitPathOrOpaque(refURL, false)
```
savil, line 132: nit: comment bool param

internal/plugin/reflike.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"go.jetpack.io/devbox/nix/flake"
+)
+
+// RefLike is like a flake ref, but in some ways more general. It can be used
+// to reference other types of files, e.g. devbox.json.
```
savil, line 12: is the only difference that flake ref refer to directories, whereas here it can refer to a specific file? If so, could we call that out more strongly?

internal/plugin/github.go
```diff
@@ -4,56 +4,56 @@ import (
...
+}
+
+func (p *githubPlugin) Fetch() ([]byte, error) {
+	// Github redirects "master" to "main" in new repos. They don't do the reverse
+	// so setting master here is better.
+	contentURL, err := url.JoinPath(
+		"https://raw.githubusercontent.com/",
+		p.Owner,
+		p.Repo,
+		lo.Ternary(p.Rev == "", "master", p.Rev),
```
savil, line 31: to use golang std lib: `cmp.Or(p.Rev, "master")`

internal/plugin/github.go
```diff
@@ -4,56 +4,56 @@ import (
...
+		p.withFilename(p.Dir),
```
savil, line 32: not sure I follow: why are we passing in `p.Dir` (a directory) as an argument for `withFilename`?

internal/plugin/github.go
```diff
@@ -4,56 +4,56 @@ import (
...
 	"go.jetpack.io/devbox/internal/cachehash"
 )
 
 type githubPlugin struct {
-	raw      string
-	org      string
-	repo     string
-	revision string
-	dir      string
+	RefLike
```
savil, line 14: instead of embedding the struct, could we make it an explicit field to distinguish methods/fields that are from RefLike versus native to `githubPlugin`?

But its tolerable if it means having to re-implement a lot of methods unnecessarily

internal/plugin/github.go
```diff
@@ -66,10 +66,10 @@ func (p *githubPlugin) FileContent(subpath string) ([]byte, error) {
 	// so setting master here is better.
 	contentURL, err := url.JoinPath(
 		"https://raw.githubusercontent.com/",
-		p.org,
-		p.repo,
-		p.revision,
-		p.dir,
+		p.Owner,
+		p.Repo,
+		lo.Ternary(p.Rev == "", "master", p.Rev),
```
savil, line 71: consider: `cmp.Or`

savil (commented):
Generally lgtm, but I was having trouble following some of the logic (see comments inline)

nix/flake/flakeref.go
gcurtis: I think this is to fix a bug where we weren't parsing github refs with slashes in them correctly. I broke out the flakeref fixes into a separate PR just so they can be tracked separately - https://github.com/jetpack-io/devbox/pull/1863.

gcurtis:
Thanks for catching the flakeref parsing bugs. Still reviewing this PR, but I broke them out into a separate PR and added a couple tests - https://github.com/jetpack-io/devbox/pull/1863/files.

gcurtis (commented):
I'm having some trouble following the ref-like syntax. It sounds like it allows specifying specific files, but I'm not sure where it's parsing that (or maybe we're not yet?).

internal/plugin/reflike.go
mikeland73: flake refs are specific to nix flakes and I wanted to avoid modifying that. The only real difference here is that there's an optional default filename the reflike referres to. In the case of plugins is `devbox.json` but I want to leave the door open to using reflikes for composing multiple projects as well if that is something we want to support in the future (In that case the filename is probably a `devbox.json`)

internal/plugin/github.go
mikeland73: different plugin types have different fields represent the path. Local plugins use `Path` while github plugins use `Dir`

Maybe a cleaner solution could be to implement a `configPath()` function for each type. (though they would still both have some duplicate code)

mikeland73:
> I'm having some trouble following the ref-like syntax. It sounds like it allows specifying specific files, but I'm not sure where it's parsing that (or maybe we're not yet?).

That's actually a bug fixed in https://github.com/jetpack-io/devbox/pull/1850. Will fix here as well.

mikeland73:
> I'm having some trouble following the ref-like syntax. It sounds like it allows specifying specific files, but I'm not sure where it's parsing that (or maybe we're not yet?).

Actually this is fine here. `RefLike` takes a `filename` and uses `withFilename` to ensure the correct root config is used.

internal/plugin/github.go
mikeland73: I can make that change. It does mean that `plugin.Repo` becomes `plugin.ref.Repo` so the code is a big uglier. (There's roughly 12 callsites in this file that would change)

mikeland73:
Cleanup PR https://github.com/jetpack-io/devbox/pull/1867

(The stacked nature makes it a bit tricky to make those changes here)

PR #1850, opened 2024-02-27
[plugins] Implement plugins v2

mikeland73:
## Summary

Apologies in advance for the size of this PR. Was not really sure how to split it up, since it's a fairly large step change.

This PR makes 2 major changes:

* Change how plugin fields are processed. Instead of being processed independently in several parts of the code (`Env`, `Scripts`, etc) they are now combined inside of `devconfig` in a simple, predictable way. 
* Plugins are now allowed to have all fields that devbox.json have. In fact the plugin struct now embeds a ConfigFile.

All fields from included plugins are added to the environment. The merge algorithm is as follows:

* includes are evaluated in order first. Base config is evaluated last.
* Slices are concatenated (init hooks, etc)
* maps are merged with later evaluation replacing previous one on conflict.
* Packages are deduped by name and concatenated. (i.e. multiple versions not allowed)

**Known issues/limitations:**

* Github plugins will be slow. When using remote imports, every devbox command does 1 or more HTTP requests to load the imports.
* central lock file. All dependencies are stored there. 
* While recursive includes are allowed, relative paths are not handled correctly. That means if project A includes plugin 1 and plugin 1 includes a relative local plugin, it will not work. Including github plugins recursively works as expected (but will be slow).
* Cycle detection is not implemented. A user can cause infinite include.
* Plugins are json only. They don't support hujson. (This is inline with current plugins, but we should build support to match devbox.json)

**Additional requires testing:**

- [x] Test create_files works as expected (for both local and github)
- [x] Test process compose works as expected
- [x] Test global push pull (config changes may have affected that)
- [x] Ensure that we are able to `add` a package that is already added by plugin.

Follow up cleanup:

* Remove tyson support (this will simplify config code)
* Plugin manager can be removed or greatly reduced.
* Fix some of the limitations (relative paths, cycle detection)
* Cache included remote plugins for performance (and to avoid github limits). 

## How was it tested?

* New example tests
* Manual testing
* Existing tests


savil:
From reading the summary, a couple of callouts:

 **Design of github plugins**
> "Github plugins will be slow. When using remote imports, every devbox command does 1 or more HTTP requests to load the imports."
"Cache included remote plugins for performance (and to avoid github limits).". 

Two concerns:
1. Reproducibility: if we are pulling from Github every time, then every change to that Github repo will be instantly reflected. Instead, could we lock the current version in time in the lockfile (perhaps resolve to a git commit)? Then `devbox update` can fetch the very latest version.
2. Denial of service: due to the github limits being hit, especially for folks using Devbox in CICD. As a mitigation, we can add support for GITHUB_TOKEN that has higher limits. In CICD, users should be recommended to set that as well.

**Plugins evaluation order**
> "includes are evaluated in order first. Base config is evaluated last."

This makes sense. How are we handling plugins that are applied on packages added in the base-config? For example, a Rust plugin that must be applied to the version of Rust included in the base config.

internal/devconfig/config.go
```diff
@@ -1,39 +1,192 @@
...
+	return !cfg.Root.Equals(&DefaultConfig().Root)
+}
+
+func LoadForTest(path string) (*Config, error) {
 	b, err := os.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
-	baseConfig, err := loadBytes(b)
+	return loadBytes(b)
```
savil, line 65: can replace entire function's implemenation with `return readFromFile(path)`

internal/devconfig/config.go
```diff
@@ -1,39 +1,192 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &Config{
+		Root: *root,
+	}, nil
+}
+
+func (c *Config) LoadRecursive(lockfile *lock.File) error {
```
savil, line 101: reviewed till here.

gcurtis (commented):
Also still reviewing. I'll finish it up this afternoon.

internal/devconfig/configfile/file.go
```diff
@@ -80,36 +78,15 @@ type Stage struct {
...
-
 func (c *ConfigFile) Bytes() []byte {
 	b := c.ast.root.Pack()
 	return bytes.ReplaceAll(b, []byte("\t"), []byte("  "))
 }
 
 func (c *ConfigFile) Hash() (string, error) {
+	if c.ast == nil {
+		return cachehash.JSON(c)
+	}
```
savil, line 89: in what scenario would `c.ast` be `nil`? 


Instead, should we be building the ast in this if-block?

internal/devpkg/package.go
```diff
@@ -100,12 +102,14 @@ func PackagesFromStringsWithOptions(rawNames []string, l lock.Locker, opts devop
...
+func PackagesFromConfig(packages []configfile.Package, l lock.Locker) []*Package {
 	result := []*Package{}
-	for _, cfgPkg := range config.Packages() {
-		pkg := newPackage(cfgPkg.VersionedName(), cfgPkg.IsEnabledOnPlatform(), l)
+	for _, cfgPkg := range packages {
+		pkg := newPackage(cfgPkg.VersionedName(), cfgPkg.IsEnabledOnPlatform, l)
 		pkg.DisablePlugin = cfgPkg.DisablePlugin
-		pkg.PatchGlibc = cfgPkg.PatchGlibc && nix.SystemIsLinux()
+		pkg.patchGlibc = sync.OnceValue(func() bool {
+			return cfgPkg.PatchGlibc && nix.SystemIsLinux()
```
savil, line 111: is changing to a function _really_ needed?

I'm not convinced because:`nix.SystemIsLinux` relies on `nix.System` which is pre-computed once when running the cobra command, so likely won't be an actual system call here

internal/plugin/includes.go
```diff
@@ -4,11 +4,22 @@ import (
 	"strings"
 
 	"go.jetpack.io/devbox/internal/devpkg"
+	"go.jetpack.io/devbox/internal/lock"
```
savil, line 7: Reviewed till here.

savil (commented):
Pausing for a bit. Left some more comments but not any fundamental design problems so far!

Question:
- is there a way to disable a builtin plugin by having an includable plugin supersede it? For example, the builtin python plugin may be doing something a user feels is undesirable and would like to either disable it completely, or override it with their own python plugin

savil (commented):
@mikeland73 nice, overall lgtm. Please re-request review for when I should look again.

Please see above comments in the comment-box and in code for:
1. questions about areas i couldn't follow why changes were made
2. suggestions for minor improvements

A concern with this design is that `scripts` must be fully embedded in the plugin's `devbox.json` file, rather than moved into their own `my_script.sh` file where they can be properly handled by tooling (e.g. syntax-hightlighting, linting, etc.). Enabling that scenario would be a big win IMO.

One idea for a follow up PR is to add special-case support for detecting script files in plugin-configs, and downloading them as well.

internal/devconfig/config.go
```diff
@@ -1,39 +1,192 @@
...
+	for _, builtIn := range builtIns {
+		includable := &Config{
+			Root:       builtIn.ConfigFile,
+			pluginData: &builtIn.PluginOnlyData,
+		}
+		if err := includable.LoadRecursive(lockfile); err != nil {
+			return errors.WithStack(err)
+		}
+		pluginData := builtIn.PluginOnlyData
+		includable.pluginData = &pluginData
```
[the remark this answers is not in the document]
mikeland73: This looks weird. This may be a mistake.

internal/devconfig/configfile/file.go
mikeland73: plugin configs don't use hujson (but will in the future). Scope of this change is already too big to include that moving part.

internal/devpkg/package.go
mikeland73: This is needed because unfortunately devbox CLI calls `devbox.Open` when populating autocomplete for scripts. (https://github.com/jetpack-io/devbox/blob/main/internal/boxcli/run.go#L54-L54). Note that this code happens before `ensureNixInstalled` is called. Arguably, we don't want to require nix for stuff like listing scripts which doesn't need nix.

So `devbox.Open` cannot call anything requiring nix, otherwise all commends will fail. Without this change, just parsing a devbox.json requires nix. 

We should probably replace `command.ValidArgs` with `command.ValidArgsFunction` but even then, I think it's useful to keep `devbox.Open` working without nix.

internal/plugin/manager.go
```diff
@@ -49,37 +43,3 @@ func (m *Manager) ApplyOptions(opts ...managerOption) {
 		opt(m)
 	}
 }
-
```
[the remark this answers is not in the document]
mikeland73: After this PR it does:

* Builds includables and creates files
* Figures out what services a plugin has.

We may want to split those out. Maybe the services stuff goes into the services package and the includable stuff moves into devconf

internal/plugin/plugin.go
```diff
@@ -35,19 +34,22 @@ var (
...
 
-	DeprecatedDescription string `json:"readme"`
+type PluginOnlyData struct {
+	CreateFiles           map[string]string `json:"create_files"`
+	DeprecatedDescription string            `json:"readme"`
 	// If true, we remove the package that triggered this plugin from the environment
 	// Useful when we want to replace with flake
 	RemoveTriggerPackage bool   `json:"__remove_trigger_package,omitempty"`
 	Version              string `json:"version"`
+	Source               Includable
```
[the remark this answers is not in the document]
mikeland73: will do

internal/plugin/reflike.go
```diff
@@ -13,23 +13,25 @@ import (
...
 
 type Includable interface {
 	CanonicalName() string
 	Hash() string
 	FileContent(subpath string) ([]byte, error)
+	LockfileKey() string
 }
 
-func parseReflike(s string) (Includable, error) {
+func parseReflike(s, projectDir string) (Includable, error) {
```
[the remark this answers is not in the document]
mikeland73: Will do as part of [cleanup](https://github.com/jetpack-io/devbox/pull/1867)

mikeland73:
> is there a way to disable a builtin plugin by having an includable plugin supersede it? For example, the builtin python plugin may be doing something a user feels is undesirable and would like to either disable it completely, or override it with their own python plugin

Currently you have to use `--disable-plugin` when adding a package and then add your own plugin. It's not perfect, but seems OK since this is a bit of an edge case.

> A concern with this design is that scripts must be fully embedded in the plugin's devbox.json file, rather than moved into their own my_script.sh file where they can be properly handled by tooling (e.g. syntax-hightlighting, linting, etc.). Enabling that scenario would be a big win IMO.

> One idea for a follow up PR is to add special-case support for detecting script files in plugin-configs, and downloading them as well.

This use case is doable, but requires a bit more work from the user.

If you want script files they must be copied over using `create_files` field. Copying over all files is a bit tricky because we need to figure out the file structure of the repo (we can do this with `git` or maybe `github` has an API, not sure. Either way, a decent amount of more work).

internal/devconfig/config.go
```diff
@@ -1,39 +1,192 @@
...
 
-// Load reads a devbox config file, and validates it.
-func Load(path string) (*Config, error) {
+func IsNotDefault(path string) bool {
+	cfg, err := readFromFile(path)
+	if err != nil {
+		return false
+	}
+	return !cfg.Root.Equals(&DefaultConfig().Root)
+}
```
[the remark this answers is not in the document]
mikeland73: will do in [cleanup](https://github.com/jetpack-io/devbox/pull/1867).

internal/devconfig/config.go
```diff
@@ -1,39 +1,192 @@
...
+		}
+		if err := includable.LoadRecursive(lockfile); err != nil {
+			return errors.WithStack(err)
+		}
+
+		included = append(included, includable)
+	}
+
+	builtIns, err := plugin.GetBuiltinsForPackages(
+		c.Root.PackagesMutator.Collection,
```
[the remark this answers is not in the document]
mikeland73: Never mind, this is used outside of the configfile package.

I'm adding `SingleFilePackages` and will make `Collection` unexported.

mikeland73:
@savil made requested changes (here and in https://github.com/jetpack-io/devbox/pull/1867). PTAL

internal/devconfig/config.go
```diff
@@ -73,11 +70,15 @@ func readFromFile(path string) (*Config, error) {
 	return loadBytes(b)
 }
 
-func LoadConfigFromURL(url string) (*Config, error) {
-	res, err := http.Get(url)
+func LoadConfigFromURL(ctx context.Context, url string) (*Config, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodHead, url, nil)
```
savil, line 74: this should be `http.MethodGet`, right?

On line 84, we are trying to read `res.Body` which `HEAD` requests don't return

internal/devconfig/configfile/file.go
```diff
@@ -130,7 +130,13 @@ func (c *ConfigFile) GetPackage(versionedName string) (*Package, bool) {
 	if i == -1 {
 		return nil, false
 	}
-	return &c.PackagesMutator.Collection[i], true
+	return &c.PackagesMutator.collection[i], true
+}
+
+// SingleFilePackages returns the packages in the config file, but not the included ones.
+// Semi-awkwardly named to avoid confusion with the Packages method on Config.
+func (c *ConfigFile) SingleFilePackages() []Package {
```
savil, line 138: `TopLevelPackages` ?

internal/devconfig/configfile/packages.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 type PackagesMutator struct {
-	// Collection contains the set of package definitions
-	Collection []Package
+	// collection contains the set of package definitions
+	collection []Package
```
savil, line 20: 👍🏾

gcurtis (approved):

internal/devconfig/config.go
mikeland73: oof, good catch. Terrible copy pasta on my part

PR #1860, opened 2024-02-28
Add builtin plugin for nodejs + Corepack support

### LINKED ISSUE #1577: [Feature]: provide a standard way to install yarn that uses configured nodejs
**Is your feature request related to a problem you're trying to solve? Please describe.**

I'd like a way to install `yarn` that uses the version of nodejs configured in `devbox.json`.

I'm not sure if this could be considered a documentation bug or not, but the [documented way](https://www.jetpack.io/devbox/docs/devbox_examples/languages/nodejs/#adding-yarn-as-your-package-manager) of installing yarn does not use the configured version of node, which can lead in some pretty surprising results.

```
$ devbox version  
0.7.0
$ devbox init
$ devbox add nodejs@20.8 yarn
# ...
$ devbox run node --version
v20.8.0
$ devbox run yarn node --version
yarn node v1.22.19
v18.18.0
✨  Done in 0.03s.
$ devbox rm yarn
$ devbox add nodePackages.yarn
# ...
$ devbox run yarn node --version
yarn node v1.22.19
v18.18.0
✨  Done in 0.03s.
```

**Describe the solution you'd like**

Ideally it would be the default when installing the `yarn` package, but that's not the way nix works under the hood. I don't know if devbox could create/replace some shims since that would alter the nix behavior.

**Describe alternatives you've considered**

As a workaround, I've set up environment variables for a global npm installation:

```
  "env": {
    "NPM_CONFIG_PREFIX": "$PWD/.npmglobal",
    "PATH": "$PWD/node_modules/.bin:$PWD/.npmglobal/bin:$PATH"
  },
```

That works, but requires the user to `npm install -g yarn` first.

**Additional context**
Add any other context or screenshots about the feature request here. Has the feature been requested before? If so, provide a link to the issue.

Lagoja:
## Summary

Builtin version of the plugin built by @jasononeil to fix #1577. This plugin configures Corepack to work with Devbox, so developers can install pnpm + nodejs via corepack.

This will require updates to documentation and examples for Nodejs as well. 

## How was it tested?

* Create an empty Devbox project, and add `nodejs@latest`. 
* You should see the readme from the plugin, describing how to use Corepack
* Create a `package.json`, and add ` "packageManager": "pnpm@8.15.3"` or "packageManager": "yarn@1.22.21"`
* Run `pnpm -v` or `yarn -v` and confirm the version matches what's in your package.json

### DIFF
--- a/plugins/builtins.go
+++ b/plugins/builtins.go
@@ -30,6 +30,7 @@ var builtInMap = map[*regexp.Regexp]string{
 	regexp.MustCompile(`^(ghc|haskell\.compiler\.(.*))$`):              "haskell",
 	regexp.MustCompile(`^mariadb(-embedded)?_?[0-9]*$`):                "mariadb",
 	regexp.MustCompile(`^mysql?[0-9]*$`):                               "mysql",
+	regexp.MustCompile(`^nodejs(-slim)?_?[0-9]*$`):                     "nodejs",
 	regexp.MustCompile(`^php[0-9]*$`):                                  "php",
 	regexp.MustCompile(`^python3[0-9]*Packages.pip$`):                  "pip",
 	regexp.MustCompile(`^(\w*\.)?poetry$`):                             "poetry",
--- a/plugins/nodejs.json
+++ b/plugins/nodejs.json
@@ -1,0 +1,15 @@
+{
+    "$schema": "https://raw.githubusercontent.com/jetpack-io/devbox/main/.schema/devbox-plugin.schema.json",
+    "version": "0.0.1",
+    "name": "nodejs",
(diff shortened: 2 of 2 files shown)


mikeland73 (approved):
Sweet!

jasononeil:
Exciting!

I just wanted to check that any explicit use of `yarn` or `pnpm` packages from the nix store still take precedence over this, and they do (I guess because the user env loads after the plugin env)

Lagoja:
> Exciting!
> 
> I just wanted to check that any explicit use of `yarn` or `pnpm` packages from the nix store still take precedence over this, and they do (I guess because the user env loads after the plugin env)

I think that should override the corepack installed binaries on the path, since as you say, the user env overrides the plugin env

Issue #1849, opened 2024-02-27
.devbox/bin/devbox symlink appears to be wrong?


johnliu:
### What happened?

I discovered this somewhat accidentally but I had a project that used the biome linter. I kept getting a warning about a broken symlink in .devbox/bin/devbox. It seemed that caused devbox to not be functional in a pure shell:

```shell
$ devbox shell --pure
Starting a devbox shell...
(devbox) bash-5.2$ devbox
bash: devbox: command not found
(devbox) bash-5.2$
```

Fixing this symlink to seemed to resolve the issue so I'm assuming the symlink isn't correctly created?

```shell
$ cd .devbox/bin
$ ln -s $(which devbox) devbox
$ cd ../..
$ devbox shell --pure
Starting a devbox shell...
(devbox) bash-5.2$ devbox
Instant, easy, predictable development environments

Usage:
  devbox [flags]
  devbox [command]

Available Commands:
  add         Add a new package to your devbox
  auth        Devbox auth commands
  completion  Generate the autocompletion script for the specified shell
  create      Initialize a directory as a devbox project using a template
  generate    Generate supporting files for your project
  global      Manage global devbox packages
  help        Help about any command
  info        Display package info
  init        Initialize a directory as a devbox project
  install     Install all packages mentioned in devbox.json
  rm          Remove a package from your devbox
  run         Run a script or command in a shell with access to your packages
  search      Search for nix packages
  services    Interact with devbox services.
  shell       Start a new shell with access to your packages
  shellenv    Print shell commands that add Devbox packages to your PATH
  update      Update packages in your devbox
  version     Print version information

Flags:
  -h, --help    help for devbox
  -q, --quiet   suppresses logs

Use "devbox [command] --help" for more information about a command.
(devbox) bash-5.2$
```

### Steps to reproduce

1. Create a fresh devbox (`devbox init`).
2. Enter a pure shell (`devbox shell --pure`).
3. Try using devbox within the pureshell.

### Command

shell

### devbox.json

```jsonc
{
  "packages": [],
  "shell": {
    "init_hook": [
      "echo 'Welcome to devbox!' > /dev/null"
    ],
    "scripts": {
      "test": [
        "echo \"Error: no test specified\" && exit 1"
      ]
    }
  }
}
```

### Devbox version

0.9.1

### Nix version

nix (Nix) 2.18.1

### What system does this bug occur on?

macOS (Apple Silicon)

johnliu:
[...]

PR #1861, opened 2024-02-28
Fixes a bug where devbox symlink isn't recognized in some systems

mohsenari:
## Summary
Symlink in pure shell seemed to be broken in some systems. This fixes the issue.

Addressed #1849 
## How was it tested?
devbox init
devbox shell --pure
devbox


savil (approved):

gcurtis:
Thanks! This explains why I was seeing weird symlinks. Previously if you ran something like `./devbox` then the symlink would point to a relative path that didn't exist.

PR #1862, opened 2024-02-28
internal/shellgen: add trace messages to flake

gcurtis:
Add a few traces to the flake that print the package being downloaded or evaluated. The user should never see these, but it makes debugging the flake easier.


savil (approved):
How do we see these?

Issue #1852, opened 2024-02-27
Curl flake with ^out,dev not adding dynamic libraries


undecided:
### What happened?

Having just worked around #1851 and deleted my /nix and reinstalled everything, suddenly my project is broken thanks to ruby's FFI being unable to find libcurl.

Looks like even with the curl line added, `.devbox/nix/profile/default/lib` (as per `$LD_LIBRARY_PATH`) does not contain libcurl

### Steps to reproduce

1. With `export DEVBOX_USE_VERSION=0.8.7` run the script `devbox run problem` - it works
2. `unset DEVBOX_USE_VERSION` and with the latest devbox, run the same script - it breaks
4. With 0.9.2-dev it also breaks (with new-style glibc-patch and `--output out,dev`, of course!)

### Command

run, shell

### devbox.json

```jsonc
{
  "packages": [
    "ruby@latest",
    "github:nixos/nixpkgs#curl^out,dev"
  ],
  "shell": {
    "scripts": {
      "problem": "gem install typhoeus ; ruby -rtyphoeus -e 'puts :hi'"
    }
  }
}
```

### Devbox version

0.9.1

### Nix version

nix (Nix) 2.18.1

### What system does this bug occur on?

Linux (x86-64)

### Debug logs

```
❯ DEVBOX_USE_VERSION=0.8.7 devbox run problem
Info: New devbox available: 0.8.7 -> v0.9.1. Please run `devbox version update`.
Ensuring packages are installed.

Installing package: curl.

[1/1] curl
[1/1] curl: Success
✓ Computed the Devbox environment.
Successfully installed typhoeus-1.4.1
Parsing documentation for typhoeus-1.4.1
Done installing documentation for typhoeus after 0 seconds
1 gem installed
hi
```

```
❯ DEVBOX_USE_VERSION=0.9.1 devbox run problem
Ensuring packages are installed.
✓ Computed the Devbox environment.
Info: Removing curl@8.6.0
Successfully installed typhoeus-1.4.1
Parsing documentation for typhoeus-1.4.1
Done installing documentation for typhoeus after 0 seconds
1 gem installed
/home/matthew/projects/third_party/devboxplay/.devbox/virtenv/ruby/gems/ffi-1.16.3/lib/ffi/dynamic_library.rb:65:in `load_library': Could not open library 'libcurl': libcurl: cannot open shared object file: No such file or directory. (LoadError)
[...]

PR #1864, opened 2024-02-28
[nix-outputs] Fix flake installable with custom outouts

mikeland73:
## Summary

Fixes https://github.com/jetpack-io/devbox/issues/1852

Ensure we initialize outputs using flake if package is a flake.

Made outputs non-pointer so we don't have to worry about initialization. 

## How was it tested?

Installed `github:nixos/nixpkgs#curl^out,dev` and verified that flake and profile contain both outputs.

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -697,5 +697,10 @@ func (p *Package) GetOutputNames() ([]string, error) {
 	if p.IsRunX() {
 		return []string{}, nil
 	}
+
+	if p.installable.Outputs != "" {
+		return strings.Split(p.installable.Outputs, ","), nil
+	}
+
 	return p.outputs.GetNames(p)
 }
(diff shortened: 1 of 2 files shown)


gcurtis (approved):

internal/devpkg/package.go
```diff
@@ -121,7 +121,7 @@ func PackageFromStringWithOptions(raw string, locker lock.Locker, opts devopt.Ad
 	pkg := PackageFromStringWithDefaults(raw, locker)
 	pkg.DisablePlugin = opts.DisablePlugin
 	pkg.PatchGlibc = opts.PatchGlibc
-	pkg.outputs = initOutputs(opts.Outputs)
+	pkg.initOutputs(opts.Outputs)
```
savil, line 124: line 117 above needs `pkg.initOutputs`, for when the `raw` is a flake-ref-with-flake-outputs

PR #1865, opened 2024-02-28
[flakes] Ensure flake outputs are always used

mikeland73:
## Summary

Ensure flake packages always parse outputs. (re-do of https://github.com/jetpack-io/devbox/pull/1864)

This approach is more general. It also supports (with some small changes) the use of devbox output flags with flakes (e.g. `devbox add github:nixos/nixpkgs#curl --outputs out,dev`)

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -106,7 +106,7 @@ func PackagesFromConfig(config *devconfig.Config, l lock.Locker) []*Package {
 		pkg := newPackage(cfgPkg.VersionedName(), cfgPkg.IsEnabledOnPlatform(), l)
 		pkg.DisablePlugin = cfgPkg.DisablePlugin
 		pkg.PatchGlibc = cfgPkg.PatchGlibc && nix.SystemIsLinux()
-		pkg.initOutputs(cfgPkg.Outputs)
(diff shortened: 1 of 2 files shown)


internal/devpkg/package.go
```diff
@@ -121,7 +121,7 @@ func PackageFromStringWithOptions(raw string, locker lock.Locker, opts devopt.Ad
 	pkg := PackageFromStringWithDefaults(raw, locker)
 	pkg.DisablePlugin = opts.DisablePlugin
 	pkg.PatchGlibc = opts.PatchGlibc
-	pkg.initOutputs(opts.Outputs)
+	pkg.outputs.selectedNames = lo.Uniq(append(pkg.outputs.selectedNames, opts.Outputs...))
```
savil, line 124: this feels brittle code for longer term, but okay for now

savil (approved):

savil:
thanks!

internal/devpkg/package.go
mikeland73: It's a weird, but I thought it was simplest. It basically merges and dedups outputs that come from the flake and the outputs field in devbox.json packages. (which I think is reasonable?) the alternative would be an error no? 

I mostly didn't want to make a call on what to do if both are set. Currently flakes don't support the `--outputs` flag, but I think that's a bug/limitation we can fix.

PR #1866, opened 2024-02-28
[outputs lockfile] make backwards compatible

savil:
## Summary

This PR enables a `devbox.lock` updated by a user with the latest Devbox that has written Outputs to continue working for their teammate using a legacy Devbox.

Note two differences:
1. The legacyDevbox teammate will not benefit from output store-paths
2. If the legacyDevbox teammate modifies the lockfile in any way, then the Outputs will be lost. Presumably, this is not a huge problem since I anticipate most users rely on source control to track their `devbox.lock` in their devbox projects (except for maybe the Devbox global, but then those are less likely to be shared).

## How was it tested?

1. Added `devbox add curl` and `devbox add prometheus --outputs cli,out` using compiled devbox. This updated `devbox.lock` file.
2. `rm -rf .devbox`
3. `/usr/local/bin/devbox shell` using released Devbox. Observe that (1) it works (2) modifies `devbox.lock` file and drops `Outputs`.

### DIFF
--- a/internal/lock/lockfile.go
+++ b/internal/lock/lockfile.go
@@ -106,10 +106,8 @@ func (f *File) Save() error {
 	// such as `devbox update` or `devbox add` or `devbox remove`.
 	for pkgName, pkg := range f.Packages {
 		for sys, sysInfo := range pkg.Systems {
-			if !isDirty && sysInfo.StorePath != "" {
+			if !isDirty && sysInfo.outputIsFromStorePath {
 				f.Packages[pkgName].Systems[sys].Outputs = nil
-			} else {
-				f.Packages[pkgName].Systems[sys].StorePath = ""
 			}
 		}
 	}
--- a/internal/lock/lockfile.go
+++ b/internal/lock/lockfile.go
@@ -96,17 +96,31 @@ func (f *File) Resolve(pkg string) (*Package, error) {
 	return f.Packages[pkg], nil
 }
 
+// TODO:
+// Consider a design change to have the File struct match disk to make this system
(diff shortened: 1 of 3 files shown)


savil:
* **#1866** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/jetpack-io/devbox/1866?utm_source=stack-comment).

savil:
@gcurtis @mikeland73 ping

internal/lock/lockfile.go
```diff
@@ -106,10 +106,8 @@ func (f *File) Save() error {
 	// such as `devbox update` or `devbox add` or `devbox remove`.
 	for pkgName, pkg := range f.Packages {
 		for sys, sysInfo := range pkg.Systems {
-			if !isDirty && sysInfo.StorePath != "" {
+			if !isDirty && sysInfo.outputIsFromStorePath {
```
mikeland73, line 109: Can this be just `!isDirty` ? 

Then we can get rid of `outputIsFromStorePath`

mikeland73: Actually is isDirty even working as expected? I would think isDirty is always true when loading a legacy lockfile (because we add the outputs).

mikeland73 (approved):
Approving, but lets have a quite chat before merging

PR #1867, opened 2024-02-29
[cleanup] Remove reflike and tyson

mikeland73:
## Summary

Stacked on https://github.com/jetpack-io/devbox/pull/1850

* Remove RefLike
* Remove experimental tyson support (v2 plugins are a reasonable substitute)

## How was it tested?

CICD


internal/plugin/reflike.go
```diff
@@ -28,20 +18,12 @@ func parseReflike(s, projectDir string) (Includable, error) {
 	if err != nil {
 		return nil, err
 	}
-	reflike := RefLike{ref, pluginConfigName, s}
 	switch ref.Type {
 	case flake.TypePath:
-		return newLocalPlugin(reflike, projectDir)
+		return newLocalPlugin(ref, projectDir)
```
savil, line 23: Could this function and file be renamed to `parseIncludable` and `includable.go`?

nix/flake/flakeref.go
```diff
@@ -67,6 +67,8 @@ type Ref struct {
 	// or "git". Note that the URL is not the same as the raw unparsed
 	// flake ref.
 	URL string `json:"url,omitempty"`
+
+	Raw string `json:"raw,omitempty"`
```
savil, line 71: could you add a comment for what Raw is, alongwith a few examples? 

That will be helpful, and also inline with the nice comments in the rest of the struct :)

savil: Alternatively, you could add a comment to point people to docblock of `parseRef` function

savil (approved):

savil (approved):

gcurtis (approved):

PR #1869, opened 2024-03-01
Fix #1783: Support non-root install in pure mode

### LINKED ISSUE #1783: Support non-root setup

I use nix-portable since I cannot use sudo/su on my work system due to security policy. devbox tries to install on /usr/local/bin, which requires root access. It also tries to install nix as root if the nix command is not available.

Even when `nix` is available via `nix-portable`, some functionalities do not seem to work as expected, such as `devbox shell --pure`:
```
Error: could not find any nix executable in PATH. Make sure Nix is installed and in PATH, then try again
```
`nix` is available as binary in PATH, so the message is wrong, and the problem has been triggered by a different issue, probably.

I modified the setup script to target ~/.local/bin and linked nix-portable to nix. It would be nice to be able to specify a different installation path and use sudo only whenever it is needed. Giving proper documentation on how to install devbox with nix-portable would be a plus.

_No response_

kadaan:
## Summary
Fix #1783: Support non-root install in pure mode

## How was it tested?
Ran `devbox shell` and `devbox shell --pure` and they both work now with my non-root install of nix on MacOS.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -1262,12 +1262,11 @@ func (d *Devbox) parseEnvAndExcludeSpecialCases(currentEnv []string) (map[string
 		// Finding nix executables in path and passing it through
 		// As well as adding devbox itself to PATH
 		// Both are needed for devbox commands inside pure shell to work
-		includedInPath, err := findNixInPATH(env)
+		nixPath, err := findNixInPATH()
 		if err != nil {
 			return nil, err
 		}
-		includedInPath = append(includedInPath, dotdevboxBinPath(d))
-		env["PATH"] = envpath.JoinPathLists(includedInPath...)
+		env["PATH"] = envpath.JoinPathLists(nixPath, dotdevboxBinPath(d))
 	}
 	return env, nil
 }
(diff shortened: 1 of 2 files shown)


savil:
thanks for sending this PR as well!

internal/devbox/devbox.go
```diff
@@ -1262,12 +1262,11 @@ func (d *Devbox) parseEnvAndExcludeSpecialCases(currentEnv []string) (map[string
 		// Finding nix executables in path and passing it through
 		// As well as adding devbox itself to PATH
 		// Both are needed for devbox commands inside pure shell to work
-		includedInPath, err := findNixInPATH(env)
+		nixPath, err := findNixInPATH()
```
gcurtis, line 1265: If `findNixInPATH` is only ever called with `os.Environ()`, then I think we should just remove it and simplify this down to:

```suggestion
		nixPath, err := exec.LookPath("nix")
		if err != nil {
				return "", errors.New("could not find any nix executable in PATH. Make sure Nix is installed and in PATH, then try again")
		}
		nixPath = filepath.Dir(nixPath)
```

kadaan:
@gcurtis Any other feedback?

gcurtis (approved):
Thanks for the changes! This looks good. I'll merge after tests finish.

savil (approved):

kadaan:
Thanks for the reviews and merging!

PR #1870, opened 2024-03-01
Fix #1868: Fix flake handling

### LINKED ISSUE #1868: devbox update fail when git flake are present
### What happened?

Running `devbox update` fails when git flakes are present in the `devbox.json`.

### Steps to reproduce

1.  Add git flake to devbox.json
2. Run devbox shell
3. Exit shell
4. Run devbox update

### Command

_No response_

### devbox.json

```jsonc
{
  "packages": [
    "git+ssh://git@bitbucket.org/foo/bar.git#baz"
  ],
}
```

### Devbox version

0.9.1

### Nix version

2.19.3

macOS (Apple Silicon)

### Debug logs

2024/02/29 17:53:38 findProjectDir: path is
2024/02/29 17:53:38 finding devbox config in dir: /source/work/joelb_test
2024/02/29 17:53:38 findProjectDir: path is
2024/02/29 17:53:38 finding devbox config in dir: /source/work/joelb_test
2024/02/29 17:53:38 findProjectDir: path is
2024/02/29 17:53:38 finding devbox config in dir: /source/work/joelb_test
Error: git+ssh://git@bitbucket.org/foo/bar.git#baz: package not found

2024/02/29 17:53:39
ExecutionID:2323d1a6d79f4977814fa853ea181384
package not found
go.jetpack.io/devbox/internal/nix.init
	<autogenerated>:1
runtime.doInit1
	/nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/runtime/proc.go:7176
runtime.doInit
	/nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/runtime/proc.go:7143
runtime.main
	/nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/runtime/proc.go:253
runtime.goexit
	/nix/store/2022s0jnrn2iyxjaikfy51w5fvifp38b-go-1.22.0/share/go/src/runtime/asm_arm64.s:1222
2024/02/29 17:53:39 findProjectDir: path is
2024/02/29 17:53:39 finding devbox config in dir: /source/work/joelb_test

kadaan:
## Summary
`devbox update` does not work with git flakes.  This adds support for git flake to `devbox update` and also ensures that the nix packages are rebuilt when an update is requested.  Without this, git flakes will not be refreshed from git.

## How was it tested?

1. Configured a devbox.json with devbox packages and flakes
2. Ran devbox shell without a .devbox folder and without packages being installed in nix
3. Saw packages get installed 
4. Ran devbox shell with a .devbox folder and with packages being installed in nix
5. Saw packages verified, but not updated or installed
6. Ran devbox update
7. Saw packages get rebuild with `nix build --refresh`

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -486,18 +486,18 @@ func (d *Devbox) packagesToInstallInStore(ctx context.Context, mode installMode)
 		for _, installable := range installables {
 			if mode == update {
 				packagesToInstall = append(packagesToInstall, pkg)
-			} else {
-				storePaths, err := nix.StorePathsFromInstallable(ctx, installable, pkg.HasAllowInsecure())
-				if err != nil {
-					return nil, packageInstallErrorHandler(err, pkg, installable)
-				}
-				isInStore, err := nix.StorePathsAreInStore(ctx, storePaths)
-				if err != nil {
-					return nil, err
-				}
-				if !isInStore {
-					packagesToInstall = append(packagesToInstall, pkg)
-				}
+				continue
+			}
+			storePaths, err := nix.StorePathsFromInstallable(ctx, installable, pkg.HasAllowInsecure())
+			if err != nil {
+				return nil, packageInstallErrorHandler(err, pkg, installable)
+			}
+			isInStore, err := nix.StorePathsAreInStore(ctx, storePaths)
+			if err != nil {
+				return nil, err
+			}
+			if !isInStore {
+				packagesToInstall = append(packagesToInstall, pkg)
 			}
 		}
 	}
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -260,7 +260,7 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 	}
 
 	if mode == install || mode == update || mode == ensure {
-		if err := d.installPackages(ctx); err != nil {
+		if err := d.installPackages(ctx, mode); err != nil {
 			return err
 		}
 	}
@@ -384,7 +384,7 @@ func resetProfileDirForFlakes(profileDir string) (err error) {
 	return errors.WithStack(os.Remove(profileDir))
 }
 
-func (d *Devbox) installPackages(ctx context.Context) error {
+func (d *Devbox) installPackages(ctx context.Context, mode installMode) error {
 	// Create plugin directories first because packages might need them
 	for _, pkg := range d.InstallablePackages() {
 		if err := d.PluginManager().Create(pkg); err != nil {
@@ -392,7 +392,7 @@ func (d *Devbox) installPackages(ctx context.Context) error {
 		}
 	}
 
-	if err := d.installNixPackagesToStore(ctx); err != nil {
+	if err := d.installNixPackagesToStore(ctx, mode); err != nil {
 		return err
 	}
 
@@ -420,8 +420,8 @@ func (d *Devbox) InstallRunXPackages(ctx context.Context) error {
 // This is done by running `nix build` on the flake. We do this so that the
 // packages will be available in the nix store when computing the devbox environment
 // and installing in the nix profile (even if offline).
-func (d *Devbox) installNixPackagesToStore(ctx context.Context) error {
-	packages, err := d.packagesToInstallInStore(ctx)
+func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode) error {
+	packages, err := d.packagesToInstallInStore(ctx, mode)
 	if err != nil {
 		return err
 	}
@@ -439,12 +439,17 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context) error {
 			return err
 		}
 
+		// --no-link to avoid generating the result objects
+		flags := []string{"--no-link"}
+		if mode == update {
+			flags = append(flags, "--refresh")
+		}
+
 		for _, installable := range installables {
 			args := &nix.BuildArgs{
 				AllowInsecure: pkg.HasAllowInsecure(),
-				// --no-link to avoid generating the result objects
-				Flags:  []string{"--no-link"},
-				Writer: d.stderr,
+				Flags:         flags,
+				Writer:        d.stderr,
 			}
 			err = nix.Build(ctx, args, installable)
 			if err != nil {
@@ -460,7 +465,7 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context) error {
 	return err
 }
 
-func (d *Devbox) packagesToInstallInStore(ctx context.Context) ([]*devpkg.Package, error) {
+func (d *Devbox) packagesToInstallInStore(ctx context.Context, mode installMode) ([]*devpkg.Package, error) {
 	// First, get and prepare all the packages that must be installed in this project
 	packages, err := d.AllInstallablePackages()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


savil:
@kadaan thanks for sending this PR! We'll try to review it later today

nix/flake/flakeref.go
```diff
@@ -103,10 +103,7 @@ func ParseRef(ref string) (Ref, error) {
 		parsed.Path = ref
 		return parsed, nil
 	}
-	parsed, fragment, err := parseURLRef(ref)
-	if fragment != "" {
-		return Ref{}, redact.Errorf("flake reference %q contains a URL fragment", ref)
-	}
```
gcurtis, line 109: Take a look at `flake.Installable`, which I think does what you need. Flake references can't have attribute paths.

kadaan:
@savil No problem.  Looks like go-lint is failing.  Maybe I missed something on my end in preparing these PRs.  Can you point me to the steps I should run to ensure that the code is formatted, linted, etc. to your standards?

internal/devpkg/pkgtype/flake.go
```diff
@@ -0,0 +1,20 @@
+package pkgtype
+
+import (
+	"go.jetpack.io/devbox/nix/flake"
+	"strings"
+)
+
+func IsFlake(s string) bool {
```
gcurtis, line 8: I think you can reuse `devpkg.isAmbiguous` here.

In some cases it's impossible to know if a package is a flake without resolving it (for example, `foo` could be an indirect flake's default package or it could be a Devbox package named `foo`).

internal/lock/resolve.go
```diff
@@ -28,6 +28,10 @@ import (
 // a newer hash than the lock file but same version. In that case we don't want
 // to update because it would be slow and wasteful.
 func (f *File) FetchResolvedPackage(pkg string) (*Package, error) {
+	if pkgtype.IsFlake(pkg) {
```
gcurtis, line 31: You'll get some false negatives here, but maybe that's okay? You can know for sure if something is not a Devbox package, but you can't always know for certain if it is a flake.

internal/devpkg/package.go
```diff
@@ -238,7 +238,7 @@ func (p *Package) FlakeInputName() string {
 			}
 		}
 	default:
-		result = p.installable.String() + "-" + p.Hash()
+		result = p.installable.Ref.URL + "-" + p.Hash()
```
gcurtis, line 241: `Ref.URL` isn't quite the same as the flakeref in URL form. It's only set when a flakeref has the type tarball/file/git. You can use `Ref.String()` to get the ref without the attribute path:

```suggestion
		result = p.installable.Ref.String() + "-" + p.Hash()
```

internal/devbox/packages.go
```diff
@@ -479,16 +484,20 @@ func (d *Devbox) packagesToInstallInStore(ctx context.Context) ([]*devpkg.Packag
...
+					return nil, packageInstallErrorHandler(err, pkg, installable)
+				}
+				isInStore, err := nix.StorePathsAreInStore(ctx, storePaths)
+				if err != nil {
+					return nil, err
+				}
+				if !isInStore {
+					packagesToInstall = append(packagesToInstall, pkg)
+				}
 			}
```
gcurtis, line 501: Replace else block with continue:

```
if mode == update {
	packagesToInstall = append(packagesToInstall, pkg)
	continue
}
```

Lagoja:
> @savil No problem. Looks like go-lint is failing. Maybe I missed something on my end in preparing these PRs. Can you point me to the steps I should run to ensure that the code is formatted, linted, etc. to your standards?

You can run `devbox run lint` in the project root to automatically run our linting + formatting

gcurtis (commented):
Thanks for the PR. I have some feedback on the flake parsing portion.

internal/devpkg/pkgtype/flake.go
kadaan: A couple issue here:
1. devpkg.isAmbiguous would need to have it's visibility changed
2. devpkg.isAmbiguous would need to move to a different package because using devpkg in the lock package causes an import cycle

nix/flake/flakeref.go
kadaan: The code that is deleted here was added as part of https://github.com/jetpack-io/devbox/commit/1dbcad274cddf480a6fab062dee2649a3b726112.  

It seems odd to throw an error when parsing a reference given that a fragment is a valid part of a flake url.  Leaving this code here causes

internal/lock/resolve.go
kadaan: Can you give an example of a package that would be a false negative?

nix/flake/flakeref.go
gcurtis: Fragments aren't valid in a flake reference, but they are valid in an "installable". They're slightly different, but the distinction is important in some cases. For example:

```
$ nix eval --expr 'builtins.parseFlakeRef "nixpkgs#hello"'
error:
       … while calling the 'parseFlakeRef' builtin

         at «string»:1:1:

            1| builtins.parseFlakeRef "nixpkgs#hello"
             | ^

       error: unexpected fragment 'hello' in flake reference 'nixpkgs#hello'
```

internal/lock/resolve.go
gcurtis: For example, `cachix` is both a Devbox package name and a flake ref:

```
$ devbox search cachix
Found 4+ results for "cachix":

* cachix  (1.7.1, 1.7, 1.6.1, 1.6, 1.5, 1.3.3, 1.4.2, 1.3.1, 1.3, 1.2)
```

```
$ nix flake metadata cachix
Resolved URL:  github:cachix/cachix
Locked URL:    github:cachix/cachix/3a6e4a049f7cadd8894015c2488a2fa969c83c41
```

When we get an ambiguous package name like that, Devbox's behavior is to assume it's a package first and fallback to treating it like a flake second.

internal/devpkg/pkgtype/flake.go
kadaan: @gcurtis One issue before this PR is that `resolve.FetchResolvedPackage` would call `searcher.ParseVersionedPackage` and that would incorrectly identify `git+ssh://git@foo/bar/baz.git#boo` as a versioned package due to the `@`.  Filtering out known Flakes and leaving ambiguous Flakes is probably good enough at this point.

kadaan: I switched to using `ParseInstallable` and `IsAmbiguous` in `pkgtype.IsFlake`.  To do so, I moved `IsAmbiguous` to package `pkgtype`.

internal/lock/resolve.go
kadaan: Perfect.  TY

kadaan:
@gcurtis Any other feedback?

gcurtis (approved):
This looks good to me. @savil or @mikeland73 I dunno if you guys want to double check the update logic.

savil:
@gcurtis will take a look and merge if all good

savil:
Looks awesome! Did some sanity testing with non-flakes to exercise the `nix build --refresh` flag to verify it doesn't do anything unexpected.

Thank you @kadaan 🙏🏾 !

kadaan:
@savil You are welcome.  Thanks for the reviews and merging!

PR #1877, opened 2024-03-05
[plugins] Fix relative plugins, add cycle detection

mikeland73:
## Summary

TSIA

## How was it tested?

Added more tests via examples.

Tested cycle manually. (including `./..` in plugin1a)

image


examples/plugins/v2-local/devbox.json
```diff
@@ -11,10 +11,14 @@
...
         "test -n \"$PLUGIN1_INIT_HOOK\" || exit 1",
+        "test -n \"$PLUGIN1A_INIT_HOOK\" || exit 1",
         "test -n \"$PLUGIN1_ENV\" || exit 1",
+        "test -n \"$PLUGIN1A_ENV\" || exit 1",
         "if [ \"$PLUGIN1_ENV2\" != \"override\" ]; then exit 1; fi;",
         "devbox run plugin_1_script",
-        "hello"
+        "devbox run plugin_1A_script",
+        "hello",
+        "cowsay 'Hello, world!'"
```
savil, line 21: Would you mind adding some commentary to the lines in this file? Its unclear what scenarios are being demonstrated and vouched for by the various plugins. I can see some combination of:
1. relative, path:, plugin-with-json-file specified in `include` below.
2. plugins having different env-vars (aside from name, are there other differences?)
3. some plugin (which one?) introducing a `cowsay` package into the devbox project
4. plugins introducing scripts that can be `devbox run`

Since these are listed under `examples/`, we should make them user-friendly to understand because its reasonable that they'll browse the devbox projects inside `examples/` for inspiration and guidance on organizing their own projects.

internal/devconfig/config.go
```diff
@@ -104,19 +114,36 @@ func loadBytes(b []byte) (*Config, error) {
...
+		pluginConfig, err := plugin.LoadConfigFromInclude(
+			includeRef, lockfile, c.absRootPath)
 		if err != nil {
 			return errors.WithStack(err)
 		}
 
+		if seen[pluginConfig.Source.Hash()] {
+			return errors.Errorf("circular include detected: %s", includeRef)
+		}
+		seen[pluginConfig.Source.Hash()] = true
```
savil, line 136: For ease of debugging, one could improve the error message by showing the cyclical-path-of-includes. May need to track a bit more state...but feels doable.

internal/devconfig/config.go
```diff
@@ -104,19 +114,36 @@ func loadBytes(b []byte) (*Config, error) {
...
 		includable := &Config{
 			Root:       pluginConfig.ConfigFile,
 			pluginData: &pluginConfig.PluginOnlyData,
 		}
-		if err := includable.LoadRecursive(lockfile); err != nil {
+		if hasDir, ok := pluginConfig.Source.(*plugin.LocalPlugin); ok {
+			includable.absRootPath = filepath.Dir(hasDir.Path())
+		}
+
+		if err := includable.loadRecursive(lockfile, maps.Clone(seen)); err != nil {
```
savil, line 146: why do we need to clone the `seen` map?

Would recommend adding the explanation as a comment, since non-obvious to me. 

EDIT: since we do `maps.Clone` in multiple places, maybe the comment belongs in the function docblock of `loadRecursive` as a warning or note

internal/devconfig/config.go
```diff
@@ -104,19 +114,36 @@ func loadBytes(b []byte) (*Config, error) {
...
+		seen[pluginConfig.Source.Hash()] = true
+
 		includable := &Config{
 			Root:       pluginConfig.ConfigFile,
 			pluginData: &pluginConfig.PluginOnlyData,
 		}
-		if err := includable.LoadRecursive(lockfile); err != nil {
+		if hasDir, ok := pluginConfig.Source.(*plugin.LocalPlugin); ok {
+			includable.absRootPath = filepath.Dir(hasDir.Path())
+		}
```
savil, line 144: the name `hasDir` indicates to me that its type is boolean. Could we call it `localPlugin` instead?

Also consider encapsulating lines 138-144 into a function such as: `initIncludableFromPlugin(pluginConfig)`

savil (commented):
Nice! Some comments and questions inline.

Consider adding a testscript for the cycle plugin scenario:

```
! devbox run script-that-triggers-cycle
! stderr 'circular include detected:'
```

savil:
some tests also appear to be failing

internal/devconfig/config.go
mikeland73: maps are reference types so if we don't clone them the same map is shared in all calls of this function (which would return false positives because it would find instances where the same plugin is imported by different paths)

internal/devconfig/config.go
```diff
@@ -104,19 +114,36 @@ func loadBytes(b []byte) (*Config, error) {
...
 
 	for _, includeRef := range c.Root.Include {
-		pluginConfig, err := plugin.LoadConfigFromInclude(includeRef, lockfile)
+		pluginConfig, err := plugin.LoadConfigFromInclude(
+			includeRef, lockfile, c.absRootPath)
 		if err != nil {
 			return errors.WithStack(err)
 		}
 
+		if seen[pluginConfig.Source.Hash()] {
```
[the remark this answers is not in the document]
mikeland73: Yes this is allowed. (I don't see any issues with it and disallowing it would make composing tricker) 

what is this graph magic.

internal/devconfig/config.go
```diff
@@ -114,36 +110,40 @@ func loadBytes(b []byte) (*Config, error) {
...
 		}
 
+		newCyclePath := fmt.Sprintf("%s -> %s", cyclePath, includeRef)
 		if seen[pluginConfig.Source.Hash()] {
-			return errors.Errorf("circular include detected: %s", includeRef)
+			// Note that duplicate includes are allowed if they are in different paths
+			// e.g. 2 different plugins can include the same plugin.
+			// We do not allow a single plugin to include duplicates.
+			return errors.Errorf(
+				"circular or duplicate include detected: %s", newCyclePath)
```
savil, line 139: do we need a newline after the colon, as in:
```
"circular or duplicate include detected:\n%s", newCyclePath)
```

testscripts/plugin/plugin.cycle.test.txt
```diff
@@ -0,0 +1,51 @@
+! exec devbox install
+stderr 'circular or duplicate include detected:'
+
+! exec devbox install
```
savil, line 4:
```
! exec devbox install -c ./duplicate
```

internal/devconfig/configfile/file.go
```diff
@@ -25,6 +25,10 @@ const (
 
 // ConfigFile defines a devbox environment as JSON.
 type ConfigFile struct {
+	// AbsRootPath is the absolute path to the devbox.json or plugin.json file
```
savil, line 28: To clarify @gcurtis's confusion about this being set to a directory, consider appending `'s directory` at the end:

```
// AbsRootPath is the absolute path to the devbox.json or plugin.json file's directory
```

savil (approved):
nice!

Issue #1788, opened 2024-02-05
Conflicting Python recommendations


frafra:
### What happened?

```
We detected extra packages you may need. To install them, run `devbox add python3 python310 poetry`
```

This is not a good suggestion, since python3 and python310 conflict: `error: An existing package already provides the following file: [...]`

### Steps to reproduce

1. Run `devbox init` in a Python project?

### Command

init

### Devbox version

0.9.0

### Nix version

2.18.0

### What system does this bug occur on?

Linux (x86-64)

Lagoja:
I think most of these recommendations are out of date, and the detection logic isn't very good. We should probably remove them

PR #1890, opened 2024-03-11
[cleanup] Remove init rec feature

mikeland73:
## Summary

This feature is not really maintained and out of date. I think value is fairly limited.

fixes https://github.com/jetpack-io/devbox/issues/1788

## How was it tested?
CICD


savil (approved):
Good

gcurtis (approved):
Nice.

PR #1891, opened 2024-03-12
[nix-cache] Add cache copy command

mikeland73:
## Summary

Note: this may not be the final UX for this command. I'm just getting it in hidden so we can build the cache feature end-to-end and will refine details later.

Adds `devbox cache copy <url>` command. It copies the current devbox nix profile to the url. `<url>` can be anything supported by https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-copy

There are two future plans for this command:

* `url` will be optional when logged in. It will be determined by API in those cases. We may want to move url to be a flag in that case (possibly the `--to` flag to mimic `nix copy`). 
* This command will support specifying the package you wish to copy. Packages can be nix installable, but also devbox packages (i.e. package@version). We could even support runx in the future.

I used `copy` to mimic `nix` command name, but I slightly prefer `upload`. That and the above changes would make this command look like:

`devbox cache upload [--to <url>] [package]`

or as simple as 

`devbox cache upload`

to upload current profile when logged in.

## How was it tested?

`devbox cache copy "s3://mike-test-nix-cache?region=us-west-2"`

### DIFF
--- a/internal/boxcli/cache.go
+++ b/internal/boxcli/cache.go
@@ -23,9 +23,8 @@ func cacheCmd() *cobra.Command {
 	}
 
 	copyCommand := &cobra.Command{
-		Use:   "copy <url>",
-		Short: "Copy nix store paths to the cache",
-		Long:  "Copies all nix packages in current project to url.",
+		Use:   "copy <uri>",
+		Short: "Copies all nix packages in current project to the cache at <uri>",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			box, err := devbox.Open(&devopt.Opts{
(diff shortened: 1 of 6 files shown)


internal/boxcli/cache.go
```diff
@@ -0,0 +1,48 @@
...
+	cacheCommand := &cobra.Command{
+		Use:               "cache",
+		Short:             "Collection of commands to interact with nix cache",
+		PersistentPreRunE: ensureNixInstalled,
+	}
+
+	copyCommand := &cobra.Command{
+		Use:   "copy <url>",
+		Short: "Copy nix store paths to the cache",
+		Long:  "Copies all nix packages in current project to url.",
```
savil, line 28: The long and the short mean quite different things. Which is it?

savil: Could both be: `"Copies all nix packages in current project to the cache at <url>"`?

savil (approved):
The PR description plan makes sense. I like the suggested changes.

Issue #1848, opened 2024-02-27
Plugins from the same repo share the same state directories


jay-aye-see-kay:
### What happened?

We have two plugins in a public but for internal use repo here https://github.com/cultureamp/devbox-extras/tree/main/plugins and noticed that their value of `{{ .Virtenv }}` is the same.

So if both of our two plugins are installed the `corepack-bin/` from one and `data/`+`process-compose.yaml` from the other are both in `.devbox/virtenv/cultureamp-devbox-extras/`. I'm not sure what happens if they both try to create a file/dir with the same name.

### Steps to reproduce

1. create a temp devbox project with the `devbox.json` defined below
2. run `ls .devbox/virtenv/cultureamp-devbox-extras/` and see their files are all together in the same dir

### devbox.json

```jsonc
{
  "packages": [
    "nodejs_18@latest",
    "github:cultureamp/devbox-extras#dynamodb_local"
  ],
  "include": [
    "github:cultureamp/devbox-extras?dir=plugins/corepack",
    "github:cultureamp/devbox-extras?dir=plugins/dynamodb_local"
  ]
}
```

### Devbox version

0.9.1

### Nix version

.devbox/virtenv/cultureamp-devbox-extras/

### What system does this bug occur on?

macOS (Apple Silicon)

jay-aye-see-kay:
Suggested fix: update `{{ .DevboxDir }}` and `{{ .Virtenv }}` variables to include the plugin name as well as the repo it comes from.

savil:
thanks for the bug report!
@mikeland73 assigning to you since this is an area you're working on

PR #1897, opened 2024-03-14
[plugins] Fix github canonical name

mikeland73:
## Summary

Fixes https://github.com/jetpack-io/devbox/issues/1848

github canonical names were `owner-repo` but this breaks when there are multiple plugins in a single repo and creates possible conflicts with other plugins.

This renames them to:

`owner.repo.name` 

where `name` is the name in the plugin.json. For backward compatibility we don't require the name to be present, if it's missing the name is the directory (with slashes replaced with hyphens)

## How was it tested?

Installed a few plugins from the same repo and inspected virtenv dir.


internal/plugin/github.go
```diff
@@ -5,22 +5,45 @@ import (
...
 )
 
 type githubPlugin struct {
-	ref flake.Ref
+	ref  flake.Ref
+	name string
+}
+
+// Github only allows alphanumeric, hyphen, underscore, and period in repo names.
+// but we clean up just in case.
```
savil, line 23: why is the "but we clean up just in case" comment needed? The regex checks for the all the cases that Github allows for, which is good. Not understanding the "but..." as if we are doing some further restrictions?

internal/plugin/github.go
```diff
@@ -5,22 +5,45 @@ import (
...
+var githubNameRegexp = regexp.MustCompile("[^a-zA-Z0-9-_.]+")
+
+func newGithubPlugin(ref flake.Ref) *githubPlugin {
+	plugin := &githubPlugin{ref: ref}
+	// For backward compatibility, we don't strictly require name to be present
+	// in github plugins. If it's missing, we just use the directory as the name.
+	name, _ := getPluginNameFromContent(plugin)
+	if name == "" {
+		name = strings.ReplaceAll(ref.Dir, "/", "-")
```
savil, line 32: what happens if the plugin's devbox.json config is in the root of the github repo? Can we default it to "root-folder" or something?

internal/plugin/github.go
```diff
@@ -5,22 +5,45 @@ import (
...
+	}
+	plugin.name = githubNameRegexp.ReplaceAllString(
+		strings.Join(lo.Compact([]string{ref.Owner, ref.Repo, name}), "."),
```
savil, line 35: is the `lo.Compact` needed? When would one of the ref.Owner, ref.Repo or name be empty?

internal/plugin/includable.go
```diff
@@ -22,8 +24,37 @@ func parseIncludable(includableRef, workingDir string) (Includable, error) {
...
+	if err := json.Unmarshal(content, &m); err != nil {
+		return "", err
+	}
+	name, ok := m["name"].(string)
+	if !ok || name == "" {
+		return "",
+			usererr.New(
+				"plugin %s is missing a required field 'name'", plugin.LockfileKey())
+	}
+	if !nameRegex.MatchString(name) {
```
savil, line 53: nit: lets move `nameRegex` to above this function since its used only here.

internal/plugin/github.go
```diff
@@ -5,22 +5,45 @@ import (
...
+
+// Github only allows alphanumeric, hyphen, underscore, and period in repo names.
+// but we clean up just in case.
+var githubNameRegexp = regexp.MustCompile("[^a-zA-Z0-9-_.]+")
+
+func newGithubPlugin(ref flake.Ref) *githubPlugin {
+	plugin := &githubPlugin{ref: ref}
+	// For backward compatibility, we don't strictly require name to be present
+	// in github plugins. If it's missing, we just use the directory as the name.
+	name, _ := getPluginNameFromContent(plugin)
```
savil, line 30: lets do a followup that checks the error to be `ErrNameOmitted` since right now it conflates malformed name with missing name.

Not blocking because I think the change could be a bit involved since you'd need to propagate the error from this function...

savil (approved):
Approving to unblock

Lagoja:
Tested and confirmed that this works as expected

internal/plugin/github.go
savil: actually, is distinguishing this scenario important for this code flow? I'm now having doubts....

internal/plugin/github.go
mikeland73: I just meant that we should never encounter weird stuff (because it would not be a valid github repo) but we sanitize it anyway just in case. (for example we really don't want a slash in there)

internal/plugin/github.go
mikeland73: If there's no name and it's in the root, the qualified name becomes `owner.repo`

internal/plugin/github.go
mikeland73: `name` can be empty if the plugin has no name and the dir is root.

PR #1898, opened 2024-03-14
Update JSON schema for plugin v2

Lagoja:
## Summary

Update JSON schema for Devbox Plugins


mikeland73 (approved):
Can we have 1 schema include the other? Ideally the plugin schema is the devbox.json plus a few other fields.

Lagoja:
It looks like that's possible: https://json-schema.org/understanding-json-schema/structuring

Can we fix that after the release?

mikeland73:
> Can we fix that after the release?

yep!

Issue #1901, opened 2024-03-15
Devbox 0.10.0 upgrade fails on `process-compose` install due to changed nix


jay-aye-see-kay:
## Description

Upgrade to devbox `0.10.0` and on first `devbox services up` we get the error `nix profile remove": error: 'nix profile' no longer supports indices ('0')` when trying to upgrade process-compose

## A workaround (there might be better ones)

- Find the process-compose package in devbox's nix profile `nix profile list --profile ~/.local/share/devbox/util/profile`
- Note the package's name (for us it was `aarch64-darwin` which seems... wrong?)
- run `nix profile remove --profile  ~/.local/share/devbox/util/profile aarch64-darwin` (replacing `aarch64-darwin` with your package name)

## Other notes

- this didn't happen on a M1 running nix `2.18.1`, the change in cli must have been introduced sometime between that and `2.20.3`

### Steps to reproduce

1. Ensure you have a new-ish `nix` install (was observed on a macOS machine running `2.20.3`)
2. Install devbox `0.9.1` or earlier, run `devbox services up` somewhere to ensure you have `process-compose`
3. upgrade to devbox `0.10.0` and run `devbox services up` somewhere to update to newer `process-compose`
4. See error upgrading `"nix profile remove": error: 'nix profile' no longer supports indices ('0')"`

### Devbox version

0.10.0

### Nix version

2.20.3

### What system does this bug occur on?

macOS (Apple Silicon)

### Debug logs

➜  temp devbox services up
✓ Computed the Devbox environment.
Upgrading process-compose to v0.85.0...
Error: error running "nix profile remove": error: 'nix profile' no longer supports indices ('0')
: exit status 1

Error: There was an internal error. Run with DEVBOX_DEBUG=1 for a detailed error message, and consider reporting it at https://github.com/jetpack-io/devbox/issues
Error: exit status 1

Lagoja:
Looks like a breaking change in 2.20 that wasn't covered in our profile upgrade. Have a PR that i'm testing now.

PR #1902, opened 2024-03-15
Fix process-compose update on nix 2.20

Lagoja:
## Summary
TSIA, fixes an issue with our process-compose update with Nix 2.20. (#1901 )

Nix 2.20 dropped support for uninstalling packages by index, while adding support for uninstalling by name. this fix uses NameOrIndex to get the right value to pass to `nix profile remove`

## How was it tested?
Verified on localhost with Nix 2.20. CI Tests should confirm that it's compatible with Nix 2.18.

### DIFF
--- a/internal/devbox/util.go
+++ b/internal/devbox/util.go
@@ -5,7 +5,6 @@ package devbox
 
 import (
 	"context"
-	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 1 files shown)


savil (approved):

savil:
Let me review #1903 as well

PR #1903, opened 2024-03-15
[process-compose] Remove old process compose using store path

mikeland73:
## Summary

Fixes https://github.com/jetpack-io/devbox/issues/1901

Alternative: https://github.com/jetpack-io/devbox/pull/1902

nix 2.20 removed index support. Use store paths instead.

### DIFF
--- a/internal/nix/profiles.go
+++ b/internal/nix/profiles.go
@@ -75,13 +75,15 @@ func ProfileInstall(ctx context.Context, args *ProfileInstallArgs) error {
 	return cmd.Run()
 }
 
-func ProfileRemove(profilePath string, indexes ...string) error {
+// ProfileRemove removes packages from a profile.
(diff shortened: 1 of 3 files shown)


internal/nix/profiles.go
```diff
@@ -75,13 +75,15 @@ func ProfileInstall(ctx context.Context, args *ProfileInstallArgs) error {
 	return cmd.Run()
 }
 
-func ProfileRemove(profilePath string, indexes ...string) error {
+// ProfileRemove removes packages from a profile.
+// WARNING, don't use indexes, they are not supported by nix 2.20+
+func ProfileRemove(profilePath string, packageNames ...string) error {
```
savil, line 80: nit: `packageNameOrStorePaths` ?

savil (approved):

PR #1904, opened 2024-03-15
check if corepack exists before running init_hook

Lagoja:
## Summary

Fixes a reported issue where init_hook would fail if Node.js version was <14

## How was it tested?

Tested using `devbox add nodejs@[12,14,18,latest]`, and running `devbox shell`.

### DIFF
--- a/plugins/nodejs.json
+++ b/plugins/nodejs.json
@@ -2,14 +2,22 @@
     "$schema": "https://raw.githubusercontent.com/jetpack-io/devbox/main/.schema/devbox-plugin.schema.json",
     "version": "0.0.2",
     "name": "nodejs",
-    "readme": "Devbox automatically configures Corepack for Nodejs. You can install Yarn or Pnpm by adding them to your `package.json` file using `packageManager`\nCorepack binaries will be installed in your local `.devbox` directory",
+    "readme": "\nIf you are using Node.js 14 or newer, you can enable Corepack by running `devbox run corepack-enable`. Corepack lets you install Yarn or Pnpm by adding them to your `package.json` file using `packageManager`\nCorepack binaries will be installed in your local `.devbox` directory\n",
     "env" : {
         "PATH": "{{ .Virtenv }}/corepack-bin/:$PATH"
     },
     "shell": {
         "init_hook": [
-            "mkdir -p {{ .Virtenv }}/corepack-bin",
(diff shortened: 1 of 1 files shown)


plugins/nodejs.json
```diff
@@ -9,7 +9,7 @@
     "shell": {
         "init_hook": [
             "mkdir -p {{ .Virtenv }}/corepack-bin",
-            "corepack enable --install-directory \"{{ .Virtenv }}/corepack-bin/\""
+            "if command -v corepack &> /dev/null; then corepack enable --install-directory \"{{ .Virtenv }}/corepack-bin/\"; fi"
```
mikeland73, line 12: nit you could do:

`command -v corepack &> /dev/null && corepack enable --install-directory \"{{ .Virtenv }}/corepack-bin/\"`

which is a bit shorter and more portable.

mikeland73 (approved):

plugins/nodejs.json
```diff
@@ -1,15 +1,23 @@
...
+        "scripts": {
+            "corepack-enable": [
+                "mkdir -p {{ .Virtenv }}/corepack-bin",
+                "command -v corepack &> /dev/null && corepack enable --install-directory \"{{ .Virtenv }}/corepack-bin/\"" 
+            ],
+            "corepack-disable": [
+                "command -v corepack &> /dev/null && corepack disable --install-directory \"{{ .Virtenv }}/corepack-bin/\"" ,
+                "rm -rf {{ .Virtenv }}/corepack-bin"
+            ]
+        }
```
mikeland73, line 21: This is probably ok, but one downside of making it a script is that it would be synced across users (so there's no way to enable it for my project so that all users get it, which would be a more devbox way to do it). We could achieve that using environment variables + init hook instead. (e.g. if env variable is set we set the path and install corepack if needed)

I'm also not sure about the semi-permanent ergonomics of this script.

mikeland73 (approved):
Let's ship this to unbreak folks, but this feels like something that should be configuration and not a script.

plugins/nodejs.json
Lagoja: To make it permanent, you can add `corepack-enable` to the init_hook. I can update that in the doc string. One problem with environment variable and init_hook is that there is no way to disable it once it gets set in your .devbox dir, outside of deleting the folder or updating path. 

If there's a better ergonomic, I'd rather get it right now than ship it again later

PR #1919, opened 2024-03-22
[plugins] Support private GitHub repositories

### LINKED ISSUE #1918: Private Github plugins

Devbox allows you to reference plugins hosted by Github, but from what I can tell they only allow public repos. Our company wants to build a plugin for internal use and mark the repo visibility as private, but we run into an authentication issue when attempting to pull.

If the user has `GITHUB_TOKEN` set in their shell (not devbox.json or devbox shell), then `token=$GITHUB_TOKEN` is appended as a URL param when resolving the Github-hosted plugin URL. For example:

- If `GITHUB_TOKEN` is not set, the URL will be: `https://raw.githubusercontent.com/<org>/<repo>/master/<plugin-dir>`
- If `GITHUB_TOKEN` is set to `gh_abcd`, the URL will be: `https://raw.githubusercontent.com/<org>/<repo>/master/<plugin-dir>?token=gh_abcd`

Currently we pull the repo and reference using the local path. This works but my concerns are:

1. Team members have to manage cloning and updating the local repo
2. The local path is dependent on where team member clone the repo to, so it can make our `devbox.json` files brittle.

jylamont:
## Summary

Fixes #1918 

Plugins stored in a private GitHub repository fail to pull due to due to a lack of authentication. This PR will add an `Authorization: token <GITHUB_TOKEN>` header to the request if the `GITHUB_TOKEN` environment variable is set.

## How was it tested?

Added test cases for `http.Request` with and without environment variable.

Manually tested on my machine:
- Added a private GitHub repo to devbox.json `"include"`
- `devbox shell`: confirmed existing 404 error
- `GITHUB_TOKEN=xxx devbox shell`: shell environment boots correctly

### DIFF
--- a/internal/plugin/github.go
+++ b/internal/plugin/github.go
@@ -2,9 +2,11 @@ package plugin
 
 import (
 	"cmp"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"strings"
 
@@ -56,12 +58,13 @@ func (p *githubPlugin) Hash() string {
 }
 
 func (p *githubPlugin) FileContent(subpath string) ([]byte, error) {
-	contentURL, err := p.url(subpath)
+	req, err := p.request(subpath)
 	if err != nil {
 		return nil, err
 	}
 
-	res, err := http.Get(contentURL)
+	client := &http.Client{}
+	res, err := client.Do(req)
 	if err != nil {
 		return nil, err
 	}
@@ -71,7 +74,7 @@ func (p *githubPlugin) FileContent(subpath string) ([]byte, error) {
 			"failed to get plugin %s @ %s (Status code %d). \nPlease make "+
 				"sure a plugin.json file exists in plugin directory.",
 			p.LockfileKey(),
-			contentURL,
+			req.URL.String(),
 			res.StatusCode,
 		)
 	}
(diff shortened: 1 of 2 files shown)


internal/plugin/github.go
```diff
@@ -91,6 +94,28 @@ func (p *githubPlugin) url(subpath string) (string, error) {
...
+	req, err := http.NewRequest(http.MethodGet, contentURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add github token to request if available
+	ghToken := os.Getenv("GITHUB_TOKEN")
+
+	if ghToken != "" {
+		authValue := fmt.Sprintf("token %s", ghToken)
```
mikeland73, line 112: Can this be `Bearer` instead of `token`? I think github supports both. Would add flexibility to use JWT tokens as well.

mikeland73 (approved):
Awesome, thanks!

Non blocker, but I think this could be simplified using `oauth2` client:


```golang
httpClient := oauth2.NewClient(ctx, oauth2.StaticTokenSource(
  &oauth2.Token{AccessToken: os.Getenv("GITHUB_TOKEN")},
))
return httpClient.Get(contentURL)
```

that way you don't have to manually set headers. I think github will ignore empty auth header as well.

jylamont:
Thanks @mikeland73, my team will really appreciate this. I'll try to incorporate your suggestions in a follow-up PR.

Issue #1914, opened 2024-03-20
Devbox should handle spaces in paths


Lagoja:
### What problem are you trying to solve?

I'm frustrated when I try to use Devbox with a path that has a space in it. For example: if my Devbox project is in the following path: 

```
~/src/repros/spaced path
```

Devbox will fail with the following error: 

```
Error: nix print-dev-env --json "path:/Users/johnlago/src/repros/spaced path/.devbox/gen/flake": exit status 1
```

With DEVBOX_DEBUG=1:
```
...
2024/03/20 16:54:06 Running print-dev-env cmd: /nix/var/nix/profiles/default/bin/nix print-dev-env path:/Users/johnlago/src/repros/spaced path/.devbox/gen/flake --extra-experimental-features ca-derivations --option experimental-features nix
✓ Computed the Devbox environment.
Error: nix print-dev-env --json "path:/Users/johnlago/src/repros/spaced path/.devbox/gen/flake": exit status 1

2024/03/20 16:54:07 Command stderr: path '/Users/johnlago/src/repros/spaced path/path:/Users/johnlago/src/repros/spaced path/.devbox/gen/flake' does not contain a 'flake.nix', searching up
error: getting status of '/Users/johnlago/src/repros/spaced path/path:/Users/johnlago/src/repros/spaced path/.devbox/gen/flake': No such file or directory
...
```

### What solution would you like?

As of Nix 2.19, flakes + print-dev-env support paths that are URL or [percent encoded](https://github.com/NixOS/nix/pull/6614). We should adapt our command to use percent encoding, so we can support spaced paths.

### Alternatives you've considered

Not using paths with spaces

Lagoja:
Related issues #1876 #1429 #971

ademlabs:
Is this in the latest release (0.12)? Still having this issue currently.

Lagoja:
Spaces in Paths should be fixed as of [Devbox 0.13.1](https://github.com/jetify-com/devbox/releases/tag/0.13.1)

deftdawg:
Unfortunately it's still broken in 0.13.4...  

I think this change was supposed to resolve it (my system is running bash):
https://github.com/jetify-com/devbox/commit/ac072048a1ccdd8bc3d87cada19690c68f8579f2#diff-38f085ff12c1d3451f41d90ca82ba777f763554116f422c12c8ca92d722f583fR59

```sh
devbox version
# 0.13.4
mkdir -p parent\ space/devbox-test
[...]

PR #1923, opened 2024-03-22
[plugins] support reading (and discarding) comments plugin.json files

savil:
## Summary

Fixes #1911

This PR cleans the json-file-content of comments before processing it as "json content" i.e. we discard the comments prior to invoking existing logic for ingesting the `plugin.json` file.

This PR relies on Devbox not programmatically modifying the `plugin.json`. For now, users hand edit the `plugin.json` files. This is different from `devbox.json` which we do programmatically modify due to `devbox add`, `devbox rm` and similar commands, and so need to keep track of the json-with-comments' AST.

## How was it tested?

To test, I added comments to the following:
1. local plugin example
2. a builtin plugin (python)

I did NOT test remote Github plugin. But I tried to modify its code path. Did i get it right?

PR #1922, opened 2024-03-22
remove extraneous print

savil:
## Summary

oops

PR #1920, opened 2024-03-22
Support Project Directory having spaces

savil:
## Summary
In aid of #1914

Added some basic support for spaces. I think this requires more extensive testing around:
1. plugins
2. flake packages

I should try to have the examples tests run in cicd with a space in their path.

## How was it tested?

```
mkdir -p `/Users/<omitted>/code/jetpack/devbox-projects/path with space`
devbox init
devbox shell
devbox add hello
devbox add cowsay
```

could run:
```
cowsay hello
```

set `export FISH=$(which fish)` and repeat the above steps

PR #1924, opened 2024-03-23
[path with spaces] run scripts for projects with paths in spaces

savil:
## Summary

Related to #1914

Ran into this scenario during further testing

## How was it tested?

move the plugins examples:
`mv devbox/examples "devbox/examples with plugins"`

`cd examples/plugins with spaces/v2-github`
`devbox run run_test` -> this would previously fail

### DIFF
--- a/internal/nix/run.go
+++ b/internal/nix/run.go
@@ -8,14 +8,15 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/cmdutil"
 	"go.jetpack.io/devbox/internal/debug"
 )
 
-func RunScript(projectDir, cmdWithArgs string, env map[string]string) error {
-	if cmdWithArgs == "" {
+func RunScript(projectDir string, cmdWithArgs []string, env map[string]string) error {
+	if len(cmdWithArgs) == 0 {
 		return errors.New("attempted to run an empty command or script")
 	}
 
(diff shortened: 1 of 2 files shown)


savil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1924?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1924** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1923** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1923?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1922** <a href="https://app.graphite.dev/github/pr/jetpack-io/devbox/1922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @savil and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

LucilleH (approved):

internal/nix/run.go
```diff
@@ -24,9 +25,13 @@ func RunScript(projectDir, cmdWithArgs string, env map[string]string) error {
 		envPairs = append(envPairs, fmt.Sprintf("%s=%s", k, v))
 	}
 
+	// Wrap in quotations since the command's path may contain spaces.
+	cmdWithArgs[0] = "\"" + cmdWithArgs[0] + "\""
+	cmdWithArgsStr := strings.Join(cmdWithArgs, " ")
```
mikeland73, line 30: You can use `fmt.Springf("%q", s)` to surround a string in quotes (in a way that will properly escape stuff in the string).

godocs:
`%q	a double-quoted string safely escaped with Go syntax`

thought, It makes me a tiny bit nervous, what if the arg is already quoted?

I do wonder if this is the wrong place to do this. Specifically, `shellgen.ScriptPath` gets called in a bunch of places then possibly used without quotes. For example, `EnvExports` will return a broken (non quoted) source line if `RunHooks` is true. (This is why you had to surround `HooksFilePath` in quotes in shellrc here https://github.com/jetpack-io/devbox/pull/1920/files )

We could rename `shellgen.ScriptPath` to `shellgen.QuotedScriptPath` instead and just make it clear it's quoted? Not sure if this is best.

PR #1932, opened 2024-03-28
[cache] Use provided nix bin cache from api

mikeland73:
## Summary

* If user is logged in, check jetpack API to see if they have nix cache available. If available use it.
* Adds new provider framework for logged-in/logged-out functionality.

Note: There's a minor issue we need to fix with the access token. It's missing the audience which is needed by the API. Will fix in https://github.com/jetpack-io/opensource

## How was it tested?

* Copied over access token with correct audience
* Added myself as a trusted user
* `devbox add hello` on org that has nix cache. Observed it attempted to use the cache.


internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -0,0 +1,99 @@
...
+	"go.jetpack.io/devbox/internal/devbox/providers/identity"
+	"go.jetpack.io/devbox/internal/nix"
+	"go.jetpack.io/devbox/internal/ux"
+	"go.jetpack.io/pkg/api"
+	"go.jetpack.io/pkg/auth"
+)
+
+type Provider struct{}
+
+var singleton *Provider = &Provider{}
```
mohsenari, line 20: why do providers need to be singletons?

internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -0,0 +1,99 @@
...
+	// * Automate making user trusted if needed
+	if !nix.IsUserTrusted(ctx) {
+		ux.Fwarning(
+			os.Stderr,
+			"In order to use a custom nix cache you must be a trusted user. Please "+
+				"add your username to nix.conf (usually located at /etc/nix/nix.conf)"+
+				" and restart the nix daemon.",
+		)
+	}
+}
```
mohsenari, line 99: if you're just showing a warning when a user is not trusted then `ensureTrustedUser` is not the best name for it. 
Maybe `checkTrustedUser` or alternatively we can return `NixCacheConfig{}` if the user is not trusted.

internal/devbox/secrets.go
```diff
@@ -18,7 +17,6 @@ func (d *Devbox) UninitializedSecrets(ctx context.Context) *envsec.Envsec {
 		},
 		IsDev:      build.IsDev,
 		Stderr:     d.stderr,
-		Store:      &jetstore.JetpackAPIStore{},
```
mohsenari, line 21: is this change relevant to the PR?

internal/nix/build.go
```diff
@@ -29,6 +30,7 @@ func Build(ctx context.Context, args *BuildArgs, installables ...string) error {
 		cmd.Args = append(cmd.Args, "--extra-substituters", args.ExtraSubstituter)
 	}
 	cmd.Env = allowUnfreeEnv(os.Environ())
+	cmd.Env = append(cmd.Env, args.Env...)
```
mohsenari, line 33: this can be `cmd.Env = append(allowUnfreeEnv(os.Environ()), args.Env...)`

mohsenari (approved):
see a couple of comments

internal/devbox/providers/nixcache/nixcache.go
mikeland73: They don't right now, but I'm hoping to add caching to avoid making multiple API calls. Will do this after testing end to end and understanding where the perf bottlenecks are.

internal/devbox/providers/nixcache/nixcache.go
mikeland73: I can change the function name. The issue is that I'm not 100% confident we can catch all cases where you are actually allowed to use the substitutor, so returning an empty cache config would be wrong.

Especially since `nix build` will still work in the bad case (it just shows a warning)

internal/devbox/secrets.go
mikeland73: This is no longer used when we updated envsec dependency.

Issue #1936, opened 2024-03-28
devbox rm <package> not working on Ubuntu 22.04


mrpastewart:
### What happened?

When I try to remove a package using devbox rm <package> specifically devbox rm php or devbox rm php@latest  it fails with

Warning: the following packages were not found in your devbox.json: php

The package is indeed in my .json file but it fails to remove it as I can still execute the php-cli

### Steps to reproduce

1.
2.
3.

### Command

rm

### devbox.json

```jsonc
{
  "$schema":  "https://raw.githubusercontent.com/jetpack-io/devbox/0.10.1/.sche>
  "packages": ["php@latest"],
  "shell": {
    "init_hook": [
      "echo 'Welcome to devbox!' > /dev/null"
    ],
    "scripts": {
      "test": [
        "echo \"Error: no test specified\" && exit 1"
      ]
    }
  }
}
```

### Devbox version

0.10.1

### Nix version

2.18.1

### What system does this bug occur on?

Linux (x86-64)

### Debug logs

2024/03/28 10:24:05 findProjectDir: path is 
2024/03/28 10:24:05 finding devbox config in dir: /home/platform/devtest
2024/03/28 10:24:05 findProjectDir: path is 
2024/03/28 10:24:05 finding devbox config in dir: /home/platform/devtest
2024/03/28 10:24:05 findProjectDir: path is 
2024/03/28 10:24:05 finding devbox config in dir: /home/platform/devtest
Warning: the following packages were not found in your devbox.json: php
2024/03/28 10:24:05 Creating files for package "php@latest" create files
2024/03/28 10:24:05 Creating file "/home/platform/devtest/.devbox/virtenv/php/process-compose.yaml" from contentPath: "php/process-compose.yaml"
2024/03/28 10:24:05 Creating file "/home/platform/devtest/.devbox/virtenv/php/flake/flake.nix" from contentPath: "php/flake.nix"
2024/03/28 10:24:05 running command: /home/platform/.nix-profile/bin/nix search path:/home/platform/devtest/.devbox/virtenv/php/flake ^ --json --extra-experimental-features ca-derivations --option experimental-features nix-command flakes fetch-closure
2024/03/28 10:24:05 running command: /home/platform/.nix-profile/bin/nix search path:/home/platform/devtest/.devbox/virtenv/php/flake#composer ^ --json --extra-experimental-features ca-derivations --option experimental-features nix-command flakes fetch-closure
[...]

PR #1940, opened 2024-03-28
[plugins] Fix rm for packages with __remove_trigger_package

mikeland73:
## Summary

Fixes issue where user could not remove package that is replaced by a flake in plugin (uses `__remove_trigger_package`)

Fixes https://github.com/jetpack-io/devbox/issues/1936

We need two different functions for packages:

* `AllPackages` which is used for instalation and lock file. Includes packages added/removed by plugins.
* `TopLevelPackages` is only what is in root devbox.json

## How was it tested?

* Tested manually
* Added php to rm unit test

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -176,7 +176,7 @@ func (d *Devbox) ConfigHash() (string, error) {
 
 	buf := bytes.Buffer{}
 	buf.WriteString(h)
-	for _, pkg := range d.ConfigPackages() {
+	for _, pkg := range d.AllPackages() {
 		buf.WriteString(pkg.Hash())
 	}
 	for _, pluginConfig := range d.cfg.IncludedPluginConfigs() {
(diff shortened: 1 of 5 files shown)


testscripts/rm/add-rm.test.txt
```diff
@@ -1,20 +1,20 @@
...
-exec devbox add hello vim cowsay
+exec devbox add hello vim cowsay php
 json.superset devbox.json all.json
 
-exec devbox rm vim hello
+exec devbox rm vim hello php
 json.superset devbox.json cowsay.json
 
-exec devbox add vim hello vim hello vim hello vim hello cowsay
+exec devbox add vim hello vim hello vim hello vim hello cowsay php php
```
savil, line 9: i assume double `php` is intentional

internal/devbox/devbox.go
```diff
@@ -1033,21 +1033,28 @@ func (d *Devbox) PackageNames() []string {
...
 }
 
-// ConfigPackages returns the packages that are defined in devbox.json
-// NOTE: the return type is different from devconfig.Packages
-func (d *Devbox) ConfigPackages() []*devpkg.Package {
+// AllPackages returns the packages that are defined in devbox.json and
+// recursively added by plugins.
+// * NOTE1: This will not return packages removed by their plugin with the
+// __remove_trigger_package field.
+// * NOTE2: the return type is different from devconfig.Packages
```
savil, line 1040: optional: can remove NOTE2. It was there because the old function name was `ConfigPackages` which may have led a reader to assume the return type is `devconfig.Packages`

savil (approved):

testscripts/rm/add-rm.test.txt
mikeland73: yep

PR #1961, opened 2024-04-04
nixcache: configure auth with `devbox cache configure`

gcurtis:
Depends on #1959 and #1960. The diff contains commits from those PRs.

---

Add a hidden `devbox cache configure` command that sets up Nix so it can authenticate with a private Devbox cache. This is a first step. Ultimately, we'd like to do this configuration automatically before building. We just need to make sure we don't re-prompt the user for sudo or slow down commands like `devbox run` or `devbox shell`.

Configuring auth goes as follows:

1. Check to see if the Nix daemon is running. If it isn't, we don't need to do anything.
2. Otherwise, we check to see if `~root/.aws/config` exists. If it does, we assume that we've already configured the root user's AWS credentials. This won't handle situations where the user already has a root AWS config and can be improved.
3. If the AWS config doesn't exist, prompt the user to escalate to root. If they say yes, relaunch `devbox cache configure`.
4. The new devbox process running as root writes an AWS config file that tells the AWS CLI/SDKs to authenticate by running `devbox cache credentials`.

Note that we need to use the default AWS profile. There's a bug in Nix where non-default profiles (specified in the S3 URI with ?profile=devbox) don't use the default credential chain. If this gets fixed, we should use a separate profile so that we don't interfere with any AWS options set by the user (although hopefully this is rare for root).


internal/devbox/packages.go
```diff
@@ -445,19 +445,19 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode
...
-				AllowInsecure:    pkg.HasAllowInsecure(),
-				Env:              nixCacheConfig.CredentialsEnvVars(),
-				ExtraSubstituter: nixCacheConfig.URI,
-				Flags:            flags,
-				Writer:           d.stderr,
+		args.ExtraSubstituter, err = d.providers.NixCache.URI(ctx)
+		if err == nil {
+			creds, err := d.providers.NixCache.Credentials(ctx)
+			if err == nil {
+				args.Env = creds.Env()
```
mikeland73, line 457: Is this needed for single user installs?

internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -106,7 +216,56 @@ func checkIfUserCanAddSubstituter(ctx context.Context) {
...
+func (c Credentials) MarshalJSON() ([]byte, error) {
+	return json.Marshal(c.aws)
+}
+
+func (c *Credentials) UnmarshalJSON(data []byte) error {
+	if c == nil {
+		c = &Credentials{}
+	}
+	return json.Unmarshal(data, &c.aws)
+}
```
mikeland73, line 248: If we embed the aws credentials in the struct we won't need these.

internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -2,91 +2,201 @@ package nixcache
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to relaunch with sudo: %w", err)
+	}
+	return nil
+}
+
+// Credentials fetches short-lived credentials that grant access to the user's
+// private cache.
+func (p *Provider) Credentials(ctx context.Context) (Credentials, error) {
+	cache := filecache.New[Credentials]("devbox/credentials")
```
mikeland73, line 152: nix, should we call this cache `devbox/providers/nixcache` ?  Same for URI

mikeland73: Also, after using this API, I realize it's weird for the templatizing to happen on `New`. Maybe we rename `New` to `ForType`

`filecache.ForType[Foo](name)`

internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -106,7 +216,56 @@ func checkIfUserCanAddSubstituter(ctx context.Context) {
...
+func (c *Credentials) UnmarshalJSON(data []byte) error {
+	if c == nil {
+		c = &Credentials{}
+	}
+	return json.Unmarshal(data, &c.aws)
+}
+
+// awsCredentials marshals AWS credentials to JSON per the schema described in
+// `aws help config-vars` under "Sourcing Credentials From External Processes".
+type awsCredentials struct {
```
mikeland73, line 252: Could we avoid the 2 extra structs? (or just have 1?)

internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -2,91 +2,201 @@ package nixcache
...
-		func() (*nixv1alpha1.GetBinCacheResponse, time.Time, error) {
-			r, err := apiClient.GetBinCache(ctx)
-			if err != nil || r.GetNixBinCacheUri() == "" {
-				return nil, time.Time{}, err
-			}
-			return r, r.GetNixBinCacheCredentials().Expiration.AsTime(), nil
-		},
-	)
+func (p *Provider) configureRoot(username string) error {
+	exe, err := os.Executable()
```
mikeland73, line 76: Keep in mind that prod devbox is wrapped. Calling unwrapped version may cause issues.

ditto on other line that looks up executable. 

I think for this feature (and just to be safe) we enforce that executable is just `devbox`? 

I remember using the unwrapped version at some point in our bin wrappers and it caused all sorts of issues (specifically, the wrapper sets some env vars that the binary expects I think).

mikeland73 (approved):
For followup:

* We should save the configure state to a place that the user can access so devbox can remember if it's already been configured. Specifically, for a non-root user in a multi user install, the cache wont work at all if they haven't configured. Since we don't want to ask for sudo every time, saving some state to non a non root location (maybe XDG config) so that we can warn the user they need to configure if they have not previously done so.

internal/devbox/packages.go
gcurtis: Yeah, and sometimes Nix runs things as the current user (not via the daemon) so we also set the env vars to be safe.

internal/devbox/providers/nixcache/nixcache.go
gcurtis: root might not have devbox in its PATH like the user does. I'll see if there's a way to get the path to the wrapper script.

internal/devbox/providers/nixcache/nixcache.go
gcurtis: The intent was to hide the raw AWS credentials and the Version field, but I can consolidate.

internal/devbox/providers/nixcache/nixcache.go
gcurtis: Added a check for `LAUNCHER_PATH` like we do for the version check.

Issue #1882, opened 2024-03-08
Disable packages installation while entering a Devbox shell


dax:
### What problem are you trying to solve?

When used with direnv and when regularly switching between branches, Devbox slows down the workflow because it tries to keep the packages in sync with the `devbox.lock` state with a `Ensuring packages are installed.`.

### What solution would you like?

Disabling (with an option to `devbox shell`, `devbox run` and `devbox shellenv`) the automatic installation of packages when `devbox.lock` is not in sync and replacing it with a warning message could be enough in many cases.
Indeed, being out of sync might have no real impact on what we want to achieve. 
For example, switching to a branch to work on some features unrelated to the package not being installed or not with the right version.

savil:
@dax thanks for filing this issue.
To clarify: are you imagining an option that you apply as a one-off when doing `devbox shell/run/shellenv`, or something you want to permanently apply to the code used by direnv in your devbox project's directory?

dax:
@savil what I have in mind would be an option (as an env variable) to permanently disable automatic installation of packages by Devbox.

savil:
#2013 is out in the new 0.10.6 release!

PR #1963, opened 2024-04-05
feat: Add recompute flag to shellenv command

dax:
## Summary

(related [issue](https://github.com/jetpack-io/devbox/issues/1882))
When used with direnv and when regularly switching between branches, Devbox slows down the workflow because it tries to keep the packages in sync with the devbox.lock state with a Ensuring packages are installed.
This PR add the `--recompute` flag to the `shellenv` command that allow to set it to `--recompute=false` and then disable the automatic installation of package in case `devbox.json` is out of sync.

## How was it tested?

Setting the `.envrc` (ie. the direnv config file) as:
```
devbox shellenv --init-hook --no-refresh-alias --recompute=false
```

`--install` has been removed and `--recompute=false` compared with what `devbox generate direnv --print-envrc` generate by default.

Then updating `devbox.json` does not trigger package installation or removal but the context (ie. the env vars) are correctly updated.

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -14,12 +14,7 @@ import (
 	"go.jetpack.io/devbox/internal/ux"
 )
 
-type globalShellEnvCmdFlags struct {
-	recompute bool
-}
-
 func globalCmd() *cobra.Command {
-	globalShellEnvCmdFlags := globalShellEnvCmdFlags{}
 	globalCmd := &cobra.Command{}
 	persistentPreRunE := setGlobalConfigForDelegatedCommands(globalCmd)
 	*globalCmd = cobra.Command{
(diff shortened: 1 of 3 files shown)


savil (commented):
Thanks! I generally like the approach. That said, looking at the implementation I think a nicer fix would be to push the flag inside `shellEnvCmd` function's implementation.

Right now, we have this awkward parameter in `func shellEnvCmd(recomputeEnvIfNeeded *bool)` because we wanted it selectively exposed for just global and not regular devbox. If we're seeking to expose the `--recompute` flag in both global and regular devbox, then no need to expose this parameter. The flag can be handled internally inside `shellEnvCmd`.

Would you wanna make that change? If you are strapped for time, then happy to approve this PR, land it and I can make a follow up fix for it.

savil:
would you mind also running `devbox run lint` to ensure the code is considered "clean"?

savil:
`devbox run fmt` should help with formatting concerns

dax:
> Would you wanna make that change?

Thank you @savil for the suggestion. I'll push an update soon (and run the linter and formatter 😄 )

dax:
@savil The PR is now ready 🙏

mikeland73 (changes requested):
I think adding this flags makes sense, but the current implementation will cause a change in behavior in global shellenv (because recompute was previously defaulted to false, and now it is defaulted to true).

I think you may be able to simply override the flag in the global case (but we need some testing to confirm, cobra internals don't always behave intuitively).

dax:
Thanks @mikeland73 for spotting the regression 🙏 
I guess I can close this PR as @savil now have submitted a cleaner one 🙏

PR #1970, opened 2024-04-08
[nix-cache] Use new api

mikeland73:
## Summary

This uses new API, and splits out the 2 calls (one to get URI, and other to get credentials).


gcurtis (commented):
Most of these changes are in https://github.com/jetpack-io/devbox/pull/1961. Could we use that to avoid conflicts?

mikeland73:
@gcurtis yeah, I'll review that. I was just hoping to unbreak the cli before shipping a release.

PR #1972, opened 2024-04-08
Change copyright

LucilleH:


Lagoja (approved):

loreto (commented):
Super nit, but technically it should be `Jetify Inc.` (see period at the end of `Inc`). It should have been like that w/ Jetpack Technologies Inc too 🤷

PR #1985, opened 2024-04-11
[offline] Allow devbox to work offline if DNS fails

mikeland73:
## Summary

Fixes https://github.com/jetify-com/devbox/issues/1977

(with another hack).

Unfortunately the way we structure our flake makes it really difficult to remove our network requests (when state is stale). Ideally, the flake can be built off of local only nix store references, but that's a much bigger project.

In the meantime, this is a quick hack that fixes issue above.

## How was it tested?

Added synthetic error that returns `connection refused` in the message. Ensures I was still able to run/shell with a stale state.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -1008,11 +1008,19 @@ func (d *Devbox) ensureStateIsUpToDateAndComputeEnv(
 	// When ensureStateIsUpToDate is called with ensure=true, it always
 	// returns early if the lockfile is up to date. So we don't need to check here
 	if err := d.ensureStateIsUpToDate(ctx, ensure); isConnectionError(err) {
+		if !fileutil.Exists(d.nixPrintDevEnvCachePath()) {
+			ux.Ferror(
+				d.stderr,
+				"Error connecting to the internet and no cached environment found. Aborting.\n",
+			)
+			return nil, err
+		}
 		ux.Fwarning(
 			d.stderr,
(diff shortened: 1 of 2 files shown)


internal/devbox/devbox.go
```diff
@@ -1007,10 +1007,13 @@ func (d *Devbox) ensureStateIsUpToDateAndComputeEnv(
 
 	// When ensureStateIsUpToDate is called with ensure=true, it always
 	// returns early if the lockfile is up to date. So we don't need to check here
-	if err := d.ensureStateIsUpToDate(ctx, ensure); err != nil && !strings.Contains(err.Error(), "no such host") {
-		return nil, err
+	if err := d.ensureStateIsUpToDate(ctx, ensure); isConnectionError(err) {
+		ux.Fwarning(
+			d.stderr,
+			"Error connecting to the internet. Will attempt to use cached environment.\n",
+		)
```
savil, line 1014: what does the user experience if there was no cached environment? 

i.e. can you `rm -rf .devbox`, turn off internet and see what you experience?

savil (approved):

internal/devbox/devbox.go
mikeland73: It fails with an ugly error. I just added a check so we only keep going if the print dev env cache is there.

There are still a few other edge case failure modes, e.g. if some package has been gc'd from nix store, but I think that's fine.

PR #1995, opened 2024-04-12
Change footer links to https

Lagoja:
## Summary
Two of our footer links were using http instead of https

## How was it tested?
yarn build


loreto (approved):
LGTM

PR #1996, opened 2024-04-12
nixcache: automatically configure nix daemon

gcurtis:
Before installing packages, check to see if the Nix daemon is already configured to use a Devbox cache when one is available. The user no longer needs to explicitly call `devbox cache configure`, however the command is still there (and hidden) to make sudo escalation possible.

There are a few known limitations in this initial implementation:

- We don't keep track of "no" responses to sudo, so Devbox will bug the user every time if they refuse.
- The Nix daemon needs to restart after adding a trusted user. We currently only support auto-restarting the daemon on macOS and assume its managed with launchd.
- We sleep rather than poll for the daemon to restart. This only happens when a trusted user is added, but it does make things a little flakier.


internal/devbox/packages.go
```diff
@@ -456,6 +458,18 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode
...
+			// the cache fails. Just proceed without it.
+			u, err := user.Current()
+			if err != nil {
+				err = redact.Errorf("lookup current user: %v", err)
+				debug.Log("error configuring cache: %v", err)
+			}
+			err = d.providers.NixCache.Configure(ctx, u.Username)
+			if err != nil {
+				debug.Log("error configuring cache: %v", err)
+			}
```
mikeland73, line 472: Should we use `ux.FWarning` and print top stderr?

debug doesn't get shown by default.

internal/nix/nix.go
```diff
@@ -412,3 +413,21 @@ func parseInsecurePackagesFromExitError(errorMsg string) []string {
 
 	return insecurePackages
 }
+
+func RestartDaemon(ctx context.Context) error {
+	if runtime.GOOS != "darwin" {
+		return nil
```
mikeland73, line 419: Should we print a warning telling user to restart? (only if non-darwin multi-user install)

internal/nix/nix.go
```diff
@@ -412,3 +413,21 @@ func parseInsecurePackagesFromExitError(errorMsg string) []string {
...
+	}
+	err := exec.CommandContext(ctx, "launchctl", "bootout", "system", "/Library/LaunchDaemons/org.nixos.nix-daemon.plist").Run()
+	if err != nil {
+		return err
+	}
+	err = exec.CommandContext(ctx, "launchctl", "bootstrap", "system", "/Library/LaunchDaemons/org.nixos.nix-daemon.plist").Run()
+	if err != nil {
+		return err
```
mikeland73, line 427: If error, should we just print error saying unable to restart daemon. Telling user they need to restart it manually to get custom cache?

internal/nix/nix.go
```diff
@@ -412,3 +413,21 @@ func parseInsecurePackagesFromExitError(errorMsg string) []string {
 
 	return insecurePackages
 }
+
+func RestartDaemon(ctx context.Context) error {
```
mikeland73, line 417: nit no need to export

mikeland73 (approved):

PR #1997, opened 2024-04-12
[plugins] Cache github plugin

mikeland73:
## Summary

Excessive github requests are slow and can also lead to API rate limiting. This caches content for 24 hours. 

Ideally we store resolved plugin data in the lock file and cache indefinitely (until user does update) but that's a bit more work so this is a stopgap.

Users can use `devbox update` to clear cache. 

cc: @Lagoja 

## How was it tested?

Added print statements to track http requests and observed them going away after content is cached.


Lagoja:
Does this also allow updating to newer versions of Github plugins by running `devbox update`?

mikeland73:
> Does this also allow updating to newer versions of Github plugins by running devbox update

Not currently, but the plan is to build that mechanism into this command as well. In order to do that we need to

* Fetch current resolved version of plugin and save to lock file
* Change update to re-resolve plugin

One caveat is that plugin semantics are a bit weird in case of upgrades. Specifically when it comes to creating files. 

For non built-in plugins the semantics are the following:

* virtenv files are always recreated.
* devbox.d files are created if they are missing. This means some deleted files may be recreated. (I think this can be improved)

For built-ins:

* virtenv files are always recreated.
* devbox.d files are only created the first time the package that triggered the plugin is added to lockfile and then never again.

I think a future upgrade path that merges how both of these behave could be:

* Save all version info in plugin. Instead of using a number/semver version, we could use a git hash (for builtins as well). This would allow folks to stay on an older version of a plugin which would allow us to do backward incompatible changes.
* virtenv files are always recreated.
* devbox.d files (if any) should show a prompt asking (add-if-missing, replace, ignore) or something like it.

internal/cachehash/hash.go
```diff
@@ -66,7 +73,18 @@ func JSONFile(path string) (string, error) {
...
 
 func newHash() hash.Hash { return sha256.New() }
+
+// Slug returns a deterministic URL slug version of the string. With the
+// following characteristics:
+//
+// * A 6 character hash is appended to avoid collisions
+// * Trims the slug to last 50 characters
+// * Removes any prefixed hashes to ensure first character is alphanumeric
+func Slug(s string) string {
```
gcurtis, line 87: Where is this called? It sounds like this might be distinct from cache key hashing, no?

gcurtis (approved):

internal/cachehash/hash.go
mikeland73: It's no longer needed (now called internally by cache). Will remove.

In a follow up PR we can also start referencing cachehash and redact from `pkg` repo.

PR #2001, opened 2024-04-15
[cache] Add cache info command

mikeland73:
## Summary

* Adds new info command
* key cache and credentials local cache to current member.

### DIFF
--- a/internal/boxcli/cache.go
+++ b/internal/boxcli/cache.go
@@ -5,6 +5,7 @@ package boxcli
 
 import (
 	"encoding/json"
+	"fmt"
 	"os/user"
 
 	"github.com/MakeNowJust/heredoc/v2"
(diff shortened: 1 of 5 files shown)


internal/boxcli/cache.go
```diff
@@ -64,7 +65,7 @@ func cacheCmd() *cobra.Command {
 	cacheCommand.AddCommand(uploadCommand)
 	cacheCommand.AddCommand(cacheConfigureCmd())
 	cacheCommand.AddCommand(cacheCredentialsCmd())
-	cacheCommand.Hidden = true
```
gcurtis, line 67: Should we add a caveat to the command help saying it's only in private beta for now?

gcurtis (approved):

internal/boxcli/cache.go
mikeland73: The upload command works without Jetify cloud (you can specify your own s3 bucket and pass in your own credentials) so I thought it would be cool to open it up. The credentials and config commands are still hidden.

PR #2007, opened 2024-04-17
[bug-fix] Fix lockfile rewrite

mikeland73:
## Summary

The source of this bug is that when calling `lockfile.Tidy()` we were passing in only top level packages and not including plugin packages as well.

This bug was also affecting dev container construction and some edge case plugin template logic. This bug only affects users that are using custom plugins with packages.

I renamed `PackageNames` to `AllPackageNames` for clarity. Also removed a redundant function in config package that was no longer needed.
 
## How was it tested?

In other project added golang monorepo plugin and logged to ensure Tidy was not recreating lockfile. Also tested adding go to that project.

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -84,7 +84,7 @@ func listGlobalCmdFunc(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	for _, p := range box.PackageNames() {
+	for _, p := range box.AllPackageNames() {
 		fmt.Fprintf(cmd.OutOrStdout(), "* %s\n", p)
 	}
 	return nil
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -49,9 +49,7 @@ func globalCmd() *cobra.Command {
 	addCommandAndHideConfigFlag(globalCmd, servicesCmd(persistentPreRunE))
 	addCommandAndHideConfigFlag(globalCmd, shellEnv)
 	addCommandAndHideConfigFlag(globalCmd, updateCmd())
-
-	// Create list for non-global? Mike: I want it :)
-	globalCmd.AddCommand(globalListCmd())
+	addCommandAndHideConfigFlag(globalCmd, listCmd())
 
 	return globalCmd
 }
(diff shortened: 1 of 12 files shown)


internal/devbox/devbox.go
```diff
@@ -1039,11 +1039,14 @@ func (d *Devbox) flakeDir() string {
...
-	// Skipping for now to avoid propagating the error value.
-	return d.cfg.PackagesVersionedNames()
+// AllPackageNames returns the all package names, including those added
+// by plugins
+func (d *Devbox) AllPackageNames() []string {
+	result := make([]string, 0, len(d.cfg.Packages()))
+	for _, p := range d.cfg.Packages() {
+		result = append(result, p.VersionedName())
+	}
+	return result
```
savil, line 1049: Are the names from `AllPackageNames` exactly matching the set of packages in `AllPackages` functions below?

`AllPackages` seems to respect `__remove_trigger_package`. Does `AllPackageNames` do that as well?

internal/plugin/plugin.go
```diff
@@ -144,7 +144,7 @@ func (m *Manager) createFile(
 		"DevboxDirRoot":        filepath.Join(m.ProjectDir(), devboxDirName),
 		"DevboxProfileDefault": filepath.Join(m.ProjectDir(), nix.ProfilePath),
 		"PackageAttributePath": attributePath,
-		"Packages":             m.PackageNames(),
+		"Packages":             m.AllPackageNames(),
```
savil, line 147: 🤔 is this one safe to do?

savil (approved):

internal/devbox/devbox.go
mikeland73: Good catch. Not sure. I'll audit.

internal/plugin/plugin.go
mikeland73: yeah this is all good.

gcurtis (approved):

internal/devbox/devbox.go
mikeland73: Definitely needed in most places.  Added back in. I think we can improve on this (some callers want different representations of this list) but for now the safest fix is to just include everything (top level, plugin packages, and also trigger removed)

internal/boxcli/global.go
```diff
@@ -61,33 +59,33 @@ func addCommandAndHideConfigFlag(parent, child *cobra.Command) {
 	_ = child.Flags().MarkHidden("config")
 }
 
-func globalListCmd() *cobra.Command {
-	return &cobra.Command{
+type listCmdFlags struct {
```
savil, line 62: why is this in this PR?

PR #2009, opened 2024-04-18
[gen] Add gen alias command

mikeland73:
## Summary

Add a command that allows creating aliases to devbox scripts. This lets users source aliases from any project they want. It's opt-in, works in POSIX and fish. Aliases make it easy to call the scripts for other projects or directories.

How it works:

```bash
eval "$(devbox gen alias)"
devbox-build
```

It builds the alias name using the project name in devbox.json.

Alternatively the user can specify a prefix

```bash
eval "$(devbox gen alias --prefix dx)"
dx-build
```

## How was it tested?

```bash
eval "$(devbox gen alias --prefix dx)"
dx-build
cd $HOME && dx-build
```

### DIFF
--- a/internal/boxcli/generate.go
+++ b/internal/boxcli/generate.go
@@ -4,6 +4,10 @@
 package boxcli
 
 import (
+	"cmp"
+	"fmt"
+	"regexp"
+
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


gcurtis (approved):
Sounds good to me!

PR #2010, opened 2024-04-18
nixcache: improve auto-configuration of cache

gcurtis:
Fix a few issues with the current auto-configuration flow for private nix caches:

- Track whether or not we've configured ~root/.aws/config in a separate state file in the user's home directory. This lets Devbox know if AWS has already been configured, even if it cannot read root's home directory.
- If the user answers no to the sudo confirmation prompt, don't ask them again. This is so we don't pester the user every time they install a package.
- Preserve XDG_STATE_HOME in sudo so that we write state files to the correct directory.
- Append a timestamp to the ~root/.aws directory when we back it up. This allows multiple backups if the setup process is run more than once.

The logic for saving state around whether or not the cache setup has already run lives in a new `setup` package. The `setup` package tracks when a task last ran, what version of Devbox it ran with, and if there was an error. This makes it easier to define tasks that only run once for a user or only occur after an upgrade.


internal/devbox/providers/nixcache/nixcache.go
```diff
@@ -33,154 +24,53 @@ func Get() *Provider {
...
-		// It looks like this is a multi-user install running a Nix
-		// daemon, so we need to configure AWS S3 authentication for the
-		// root user.
-		if err := p.sudoConfigureRoot(ctx, username); err != nil {
-			return err
+	// Otherwise, ask the user to confirm if it's okay to sudo.
+	const sudoPrompt = "Devbox requires root to configure the Nix daemon to use your organization's Devbox cache. Allow sudo?"
+	for _, t := range setupTasks {
+		err := setup.ConfirmRun(ctx, t.key, t.task, sudoPrompt)
+		if errors.Is(err, setup.ErrUserRefused) {
```
mikeland73, line 64: Should we differentiate between the first errUserRefused and subsequent (no interaction) times?

reasoning: The first time the user says no, we may want to tell them how to configure later if they want.

internal/setup/setup.go
```diff
@@ -0,0 +1,212 @@
...
+		return
+	}
+
+	err = os.WriteFile(path, data, 0o644)
+	if errors.Is(err, os.ErrNotExist) {
+		err = os.MkdirAll(filepath.Dir(path), 0o755)
+		if err == nil {
+			err = os.WriteFile(path, data, 0o644)
+		}
+	}
```
mikeland73, line 176: `MkdirAll` doesn't error if path already exists or if only partial path exists so you can simplify:

```golang 
err := os.MkdirAll(filepath.Dir(path), 0o755)
if err == nil {
  err = os.WriteFile(path, data, 0o644)
}
```

internal/devbox/providers/nixcache/setup.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// Ensure the XDG state directory exists before sudoing, otherwise it
+	// will be owned by root. It's used by the setup package to remember
+	// user responses to the confirmation prompt.
+	err = os.MkdirAll(xdg.StateSubpath("devbox"), 0o700)
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.CommandContext(ctx, "sudo", "--preserve-env=XDG_STATE_HOME", "--", exe, "cache", "configure", "--user", username)
```
mikeland73, line 151: I don't think `XDG_STATE_HOME` is set by default. If this is empty, will it still configure the right dir?

internal/devbox/providers/nixcache/setup.go
```diff
@@ -0,0 +1,220 @@
...
+// obtaining a token from `devbox cache credentials`.
+type awsSetupTask struct {
+	// username is the OS username that the Nix daemon should sudo as when
+	// running `devbox cache credentials`.
+	username string
+}
+
+func (a *awsSetupTask) NeedsRun(ctx context.Context, lastRun setup.RunInfo) bool {
+	return lastRun.Time.IsZero()
+}
```
mikeland73, line 67: Is aws setup needed for non-daemon? We can just pass env variables into build no? What if user doesn't even have sudo access?

internal/devbox/providers/nixcache/setup.go
```diff
@@ -0,0 +1,220 @@
...
+
+	flag := os.O_WRONLY | os.O_CREATE | os.O_EXCL
+	perm := fs.FileMode(0o644)
+	config, err := os.OpenFile(configPath, flag, perm)
+	if errors.Is(err, os.ErrNotExist) {
+		if err = os.Mkdir(filepath.Dir(configPath), 0o755); err != nil {
+			return redact.Errorf("create ~root/.aws directory: %v", err)
+		}
+		config, err = os.OpenFile(configPath, flag, perm)
+	}
```
mikeland73, line 103: See other comment about simplififying this pattern by calling `MkDirAll` first.

mikeland73 (approved):
Approving because it's not a regression, but my main comment is about not needing sudo for single user setups. I assume that if there's no daemon, `nix build` can be passed environment variables directly in which case we don't need either setup.

internal/devbox/providers/nixcache/nixcache.go
gcurtis: Good idea. Added `ErrAlreadyRefused` to distinguish them.

internal/devbox/providers/nixcache/setup.go
gcurtis: This is just so that we continue to use the custom state directory if one happens to be set. Otherwise, we'd default to using `~/.local/state` because `XDG_STATE_HOME` would be lost in sudo.

internal/devbox/providers/nixcache/setup.go
gcurtis: Added a check for the Nix daemon here.

internal/devbox/providers/nixcache/setup.go
gcurtis: I simplified the other one, but I wanted to avoid `os.MkdirAll` here. We shouldn't be creating anything at or above the user's home directory, and I wanted to be extra cautious since this runs as root. I added a comment explaining why.

PR #2011, opened 2024-04-18
[docker] Updated devbox install links for docker images

mohsenari:
## Summary
Updated devbox install link for docker image.
After merging this we will release a new docker image, and that should upgrade the launcher version to 0.2.2. Currently, docker images have 0.2.1 or 0.2.0 for launchers due to devbox image being built and pushed to dockerhub before launcher changes.

## How was it tested?
N/A


mikeland73 (approved):

mikeland73:
We should remember to publish a new image every time we publish a new launcher.

PR #2012, opened 2024-04-19
[alias] Add no-prefix flag

mikeland73:
## Summary

Adds `--no-prefix` flag to generate aliases for scripts without a prefix.

## How was it tested?

```bash
devbox gen alias --no-prefix
build
```

### DIFF
--- a/internal/boxcli/generate.go
+++ b/internal/boxcli/generate.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
 	"github.com/spf13/cobra"
 
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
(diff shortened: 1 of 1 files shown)


internal/boxcli/generate.go
```diff
@@ -210,7 +216,7 @@ func genAliasCmd() *cobra.Command {
 			}
 			re := regexp.MustCompile("[^a-zA-Z0-9_-]+")
 			prefix := cmp.Or(flags.prefix, box.Config().Root.Name)
-			if prefix == "" {
+			if prefix == "" && !flags.noPrefix {
```
savil, line 219:
```
return usererr.New(
					"To generate aliases, you must specify a prefix or set a name " +
						"in devbox.json, or use `--no-prefix` flag")
```
?

savil (approved):

PR #2013, opened 2024-04-20
[shellenv] add --recompute flag with default=true, while keep global shellenv's recompute flag with default=false

savil:
## Summary

Credit to @dax and #1963 for pointing out the need for this, and pioneering this solution.
This PR merely nudges the code to be a bit neater.

## How was it tested?

```
% git diff
diff --git a/internal/boxcli/shellenv.go b/internal/boxcli/shellenv.go
index 991c2bec..8076e125 100644
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -78,6 +78,9 @@ func shellEnvFunc(
        cmd *cobra.Command,
        flags shellEnvCmdFlags,
 ) (string, error) {
+       fmt.Printf("shellenv --recompute flag is %v\n", flags.recompute)
+       return "", nil
+
        env, err := flags.Env(flags.config.path)
        if err != nil {
                return "", err
```

compiled via `devbox run build`, and then:

```
 % devbox shellenv
shellenv --recompute flag is true

hash -r
 % devbox global shellenv
shellenv --recompute flag is false

hash -r
```

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -31,8 +31,12 @@ func globalCmd() *cobra.Command {
 	shellEnv := shellEnvCmd()
 	// For `devbox shellenv` the default value of recompute is true.
 	// Change the default value to false for `devbox global shellenv` only.
-	shellEnv.Flag("recompute").DefValue = "false"
+	shellEnv.Flag("recompute").DefValue = "false" // Needed for help text
 	if err := shellEnv.Flag("recompute").Value.Set("false"); err != nil {
+		// This will never panic because internally it just does
+		// `strconv.ParseBool("false")` which is always valid.
+		// If this were to change, we'll immediately detect this during development
+		// since this code always runs on any devbox command (and will fix it).
 		panic(errors.WithStack(err))
 	}
 
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -14,12 +14,7 @@ import (
 	"go.jetpack.io/devbox/internal/ux"
 )
 
-type globalShellEnvCmdFlags struct {
-	recompute bool
-}
-
 func globalCmd() *cobra.Command {
-	globalShellEnvCmdFlags := globalShellEnvCmdFlags{}
 	globalCmd := &cobra.Command{}
 	persistentPreRunE := setGlobalConfigForDelegatedCommands(globalCmd)
 	*globalCmd = cobra.Command{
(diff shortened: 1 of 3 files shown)


savil:
* **#2013** <a href="https://app.graphite.dev/github/pr/jetify-com/devbox/2013?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @savil and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

savil:
@mikeland73 @gcurtis ping

internal/boxcli/global.go
```diff
@@ -33,11 +28,13 @@ func globalCmd() *cobra.Command {
...
 
-	shellEnv := shellEnvCmd(&globalShellEnvCmdFlags.recompute)
-	shellEnv.Flags().BoolVarP(
-		&globalShellEnvCmdFlags.recompute, "recompute", "r", false,
-		"Recompute environment if needed",
-	)
+	shellEnv := shellEnvCmd()
+	// For `devbox shellenv` the default value of recompute is true.
+	// Change the default value to false for `devbox global shellenv` only.
+	shellEnv.Flag("recompute").DefValue = "false"
```
mikeland73, line 34: I don't think this is strictly needed , only for internal consistency. maybe add a comment?

internal/boxcli/global.go
```diff
@@ -33,11 +28,13 @@ func globalCmd() *cobra.Command {
...
+	if err := shellEnv.Flag("recompute").Value.Set("false"); err != nil {
+		panic(errors.WithStack(err))
+	}
```
mikeland73, line 37: Maybe add comment:

```
// This will never panic because internally it just does `strconv.ParseBool("false")` which is always valid
```

mikeland73 (approved):

internal/boxcli/global.go
savil: The `DefValue` modifies the default value printed in the `--help`. Without this line, the help for the flag looks like:
```
% devbox global shellenv --help
...
  -r, --recompute            Recompute environment if needed (default true)
```

savil: Added comment to explain this need.

PR #2014, opened 2024-04-22
Add trailing slash redirection to docs

LucilleH:
## Summary
TSIA. Mitigating Google crawling issues

## How was it tested?
On vercel preview


loreto (approved):
LGTM

Do we have other docusarus generated docs that need the same change?

PR #2017, opened 2024-04-23
[gen] Implement gen prod dockerfile

mikeland73:
## Summary

Simple implementation for generating a prod dockerfile from a devbox project. 

Notes:

* Flag to create this is hidden.
* Currently this requires a `start` script. There are also optional `install` and `build` scripts.

Follow ups:
* In addition to `start` script I want to support services but I was having trouble getting process compose to start as Dockerfile CMD. Will debug further and add in follow up.
* We can greatly optimize the Dockerfile by downloading devbox dependencies first before coping source code. This will likely require a new devbox command or flag. 

## How was it tested?

Tested using this project https://github.com/mikeland73/hello-world-server

```bash
devbox generate dockerfile --for prod
docker build . -t hello-world
docker run -p 8080:8080 hello-world
```


internal/devbox/generate/tmpl/prod.Dockerfile.tmpl
```diff
@@ -0,0 +1,28 @@
...
+those dependencies may also pull in other local dependencies and so on. Another
+sulution would be to add a new flag `devbox install --skip-errors` that would 
+just try to install what it can, and ignore the rest.
+
+A hack to make this simpler is to install from the lockfile instead of the json.
+*/}}
+
+COPY --chown=${DEVBOX_USER}:${DEVBOX_USER} . .
+
+RUN devbox install
```
mohsenari, line 22: There was an issue with using `devbox install` command in dockerfile. IIRC it didn't activate the python plugin so the subsequent python commands were not running in a virtual environment.

mohsenari: this is the issue I'm talking about:
https://github.com/jetify-com/devbox/issues/1122

can you test the steps in the issue to make sure we don't regress?

mohsenari (approved):

internal/devbox/generate/tmpl/prod.Dockerfile.tmpl
```diff
@@ -0,0 +1,28 @@
...
+
+RUN {{ .DevboxRunInstall }}
+
+RUN {{ .DevboxRunBuild }}
+
+CMD [{{ .Cmd }}]
```
LucilleH, line 28: It think we can optimize this image a bit. Something like:
- gc nix store (clean up)
- devbox shellenv (so cmd has the right env)
- separate install layer

mikeland73: @LucilleH for sure! I think some of those might be doable without devbox changes (but not all).

> gc nix store (clean up)

I can try this, but in theory the nix store should be pretty clean no? We haven't installed anything we don't need.

> devbox shellenv

This can help a bit but not without changes. Ideally, we do shellenv and then call the raw start command (without using `devbox run`). A small concern is the init hook not running multiple times (I think that's probably fine). So in practice it would look like:

```
RUN eval $(devbox shellenv --init-hook)
RUN devbox run install --raw
RUN devbox run build --raw
CMD ["devbox", "run", "start", "--raw"] 
```

Where `--raw` would be a new flag that indicates run the raw script without starting a devbox env (which would be wasteful) 

> separate install layer

This is the biggest win. But see my comment about local dependencies. I think there's 2 huge optimizations here:

* Do `devbox install` before copying source code. This makes a very cachable base layer.
* Split stages so that we copy over only nix stuff we need into final image. 

(1) is easy, but requires a new flag/command so we can ignore errors of missing local dependencies.
(2) is harder, but we did something similar in early devbox days.

My preference is to merge this and optimize in follow up. Thoughts?

internal/devbox/generate/tmpl/prod.Dockerfile.tmpl
mikeland73: I'm a bit confused by that issue. I understand that `install` doesn't run init hooks, but I would assume the line below

`RUN devbox shellenv --init-hook >> ~/.profile`

Would do the initialization.

Furthermore, actually activating the virtual environment is a manual process so neither install/run would do it.

Do you know why shellenv is not working?

mikeland73: FWIW, since the prod dockerfile always ends in `devbox run start` it will always run the init hook which creates the virtenv. To activate it, the user would need to add that to their start command. Maybe we can add an env variable that automatically activates it.

mikeland73:
Gonna merge this and improve in follow up.

PR #2018, opened 2024-04-24
[telemetry] Add uid, org_id, and nix build event

mikeland73:
## Summary

This PR improves segment logging by:

* Adding orgID if user is logged in.
* Using userID if user is logged in.
* Removing anonymousID if userID is non zero. (Segment will remove userID if we pass an anonymousID)
* Add new event to track `nix build`

This allows us to:

* Use orgID to segment data, potentially allowing individual orgs to export/use their own data as needed.
* Track logged in users.
* Track which packages are slow to build.

Private Notes:

* We still respect `DO_NOT_TRACK` even if user is logged in. 
* userID and orgID data is only sent if user is logged in.

cc: @Lagoja 

## How was it tested?

Logged to development segment/amplitude, built a quick table showing nix average build times:

image


internal/telemetry/telemetry.go
```diff
@@ -61,16 +63,30 @@ func Start() {
 	const deviceSalt = "64ee464f-9450-4b14-8d9c-014c0012ac1a"
 	deviceID, _ = machineid.ProtectedID(deviceSalt)
 
-	username := os.Getenv(envir.GitHubUsername)
-	if username != "" {
+	started = true
+}
+
+func userID(ctx context.Context) string {
+	tok, err := identity.Get().GenSession(ctx)
```
gcurtis, line 70: Is there a way to never refresh the token here and in `orgID`?

gcurtis (approved):
Is there a way to get the current username without refreshing the token? I think that would also let us remove the context from the telemetry functions, which should probably still run even if the context is cancelled (to report things like timeouts).

internal/telemetry/telemetry.go
mikeland73: yeah, our auth library has a `GetSessions` which returns `refreshableTokenSource` which have a Peek() function. I'll switch to that.

PR #2021, opened 2024-04-24
[easy][cicd] Bump devbox installer and switch to macos-13

mikeland73:
## Summary

The devbox installer has compatibility issues with macos-14 so changing from latest to 13 until that is fixed


LucilleH (approved):

.github/workflows/cli-tests.yaml
```diff
@@ -47,6 +47,7 @@ env:
   HOMEBREW_NO_ENV_HINTS: 1
   HOMEBREW_NO_INSTALL_CLEANUP: 1
   DEVBOX_DEBUG: 1
+  WAIT_FOR_NIX_VERSION: 60
```
LucilleH, line 50: We still need this?

mikeland73: nope, will remove

PR #2025, opened 2024-04-25
[easy][deps] Update pkg, use new `DEVBOX_API_TOKEN` name

mikeland73:
## Summary

TSIA


internal/devbox/providers/identity/identity.go
```diff
@@ -4,12 +4,12 @@ import (
...
 	"os"
 
+	"go.jetify.com/typeid"
 	"go.jetpack.io/devbox/internal/build"
 	"go.jetpack.io/pkg/api"
 	"go.jetpack.io/pkg/auth"
 	"go.jetpack.io/pkg/auth/session"
 	"go.jetpack.io/pkg/ids"
-	"go.jetpack.io/typeid"
 	"golang.org/x/oauth2"
```
mohsenari, line 13: Not for this PR, but when should we just update all these dependencies to use `go.jetify.com` instead?

mohsenari (approved):

PR #2032, opened 2024-04-30
[nix] Build all packages in same nix command

mikeland73:
## Summary

This changes how `installNixPackagesToStore` calls `nix.Build` so that it installs all installables in same nix command.

It does this by grouping all installables into two groups, one for allow-insecure and other for default and runs `build` on both groups (if not empty)

I was able to remove `packageInstallErrorHandler`  because it is already called in `packagesToInstallInStore` so we will never actually try to build something that can't build on the system, or if it requires `--allow-insecure` we will notice it then.

This does change our segment logging to group packages together for install times, so we will need to do some more work on the metrics side to determine slow packages. 

## How was it tested?

* Installed multiple packages not present in my nix store at once, including a combination of secure and insecure ones.
* Tried to install something that doesn't build on my system, saw correct error message

## Perf test

Anecdotal test:

Using the following
image

Before 
image

After 
image

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -430,79 +430,88 @@ func (d *Devbox) InstallRunXPackages(ctx context.Context) error {
 // and installing in the nix profile (even if offline).
 func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode) error {
 	packages, err := d.packagesToInstallInStore(ctx, mode)
-	if err != nil {
+	if err != nil || len(packages) == 0 {
 		return err
 	}
 
-	stepNum := 0
-	total := len(packages)
+	// --no-link to avoid generating the result objects
+	flags := []string{"--no-link"}
+	if mode == update {
+		flags = append(flags, "--refresh")
+	}
+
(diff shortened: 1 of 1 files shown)


savil:
For the scenario where a user clones a repo and does `devbox install`, this PR's change will mean they have to wait a long time without much UI information about what is being installed.  What does the UX look like?

Alternatively, we could preserve the old-behavior that prints output for each package, while having a `--parallel` flag or `DEVBOX_INSTALL_PARALLEL` env-var setting that runs it in parallel.

gcurtis (approved):
LGTM.

@savil I think the user will see progress via the `nix build` command because it's hooked up to stderr.

mikeland73:
@savil like @gcurtis mentioned, this does show output the entire time.

One thing this removed is the package by package manual output we used to do, specifically `fmt.Sprintf("[%d/%d] %s", stepNum, total, pkg)` 

in it's place we now do:

```
ux.Finfo(
		d.stderr,
		"Installing the following packages to the nix store: %s\n",
		strings.Join(packageNames, ", "),
	)
```

PR #2040, opened 2024-05-07
[process-compose] Bump version

mikeland73:
## Summary

Bumps to v1.5.0

Note: Because process-compose flake is not in any cache installing a new version can be pretty slow (it has to build which means it has to download dependencies in order to build). A few solutions:

* We could use nixpkgs (which lags the repo version by a few versions)
* We could use runx to install it (not nix, but it gets the latest version)
* We could store in our own cache.

## How was it tested?

Tested running services on a few examples.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -55,7 +55,7 @@ const (
 
 	// shellHistoryFile keeps the history of commands invoked inside devbox shell
 	shellHistoryFile            = ".devbox/shell_history"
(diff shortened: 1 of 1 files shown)


Lagoja:
I'm open to doing this via Nixpkgs if it simplifies things, even if we're a little bit behind

mikeland73:
@Lagoja lets do that as a follow up, it is gonna take a bit more work to turn utilities into a devbox project (but that's a worthwile change)

Lagoja (approved):

PR #2042, opened 2024-05-09
[perf] Skip cache check if store path exists locally

mikeland73:
## Summary

If store path exists locally, we know we can use that store path in `fetchClosure`. This helps speed up almost any operation where state is not up to date and we need to check if some store path is in cache.

## How was it tested?

* added print statement

`devbox run echo 5`

### DIFF
--- a/internal/devpkg/narinfo_cache.go
+++ b/internal/devpkg/narinfo_cache.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	"go.jetpack.io/devbox/internal/boxcli/featureflag"
+	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/lock"
 	"go.jetpack.io/devbox/internal/nix"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


internal/devpkg/narinfo_cache.go
```diff
@@ -145,6 +148,11 @@ func (p *Package) fetchNarInfoStatus(outputName string) (bool, error) {
 
 	outputInCache := map[string]bool{} // key = output name, value = in cache
 	for _, output := range outputs {
+		// if store path exists locally, then it is equivalent to being in the cache.
```
savil, line 151: I think this is correct, _assuming_ the nix order for resolving a store-path is is:
1. check local /nix/store to see if already built
2. check binary cache(s) configured to see if binaries are present
3. build derivation locally

I say _assuming_ because looking at the docs for `builtins.fetchClosure` it doesn't say that it will check local nix-store and skip the binary cache if its present:
https://nixos.org/manual/nix/stable/language/builtins.html#builtins-fetchClosure

savil (approved):

internal/devpkg/narinfo_cache.go
mikeland73: I assume it works this way, because otherwise we would be re-downloading everything every time the flake is evaluated. Currently we `build` everything first (which stores it int the nix store) and then we evaluate the flake using print-dev-env.

Also, see this note:

> fetchClosure is similar to [builtins.storePath](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-storePath) in that it allows you to use a previously built store path in a Nix expression. However, fetchClosure is more reproducible because it specifies a binary cache from which the path can be fetched

`builtins.storePath`  always checks the local store first before using substituters, so I assume `fetchClosure` does the same before trying to fetch from remote.

cc: @gcurtis let us know if this is your understanding as well.

gcurtis: That's also my understanding.

gcurtis (approved):

PR #2043, opened 2024-05-09
internal/setup: move sudo into SudoDevbox function + fix macOS CI

gcurtis:
Move the code that calls `sudo devbox` into the `setup` package and clean it up a bit. Fix some issues that were preventing the cache from being configured in non-interactive environments (such as CI).

- Don't re-prompt the user for confirmation in the sudo process.
- Don't call NeedsRun a second time in the sudo process.
- Don't prompt the user when stdin isn't a tty. This fixes cache configuration on macOS in CI.
- Preserve Devbox environment variables such as DEVBOX_DEBUG, DEVBOX_API_TOKEN, DEVBOX_PROD, and DEVBOX_USE_VERSION. This addresses cases where the sudo devbox process would use the wrong auth or API endpoint.
- Combine the nixcache aws and nix setup tasks into one task. There was a lot of overlap and this simplifies the setup code.

Reran the `cache-upload` workflow against this branch and it looks like the macOS runner is able to use the cache now - https://github.com/jetify-com/devbox/actions/runs/9022662173/job/24792707507#step:5:828

### DIFF
--- a/internal/devbox/providers/nixcache/setup.go
+++ b/internal/devbox/providers/nixcache/setup.go
@@ -9,83 +9,96 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"strings"
 	"time"
+	"unicode"
 
 	"go.jetpack.io/devbox/internal/debug"
 	"go.jetpack.io/devbox/internal/envir"
 	"go.jetpack.io/devbox/internal/nix"
 	"go.jetpack.io/devbox/internal/redact"
 	"go.jetpack.io/devbox/internal/setup"
-	"go.jetpack.io/devbox/internal/xdg"
 )
 
-// nixSetupTask adds the user to Nix's trusted-users list so that they can use
-// their private Devbox cache with the Nix daemon.
-type nixSetupTask struct {
-	// username is the OS username to trust.
-	username string
+func (p *Provider) Configure(ctx context.Context, username string) error {
+	return p.configure(ctx, username, false)
 }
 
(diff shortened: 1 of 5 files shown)


internal/devbox/providers/nixcache/setup.go
```diff
@@ -144,35 +156,34 @@ credential_process = %s -u %s -i -- %s cache credentials
...
+		"DEVBOX_API_TOKEN",
+		"DEVBOX_PROD",
+		"DEVBOX_USE_VERSION",
+		"XDG_CACHE_HOME",
+		"XDG_CONFIG_DIRS",
+		"XDG_CONFIG_HOME",
+		"XDG_DATA_DIRS",
+		"XDG_DATA_HOME",
+		"XDG_RUNTIME_DIR",
+		"XDG_STATE_HOME",
```
mikeland73, line 174: Are there no issues with propagating all of these?

internal/devbox/providers/nixcache/setup.go
```diff
@@ -144,35 +156,34 @@ credential_process = %s -u %s -i -- %s cache credentials
...
+	}
+	sb := strings.Builder{}
+	for _, name := range envs {
+		val := os.Getenv(name)
+		if val != "" {
+			sb.WriteString(name)
+			sb.WriteByte('=')
+			sb.WriteString(val)
```
mikeland73, line 182: Should this be quoted?

mikeland73 (approved):

internal/devbox/providers/nixcache/setup.go
gcurtis: Fixed.

PR #2053, opened 2024-05-14
nixcache: fix setup error on multi-user systemd Nix installs

gcurtis:


internal/devbox/packages.go
```diff
@@ -444,7 +444,11 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode
 		Flags:  flags,
 		Writer: d.stderr,
 	}
-	args.ExtraSubstituter, _ = d.providers.NixCache.URI(ctx)
+	args.ExtraSubstituter, err = d.providers.NixCache.URI(ctx)
+	if err != nil {
+		debug.Log("error getting nix cache URI, assuming user doesn't have a private cache: %v", err)
```
mikeland73, line 449: nit, cache could be private or public.

Lagoja (approved):

internal/devbox/providers/nixcache/setup.go
```diff
@@ -88,7 +89,9 @@ func (s *setupTask) Run(ctx context.Context) error {
 	}
 
 	err = nix.IncludeDevboxConfig(ctx, s.username)
-	if err != nil {
+	if errors.Is(err, nix.ErrUnknownServiceManager) {
+		ux.Fwarning(os.Stderr, "Devbox configured Nix to use your organization's private cache. Please restart the Nix daemon and re-run Devbox.\n")
+	} else if err != nil {
```
mikeland73, line 94: Would this show up in CICD even if they don't need to do it?

gcurtis: Fixed.

mikeland73 (approved):

PR #2065, opened 2024-05-20
[debug] Add cache credentials flag and more perf profiling

mikeland73:
## Summary

Credentials command is hidden, this flag can help debugging.

### DIFF
--- a/internal/boxcli/cache.go
+++ b/internal/boxcli/cache.go
@@ -109,7 +109,7 @@ func cacheCredentialsCmd() *cobra.Command {
 				return err
 			}
 
-			if flags.format == "env" {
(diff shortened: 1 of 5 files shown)


internal/boxcli/cache.go
```diff
@@ -103,6 +108,14 @@ func cacheCredentialsCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			if flags.format == "env" {
```
gcurtis, line 112: Just to future-proof ourselves, what about "sh" (following brew and shellcheck's convention) to differentiate from fish or other shells.

gcurtis (approved):

PR #2073, opened 2024-05-22
chore(dev.Dockerfile.tmpl): Remove superfluous step number

silvestre:
### Issue

For some reason only the comment about copying the local flake dirs got a step number. Maybe a leftover?

### Fix

Just remove it, as no other step is numbered.

## How was it tested?

It wasn't as it is only a comment change. 😄


Lagoja (approved):

mohsenari:
Thanks. It was indeed leftover from some changes we did on the template file.

PR #2076, opened 2024-05-23
[perf] Improve install and ensure perf

mikeland73:
## Summary

This improves install and ensure perf by making the following changes:

* Batch all store path existence checks. (previously we did one by one, each `nix ls` call is about 50ms)
* If lockfile has store paths we use them (instead of trying to get them from nix)
* Fix regression in `FillNarInfoCache`

## How was it tested?

Installed large devbox.json (30+ packages) and used `DEVBOX_PRINT_EXEC_TIME` and observed almost all "prep" time disappear. prep time is the work devbox does before actually building the packages.

### DIFF
--- a/internal/nix/store.go
+++ b/internal/nix/store.go
@@ -60,6 +60,9 @@ func StorePathsFromInstallable(ctx context.Context, installable string, allowIns
 // StorePathsAreInStore a map of store paths to whether they are in the store.
 func StorePathsAreInStore(ctx context.Context, storePaths []string) (map[string]bool, error) {
 	defer debug.FunctionTimer().End()
+	if len(storePaths) == 0 {
+		return map[string]bool{}, nil
+	}
 	args := append([]string{"path-info", "--offline", "--json"}, storePaths...)
 	cmd := commandContext(ctx, args...)
 	debug.Log("Running cmd %s", cmd)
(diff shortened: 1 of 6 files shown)


internal/nix/store.go
```diff
@@ -48,49 +48,77 @@ func StorePathsFromInstallable(ctx context.Context, installable string, allowIns
...
 
 // parseStorePathFromInstallableOutput parses the output of `nix store path-from-installable --json`
 // This function is decomposed out of StorePathFromInstallable to make it testable.
-func parseStorePathFromInstallableOutput(installable string, output []byte) ([]string, error) {
-	// Newer nix versions (like 2.20)
+func parseStorePathFromInstallableOutput(output []byte) (map[string]any, error) {
+	// Newer nix versions (like 2.20) have output of the form
+	// {"<store-path>": {}}
+	// if a store path is used as an installable, the keys will be present even if invalid but
+	// the values will be null.
```
savil, line 100: oh good find!

internal/nix/store.go
```diff
@@ -48,49 +48,77 @@ func StorePathsFromInstallable(ctx context.Context, installable string, allowIns
...
-		Valid bool   `json:"valid"`
-	}
+	var out2 []pathInfoLegacy
+
 	if err := json.Unmarshal(output, &out2); err == nil {
-		res := []string{}
+		res := map[string]any{}
 		for _, outValue := range out2 {
-			res = append(res, outValue.Path)
+			if outValue.Valid {
```
savil, line 114: 👍🏾

savil (approved):
LGTM!

PR #2081, opened 2024-05-24
[easy][bug] Fix install if package has no storepaths in lockfile

mikeland73:
## Summary

Fixes issue introduced in https://github.com/jetify-com/devbox/blame/main/internal/devbox/packages.go#L558-L558

## How was it tested?

Removed store paths from lock file, ensured package was still installed.


savil (approved):
Bugs like this really make me wish golang would just introduce a nillable-type. Like the one single lesson they could have learnt from C or Java, but have still failed to.

PR #2082, opened 2024-05-24
[cache] Fallback for failed build with cache

mikeland73:
## Summary

* If we fail to build with caches, try again without them.
* Added logging when build fails and user has substituters.

## How was it tested?

Ran install on devbox project that installs cached terraform. Commented out `--extra-substituters` flag to force it to fail. Observed output indicating substituter error, and fallback to building from source.

image

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -16,7 +16,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
 	"go.jetpack.io/devbox/internal/devbox/devopt"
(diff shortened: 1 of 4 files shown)


internal/devbox/packages.go
```diff
@@ -516,7 +529,6 @@ func (d *Devbox) installNixPackagesToStore(ctx context.Context, mode installMode
 		args.AllowInsecure = allowInsecure
 		err = nix.Build(ctx, args, installables...)
 		if err != nil {
-			color.New(color.FgRed).Fprintf(d.stderr, "Fail\n")
```
savil, line 519: is deleting this a good idea for non-nixcache scenarios (i.e. the common scenario)?

savil (approved):
+1 to sharing the output. Very helpful.

internal/devbox/packages.go
mikeland73: Yeah, we print error a few times (once for nix stderr, and once for devbox failing) so this was a third error (and not very informative because it was just `fail`). The nix output tends to be the most informative.

PR #2092, opened 2024-05-28
Add name/description to devbox json schema.

vhsantos:
## Summary

The current JSON schema does not include the **name** and **description** fields. This update adds these fields as optional properties to the Devbox JSON schema.

## How was it tested?

* Used an online JSON schema validator to ensure that the updated schema is valid.
* Ensured that the schema allows optional name and description fields.

### DIFF
--- a/.schema/devbox.schema.json
+++ b/.schema/devbox.schema.json
@@ -9,6 +9,14 @@
             "description": "The schema version of this devbox.json file.",
             "type": "string"
         },
+        "name": {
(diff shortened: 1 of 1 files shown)


LucilleH:
cc @Lagoja @loreto

Lagoja:
@loreto seems ok to me since the schema keeps those fields optional, and this could be useful metadata for products like Jetify Cloud and others.

loreto (approved):
LGTM

PR #2096, opened 2024-05-28
nixcache: don't check Jetify caches if they're not configured

gcurtis:
Don't check the Jetify Nix caches for store paths if Nix isn't configured to use them. This can happen if the user says "no" to the confirmation prompt for cache setup. Otherwise, `nix build` prints an error about the path missing:

	$ devbox add mongodb@6.0.15
	Info: Adding package "mongodb@6.0.15" to devbox.json
	? You're logged into a Devbox account that now has access to a Nix cache. Allow Devbox to configure Nix to use the new cache (requires sudo)? No
	Info: Skipping cache setup. Run `devbox cache configure` to enable the cache at a later time.
	Info: Installing the following packages to the nix store: mongodb@6.0.15
	error: path '/nix/store/n7fkj8vif9yc1g063r0z88j5dh3i7p8a-mongodb-6.0.15' is required, but there is no substituter that can build it

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"io/fs"
 	"os"
-	"os/user"
 	"path/filepath"
 	"runtime/trace"
 	"slices"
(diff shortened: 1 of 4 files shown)


mikeland73 (approved):
There's another callsite for `CachedReadCaches` that is used for retry logic but I don't think we need any changes there.

PR #2098, opened 2024-05-29
[bug] Fix parseStorePathFromInstallableOutput

mikeland73:
## Summary

Fix regression introduced in https://github.com/jetify-com/devbox/pull/2076

I misunderstood the API. A null value (in modern API) or a valid: false (in legacy API) occur when store path doesn't exist. Whether the store path itself is a valid store path doesn't matter, obviously if it is not a valid path, it won't be present in the store, but if it's valid and not installed, it will return null value or valid: false. The users of this function do not want to ignore these paths.

Edit: This bug is a bit less bad than initially thought. Because we install non-cached packages in flake, so anything missing would be installed then.

## How was it tested?

Unit test. 

This is tricky to test. I added a test script that can hopefully catch this.

Edit: This test doesn't actually catch this because the package ends up getting installed later on (in the flake). I still think this test is good to have.

### DIFF
--- a/testscripts/lockfile/nopaths.txt
+++ b/testscripts/lockfile/nopaths.txt
@@ -1,0 +1,23 @@
+# Test installing a package without outputs in the store path. 
+# NOTE: Purposefully using a weird version to ensure it is not already in store.
+
+exec devbox run curl --version | grep -o 'curl\s7\.87\.0'
+stdout 'curl 7.87.0'
+
+-- devbox.json --
+{
+  "packages": ["curl@7.87.0"],
+}
+
+-- devbox.lock --
+{
+  "lockfile_version": "1",
+  "packages": {
+    "curl@7.87.0": {
+      "last_modified": "2023-02-26T03:47:33Z",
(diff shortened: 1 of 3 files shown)


testscripts/lockfile/nopaths.txt
```diff
@@ -0,0 +1,23 @@
+# Test installing a package without outputs in the store path. 
+# NOTE: Purposefully using a weird version to ensure it is not already in store.
```
savil, line 2: once I run this testscript locally, it will then be in my store indefinitely (or until I run `nix store gc`) So, in that case the testscript won't be testing that its missing from local nix store.

so, the goal of this testscript is then to verify that one can do `devbox install` (equivalent) of "a package that doesn't have ooutputs in store path" ? Feels reasonable.

savil (approved):

testscripts/lockfile/nopaths.txt
mikeland73: yeah, this test mostly just tests that devbox install/run/shell work if the lockfile is missing store paths. I think it is useful (even though it would not catch this particular case where this function was removing store paths that were not installed) 

The unit test you had originally (with valid: false) does catch it! I just misunderstood the API, and broke your test.

PR #2101, opened 2024-05-29
Refactor the list command

josh-d2:
## Summary
Refactor the list command into its own file. Also update the description to be more accurate.

## How was it tested?
1. Verified the `list` command is updated with the new message in the `--help` command output
2. Verified the `list` command outputs the same list of packages before and after this change (compared my local `devbox` install output to the output in my `devbox shell` with these changes).

### DIFF
--- a/internal/boxcli/list.go
+++ b/internal/boxcli/list.go
@@ -22,7 +22,7 @@ func listCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "list",
 		Aliases: []string{"ls"},
-		Short:   "List the packages in your current devbox project",
+		Short:   "List installed packages",
 		PreRunE: ensureNixInstalled,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


Lagoja (approved):

internal/boxcli/list.go
```diff
@@ -0,0 +1,43 @@
...
+type listCmdFlags struct {
+	config configFlags
+}
+
+func listCmd() *cobra.Command {
+	flags := listCmdFlags{}
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List the packages in your current devbox project",
```
mikeland73, line 25: Since this is also used for global, maybe a more generic:

`List installed packages`

mikeland73 (approved):
🎉 🎉 🎉

internal/boxcli/list.go
josh-d2: Updated!

PR #2102, opened 2024-05-29
[perf] Add flag to fix missing store paths

mikeland73:
## Summary

Adds new `--tidy-lockfile` flag that can be used with install. It fills in any missing store paths in lockfile, improving future performance. 

I'm not 100% convinced this works in all cases. My two biggest concerns would be:

* I'm setting all outputs to default.
* re derivation name parsing: Not 100% sure we can rely on versions never having dashes.

## How was it tested?

* unit test
* Manually removed store paths form lockfile and tried to install. Got warning, ran with flag.

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -648,6 +648,9 @@ func (d *Devbox) moveAllowInsecureFromLockfile(writer io.Writer, lockfile *lock.
 func (d *Devbox) FixMissingStorePaths(ctx context.Context) error {
 	packages := d.InstallablePackages()
 	for _, pkg := range packages {
+		if pkg.IsRunX() {
+			continue
+		}
 		existingStorePaths, err := pkg.GetResolvedStorePaths()
 		if err != nil {
 			return err
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -335,6 +335,7 @@ func (d *Devbox) updateLockfile(recomputeState bool) error {
 // - the generated flake
 // - the nix-profile
 func (d *Devbox) recomputeState(ctx context.Context) error {
+	defer debug.FunctionTimer().End()
 	if err := shellgen.GenerateForPrintEnv(ctx, d); err != nil {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


mikeland73:
@Lagoja @gcurtis made updates we talked about:

* Better copy when missing store paths.
* Only show message for devbox packages (i.e. pkg@version)
* Fix store paths on update

Lagoja:
Running this, I get: 

```
Warning: Outputs for libiconv@latest are not in lockfile. Fetching store paths from nix
To fix this issue and improve performance, please run `devbox install --fix-missing-store-paths
```

Are we fetching the store paths at the time that I get this message? Or do I need to run the command with the flag

internal/devpkg/package.go
```diff
@@ -742,13 +742,16 @@ func (p *Package) GetStorePaths(ctx context.Context, w io.Writer) ([]string, err
...
-	ux.Fwarning(
-		w,
-		"Outputs for %s are not in lockfile. Fetching store paths from nix, this may take a while\n",
-		p.Raw,
-	)
+	if p.IsDevboxPackage {
+		// No fast path, we need to query nix.
+		ux.Fwarning(
+			w,
+			"Outputs for %s are not in lockfile. Fetching store paths from nix\n"+
```
Lagoja, line 749:
```suggestion
			"Outputs for %s are not in lockfile.+
```

Lagoja (approved):

mikeland73:
@Lagoja I need to hide the message when using the flag. Will do.

internal/devbox/packages.go
```diff
@@ -643,3 +644,46 @@ func (d *Devbox) moveAllowInsecureFromLockfile(writer io.Writer, lockfile *lock.
...
+			if len(storePaths) == 0 {
+				return fmt.Errorf("no store paths found for package %s", pkg.Raw)
+			}
+			for _, storePath := range storePaths {
+				parts := nix.NewStorePathParts(storePath)
+				outputs = append(outputs, lock.Output{
+					Path: storePath,
+					Name: parts.Output,
+					// Ugh, not sure this is true, but it's more true than not.
+					Default: true,
```
gcurtis, line 680: What's the downside if this ends up being wrong?

gcurtis (approved):
Should we name the flag `--tidy-lockfile`? Are missing store paths in the lockfile a problem? I thought it just made things slower.

internal/devbox/packages.go
mikeland73: not much. We may install a bit more than we're supposed to (on the other hand, if we forget to add this value, that's bad).

mikeland73:
> Should we name the flag --tidy-lockfile? Are missing store paths in the lockfile a problem? I thought it just made things slower.

I like that name. Will change. It gives us flexibility to do more stuff.

missing store paths from lockfile can be a big perf drag, especially if all packages are in cache. If the store paths are missing we end up having to download nixpkgs every time (which adds 40s for each nixpkgs). Particularly painful on CICD because env is fresh every time.

PR #2111, opened 2024-05-31
Add new --pcflags option to devbox services up

josh-d2:
## Summary
Add new `--pcflags` option which provides flags directly to process-compose
## Motivation
https://github.com/jetify-com/devbox/issues/2105
## Usage
The `--pcflags` flag must be provided in front of each flag as shown below:
```
devbox services up ${service} \
--pcflags ${flag1} \
--pcflags ${flag2} \
...
```
## How was it tested?
1. Added debugging print statement in `manager.go`, right before providing the flags to process-compose, to ensure that the flags are being passed down as expected to process-compose.
2. Verified that `devbox services up nginx` shows the process-compose UI but `devbox serviecs up nginx --pcflags -t=false` does not show the UI.

### DIFF
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -239,7 +239,7 @@ func startProcessManager(
 		args,
 		devopt.ProcessComposeOpts{
 			Background: flags.background,
-			Flags:      flags.processComposeFlags,
+			ExtraFlags: flags.processComposeFlags,
 		},
 	)
 }
--- a/internal/boxcli/services.go
+++ b/internal/boxcli/services.go
@@ -17,8 +17,9 @@ type servicesCmdFlags struct {
 }
 
 type serviceUpFlags struct {
-	background         bool
-	processComposeFile string
+	background          bool
+	processComposeFile  string
+	processComposeFlags []string
 }
 
 type serviceStopFlags struct {
@@ -35,6 +36,8 @@ func (flags *serviceUpFlags) register(cmd *cobra.Command) {
 	)
 	cmd.Flags().BoolVarP(
 		&flags.background, "background", "b", false, "run service in background")
+	cmd.Flags().StringArrayVar(
+		&flags.processComposeFlags, "pcflags", []string{}, "pass flags directly to process compose")
 }
 
 func (flags *serviceStopFlags) register(cmd *cobra.Command) {
@@ -151,6 +154,8 @@ func startServices(cmd *cobra.Command, services []string, flags servicesCmdFlags
 		Environment: flags.config.environment,
 		Env:         env,
 		Stderr:      cmd.ErrOrStderr(),
+		// TODO: Should we allow these options to be passed here like we do for service up?
+		ProcessComposeOpts: &devopt.ProcessComposeOpts{Background: true},
 	})
 	if err != nil {
 		return errors.WithStack(err)
@@ -178,6 +183,7 @@ func stopServices(
 	if err != nil {
 		return errors.WithStack(err)
 	}
+
 	if len(services) > 0 && flags.allProjects {
 		return errors.New("cannot use both services and --all-projects arguments simultaneously")
 	}
@@ -199,6 +205,8 @@ func restartServices(
 		Environment: flags.config.environment,
 		Env:         env,
 		Stderr:      cmd.ErrOrStderr(),
+		// TODO: Should we allow these options to be passed here like we do for service up?
+		ProcessComposeOpts: &devopt.ProcessComposeOpts{Background: true},
 	})
 	if err != nil {
 		return errors.WithStack(err)
@@ -218,11 +226,15 @@ func startProcessManager(
 		return err
 	}
 	box, err := devbox.Open(&devopt.Opts{
-		Dir:                      servicesFlags.config.path,
-		Env:                      env,
-		Environment:              servicesFlags.config.environment,
-		CustomProcessComposeFile: flags.processComposeFile,
-		Stderr:                   cmd.ErrOrStderr(),
+		Dir:         servicesFlags.config.path,
+		Env:         env,
+		Environment: servicesFlags.config.environment,
+		Stderr:      cmd.ErrOrStderr(),
+		ProcessComposeOpts: &devopt.ProcessComposeOpts{
+			CustomFile: flags.processComposeFile,
+			Flags:      flags.processComposeFlags,
+			Background: flags.background,
+		},
 	})
 	if err != nil {
 		return errors.WithStack(err)
@@ -232,7 +244,5 @@ func startProcessManager(
 		cmd.Context(),
 		servicesFlags.runInCurrentShell,
 		args,
-		flags.background,
-		flags.processComposeFile,
 	)
 }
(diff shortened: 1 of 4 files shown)


internal/boxcli/services.go
```diff
@@ -218,11 +226,15 @@ func startProcessManager(
...
-		CustomProcessComposeFile: flags.processComposeFile,
-		Stderr:                   cmd.ErrOrStderr(),
+		Dir:         servicesFlags.config.path,
+		Env:         env,
+		Environment: servicesFlags.config.environment,
+		Stderr:      cmd.ErrOrStderr(),
+		ProcessComposeOpts: &devopt.ProcessComposeOpts{
+			CustomFile: flags.processComposeFile,
+			Flags:      flags.processComposeFlags,
+			Background: flags.background,
```
mikeland73, line 236: I lean towards leaving on the `CustomProcessComposeFile` as an option here. It's usually best to pass arguments to individual methods and only use these more global options for things that affect many or all devbox methods.

In this case, the custom file affects the list of services so this is pretty wide ranging. The flags and background would only affect services up.

internal/boxcli/services.go
```diff
@@ -232,7 +244,5 @@ func startProcessManager(
 		cmd.Context(),
 		servicesFlags.runInCurrentShell,
 		args,
-		flags.background,
-		flags.processComposeFile,
```
mikeland73, line 236: `flags.processComposeFile` can be removed safely (since it is passed as a devbox option), but I think we want to keep background here.

internal/boxcli/services.go
```diff
@@ -199,6 +205,8 @@ func restartServices(
 		Environment: flags.config.environment,
 		Env:         env,
 		Stderr:      cmd.ErrOrStderr(),
+		// TODO: Should we allow these options to be passed here like we do for service up?
+		ProcessComposeOpts: &devopt.ProcessComposeOpts{Background: true},
```
mikeland73, line 209: I lean towards only implementing the passthrough flags for up. 

For other commands, we don't always start the process compose server so that flags would be ignored sometimes.

internal/devbox/devbox.go
```diff
@@ -549,7 +549,11 @@ func (d *Devbox) Services() (services.Services, error) {
 		return nil, err
 	}
 
-	userSvcs := services.FromUserProcessCompose(d.projectDir, d.customProcessComposeFile)
+	customFile := ""
+	if d.processComposeOpts != nil {
+		customFile = d.processComposeOpts.CustomFile
+	}
+	userSvcs := services.FromUserProcessCompose(d.projectDir, customFile)
```
mikeland73, line 556: To avoid code like this, it can be nicer to use structs instead of pointers. Golang defaults all types to a zero value, including structs. The string zero value is "" and the struct zero value is a struct with all fields having zero value. 

```golang
type child struct {
  Data string
}
type parent struct
  c child
}

p := parent{}
fmt.Println(p.c.Data) // prints out "", no nil pointer issues.
```

On the other hand, all pointer types zero value is nil. So:

```golang
type child struct {
  Data string
}
type parent struct
  c *child
}

p := parent{}
fmt.Println(p.c.Data) // panic! nil pointer
```

More Golang pecuilarities:

Slices and maps have zero value nil as well, but they can be used in most read cases like an empty slice/map. e.g.

```golang
var mySlice []string
fmt.Println(mySlice == nil) // true
fmt.Println(len(mySlice)) // 0

var myMap map[string]string
fmt.Println(myMap == nil) // true
fmt.Println(len(myMap)) // 0
fmt.Println(myMap["foo"]) // "", no panic

myMap["bar"] = "bar2" // will panic! can't add value to nil map
```

Doing `range` over nil slices and maps is also allowed.

Finally, Golang allows method calls on nil receivers.

```golang
func (f *foo) MyMethod() {
  if f == nil {
    fmt.Println("I'm nil and it's fine")
  }
}

var f *foo // initializes f to be a nil pointer
f.MyMethod() // this is fine
```

but using a field of a nil pointer will cause a panic:

```golang
var f *foo
f.MyField // panics!
```

internal/services/manager.go
```diff
@@ -150,13 +154,16 @@ func StartProcessManager(
 		flags = append(flags, "-f", s.ProcessComposePath)
 	}
 
-	if processComposeBackground {
-		flags = append(flags, "-t=false")
-		cmd := exec.Command(processComposeBinPath, flags...)
+	if len(processComposeConfig.Flags) > 0 {
```
mikeland73, line 157: `if` not needed

internal/services/manager.go
```diff
@@ -40,6 +40,12 @@ type globalProcessComposeConfig struct {
 	File      *os.File `json:"-"`
 }
 
+type ProcessComposeOpts struct {
```
mikeland73, line 43: Nice!

internal/devbox/devbox.go
```diff
@@ -761,25 +765,42 @@ func (d *Devbox) StartProcessManager(
...
+	return services.StartProcessManager(
+		d.stderr,
+		requestedServices,
+		svcs,
+		d.projectDir,
+		*servicesProcessComposeConfig,
+	)
+}
+
+func (d *Devbox) configureProcessCompose(ctx context.Context) (*services.ProcessComposeOpts, error) {
```
mikeland73, line 784: If you don't use a pointer you can avoid the nil checks later on

mikeland73 (changes requested):
This looks pretty good, probably the main blocker is reverting the change to `devopt.Opts` and keeping the scope to be `services up` only. I think PR will get simpler after that.

For some context, I don't think we want people using `services start` or `restart` very much to start services. @Lagoja could give more context (I don't fully remember why, but `up` is the preferred API)

internal/devbox/devbox.go
josh-d2: Ahhh interesting about being able to call methods on nil receivers but not accessing fields on a nil...thanks for the explanation!

internal/services/manager.go
```diff
@@ -40,6 +40,12 @@ type globalProcessComposeConfig struct {
...
+	BinPath    string
+	Flags      []string
```
mikeland73, line 45: nit, rename to `ExtraFlags`

internal/devbox/devopt/devboxopts.go
```diff
@@ -19,6 +19,11 @@ type Opts struct {
 	Stderr                   io.Writer
 }
 
+type ProcessComposeOpts struct {
+	Flags      []string
```
mikeland73, line 23: nit, rename to `ExtraFlags`

internal/devbox/devbox.go
```diff
@@ -731,18 +731,25 @@ func (d *Devbox) StartProcessManager(
...
 ) error {
 	if !runInCurrentShell {
 		args := []string{"services", "up", "--run-in-current-shell"}
 		args = append(args, requestedServices...)
-		if processComposeFileOrDir != "" {
-			args = append(args, "--process-compose-file", processComposeFileOrDir)
+
+		// TODO: Here we're attempting to reconstruct arguments from the original command, so that we can reinvoke it in devbox shell.
+		// 		 Instead, we should consider refactoring this so that we can preserve and re-use the original command string,
+		//		 because the current approach is fragile and will need to be updated each time we add new flags.
```
mikeland73, line 742: Nice, thanks!

mikeland73 (approved):
lgtm!

PR #2118, opened 2024-06-03
[docker] Added fix to run image release on tags creation + backfill action

mohsenari:
## Summary
Github action for releasing docker images to dockerhub didn't run when we created a new tag (0.11.0) this fixes that issue.
Plus I added a new github action that can only be triggered manually. It takes inputs for tag/version name and releases docker image to dockerhub at that specific tag. This is to backfill older versions of devbox to dockerhub as well.

## How was it tested?
Will have to be tested via github actions run on a tag creation. But backfill will be tested manually


mikeland73 (changes requested):
I'm a bit confused by this action. It doesn't seem like the version is added to the image at all? Is this just meant to refresh the latest image? (`DEVBOX_USE_VERSION` is not set, so it looks like no matter what the docker image will use the latest version)

Also, are the actions identical? You should just add `workflow_dispatch` to existing one and can have `ref` defaulted correctly.

mohsenari:
> I'm a bit confused by this action. It doesn't seem like the version is added to the image at all? Is this just meant to refresh the latest image? (`DEVBOX_USE_VERSION` is not set, so it looks like no matter what the docker image will use the latest version)

The repo checkout action checks out the code at the provided tag. But good point on not having `DEVBOX_USE_VERSION` I added that to the build args for dockerfile and the action and tested locally to make sure it build the provided version.

> 
> Also, are the actions identical? You should just add `workflow_dispatch` to existing one and can have `ref` defaulted correctly.

The actions are similar but not identical. Frankly, I don't know how I can branch the logic in a YAML file to do the current tag and update latest if inputs aren't provided. This action can be deleted once we successfully backfill the older devbox versions in dockerhub

mikeland73 (commented):
What I had in mind was something like https://github.com/jetify-com/devbox/commit/854361958d374a155f01dd8decd9ac3353f2f7fc

I'm still a bit confused about what we're doing with versions. Currently, every time a tag is pushed we publish a latest without a version. I guess this is fine (it basically helps update the launcher, etc)

This does mean we don't really need to "backfill" anything. We've never actually published individual tags. I think just adding `workflow_dispatch` to the existing job is probably enough.

For future thought:
Ideally we publish latest only for newest tag (there's a latest=auto on the docker meta action, but I'm not sure how it works). In addition we also publish each individual tag with the env variable. Ideally we also pre-install the binary on the docker image so it doesn't need to be installed after the container is initialized.

mikeland73 (approved):

PR #2122, opened 2024-06-04
Moved process-compose into a generic utilities Devbox project

josh-d2:
## Summary
Moved process-compose into a generic utilities Devbox project.

This PR deletes `addDevboxUtilityPackage` and `removeDevboxUtilityPackage`, since they're no longer used.

## Motivation
https://github.com/jetify-com/devbox/issues/2106

## How was it tested?
1. Manually deleted `~/.local/share/devbox/util` directory and verified that it is re-created when starting a service, and that process-compose is correctly added as a dependency with the binary present at the expected path.
2. Ensured that `devbox services up nginx` starts the service and that the process-compose UI is displayed correctly.


internal/devbox/util.go
```diff
@@ -107,3 +45,63 @@ func utilityBinPath() (string, error) {
 	}
 	return filepath.Join(nixProfilePath, "bin"), nil
 }
+
+// TODO: Commented these out so that lint doesn't complain that they're unused, but leaving them here in case we ever
```
mikeland73, line 49: I think it's fine to remove. Keeping around means they will show up in grep, etc

If we ever need to bring bag, Adding a comment to PR description can make it easier to find.

internal/devbox/process_compose.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/devbox/devopt"
+	"go.jetpack.io/devbox/internal/xdg"
+)
+
+const processComposeVersion = "1.5.0"
+
+var processComposeConfigPath string
+
+func InitializeProcessCompose(ctx context.Context, stderr io.Writer) (string, error) {
```
mikeland73, line 21: function name should be camelCase since it is not used outside the package.

internal/devbox/process_compose.go
```diff
@@ -0,0 +1,71 @@
...
+	pcDevboxProjectPath, err := ensurePcDevboxProject()
```
mikeland73, line 22: Super nit: golang style guide says letters of acronyms should be all upper case (so this would be `ensurePCDevboxProject`).

That said, I think we should use a more generic name for this project. e.g. `ensureDevboxUtilitiesProject`

internal/devbox/process_compose.go
```diff
@@ -0,0 +1,71 @@
...
+	if err = box.Add(ctx, []string{"process-compose@" + processComposeVersion}, devopt.AddOpts{}); err != nil {
+		return "", err
+	}
+
+	err = box.Install(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return utilityLookPath("process-compose")
```
mikeland73, line 44: Are you sure this is working? I think this may be finding the old process-compose.

internal/devbox/process_compose.go
```diff
@@ -0,0 +1,71 @@
...
+	return pcDevboxProjectDir, nil
+}
+
+func createPcDevboxProjectDir() (string, error) {
+	path := xdg.DataSubpath(filepath.Join("devbox/util/process-compose"))
+	if err := os.MkdirAll(path, 0o755); err != nil {
+		return "", errors.WithStack(err)
+	}
+	return path, nil
+}
```
mikeland73, line 71: I think we should make this more generic. Since a single devbox project can contain multiple utilities, I think we can just use `xdg.DataSubpath(filepath.Join("devbox/util"))` as the root for that project.

mikeland73 (changes requested):
RC mostly because I'm not sure `utilityLookPath` is working as expected. I think it is returning the previous process compose binary.

internal/devbox/process_compose.go
josh-d2: Got it...I thought this was a specific project just for process-compose, but that makes sense that it could be more generic for additional dependencies in the long-term. I'll update the naming to be more generic here.

internal/devbox/process_compose.go
josh-d2: Yup you're right. I manually deleted the process-compose utility and then re-ran it and it failed. I'll fix it to point to the correct binary and update the PR.

internal/devbox/util.go
```diff
@@ -5,73 +5,70 @@ package devbox
...
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
 
-	for _, installable := range installables {
+	// Add all utilities here.
+	utilities := []string{
+		"process-compose@" + processComposeVersion,
+	}
```
mikeland73, line 40: I think there's a chance that in the future we may want to install as needed. But since we only have one, this if fine.

mikeland73 (approved):
Nice! this really simplifies code and should perform better as well.

nit: remove `processComposeTargetVersion`

PR #2123, opened 2024-06-05
Permissions-Policy is Only Served on HTML Content-Type

gitworkflows:
The application configurable Permissions-Policy is only served on responses
with an HTML related Content-Type.

This has been assigned the CVE identifier https://github.com/advisories/GHSA-fwhr-88qx-h9g7.

Versions Affected: >= 6.1.0
Not affected: < 6.1.0
Fixed Versions: 6.1.7.8, 7.0.8.4, and 7.1.3.4

Impact

Responses with a non-HTML Content-Type are not serving the configured Permissions-Policy. There are certain non-HTML Content-Types that would benefit from having the Permissions-Policy enforced.

Releases

The fixed releases are available at the normal locations.


gcurtis:
Appreciate the PR. This was merged in #2127.

PR #2130, opened 2024-06-06
[perf] Install all packages to profile in a single nix command

mikeland73:
## Summary

This change installs all packages into profile with same command. If we fail due to priority conflict, we fallback to the previous one by one installation.

This significantly improves performance on projects with many packages when all packages are already in store. This aims to solve a common CICD case where nothing has changed, all packages are in store, but we need to create the .devbox directory and nix profile from scratch.

cc: @Lagoja this helps address some reported issues.

## How was it tested?

Happy path:
On project with 30+ packages, delete `.devbox` directory and ran `devbox install`. This change reduced total time from 10s to 2s.

Sad path:
On new project added `curl`, `ruby`, `bundler`, `go`. `ruby` and `bundler` conflict. I ran `devbox install` and it correctly fell back to one by one. I inspected the profile to ensure nix did not install anything twice.

### DIFF
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -2,6 +2,7 @@ package devbox
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


josh-d2 (approved):
Nice, lgtm!

gcurtis (approved):

savil (commented):
Can you test to make sure the insecure-package scenario is properly handled?

PR #2136, opened 2024-06-10
[docker] fix docker release image

mohsenari:
## Summary
This PR addresses the issue of our previous approach to only publish the latest version of devbox with `:latest` tag in dockerhub. It achieves 2 things:
1. fixes the previously broken github action to release a docker image after every tag creation.
2. updates the dockerfile so that devbox binary doesn't automatically upgrade to latest version upon first use. This allows docker pulls have fixed versions.

## How was it tested?
Already tested on releasing 0.10.7 version to dockerhub as test.
- `docker pull jetpackio/devbox:0.10.7`
- `docker run jetpackio/devbox:0.10.7`
- confirm output shows devbox version 0.10.7 and doesn't upgrade to latest version.
- same for devbox-root-user

### DIFF
--- a/.github/workflows/docker-image-release.yml
+++ b/.github/workflows/docker-image-release.yml
@@ -55,7 +55,7 @@ jobs:
           context: ./internal/devbox/generate/tmpl/
           file: ./internal/devbox/generate/tmpl/DevboxImageDockerfile
           build-args: |
-            DEVBOX_USE_VERSION=${{ inputs.tag }}
+            DEVBOX_VERSION=${{ inputs.tag }}
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
@@ -65,7 +65,7 @@ jobs:
           context: ./internal/devbox/generate/tmpl/
           file: ./internal/devbox/generate/tmpl/DevboxImageDockerfileRootUser
           build-args: |
-            DEVBOX_USE_VERSION=${{ inputs.tag }}
+            DEVBOX_VERSION=${{ inputs.tag }}
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.metaroot.outputs.tags }}
--- a/.github/workflows/docker-image-release.yml
+++ b/.github/workflows/docker-image-release.yml
@@ -9,7 +9,7 @@ on:
       tag:
         description: 'tag name'
         required: true
-        default: 'undefined'
+        default: ''
         type: string
 
 jobs:
@@ -18,8 +18,6 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v4
-        with:
-          ref: ${{ inputs.tag || github.ref }}
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
       - name: Docker meta
@@ -29,9 +27,7 @@ jobs:
           images: |
             jetpackio/devbox
           tags: |
-            type=semver,pattern={{version}},value=${{ inputs.tag }}
-          flavor: |
-            latest=${{ !inputs.tag && true || false }}
+            type=raw,value=${{ inputs.tag || github.ref_name }}
       - name: Docker meta root
         id: metaroot
         uses: docker/metadata-action@v5
@@ -39,9 +35,7 @@ jobs:
           images: |
             jetpackio/devbox-root-user
           tags: |
-            type=semver,pattern={{version}},value=${{ inputs.tag }}
-          flavor: |
-            latest=${{ !inputs.tag && true || false }}
+            type=raw,value=${{ inputs.tag || github.ref_name }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Login to Docker Hub
@@ -55,7 +49,7 @@ jobs:
           context: ./internal/devbox/generate/tmpl/
           file: ./internal/devbox/generate/tmpl/DevboxImageDockerfile
           build-args: |
-            DEVBOX_VERSION=${{ inputs.tag || github.ref }}
+            DEVBOX_USE_VERSION=${{ inputs.tag || github.ref_name }}
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
@@ -65,7 +59,7 @@ jobs:
           context: ./internal/devbox/generate/tmpl/
           file: ./internal/devbox/generate/tmpl/DevboxImageDockerfileRootUser
           build-args: |
-            DEVBOX_VERSION=${{ inputs.tag || github.ref }}
+            DEVBOX_USE_VERSION=${{ inputs.tag || github.ref_name }}
           push: true
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.metaroot.outputs.tags }}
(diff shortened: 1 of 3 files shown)


internal/devbox/generate/tmpl/DevboxImageDockerfile
```diff
@@ -1,7 +1,7 @@
 FROM debian:stable-slim
 
 # Optional arg to install custom devbox version
-ARG DEVBOX_USE_VERSION
+ARG DEVBOX_VERSION
```
mohsenari, line 4: This has to be different from `ENV DEVBOX_USE_VERSION` otherwise docker ignores the ENV

internal/devbox/generate/tmpl/DevboxImageDockerfile
```diff
@@ -21,8 +21,21 @@ RUN . ~/.nix-profile/etc/profile.d/nix.sh
...
-CMD ["devbox", "version"]
+# run a devbox command to make launcher download devbox binary
+RUN devbox version
+
+# replace binary with launcher
+USER root
+RUN if [ $(uname -m) = "aarch64" ]; \
+    then mv -f /home/devbox/.cache/devbox/bin/${DEVBOX_VERSION}_linux_arm64/devbox /usr/local/bin; \
+    else mv -f /home/devbox/.cache/devbox/bin/${DEVBOX_VERSION}_linux_amd64/devbox /usr/local/bin; \
+    fi
```
mohsenari, line 38: We publish images for both `arm64` and `amd64` in the same github action. The devbox binary location changes based on cpu architecture. So this conditional `RUN` command is necessary.

.github/workflows/docker-image-release.yml
```diff
@@ -29,19 +29,19 @@ jobs:
           images: |
             jetpackio/devbox
           tags: |
-            type=semver,pattern={{version}}
+            type=semver,pattern={{version}},value=${{ inputs.tag }}
           flavor: |
-            latest=${{ inputs.tag && true || false }}
+            latest=${{ !inputs.tag && true || false }}
```
savil, line 34: this is saying: latest is true if `inputs.tag` is falsy. Is that correct?

Do we need the `&& true` anymore? My guess is 'no'

.github/workflows/docker-image-release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           context: ./internal/devbox/generate/tmpl/
           file: ./internal/devbox/generate/tmpl/DevboxImageDockerfile
           build-args: |
-            DEVBOX_USE_VERSION=${{ inputs.tag }}
+            DEVBOX_VERSION=${{ inputs.tag || github.ref }}
```
savil, line 58: A couple questions, since i'm not very familiar with this code from before. Trying to follow along the logic.
1. Why do we add the `github.ref` here? what does it signify?
2. ~Why change from `DEVBOX_USE_VERSION` to `DEVBOX_VERSION`?~ Ah, you've answered below. Cool.

internal/devbox/generate/tmpl/DevboxImageDockerfile
```diff
@@ -21,8 +21,21 @@ RUN . ~/.nix-profile/etc/profile.d/nix.sh
...
 RUN wget --quiet --output-document=/dev/stdout https://get.jetify.com/devbox   | bash -s -- -f
 RUN chown -R "${DEVBOX_USER}:${DEVBOX_USER}" /usr/local/bin/devbox
+# stop launcher from auto updating
+RUN mkdir -p $HOME/.config/devbox/ && echo "false" > $HOME/.config/devbox/auto_update
 
-CMD ["devbox", "version"]
+# run a devbox command to make launcher download devbox binary
+RUN devbox version
+
+# replace binary with launcher
```
savil, line 33: This should be `replace launcher with binary`, right?

savil (approved):
Approving to unblock, thanks for fixing this! One question about `github.ref`, the need for which I didn't quite follow.

mikeland73 (changes requested):
@mohsenari let's chat about this. I'm not confident removing the launcher won't have side effects. Pretty much 100% of users use the launcher. 

I tried doing `DEVBOX_USE_VERSION=0.10.7 devbox version` on our latest image and it worked as expected. Is this not working for you?

.github/workflows/docker-image-release.yml
```diff
@@ -29,19 +29,19 @@ jobs:
...
         id: metaroot
         uses: docker/metadata-action@v5
         with:
           images: |
             jetpackio/devbox-root-user
           tags: |
-            type=semver,pattern={{version}}
+            type=semver,pattern={{version}},value=${{ inputs.tag }}
           flavor: |
-            latest=${{ inputs.tag && true || false }}
```
mikeland73, line 44: Did you ever try removing this (which would cause it to default do `latest=auto`?

The current setting (before and after this PR) are both wrong. It would be possible to create a tag that is not latest and we would certainly not want to update the `latest` image in that case.

.github/workflows/docker-image-release.yml
mikeland73: because this workflow is only triggered automatically with a tag push`github.ref` would be a tag (version).

It is a bit fragile, if we trigger it for another reason or with an invalid tag then it would create a broken image.

A better solution may be to only add `DEVBOX_USE_VERSION` for publishing non-latest (which is why I think the previous implementation with `DEVBOX_USE_VERSION=${{ inputs.tag }}` is less fragile).

.github/workflows/docker-image-release.yml
mohsenari: I looked into documentation for `auto`. It adds latest tag if the code is from the `master` branch. There is also a setting to configure `tag` to have `latest` value if the code is on the default branch (`main` in our case). Both cases don't really apply to our use case. We want to have latest when the workflow is triggered by a tag creation and not have latest when we put a custom tag in the input and trigger the workflow manually.
Initially, I had 2 different yaml files for them but you suggested merging them since they were very similar. 
How is the setting in this PR wrong? it's saying if `input.tag` is not available (meaning the workflow was triggered by a tag creation event) latest is true. Which is what we want.

mohsenari:
Updated the PR based on huddle discussion with @mikeland73 
The new approach is to set `latest` tag to false and remove replacing the launcher logic. In a follow up PR I'll add a job to update latest tag as well.

.github/workflows/docker-image-release.yml
```diff
@@ -29,19 +27,15 @@ jobs:
           images: |
             jetpackio/devbox
           tags: |
-            type=semver,pattern={{version}}
-          flavor: |
-            latest=${{ inputs.tag && true || false }}
```
mikeland73, line 34: I think we want `latest=false` no?

.github/workflows/docker-image-release.yml
```diff
@@ -29,19 +27,15 @@ jobs:
...
       - name: Docker meta root
         id: metaroot
         uses: docker/metadata-action@v5
         with:
           images: |
             jetpackio/devbox-root-user
           tags: |
-            type=semver,pattern={{version}}
-          flavor: |
-            latest=${{ inputs.tag && true || false }}
```
mikeland73, line 44: ditto, latest=false

mikeland73 (changes requested):

mikeland73 (commented):
the default is latest=auto, so I think we want to explicitly make it false

mikeland73 (approved):

PR #2142, opened 2024-06-12
[Remove Nixpkgs] remove feature flag

savil:
## Summary

We've had this feature launched for quite a while. Initially, I was reluctant to remove the feature-flag so that we could debug code paths for users on old versions of nix (which cannot benefit from the binary cache. IIRC nix < 1.17). However, in the time since I relied on this exactly once for some perf-related task, and not otherwise.

At this point, I think we should remove it, and even print messages to users about updating their nix version if needed.

## How was it tested?

It compiles. I could do `devbox shell` in this project.


internal/shellgen/tmpl/flake.nix.tmpl
```diff
@@ -41,15 +35,42 @@
...
+            (builtins.trace "evaluating {{.}}" {{.}})
             {{- end }}
             {{- end }}
           ];
         };
-      }
-    );
-}
+      };
+ }
```
savil, line 76: this file was created via 
`mv flake_remove_nixpkgs.nix.tmpl flake.nix.tmpl`

internal/devbox/update.go
```diff
@@ -143,39 +142,36 @@ func (d *Devbox) mergeResolvedPackageToLockfile(
...
+	}
+	if updated {
+		// if we are updating the system info, then we should also update the other fields
+		useResolvedPackageInLockfile(lockfile, pkg, resolved, existing)
 
-			ux.Finfo(d.stderr, "Updated system information for %s\n", pkg)
-			return nil
-		}
+		ux.Finfo(d.stderr, "Updated system information for %s\n", pkg)
+		return nil
```
savil, line 174: to review this file, press Hide Whitespace in Github

mikeland73 (approved):

PR #2147, opened 2024-06-14
[RFC][Devbox] computeEnv: remove __ETC_PROFILE_NIX_SOURCED=1

savil:
## Summary

This was introduced in #635 (cc @ipince). I don't think its needed anymore. 
It seems to be used in the `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` script, at the very top to prevent re-execution:
```
  1 # Only execute this file once per shell.
  2 if [ -n "${__ETC_PROFILE_NIX_SOURCED:-}" ]; then return; fi
  3 __ETC_PROFILE_NIX_SOURCED=1
  4
```
Perhaps it has other use-cases as well? RFC.

From my casual usage of a Devbox binary built with this change, it all seems to still work fine.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -892,10 +892,6 @@ func (d *Devbox) computeEnv(ctx context.Context, usePrintDevEnvCache bool) (map[
 		env[key] = val.Value.(string)
 	}
 
-	// These variables are only needed for shell, but we include them here in the computed env
-	// for both shell and run in order to be as identical as possible.
-	env["__ETC_PROFILE_NIX_SOURCED"] = "1" // Prevent user init file from loading nix profiles
-
 	slog.Debug("nix environment PATH", "path", env)
 
 	env["PATH"] = envpath.JoinPathLists(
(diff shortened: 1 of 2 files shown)


gcurtis (approved):
I can't remember why this was explicitly added either. Does `nix-daemon.sh` still only run once in `devbox shell` and `devbox shell --pure`? I think a good way to check is to see if there are duplicate paths in `NIX_PROFILES`.

savil:
> I can't remember why this was explicitly added either. Does `nix-daemon.sh` still only run once in `devbox shell` and `devbox shell --pure`? I think a good way to check is to see if there are duplicate paths in `NIX_PROFILES`.


My observations. 
```
# start with devbox global 
 % env | grep NIX_PROFILE
NIX_PROFILES=/nix/var/nix/profiles/default /Users/savil/.nix-profile

# go to devbox repo and start the shell
% cd ~/code/jetpack/devbox
% devbox shell
Starting a devbox shell...
 % env | grep NIX_PROFILE
NIX_PROFILES=/nix/var/nix/profiles/default /Users/savil/.nix-profile
```
no duplicates

gcurtis:
Thanks for double checking!

PR #2149, opened 2024-06-14
[devbox] computeEnv: move devboxEnvPath variable earlier in the funct…

savil:
…ion to be accurate

## Summary

This TODO was in the code, which I'm removing. I move `devboxEnvPath` variable to be higher in the function so it is more accurate to its _meaning_.

## How was it tested?

testscript tests should pass
did a sanity test of using the binary in the devbox repo which `devbox shell`

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -892,12 +892,12 @@ func (d *Devbox) computeEnv(ctx context.Context, usePrintDevEnvCache bool) (map[
 		env[key] = val.Value.(string)
 	}
 
-	slog.Debug("nix environment PATH", "path", env)
+	// devboxEnvPath starts with the initial PATH from print-dev-env, and is
+	// transformed to be the "PATH of the Devbox environment"
+	devboxEnvPath := env["PATH"]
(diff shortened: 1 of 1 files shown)


internal/devbox/devbox.go
```diff
@@ -892,12 +892,12 @@ func (d *Devbox) computeEnv(ctx context.Context, usePrintDevEnvCache bool) (map[
...
+	// devboxEnvPath starts with the initial PATH from print-dev-env, and is
+	// transformed to be the "PATH of the Devbox environment"
+	devboxEnvPath := env["PATH"]
+	slog.Debug("nix environment PATH", "path", devboxEnvPath)
 
-	env["PATH"] = envpath.JoinPathLists(
-		nix.ProfileBinPath(d.projectDir),
-		env["PATH"],
-	)
+	devboxEnvPath = envpath.JoinPathLists(nix.ProfileBinPath(d.projectDir), devboxEnvPath)
```
savil, line 900: ugh, this won't work. `d.configEnvs` can affect PATH

PR #2150, opened 2024-06-15
[devbox global] add `--omit-nix-env` flag for shellenv/shell/run commands

savil:
…vars

## Summary

In this PR, we change `devbox global` shell environment to omit the env-vars from `nix print-dev-env`. Instead, we rely on the `nix profile` that Devbox manages to introduce the global packages into `PATH`. As before, `nix profile` continues to be generated from the `buildInputs` from `nix print-dev-env <devbox-generated-flake>`.

The motivation is:
1. `devbox global` adds a bunch of nix stdenv packages to the top of your `$PATH` variable, which can cause conflicts when trying to compile or build projects on your machine. 
2. `devbox global` also sets a global `PYTHONPATH` variable that interferes with other packages, as well as python scripts that are installed/running on the host. This global `PYTHONPATH` is unnecessary because any python binaries installed by the user are already wrapped with the `PYTHONPATH`

Implementation notes:
- Sets `Devbox.envForPackageBins` boolean setting that controls whether the Devbox Environment is optimized for executing package binaries, or for developing software using the packages.
   - [x] see if this can be passed down as a function parameter (see #2159)
- Adds `flagDefaultOptions` to `shellEnvCmd`. This is a nicer way of overriding the defaults in global, even for the `--recompute` flag that we had done before.
- Refactors code into a `devbox.execPrintDevEnv` function to remove some of the complexity from the `devbox.computeEnv` function (thanks to finger-wagging by our linter tool ;) ).

TODO:
- [x] Verify `omitNixEnv` is still properly handling the useNixPrintDevEnvCache call paths

## How was it tested?

CICD tests should pass

TODO:
- [x] Ensure the python global scenario works
- [x] Ensure we can use common packages (fzf, ripgrep, etc.) in `devbox global`
- [x] Thanks to @gcurtis vetted this PR helps resolve issues with a couple of repos:
   (rust: https://github.com/zed-industries/zed.git) and (golang with cgo: [github.com/Code-Hex/vz/example/macOS](http://github.com/Code-Hex/vz/example/macOS))

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -15,7 +15,9 @@ import (
 
 type shellEnvCmdFlags struct {
 	envFlag
-	config            configFlags
+	config configFlags
+	// TODO: is this flag needed in other commands?
+	envForPackageBins bool
 	install           bool
 	noRefreshAlias    bool
 	preservePathStack bool
@@ -24,11 +26,37 @@ type shellEnvCmdFlags struct {
 	runInitHook       bool
 }
 
-func shellEnvCmd() *cobra.Command {
+type shellenvFlagDefaults struct {
+	envForPackageBins bool
+	recomputeEnv      bool
+}
+
+type shellenvFlagDefault func(*shellenvFlagDefaults)
+
+func withEnvForPackageBins(envForPackageBins bool) shellenvFlagDefault {
+	return func(o *shellenvFlagDefaults) {
+		o.envForPackageBins = envForPackageBins
+	}
+}
+
+func withRecompute(recompute bool) shellenvFlagDefault {
+	return func(o *shellenvFlagDefaults) {
+		o.recomputeEnv = recompute
+	}
+}
+
+func shellEnvCmd(opts ...shellenvFlagDefault) *cobra.Command {
+	defaults := shellenvFlagDefaults{
+		recomputeEnv: true,
+	}
+	for _, opt := range opts {
+		opt(&defaults)
+	}
+
 	flags := shellEnvCmdFlags{}
 	command := &cobra.Command{
 		Use:     "shellenv",
-		Short:   "Print shell commands that add Devbox packages to your PATH",
+		Short:   "Print shell commands that create a Devbox Environment in the shell",
 		Args:    cobra.ExactArgs(0),
 		PreRunE: ensureNixInstalled,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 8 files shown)


internal/boxcli/shellenv.go
```diff
@@ -24,11 +26,37 @@ type shellEnvCmdFlags struct {
...
+	return func(o *shellenvFlagDefaults) {
+		o.envForPackageBins = envForPackageBins
+	}
+}
+
+func withRecompute(recompute bool) shellenvFlagDefault {
+	return func(o *shellenvFlagDefaults) {
+		o.recomputeEnv = recompute
+	}
+}
```
mikeland73, line 46: I think the `WithOpt` pattern is nice for APIs, but feels a bit heavy handed for an internal function?

internal/boxcli/shellenv.go
```diff
@@ -15,7 +15,9 @@ import (
 
 type shellEnvCmdFlags struct {
 	envFlag
-	config            configFlags
+	config configFlags
+	// TODO: is this flag needed in other commands?
+	envForPackageBins bool
```
mikeland73, line 20: This flag would be appropriate for anything that instantiates print-dev-env (shell, run, shellenv).

I'm not so sure about the name

internal/boxcli/shellenv.go
```diff
@@ -85,6 +116,7 @@ func shellEnvFunc(
 	box, err := devbox.Open(&devopt.Opts{
 		Dir:               flags.config.path,
 		Environment:       flags.config.environment,
+		EnvForPackageBins: flags.envForPackageBins,
```
mikeland73, line 119: As you already mentioned, I think it might be best to move this to each individual function that needs it.

internal/boxcli/shellenv.go
mikeland73: maybe we go for something less descriptive and just document it:

`--profile-only-mode`

internal/boxcli/shellenv.go
mikeland73: (That said, this is much better than what it is replacing)

savil: Fair point. I was seeking a way to make it optional to override the flag defaults, without forcing each caller to specify it.

The alternative would be the following, which isn't that bad either:
```
func shellEnvCmd(flagDefaults shellEnvFlagDefaults) *cobra.Command {
  ...
}

// devbox commands
shellEnvCmd(&shellEnvFlagDefaults{
     recompute: true,
     envForPackageBins: false,
})
runCmd(&runFlagDefaults{}) // relies on zero-value of envForPackageBins being false
shellCmd(&shellFlagDefaults{}) // relies on zero-value of envForPackageBins being false

// global commands 
shellEnvCmd(&shellEnvFlagDefaults{
     recompute: false,
     envForPackageBins: true,
})
runCmd(&runFlagDefaults{
   envForPackageBins: true,
})
shellCmd(&shellFlagDefaults{
  envForPackageBins: true,
})

```

WDYT?

internal/boxcli/shellenv.go
savil: yeah, that works too. I usually avoid `profile` in names, because its a nix concept that our users are not exposed to or explained in our docs, but if its a hidden flag then it doesn't matter.

internal/boxcli/shellenv.go
savil: I'll give it a shot, but I fear it may add a lot of complexity/noise in the code.

internal/boxcli/shellenv.go
savil: done

internal/boxcli/shellenv.go
savil: changed to `--omit-nix-env` since that is more accurate

internal/boxcli/shellenv.go
savil: see #2159

mikeland73 (approved):

PR #2157, opened 2024-06-17
boxcli: devbox cache enable

gcurtis:
Don't merge until the new cloud.jetify.com/team/cache URL is live.

---

Add a `devbox cache enable` subcommand that:

1. Opens a browser and performs the login flow is the user isn't logged in.
2. Configures Nix to use the Jetify cache if it isn't already.

The login flow redirects to the new cloud.jetify.com/team/cache URL to let the user know the cache is enabled.

### DIFF
--- a/internal/boxcli/cache.go
+++ b/internal/boxcli/cache.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"go.jetpack.io/devbox/internal/devbox"
 	"go.jetpack.io/devbox/internal/devbox/devopt"
+	"go.jetpack.io/devbox/internal/devbox/providers/identity"
 	"go.jetpack.io/devbox/internal/devbox/providers/nixcache"
 	nixv1alpha1 "go.jetpack.io/pkg/api/gen/priv/nix/v1alpha1"
 )
@@ -72,6 +73,7 @@ func cacheCmd() *cobra.Command {
 	cacheCommand.AddCommand(uploadCommand)
 	cacheCommand.AddCommand(cacheConfigureCmd())
 	cacheCommand.AddCommand(cacheCredentialsCmd())
+	cacheCommand.AddCommand(cacheEnableCmd())
 	cacheCommand.AddCommand(cacheInfoCmd())
 
 	return cacheCommand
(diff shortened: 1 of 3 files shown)


internal/boxcli/cache.go
```diff
@@ -143,6 +145,49 @@ func cacheCredentialsCmd() *cobra.Command {
...
+
+			For more about how Devbox configures Nix, see "devbox cache configure -h".
+		`),
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			auth, err := identity.AuthClient(identity.AuthRedirectCache)
+			if err != nil {
+				return err
+			}
+			sess, _ := auth.GetSessions()
```
mikeland73, line 164: You probably want `GetSession`? 

`GetSessions` returned unrefreshed sessions. It's more useful for trying to figure out which session to use

internal/devbox/providers/identity/identity.go
```diff
@@ -16,6 +17,16 @@ import (
...
 )
 
+// Common redirect URLs for use with [AuthClient].
+var (
+	// AuthRedirectDefault redirects to a generic success page.
+	AuthRedirectDefault = build.SuccessRedirect()
+
+	// AuthRedirectCache redirects to the "Cache" tab in the dashboard for
+	// the authenticated organization.
+	AuthRedirectCache = path.Join(build.DashboardHostname(), "team", "cache")
```
mikeland73, line 27: See comment about url in other PR.

internal/boxcli/cache.go
mikeland73:
```golang
sess, err := auth.GetSession()
needLogin := errors.Is(err, auth.ErrNotLoggedIn)
```

mikeland73: There's also `identity.GenSession` which also uses the API token, but not sure that makes sense for this command? 

Long term I think having an API token should be a drop in replacement for logging in.

mikeland73 (approved):

internal/boxcli/cache.go
gcurtis: I think we just want to know if the user has logged in before. We don't necessarily need a refreshed token.

I had the same question about whether or not to use the API token. I decided to ignore it, which is why I didn't use `identity.Peek`. But now I'm having second thoughts and wondering if we should just remove `devbox cache configure` and make `devbox cache enable` do everything. Is there ever a situation where it makes sense to configure the cache without having an account or API token?

What if `devbox cache enable` did the following:

1. If there's a `DEVBOX_API_TOKEN`, skip to step 3.
2. If the user has never logged in before (no sessions), go through the login flow.
3. Configure Nix.

PR #2159, opened 2024-06-18
[Devbox] introduce devopt.EnvOptions

savil:
## Summary

From the EnvOptions docblock:
```
// EnvOptions configure the Devbox Environment in the `computeEnv` function.
// - These options are commonly set by flags in some Devbox commands
// like `shellenv`, `shell` and `run`.
// - The struct is designed for the "common case" to be zero-initialized as `EnvOptions{}`.
```

This gets rid of the pseudo-global state in the `Devbox` struct where we were setting `pure`, `preservePathStack` and `omitNixEnv` values.

## How was it tested?

TODO:
- [x] CICD should pass
- [x] Ensure `devbox global` and `devbox` omitNixEnv works as expected
- [x] Ensure `pure` works as expected for `devbox shell --pure`.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -59,14 +59,11 @@ const (
 type Devbox struct {
 	cfg                      *devconfig.Config
 	env                      map[string]string
-	omitNixEnv               bool
 	environment              string
 	lockfile                 *lock.File
 	nix                      nix.Nixer
 	projectDir               string
 	pluginManager            *plugin.Manager
-	preservePathStack        bool
-	pure                     bool
 	customProcessComposeFile string
 
 	// This is needed because of the --quiet flag.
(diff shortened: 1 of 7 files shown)


internal/devbox/devbox.go
```diff
@@ -228,15 +222,22 @@ func (d *Devbox) Shell(ctx context.Context) error {
...
 	}
 
 	return shell.Run()
 }
 
-func (d *Devbox) RunScript(ctx context.Context, cmdName string, cmdArgs []string) error {
+// runDevboxServicesScript invokes RunScript with the envOptions set to the appropriate
+// defaults for the `devbox services` scenario.
+// TODO: move to services.go
+func (d *Devbox) runDevboxServicesScript(ctx context.Context, cmdArgs []string) error {
```
mikeland73, line 236: `RunScriptWithDefaultOptions` ?

mikeland73 (approved):
Nice!

internal/devbox/devbox.go
savil: Actually, I'm going to leave this as is, because "default options" may not be correct in all scenarios. I'll modify this in a follow up to be even more Services-specific.

PR #2162, opened 2024-06-19
[easy][lock-tidy] Don't tidy flakes

mikeland73:
## Summary

TSIA

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -145,6 +145,7 @@ func newPackage(raw string, isInstallable func() bool, locker lock.Locker) *Pack
 	// assume a Devbox package.
 	parsed, err := flake.ParseInstallable(raw)
 	if err != nil || pkgtype.IsAmbiguous(raw, parsed) {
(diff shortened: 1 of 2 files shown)


internal/devpkg/package.go
```diff
@@ -145,6 +145,7 @@ func newPackage(raw string, isInstallable func() bool, locker lock.Locker) *Pack
 	// assume a Devbox package.
 	parsed, err := flake.ParseInstallable(raw)
 	if err != nil || pkgtype.IsAmbiguous(raw, parsed) {
+		// TODO: This sets runx packages as devbox packages. Not sure if that's what we want.
```
savil, line 148: oooh :D
good thing runx isn't public yet, so we can land this

savil (approved):

internal/devpkg/package.go
mikeland73: This hasn't broken anything for us (yet) even though we use runx. But I think in the future this confusion could lead to bugs.

PR #2172, opened 2024-06-25
internal/devconfig: move project directory search into `devbox.Find`

gcurtis:
Some cleanup when I looked into supporting different config extensions (e.g., `devbox.jsonc` to make editors okay with comments).

Instead of first searching for a project directory and then calling `devbox.Open` on that path, add a `devbox.Find` method that performs the search and open as a single step.

- Add separate `Find` and `Open` functions to make it clearer when we're recursively   searching for a config vs. not.
- Try just reading files instead of performing a separate `os.Stat`.
- Fix a bug where specifying any filename other than `devbox.json` with the `-C` flag would fail with a bad error message.

Changes to user error messages:

```diff
 $ devbox add go

-Error: No devbox.json found in this directory, or any parent directories. Did you run `devbox init` yet?
+Error: no devbox.json found in the current directory (or any parent directories). Did you run `devbox init` yet?

 $ devbox -c badpath add go
-Error: stat /var/folders/79/1yc1ywp10w9f2xnr_rpp_ff00000gn/T/tmp.bU25JVWovO/badpath: no such file or directory
+
+Error: the devbox config path "badpath" does not exist.

 $ mkdir child
 $ devbox -c child add go

-Error: No devbox.json found in child. Did you run `devbox init` yet?
+Error: no devbox.json found in "child". Did you run `devbox init` yet?
```


.golangci.yml
```diff
@@ -38,6 +38,8 @@ linters-settings:
       - name: bare-return
       - name: bool-literal-in-expr
       - name: cognitive-complexity
+        exclude:
+          - "**_test.go"
```
mikeland73, line 42: 👍

internal/devconfig/config.go
```diff
@@ -63,8 +78,106 @@ func IsDefault(path string) bool {
...
-	return readFromFile(path)
+// Open loads a Devbox config from a file or project directory. If path is a
+// directory, Open looks for a well-known config name (such as devbox.json)
+// within it. The error will be [ErrNotFound] if path is a valid directory
+// without a config file.
+//
+// Open does not recursively search outside of path. See [Find] to load a config
+// by walking up the directory tree.
+func Open(path string) (*Config, error) {
+	start := time.Now()
```
mikeland73, line 89: If you want to profile performance, you can use `defer debug.FunctionTimer().End()`

mikeland73: Maybe we can log that to slog as well?

mikeland73 (approved):

savil (approved):

PR #2173, opened 2024-06-26
[ux] Improve refresh messaging

mikeland73:
## Summary

A few improvements:

* Use same refresh command. Previously we were using two different ways of refreshing the environment.
* Display `refresh-global` even if alias is not set (since it will likely be set by the end of the command).
* Don't show the refresh message will calling refresh.

## How was it tested?

* Tested refresh-global on zsh.
* TODO: test refresh-global on fish.
* Calling refresh-global does not show refresh message.

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -37,6 +37,8 @@ import (
 	"go.jetpack.io/devbox/internal/ux"
 )
 
+var StateOutOfDateMessage = "Your devbox environment may be out of date. Run %s to update it.\n"
+
 // packages.go has functions for adding, removing and getting info about nix
 // packages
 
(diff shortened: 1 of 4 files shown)


internal/devbox/packages.go
```diff
@@ -37,6 +37,8 @@ import (
 	"go.jetpack.io/devbox/internal/ux"
 )
 
+var StateOutOfDateMessage = "Your devbox environment may be out of date. Run %s to update it.\n"
```
savil, line 40: const?

internal/devbox/packages.go
```diff
@@ -285,9 +287,10 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 	// be out of date after the user installs something. If have direnv active
 	// it should reload automatically so we don't need to refresh.
 	if d.IsEnvEnabled() && !upToDate && !d.IsDirenvActive() {
-		ux.Fwarning(
+		ux.FHidableWarning(
```
savil, line 290: TIL about FHidableWarning

savil (approved):
thanks!

PR #2174, opened 2024-06-26
[devbox.json] support env_from dotenv files

mohsenari:
## Summary
NOTE: Docs update PR will follow.
As per request from users, I added support for `env_from` field being able to take a  path to `*.env` file and have those env variables available in devbox shell and run.

If a duplicate env variable is specified in both `.env` and in the `"env"` section of devbox.json, the env section from devbox.json takes priority and overwrites. This I believe is the correct behavior but I'm open to discussion.

## How was it tested?
see test file at `testscripts/run/envfrom.test.txt`

### DIFF
--- a/internal/devconfig/configfile/env.go
+++ b/internal/devconfig/configfile/env.go
@@ -1,6 +1,63 @@
 package configfile
 
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
 func (c *ConfigFile) IsEnvsecEnabled() bool {
-	// envsec for legacy.
-	return c.EnvFrom == "envsec" || c.EnvFrom == "jetpack-cloud"
+	// envsec for legacy. jetpack-cloud for legacy
+	return c.EnvFrom == "envsec" || c.EnvFrom == "jetpack-cloud" || c.EnvFrom == "jetify-cloud"
+}
+
+func (c *ConfigFile) IsdotEnvEnabled() bool {
+	// filename has to end with .env
+	return filepath.Ext(c.EnvFrom) == ".env"
+}
+
+func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
+	// This check should never happen because we call IsdotEnvEnabled
+	// before calling this method. But having it makes it more robust
+	// in case if anyone uses this method without the IsdotEnvEnabled
+	if !c.IsdotEnvEnabled() {
+		return nil, fmt.Errorf("env file does not have a .env extension")
+	}
+
+	file, err := os.Open(c.EnvFrom)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file: %s", c.EnvFrom)
+	}
+	defer file.Close()
+
+	envMap := map[string]string{}
+
+	// Read the file line by line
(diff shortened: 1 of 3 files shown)


Lagoja:
Do `.env` files (and our parser) also support the following format: 

```
export S3_BUCKET=YOURS3BUCKET
export SECRET_KEY=YOURSECRETKEYGOESHERE
```

mohsenari:
> Do `.env` files (and our parser) also support the following format:
> 
> ```
> export S3_BUCKET=YOURS3BUCKET
> export SECRET_KEY=YOURSECRETKEYGOESHERE
> ```

@Lagoja not common but `.env` file format seems to have export in front of each line in [some cases](https://hexdocs.pm/dotenvy/0.2.0/dotenv-file-format.html) .
But because `.env` doesn't have an official standard, some [others specifications](https://smartmob-rfc.readthedocs.io/en/latest/2-dotenv.htmlhttps://smartmob-rfc.readthedocs.io/en/latest/2-dotenv.html) don't support the export.
So ultimately, it's up to us if we want to add it in our parser. Currently, the parser doesn't support it.

Lagoja:
I think I'm ok to add it later, current functionality is probably good for now

Lagoja (approved):

internal/devconfig/configfile/env.go
```diff
@@ -1,6 +1,63 @@
...
+
+func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
+	// This check should never happen because we call IsdotEnvEnabled
+	// before calling this method. But having it makes it more robust
+	// in case if anyone uses this method without the IsdotEnvEnabled
+	if !c.IsdotEnvEnabled() {
+		return nil, fmt.Errorf("env file does not have a .env extension")
+	}
+
+	file, err := os.Open(c.EnvFrom)
```
mikeland73, line 29: This will fail if devbox is call from any other directory that is not where the devbox.json is. I don't think it makes sense to make this relative from the working directory, it should be relative from the config file directory.

JoyceBabu:
> If a duplicate env variable is specified in both .env and in the "env" section of devbox.json, the env section from devbox.json takes priority and overwrites. This I believe is the correct behavior but I'm open to discussion.

There are two scenarios that need to be addressed:

1. **Devbox-specific override of project-specific environment variables:**
    For example, if I have a .env file containing the default environment variables for the project, the current behavior allows overriding these defaults using the env section in devbox.json.

2. **Overriding devbox defaults on a per-environment basis:**
    How can developers override the project defaults on their individual machines? For instance, if I have an environment variable PHP_MODULES_DIR that defaults to a path on Linux, I want to override it on my MacBook. How can I achieve this?

It would be ideal to have a way to override `devbox.json` itself by creating a `devbox.override.json`, similar to how `composer.override.yaml` works in Docker Compose.

internal/devconfig/configfile/env.go
varunpalekar: I agree with this, my project failed due to it.

mohsenari: @varunpalekar thanks for reminding on the issue. I made a [PR](https://github.com/jetify-com/devbox/pull/2355) to fix this, once merged it will be available in the next version of devbox.

PR #2183, opened 2024-07-09
[example] add R example

ignatiusm:
## Summary

This adds README and example R script (along with devbox.lock and devbox.json)

## How was it tested?

I tested the example repo on both `aarch64-darwin` and `x86_64-linux` by `cd`'ing into `examples/data_science/R` directory; running `devbox shell` then `Rscript src/examplePlot.R`. I considered the test to have passed if an Rplots.pdf file was produced with a scatterplot image.

This *should* do for most R users. However there is a case that I haven't handled: Sometimes users will want to install packages that are not on CRAN and therefore not on nixpkgs/nixhub.io (e.g. installing packages from github using the `R` package `devtools`). I'm not sure of an easy way to address this yet.

Addresses https://github.com/jetify-com/devbox/issues/2182

### DIFF
--- a/examples/data_science/R/.gitignore
+++ b/examples/data_science/R/.gitignore
@@ -1,0 +1,2 @@
+.RData
+.Rhistory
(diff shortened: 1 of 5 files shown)


Lagoja (approved):

Lagoja:
This is awesome, thank you for contributing! I can add a similar page to our documentation once this is merged in

PR #2189, opened 2024-07-11
[lockfile] Fix built-in plugin issues

mikeland73:
## Summary

Addresses https://github.com/jetify-com/devbox/issues/2187 and other issues.

Fixes:

* Adding a package with built-in plugin was not triggering file creation until shell/run. Now we do it on add instead.
* plugin version was not getting added to lockfile on `add` and instead was getting added on next `ensureState`. Fixed this by ensuring plugin version added during add.

## How was it tested?

* Installed nodejs, observed plugin version was in lockfile. 
* Installed nginx and observed devbox.s files were created right away


internal/devbox/packages.go
```diff
@@ -404,6 +414,9 @@ func (d *Devbox) installPackages(ctx context.Context, mode installMode) error {
 		if err := d.PluginManager().CreateFilesForConfig(pluginConfig); err != nil {
 			return err
 		}
+		if err := d.PluginManager().UpdateLockfileVersion(pluginConfig); err != nil {
+			return err
+		}
```
savil, line 419: I don't know if this is the best place to add this, because we'll be updating the lockfile _before_ the packages are actually installed in the nix store. In the event of an installation error, then the lockfile would be (wrongfully) updated.

In `ensureStateIsUpToDate`, we:
- call `d.installPackages()`
- ...// other operations that may also error
- call  `d.updateLockfile`

Could or should we move this `pluginManager.updateLockfileVersion` to `d.updateLockfile()`?

internal/devbox/packages.go
```diff
@@ -270,6 +270,16 @@ func (d *Devbox) ensureStateIsUpToDate(ctx context.Context, mode installMode) er
 		ux.Finfo(d.stderr, "Ensuring packages are installed.\n")
 	}
 
+	if mode != ensure {
+		// Reload includes because added/removed packages might change plugins. Cases:
+		// * New package adds built in. We wanna make sure the plugin is in config.
+		// * Remove built-in that installs multiple packages (e.g. nginx). We wanna clear them up so
```
savil, line 276: .. Remove built-in _plugin_ ...

internal/devbox/packages.go
mikeland73: yeah that makes sense. I put it here because I wanted to keep the existing logic intact (i.e. `d.PluginManager().CreateFilesForConfig` was previously doing this).

I think putting at the end is good. Another question is if we should move `CreateFilesForConfig` to the end as well, but at least for this PR, I'm inclined to leave as is.

savil (approved):

PR #2203, opened 2024-07-26
searcher: set custom user agent in search client

gcurtis:
The format is `Devbox/$VERSION ($GOOS; $GOARCH)`.


gcurtis:
cc @Lagoja

mikeland73 (approved):

PR #2214, opened 2024-08-01
testscripts: python tests for numpy and psycopg2

gcurtis:
Add tests that install Python with Devbox and then run scripts that import numpy
and psycopg2. They currently fail on Linux due to linker errors related to missing shared libraries.

To run the tests on Linux with Docker, run:

	devbox run -e DEVBOX_RUN_FAILING_TESTS=1 docker-testscripts -test.run '^TestScripts$/python'

The new docker-testscripts script in devbox.json runs testscripts in a Docker
container to make it easier to test various Linux distros and architectures.

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -39,6 +39,27 @@
       "test-projects-only": "DEVBOX_RUN_PROJECT_TESTS=1 go test -v -timeout ${DEVBOX_GOLANG_TEST_TIMEOUT:-30m} ./... -run \"TestExamples|TestScriptsWithProjects\"",
       "update-examples":    "devbox run build && go run testscripts/testrunner/updater/main.go",
       "tidy":               "go mod tidy",
+
+      // docker-testscripts runs the testscripts with Docker to exercise
+      // Linux-specific tests. It invokes the test binary directly, so any extra
+      // test runner flags must have their "-test." prefix.
+      //
+      // For example, to only run Python tests:
+      //
+      //   devbox run docker-testscripts -test.run ^TestScripts$/python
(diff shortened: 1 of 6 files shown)


testscripts/ld/numpy/devbox.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "shell": {
+    "init_hook": [
+      // Make sure we're using a new virtualenv in case the OS or Python version
+      // changed.
+      "export VENV_DIR=$(mktemp -d)",
+      "python -m venv $VENV_DIR",
+    ],
+    "scripts": {
+      "test": [
```
mikeland73, line 13: for these to run automatically theyneed to be called `run_test`. Maybe want to call it `run_test_disabled` ? Or add an env var so it exits 0 by default if env var is not set

devbox.json
```diff
@@ -39,6 +39,11 @@
       "test-projects-only": "DEVBOX_RUN_PROJECT_TESTS=1 go test -v -timeout ${DEVBOX_GOLANG_TEST_TIMEOUT:-30m} ./... -run \"TestExamples|TestScriptsWithProjects\"",
       "update-examples":    "devbox run build && go run testscripts/testrunner/updater/main.go",
       "tidy":               "go mod tidy",
+
+      "test-linux-amd64": [
```
mikeland73, line 43: it feels a bit weird to have this in the top level devbox.json, maybe we add a devbox.json to testscripts? It could call out to this devbox in order to build the linux devbox

not a blocker

mikeland73 (approved):

PR #2228, opened 2024-08-21
git: add .gitattributes and remove trace.out

gcurtis:


.gitattributes
```diff
@@ -0,0 +1,9 @@
+*.go diff=golang
+*.sh diff=bash
+*.md diff=markdown
+*.py diff=python
+*.sql diff=sql
+
+devbox.json linguist-language=json5
```
savil, line 7: should this be `jsonc` instead?

It appears that json5 is a more extensive version of jsonc that also supports unquoted keys and single quoted strings

That said, json5 seems to have more official support coming, whereas jsonc is a more informal spec.

savil (approved):

.gitattributes
gcurtis: `jsonc` shows trailing commas as errors in the diff, so I went with `json5`. Maybe I should update devbox to stop adding trailing commas when there are comments so it's closer to `jsonc`.

PR #2233, opened 2024-08-23
fix: make direnv watch devbox.lock

mjgallag:

### DIFF
--- a/internal/devbox/generate/tmpl/envrcContent.tmpl
+++ b/internal/devbox/generate/tmpl/envrcContent.tmpl
@@ -1,5 +1,5 @@
 use_devbox() {
-    watch_file devbox.json
+    watch_file devbox.json devbox.lock
     eval "$(devbox shellenv --init-hook --install --no-refresh-alias{{ if .EnvFlag }} {{ .EnvFlag }}{{ end }})"
 }
 use devbox


savil (approved):
Thanks! LGTM. Will just cross-check with others as well

mikeland73:
LGTM

mjgallag:
Unfortunately because this gets to .envrc via eval, and not a change to .envrc itself, it won’t bust direnv’s cache. Perhaps something to track in devbock.lock at some point?

mikeland73:
Unfortunately because this change gets to .envrc via eval, and not a change to .envrc itself, it won’t bust direnv’s cache. Perhaps something to track in devbox.lock at some point?

@mjgallag which cache is this? Does direnv track which files it watches and doesn't add new ones if . envrc doesn't change?

Our goal was to generate .envrc once and avoid changing it (I think it has to be approved every time it changes?). We could add a version to the generated .envrc file to force direnv to pick up the changes. (and detect a version missmatch)

mjgallag:
@mikeland73 tough call on the best flow. I think putting a version in the generated .envrc would mean devbox users would have to upgrade to a new version of devbox and then be told to run generate again in order to get the change. And you do need to allow again when it changes. If change is just shipped as is it wouldn't take a effect immediately after devbox upgrade, it would only take effect if devbox.json changes, .envrc changes (and they run allow) or the cd out and back into the directory. direnv will cache between prompts because files it is watching have not changed. I believe it stores a checksum in DIRENV_WATCHES var for each watched file so as not to have to recalculate every prompt. In any case not something that needs to be solved with this pull and maybe this won't change much anyway.

mikeland73:
@mjgallag got it. From:

>  it wouldn't take a effect immediately after devbox upgrade, it would only take effect if devbox.json changes, .envrc changes (and they run allow) or they cd out and back into the directory.

sounds like doing cd or opening a new terminal and cd'ing to project will cause direnv to pick up the new .envrc right? This sounds OK to me.

Will test and merge if everything looks good.

mjgallag:
@mikeland73 yeah correct, so given this file is unlikely to change often i agree its OK. Thanks for taking a look. This should mean after running devbox update next prompt has the new package versions, rather than having to cd in and out to get.

savil:
While not perfect, it does sound like an improvement from the status quo so 👍

PR #2237, opened 2024-08-29
patchpkg: start moving glibc-patch.bash to Go

gcurtis:
Start porting the glibc patching script (which has expanded to patching other libraries too) over to Go.

To make this change more incremental, this commit only ports copying the original package's files to the patched package (basically the `cp` command). Devbox still runs the embedded bash script until the rest of it is ported.

The general patching steps are:

- When generating the `glibc-patch.nix` flake, Devbox gets the absolute path to itself and adds it as a flake input. This makes devbox available in the flake's build.
- Instead of running `glibc-patch.bash` as the flake's builder, run `devbox patch <pkg>`.

### DIFF
--- a/internal/boxcli/patch.go
+++ b/internal/boxcli/patch.go
@@ -1,0 +1,22 @@
+package boxcli
+
+import (
+	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


savil (approved):
This is neat

PR #2248, opened 2024-09-03
patchpkg: port ELF patching to Go and use RPATH

gcurtis:
Move the patching functionality from glibc-patch.bash to Go while also updating it to set RPATH instead of RUNPATH.

### DIFF
--- a/internal/patchpkg/builder.go
+++ b/internal/patchpkg/builder.go
@@ -179,6 +179,8 @@ func (d *DerivationBuilder) needsGlibcPatch(file *bufio.Reader, filePath string)
 		return false
 	}
 
+	// ELF binaries are identifiable by the first 4 magic bytes:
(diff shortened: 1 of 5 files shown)


internal/patchpkg/builder.go
```diff
@@ -142,6 +171,21 @@ func (d *DerivationBuilder) copySymlink(pkg, out *packageFS, path string) error
...
+	}
+	if path.Dir(filePath) != "bin" {
+		return false
+	}
+
+	magic, err := file.Peek(4)
+	if err != nil {
+		return false
+	}
+	return magic[0] == 0x7F && magic[1] == 'E' && magic[2] == 'L' && magic[3] == 'F'
```
mikeland73, line 186: What is this magic, add comment

mikeland73 (approved):

PR #2250, opened 2024-09-04
devpkg: auto-patch python

gcurtis:
Add a `devbox add --patch <mode>` flag that replaces `--patch-glibc` and a corresponding `patch` JSON field that replaces `patch_glibc`. The new name reflects the new patching behavior, which affects more than just glibc.

The new patch flag/field is a string instead of a bool. Valid values are `auto`, `always` and `never`. The default is `auto`.

    devbox add --patch <auto/always/never>

- `auto` - let Devbox decide if a package should be patched. Currently only enables patching for Python or if `patch_glibc` is true in the config.
- `always` - always attempt to patch a package. Corresponds to the `--patch-glibc=true` behavior.
- `never` - never patch a package, even if `auto` would.

If a config has an existing package with the `"patch_glibc": true` field, it's interpreted as `"patch": "always"` but the config itself isn't modified. However, if the user runs a command that does write to the config, then `patch_glibc` will be migrated to `patch`.

Example `devbox.json`:

```json5
{
  "packages": {
    "ruby": {
      "version": "latest",
      "patch":   "always"
    },
    "python": {
      "version": "latest"

      // No patch field implies "auto".
      // "patch": "auto"
    }
  }
}
```

### DIFF
--- a/internal/devconfig/configfile/ast.go
+++ b/internal/devconfig/configfile/ast.go
@@ -173,7 +173,7 @@ func (c *configAST) removePackageElement(arr *hujson.Array, name string) {
 
 // setPackageBool sets a bool field on a package.
 func (c *configAST) setPackageBool(name, fieldName string, val bool) {
-	pkgObject := c.FindPkgObject(name)
+	pkgObject := c.findPkgObject(name)
 	if pkgObject == nil {
 		return
 	}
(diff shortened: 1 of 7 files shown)


internal/devpkg/package.go
```diff
@@ -238,6 +238,12 @@ func (p *Package) IsInstallable() bool {
 }
 
 func (p *Package) PatchGlibc() bool {
+	if p.CanonicalName() == "python" {
+		if p.patchGlibc != nil && !p.patchGlibc() {
```
gcurtis, line 242: I think this actually still needs a tweak to distinguish between missing and false.

Lagoja:
1. Should we rename this flag? 
2. Would enabling this flag for other non-Python packages have a similar effect?

savil (approved):

internal/devpkg/package.go
mikeland73: Yeah, looks like function returns false in both cases https://github.com/jetify-com/devbox/blob/main/internal/devpkg/package.go#L113C1-L115C5

mikeland73:
Should we add a feature flag for the opt out? Or just go for it?

internal/devpkg/package.go
mikeland73: https://github.com/jetify-com/devbox/pull/2251

mikeland73 (approved):
Approve pending opt-out option. 

I think we should either feature flag it, or make sure the opt out works before releasing.

Lagoja:
I would suggest that we do a beta release, test it a bunch internally + with the community, and then release. I think an opt-in/flag is too hard for users to discover and activate. I would be open to an opt-out, maybe with a differently named flag.

gcurtis:
@Lagoja

> 1. Should we rename this option from `patch_glibc`?

That's probably a good idea since it does a bit more now. What about `autopatch` or just `patch`? I'll leave `patch_glibc` in as an alias for backwards compatibility.

> 2. Would enabling this option for other non-Python packages have a similar effect?

I haven't tested it on any other packages, but the patching logic is pretty general. It'll look for any ELF binaries in the package's `bin` directory, update glibc, and convert `RUNPATH` to `RPATH` on all of them.

The stuff with restoring build inputs is specific to Python and will just be a no-op on other packages.

gcurtis:
@savil @mikeland73 I made some pretty substantial changes if you want to take another look. I switched things over to use a new `--patch auto/always/never` flag instead.

internal/devconfig/configfile/ast.go
```diff
@@ -216,7 +216,72 @@ func (c *configAST) appendAllowInsecure(name, fieldName string, whitelist []stri
 	c.appendStringSliceField(name, fieldName, whitelist)
 }
 
-func (c *configAST) FindPkgObject(name string) *hujson.Object {
+// removePatch removes the patch field from the named package.
+func (c *configAST) removePatch(name string) {
```
mikeland73, line 220: Could we generalize these similar to `setPackageBool` ? 

`removePackageField` for removing.
`setPackageString` for adding. 

`setPatch` could wrap these (remove `patch_glibc` add `patch`) or you could not special case at all and have the code in packages.go do that logic). That keeps the ast more logic agnostic.

Honestly, you could also not remove `patch_glibc`. It doesn't really do harm and would allow us to have simpler code.

mikeland73 (approved):
LGTM

in a follow up could we add a warning if `patch_glibc` is set in config?

`patch_glibc` is deprecated and will be removed in a future version. Use `patch: always` instead.

internal/devconfig/configfile/ast.go
gcurtis: Making them more general sounds good to me. I'd like to do in a follow-up PR though just to get the patching stuff out the door.

PR #2251, opened 2024-09-05
[RFC][goutil] Java style Optional construct

mikeland73:
## Summary

Could be useful. Mostly AI generated.

### DIFF
--- a/internal/goutil/optional.go
+++ b/internal/goutil/optional.go
@@ -1,0 +1,89 @@
+package goutil
+
+import (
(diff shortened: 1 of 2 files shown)


mikeland73:
@gcurtis let me know if you have any interest in this. Otherwise, I'll close.

gcurtis:
@mikeland73 I think I ended up not needing this. I ended up making the new JSON field an enum instead of a bool, which also lets us distinguish between null and the zero value.

PR #2258, opened 2024-09-09
patchpkg: patch missing Python refs on darwin

gcurtis:
This enables the part of Python patching that restores some build dependencies on macOS. This fixes some `pip install` build errors for packages without needing to manually add dev dependencies on packages like openssl.

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -187,8 +187,6 @@ func patchGlibcFunc(canonicalName string, mode configfile.PatchMode) func() bool
 		case configfile.PatchNever:
 			patch = false
 		}
-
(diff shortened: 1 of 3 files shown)


mikeland73 (approved):

internal/devpkg/package.go
```diff
@@ -189,7 +189,7 @@ func patchGlibcFunc(canonicalName string, mode configfile.PatchMode) func() bool
 		}
 
 		// Check nix.SystemIsLinux() last because it's slow.
-		return patch && nix.SystemIsLinux()
+		return patch
```
savil, line 192: need to update comment

savil (approved):

PR #2259, opened 2024-09-09
[Python] Change `.venv` script to be more compatible with IDEs

Lagoja:
## Summary

Moves `.venv` to the Devbox Project Root, and adds some checks to ensure we warn the user if we are going to squash a user created venv. Moving `.venv` to the project root makes it more likely that Python Extensions and IDEs will pick up the devbox managed python, instead of a system python

Fixes DEV-105

### DIFF
--- a/plugins/pip/venvShellHook.sh
+++ b/plugins/pip/venvShellHook.sh
@@ -1,4 +1,5 @@
 STATE_FILE="$DEVBOX_PROJECT_ROOT/.devbox/venv_check_completed"
+echo $STATE_FILE
 
 is_valid_venv() {
     [ -f "$1/bin/activate" ] && [ -f "$1/bin/python" ]
@@ -10,10 +11,31 @@ is_devbox_python() {
         echo "DEVBOX_PACKAGES_DIR is not set. Unable to check for Devbox Python."
         return 1
     fi
-    local python_path=$(readlink "$1/bin/python")
-    echo $python_path
-    echo $DEVBOX_PACKAGES_DIR
-    [[ $python_path == $DEVBOX_PACKAGES_DIR/bin/python* ]]
+    local python_path="$1/bin/python"
+    local link_target
+
+    while true; do
+        if [ ! -L "$python_path" ]; then
+            echo $python_path
+            # Not a symlink, we're done
+            break
+        fi
+
+        link_target=$(readlink "$python_path")
+        echo "Checking symlink: $link_target"
+
+        if [[ "$link_target" == /* ]]; then
+            # Absolute path, we're done
+            python_path="$link_target"
+            break
+        elif [[ "$link_target" == python* ]] || [[ "$link_target" == ./* ]] || [[ "$link_target" == ../* ]]; then
+            # Relative path or python symlink, continue resolving
+            python_path=$(dirname "$python_path")/"$link_target"
+        else
+            # Unexpected format, stop here
+            break
+        fi
+    done
 }
 
 # Function to check Python version
--- a/plugins/pip/venvShellHook.sh
+++ b/plugins/pip/venvShellHook.sh
@@ -1,8 +1,71 @@
-#!/bin/sh
+STATE_FILE="$DEVBOX_PROJECT_ROOT/.devbox/venv_check_completed"
 
-if ! [ -d "$VENV_DIR" ]; then
-    echo "Creating new venv environment in path: '${VENV_DIR}'"
-    python3 -m venv "$VENV_DIR"
-    echo "You can activate the virtual environment by running '. \$VENV_DIR/bin/activate' (for fish shell, replace '.' with 'source')" >&2
+is_valid_venv() {
+    [ -f "$1/bin/activate" ] && [ -f "$1/bin/python" ]
+}
+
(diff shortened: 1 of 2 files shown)


loreto:
> Question: Do we care about preserving a user's pre-existing `.venv` if they created it with a different tool? Or should we just always squash it.
Preserve it.

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,92 @@
-#!/bin/sh
```
gcurtis, line 1: Why get rid of the shebang?

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,92 @@
...
+set -eu
+STATE_FILE="$DEVBOX_PROJECT_ROOT/.devbox/venv_check_completed"
+echo $STATE_FILE
```
gcurtis, line 3: Was this left in from debugging?

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,92 @@
...
+}
+
+# Check if we've already run this script
+if [ -f "$STATE_FILE" ]; then
+    # "We've already run this script. Exiting..."
+    exit 0
 fi
 
+# Check Python version
+if ! check_python_version; then
```
gcurtis, line 60: Are we just checking the version to see if venv is available? If so, it might be easier to just try importing it.

```suggestion
if ! python -c "import venv" &>/dev/null; then
```

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,92 @@
...
+if [ -d "$VENV_DIR" ]; then
+    if is_valid_venv "$VENV_DIR"; then
+        if ! is_devbox_python "$VENV_DIR"; then
+            echo "\033[1;33mWARNING: Virtual environment at $VENV_DIR doesn't use Devbox Python.\033[0m"
+            echo "Virtual environment: $VENV_DIR"
+            read -p "Do you want to overwrite it? (y/n) " -n 1 -r
+            echo
+            if [[ $REPLY =~ ^[Yy]$ ]]; then
+                echo "Overwriting existing virtual environment..."
+                rm -rf "$VENV_DIR"
```
gcurtis, line 76: I think you can use `python3 -m venv --clear` to overwrite.

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,92 @@
...
-if ! [ -d "$VENV_DIR" ]; then
-    echo "Creating new venv environment in path: '${VENV_DIR}'"
-    python3 -m venv "$VENV_DIR"
-    echo "You can activate the virtual environment by running '. \$VENV_DIR/bin/activate' (for fish shell, replace '.' with 'source')" >&2
+is_valid_venv() {
+    [ -f "$1/bin/activate" ] && [ -f "$1/bin/python" ]
+}
+
+# Function to check if Python is a symlink to a Devbox Python
+is_devbox_python() {
```
gcurtis, line 10: There's also the `test -ef` flag to check if two files are the same. So you could do:

```sh
if [ "$1/bin/python" -ef "$DEVBOX_PACKAGES_DIR/bin/python" ]; then
```

plugins/pip/venvShellHook.sh
Lagoja: Ah that's a much nicer approach

Lagoja: Had to tweak this because for some reason it was failing the automated tests (but not when run locally)

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,54 @@
...
+    [ -f "$1/bin/activate" ] && [ -f "$1/bin/python" ]
+}
+
+is_devbox_venv() {
+    [ "$1/bin/python" -ef "$DEVBOX_PACKAGES_DIR/bin/python" ]
+}
+
+create_venv() {
+    python -m venv "$VENV_DIR" --clear
+    echo "*\n.*" >> "$VENV_DIR/.gitignore"
```
mikeland73, line 15: nice

mikeland73 (approved):

plugins/pip/venvShellHook.sh
```diff
@@ -1,8 +1,54 @@
...
+if [ -f "$STATE_FILE" ]; then
+    # "We've already run this script. Exiting..."
+    exit 0
 fi
 
+# Check that Python version supports venv
+if ! python -c 'import venv' 1> /dev/null 2> /dev/null; then
+    echo "\033[1;33mWARNING: Python version must be > 3.3 to create a virtual environment.\033[0m"
+    touch "$STATE_FILE"
+    exit 1
```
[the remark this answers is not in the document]
Lagoja: Yes it should, because the python in the venv is a symlink to the python in our Devbox profile, which is a symlink to the nix store. I'll test a few times to make sure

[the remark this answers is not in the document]
Lagoja: Confirmed, once the `.venv` is created, updating python automatically updates the python linked in the `.venv` folder

PR #2265, opened 2024-09-12
Chore/switch pip plugin reference to python

djgoku:
## Summary

Remove pip plugin and use Python plugin.

## How was it tested?

`devbox run lint`
`devbox run test`
`go mod tidy`
`devbox shell` testing new python plugin (`cp plugins/python.json plugins/plugin.json` and add `"include": ["path:./plugins/"],` to devbox.json. Though this is the output I got (not the Poetry message):

```
devbox % ./dist/devbox shell                       
Starting a devbox shell...
Virtual environment directory doesn't exist. Creating new one...

Poetry could not find a pyproject.toml file in /Users/dj_goku/dev/github/djgoku/devbox or its parents
```


Lagoja:
Made two small tweaks: 
1. Updated the packages listed in the docs snippet
2. I found a bug in how the UV_PYTHON variable was set, I so I changed that as well

Lagoja (approved):

PR #2271, opened 2024-09-13
patchpkg: patch libstdc++

gcurtis:
This fixes an issue with TensorFlow not being able to find libstdc++.so at runtime.

### DIFF
--- a/internal/patchpkg/patch.go
+++ b/internal/patchpkg/patch.go
@@ -12,56 +12,88 @@ import (
 	"strings"
 )
 
-// glibcPatcher patches ELF binaries to use an alternative version of glibc.
-type glibcPatcher struct {
+// libPatcher patches ELF binaries to use an alternative version of glibc.
+type libPatcher struct {
 	// ld is the absolute path to the new dynamic linker (ld.so).
 	ld string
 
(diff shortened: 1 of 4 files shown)


internal/patchpkg/patch.go
```diff
@@ -12,56 +12,80 @@ import (
...
+	// Verify that we can find a directory with libstdc++.so in it.
+	glob := "lib*/libstdc++.so*"
+	matches, _ := fs.Glob(pkg, glob)
+	if len(matches) == 0 {
+		return fmt.Errorf("cannot find libstdc++.so file matching %q", glob)
+	}
+	for i := range matches {
+		matches[i] = path.Dir(matches[i])
+	}
+	slices.Sort(matches) // pick the shortest name: lib < lib32 < lib64 < libx32
```
savil, line 76: why this heuristic?

savil (approved):

internal/patchpkg/patch.go
gcurtis: Added a comment.

- lib is usually a symlink to the correct arch (e.g., lib -> lib64)
- *.so is usually a symlink to the correct version (e.g., foo.so -> foo.so.2)

PR #2272, opened 2024-09-13
[Services] Pass shutdown signals from Devbox to process-compose

Lagoja:
## Summary

This PR attempts to fix an issue where process-compose can terminate without cleaning up it's running processes. This occurs if you are running `devbox services up` in a terminal, and then close the running terminal without first exiting process-compose.

With this PR, Devbox will now explicitly propagate SIGTERM, SIGINT, SIGHUB, and SIGTSTP signals to process compose so it that it can run it's shutdown process and cleanup any running processes.

## How was it tested?

1. Start `devbox services up` in a project that runs services
2. Verify process-compose is running
3. Close the terminal without exiting process-compose
4. Verify that process-compose and all running services have terminated

### DIFF
--- a/internal/services/manager.go
+++ b/internal/services/manager.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


Lagoja:
Closing for now until we can verify this is the issue

gcurtis:
Possibly related to F1bonacc1/process-compose#246.

PR #2277, opened 2024-09-16
[sandboxes] devbox ls: print versions

### LINKED ISSUE #2274: The list command should show the installed version

`devbox list` and `devbox global list`  show the version that a package is pinned to but not the exact version that is installed.  It is common for people to post the output of `list` for debugging purposes, but only seeing `latest` isn't super helpful there.

It would be great to expand the list command to include the package name, the pinned version, and the installed version.  So instead of this:
```
* neovim@latest
* syncthing@latest
```

We could see something like this:
```
* neovim@latest - 0.10.1
* syncthing@latest - 1.27.12
```

The version information is available in devbox.lock.

Looking at devbox.lock every time I want to see the exact version installed.

savil:
## Summary

For packages without an explicit version, this PR will print the version for
the user in `devbox ls`.

Fixes #2274 

## How was it tested?
```
% devbox ls
* go@latest - 1.23.0
* runx:golangci/golangci-lint@latest - 1.60.2
* runx:mvdan/gofumpt@latest - 0.7.0
* cowsay@3.8.2
```
Notice that versions are printed for go@latest and the runx packages, but 
not for cowsay since it was added via `devbox add cowsay@3.8.2`

Also, `devbox ls -c examples/flakes/php` and `devbox ls -c examples/flakes/remote`
```
% devbox ls -c examples/flakes/remote
* github:nixos/nixpkgs/5233fd2ba76a3accb5aaa999c00509a11fd0793c#hello
* github:nixos/nixpkgs/5233fd2ba76a3accb5aaa999c00509a11fd0793c#cowsay
* github:F1bonacc1/process-compose/v0.43.1
```


internal/boxcli/list.go
```diff
@@ -32,8 +33,26 @@ func listCmd() *cobra.Command {
...
-			for _, p := range box.AllPackageNamesIncludingRemovedTriggerPackages() {
-				fmt.Fprintf(cmd.OutOrStdout(), "* %s\n", p)
+
+			for _, pkg := range box.AllPackagesIncludingRemovedTriggerPackages() {
+				resolvedVersion, err := pkg.ResolvedVersion()
+				if err != nil {
+					// Continue to print the package even if we can't resolve the version
+					// so that the user can see the error for this package, as well as get the
+					// results for the other packages
+					resolvedVersion = "<error resolving version>"
```
gcurtis, line 43: Maybe omit the version from the list and print it on stderr at the end?

gcurtis (approved):

internal/devpkg/package.go
```diff
@@ -830,3 +830,15 @@ func packageInstallErrorHandler(err error, pkg *Package, installableOrEmpty stri
 
 	return usererr.WithUserMessage(err, "error installing package %s", pkg.Raw)
 }
+
+func (p *Package) ResolvedVersion() (string, error) {
+	if err := p.resolve(); err != nil {
+		return "", err
+	}
```
mikeland73, line 837: If package is not in lockfile this is just current resolved version, but that's probably ok.

internal/devbox/devbox.go
```diff
@@ -848,6 +848,11 @@ func (d *Devbox) AllPackageNamesIncludingRemovedTriggerPackages() []string {
 	return result
 }
 
+func (d *Devbox) AllPackagesIncludingRemovedTriggerPackages() []*devpkg.Package {
```
mikeland73, line 851: Change `AllPackageNamesIncludingRemovedTriggerPackages` to use this?

internal/boxcli/list.go
```diff
@@ -32,8 +33,26 @@ func listCmd() *cobra.Command {
...
+				}
+				msg := ""
+
+				// Print the resolved version, unless the user has specified a version already
+				if strings.HasSuffix(pkg.Versioned(), "latest") && resolvedVersion != "" {
+					// Runx packages have a "v" prefix (why?). Trim for consistency.
```
mikeland73, line 49: technically semver has (or can have?) a prefix `v`

mikeland73 (approved):

loreto:
@savil Is this ready to be merged?

Issue #2196, opened 2024-07-18
NIXPKGS_ALLOW_UNFREE env var not working


eyevanovich:
### What happened?

When trying to pull in a NUR flake that is unfree I am unable to get Devbox to accept NIXPKGS_ALLOW_UNFREE. This is due to the fact that the `--impure` flag is not being included during `nix print-dev-env` when using `devbox shell`.

More context can be found in [this discord thread](https://discord.com/channels/903306922852245526/1263573348852695194)

### Steps to reproduce

1. devbox init
2. debox add github:goreleaser/nur/ec67c3ebe81bd5614190e3dd4600089859d5b179#goreleaser-pro
3. NIXPKGS_ALLOW_UNFREE=1 devbox shell

### Command

shell

### devbox.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.12.0/.schema/devbox.schema.json",
  "packages": [
    "go@1.22.5",
    "tilt@0.33.10",
    "wget@1.21.4",
    "pre-commit@3.7.1",
    "python@3.12.3",
    "gnumake@4.4.1",
    "go-task@3.37.2",
    "_1password@2.28.0",
    "curl@8.7.1",
    "go-migrate@4.17.1",
    "air@1.52.1",
    "gnused@4.9",
    "github:goreleaser/nur/ec67c3ebe81bd5614190e3dd4600089859d5b179#goreleaser-pro"
  ],
  "env": {
    "GOBIN": "$PWD/bin",
    "PATH": "$PWD/bin:$PATH",
    "NIXPKGS_ALLOW_UNFREE": "1"
  },
  "shell": {
    "init_hook": [
      "task --silent env-setup"
    ],
    "scripts": {}
  }
}
```

### Devbox version

0.12.0

### Nix version

nix (Nix) 2.21.2

### What system does this bug occur on?

macOS (Apple Silicon)

snipebin:
Similar issue with postgres timescaledb extension:

```shell
› NIXPKGS_ALLOW_UNFREE=1 devbox shell      
Info: Ensuring packages are installed.
✓ Computed the Devbox environment.
[...]

PR #2279, opened 2024-09-17
Aadd `--impure` flag when calling print-dev-env

imfede:
## Summary

Add `--impure` to the `print-dev-env` command, allowing flakes to access environment variables also in this step. Without this change, a flake that needs an env var to work will correctly build but then a shell could not be created.

This also solves #2196, signifying that other people encountered a similar problem.

## How was it tested?

Both by
```
devbox run test
```

and manually, building the tool and locally checking that a flake would pick up variables (with `builtins.getEnv`) when activating a shell.

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/redact"
 	"golang.org/x/mod/semver"
@@ -74,7 +75,11 @@ func (*Nix) PrintDevEnv(ctx context.Context, args *PrintDevEnvArgs) (*PrintDevEn
 	}
 
 	if len(data) == 0 {
-		cmd := command("print-dev-env", "--json", "--impure",
+		optionalImpureFlag := ""
+		if featureflag.ImpurePrintDevEnv.Enabled() {
+			optionalImpureFlag = "--impure"
+		}
+		cmd := command("print-dev-env", "--json", optionalImpureFlag,
 			"path:"+flakeDirResolved,
 		)
 		slog.Debug("running print-dev-env cmd", "cmd", cmd)
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -75,13 +75,11 @@ func (*Nix) PrintDevEnv(ctx context.Context, args *PrintDevEnvArgs) (*PrintDevEn
 	}
 
 	if len(data) == 0 {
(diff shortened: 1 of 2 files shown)


mikeland73:
@imfede, thanks for the contribution!

IIRC we didn't make this `impure` in order to use the nix evaluation cache. That said, devbox uses its own caching as well, so we might not need that.

Could you add a feature flag to this and default it to disabled? That way you could start using it right away and we can test if it has any perf side-effects.

mikeland73 (dismissed):
Please put this behind a disabled feature flag.

See [featureflag](https://github.com/jetify-com/devbox/tree/main/internal/boxcli/featureflag) package.

To enable for yourself you can use 

`DEVBOX_FEATURE_[NAME]=1`

After we test a bit, if we don't see a performance issue we'll remove the flag and make it permanent.

Thanks again for contribution!

imfede:
Thanks for the review!

My golang skills are practically non-existant, so I hope the end result is good enough. Let me know how can I improve on it if needed :)

internal/nix/nix.go
```diff
@@ -74,7 +75,11 @@ func (*Nix) PrintDevEnv(ctx context.Context, args *PrintDevEnvArgs) (*PrintDevEn
 	}
 
 	if len(data) == 0 {
-		cmd := command("print-dev-env", "--json",
+		optionalImpureFlag := ""
+		if featureflag.ImpurePrintDevEnv.Enabled() {
+			optionalImpureFlag = "--impure"
+		}
+		cmd := command("print-dev-env", "--json", optionalImpureFlag,
```
savil, line 82: If I do the following steps, then I get an error here:
```
# build binary in dist/devbox
 % devbox run build
 
 # clear local state
 % rm -rf .devbox
 
 # install the devbox-repo's own packages
 % dist/devbox install
Error: nix: command error: nix --extra-experimental-features ca-derivations --option experimental-features 'nix-command flakes fetch-closure' print-dev-env --json  path:/Users/savil/code/jetpack/devbox/.devbox/gen/flake: unexpected argument 'path:/Users/savil/code/jetpack/devbox/.devbox/gen/flake': exit code 1

```

I think the issue is with the empty argument.

I think you could do:
```
		cmd := command("print-dev-env", "--json")
		if featureflag.ImpurePrintDevEnv.Enabled() {
			cmd.Args = append(cmd.Args, "--impure")
		}
		cmd.Args = append(cmd.Args, "path:"+flakeDirResolved)
```

imfede: Weird, I thought I had tested it without the feature flag active. 

Thanks for spotting this, I have pushed a fix following your suggestion.

savil (approved):
LGTM thanks for sending this!

imfede:
Thanks for the reviews! Out of information, do you have a plan on when this could be released?

PR #2286, opened 2024-09-20
refactor: add bash shebang for shellcheck

mjgallag:
https://www.shellcheck.net/wiki/SC2148 (tips depend on target shell)
https://direnv.net/#how-it-works (.envrc is loaded into a bash sub-shell)

### DIFF
--- a/internal/devbox/generate/tmpl/envrc.tmpl
+++ b/internal/devbox/generate/tmpl/envrc.tmpl
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # Automatically sets up your devbox environment whenever you cd into this
 # directory via our direnv integration:
 


mikeland73 (approved):

mikeland73:
@mjgallag this LGTM! 

One caveat is that this file is not replaced automatically (because it would require re-allow from direnv).

Folks that already have .envrc would need to do: `devbox gen direnv --force` and `direnv allow` 

(We've talked about versioning this file in the past, but that's outside the scope of this change)

gcurtis:
Not as familiar with the envrc stuff so this might not matter, but is the file generated from this template ever executed directly (e.g., `./envrc`)? Although it's unlikely, that would fail if there's no bash in `/bin/bash`. Another option is to add a shellcheck directive with `# shellcheck shell=bash`.

mjgallag:
@gcurtis direnv uses the bash path that is set at build time:
https://github.com/direnv/direnv/blob/ae5a61848a06da4251b91fd084bffa9a5732fb89/internal/cmd/rc.go#L271
https://github.com/NixOS/nixpkgs/blob/aaa7fb5840f278da8010fe7466c3d952f7e536da/pkgs/by-name/di/direnv/package.nix#L17-L19
https://github.com/Homebrew/homebrew-core/blob/fb3c5a4fcbd4d4974499ed9b46304d8d0881cbc3/Formula/d/direnv.rb#L24

I could envision maybe running `. ./envrc` or `./envrc` directly as a convenience when testing changes to it, but the shebang shouldn't have any effect as far as `direnv` is concerned. That being said I don't have a strong preference here between `#!/bin/bash`, [`#!/usr/bin/env bash`](https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash) or `# shellcheck shell=bash`.

gcurtis (approved):

PR #2292, opened 2024-09-23
internal/devbox: handle whitespace in paths

### LINKED ISSUE #2290: devbox fails to start if there is whitespace in the path
### What happened?

Initialized devbox in a directory with a space character and it crashed:

```
scott@scott-XPS-15-9570:~$ mkdir foo\ bar
scott@scott-XPS-15-9570:~$ cd foo\ bar/
scott@scott-XPS-15-9570:~/foo bar$ devbox init
scott@scott-XPS-15-9570:~/foo bar$ devbox shell
Error: nix: command error: nix --extra-experimental-features ca-derivations --option experimental-features 'nix-command flakes fetch-closure' print-dev-env --json 'path:/home/scott/foo bar/.devbox/gen/flake': 'path:/home/scott/foo bar/.devbox/gen/flake' is not a valid URL: exit code 1

```

I changed the name of the directory and it was fine:
```
scott@scott-XPS-15-9570:~/foo bar$ cd ..
scott@scott-XPS-15-9570:~$ mv foo\ bar/ foobar
scott@scott-XPS-15-9570:~$ cd foobar/
scott@scott-XPS-15-9570:~/foobar$ devbox shell
Starting a devbox shell...
(devbox) scott@scott-XPS-15-9570:~/foobar$ 

```

### Steps to reproduce

See Above

### Command

_No response_

### devbox.json

```jsonc
{
    "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.13.0/.schema/devbox.schema.json",
    "packages": [],
    "shell": {
      "init_hook": [
      ],
      "scripts": {
        "test": [
        ]
      }
    }
  }

```
```


### Devbox version

0.13.0

### Nix version

nix (Nix) 2.18.1

Linux (x86-64)

### Debug logs

_No response_

gcurtis:
Make sure the flake ref is URL-encoded and quote the rcfile in the shellrc template.

Fixes #2290.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -270,7 +270,16 @@ func (d *Devbox) RunScript(ctx context.Context, envOpts devopt.EnvOptions, cmdNa
 	// better alternative since devbox run and devbox shell are not the same.
 	env["DEVBOX_SHELL_ENABLED"] = "1"
 
-	// wrap the arg in double-quotes, and escape any double-quotes inside it
+	// wrap the arg in double-quotes, and escape any double-quotes inside it.
+	//
+	// TODO(gcurtis): this breaks quote-removal in parameter expansion,
+	// command substitution, and arithmetic expansion:
+	//
+	//	$ unset x
+	//	$ echo ${x:-"my file"}
+	//	my file
+	//	$ devbox run -- echo '${x:-"my file"}'
+	//	"my file"
 	for idx, arg := range cmdArgs {
 		cmdArgs[idx] = strconv.Quote(arg)
 	}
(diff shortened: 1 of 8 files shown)


savil (approved):

savil:
Awesome!

I think this solves https://github.com/jetify-com/devbox/issues/1914, and https://github.com/jetify-com/devbox/issues/971 as well?  Could you close them if so?

Lagoja:
@savil last time you looked into this issue -- did we discover a compatibility risk with older versions of Nix?

savil:
@Lagoja I think this was my revert commit https://github.com/jetify-com/devbox/pull/1935. Unfortunately, I forgot to document the reason I had to revert :(

But digging some more, I believe the prior approach relied on using percent-encoded paths which were only supported on nix 2.19. However, I think @gcurtis 's `flake.Ref` handles this internally (how?), as per his claim here (see [internal discussion](https://jetify-com.slack.com/archives/C040L5DFHFT/p1710866310917689))

@gcurtis how does flake.Ref handle spaces internally so they work with older nix versions?

gcurtis:
@savil `flake.Ref.String` uses `url.PathEscape` to encode the file path. I think in the reverted PR we were calling `flake.ParseRef("path:" + flakeDirResolved)`, so it assumed that  `flakeDirResolved` was already encoded (the `path:` makes it a URL).

I added a test and ran it back to Nix 2.17 and everything seemed to work. I also hardened the shell quoting using the same logic we use for environment variables.

PR #2296, opened 2024-09-24
patchpkg: patch python to use devbox CUDA libs

gcurtis:
Automatically patch python to use any `cudaPackages.*` packages that are in devbox.json. This will only work if the CUDA drivers are already installed on the host system.

The patching process is:

1.  When generating the patch flake, look for the system’s `libcuda.so` library (installed by the driver) and copy it into the flake’s directory.
2.  Nix copies the flake’s source directory (and therefore libcuda.so) into the nix store when building it.
3.  The flake calls `devbox patch` which adds a `DT_NEEDED` entry to the python binary for `libcuda.so`. It also adds the lib directories of any other `cudaPackages.*` packages that it finds in the environment.

### DIFF
--- a/internal/patchpkg/builder.go
+++ b/internal/patchpkg/builder.go
@@ -17,6 +17,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"strings"
 )
 
 //go:embed glibc-patch.bash
@@ -42,6 +43,10 @@ type DerivationBuilder struct {
 
 	RestoreRefs bool
 	bytePatches map[string][]fileSlice
+
+	// src contains the source files of the derivation. For flakes, this is
+	// anything in the flake.nix directory.
+	src *packageFS
 }
 
 // NewDerivationBuilder initializes a new DerivationBuilder from the current
@@ -79,6 +84,9 @@ func (d *DerivationBuilder) init() error {
 			return fmt.Errorf("patchpkg: can't patch gcc using %s: %v", d.Gcc, err)
 		}
 	}
+	if src := os.Getenv("src"); src != "" {
+		d.src = newPackageFS(src)
+	}
 	return nil
 }
 
@@ -95,6 +103,11 @@ func (d *DerivationBuilder) Build(ctx context.Context, pkgStorePath string) erro
 }
 
 func (d *DerivationBuilder) build(ctx context.Context, pkg, out *packageFS) error {
+	// Create the derivation's $out directory.
+	if err := d.copyDir(out, "."); err != nil {
+		return err
+	}
+
 	if d.RestoreRefs {
 		if err := d.restoreMissingRefs(ctx, pkg); err != nil {
 			// Don't break the flake build if we're unable to
@@ -103,11 +116,18 @@ func (d *DerivationBuilder) build(ctx context.Context, pkg, out *packageFS) erro
 			slog.ErrorContext(ctx, "unable to restore all removed refs", "err", err)
 		}
 	}
+	if err := d.findCUDA(ctx, out); err != nil {
+		slog.ErrorContext(ctx, "unable to patch CUDA libraries", "err", err)
+	}
 
 	var err error
 	for path, entry := range allFiles(pkg, ".") {
 		if ctx.Err() != nil {
 			return ctx.Err()
+		}
+		if path == "." {
+			// Skip the $out directory - we already created it.
+			continue
 		}
 
 		switch {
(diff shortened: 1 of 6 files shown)


internal/patchpkg/builder.go
```diff
@@ -302,6 +322,65 @@ func (d *DerivationBuilder) findRemovedRefs(ctx context.Context, pkg *packageFS)
...
+		if err != nil {
+			return fmt.Errorf("get absolute path to library: %v", err)
+		}
+		d.glibcPatcher.needed = append(d.glibcPatcher.needed, need)
+
+		slog.DebugContext(ctx, "added DT_NEEDED entry for system CUDA library", "path", need)
+	}
+
+	slog.DebugContext(ctx, "looking for nix CUDA libraries in $patchDependencies")
+	deps := os.Getenv("patchDependencies")
```
savil, line 357: should we error or skip below loop if this is empty?

internal/patchpkg/builder.go
```diff
@@ -302,6 +322,65 @@ func (d *DerivationBuilder) findRemovedRefs(ctx context.Context, pkg *packageFS)
...
+	for _, pkg := range strings.Split(deps, " ") {
+		slog.DebugContext(ctx, "checking for nix CUDA libraries in package", "pkg", pkg)
```
savil, line 359: the env-var is a more general "patchDependencies" but this loop has log-messages that are specific to CUDA. Could we update them to be more general as well?

internal/patchpkg/search.go
```diff
@@ -80,3 +82,97 @@ func searchEnv(re *regexp.Regexp) string {
...
+	// Common direct paths to try first.
+	linkerNames := []string{
+		"/usr/lib/x86_64-linux-gnu/libcuda.so", // Debian
+		"/usr/lib64/libcuda.so",                // Red Hat
+		"/usr/lib/libcuda.so",
+	}
+	for _, path := range linkerNames {
+		// Return what the link is pointing to because the dynamic
+		// linker will want libcuda.so.1, not libcuda.so.
+		soname, err := os.Readlink(path)
```
savil, line 117: so, unlike `readlink -f` this does one hop and not all the way to the end (i.e. the "real name" libcuda)?

internal/shellgen/flake_plan.go
```diff
@@ -162,5 +253,46 @@ func (g *glibcPatchFlake) fetchClosureExpr(pkg *devpkg.Package) (string, error)
 }
 
 func (g *glibcPatchFlake) writeTo(dir string) error {
+	wantCUDA := slices.ContainsFunc(g.Dependencies, func(dep string) bool {
```
savil, line 256: reviewed till here. Looking good so far! Will resume soon.

internal/shellgen/tmpl/glibc-patch.nix.tmpl
```diff
@@ -34,9 +33,19 @@
           }
         else null) args;
 
+      selectDefaultOutputs = drv: selectOutputs drv (drv.meta.outputsToInstall or [ drv.out ]);
+      selectAllOutputs = drv: drv.all;
+      selectOutputs = drv: builtins.map (output: drv.${output});
```
savil, line 38: how are these used?

savil (approved):
LGTM, modulo one question below

internal/patchpkg/search.go
gcurtis: Yeah, it's getting the middle name. It goes `libcuda.so` → `libcuda.so.1` → `libcuda.so.555.123.123`.

internal/shellgen/tmpl/glibc-patch.nix.tmpl
gcurtis: I can comment with some examples. They're functions that take a derivation and a list of output names, and return those outputs. So

```
selectOutputs nixpkgs#foo [ "out", "lib", "dev" ]
```

will return ` [ nixpkgs#foo.out nixpkgs#foo.lib nixpkgs#foo.dev ]`.

gcurtis: Thery're used by the template to install multiple outputs from a package. It basically generates `fmt.Sprintf("selectOutputs %s %q", ..., installable.SplitOutputs())`.

internal/patchpkg/builder.go
```diff
@@ -302,6 +322,65 @@ func (d *DerivationBuilder) findRemovedRefs(ctx context.Context, pkg *packageFS)
 	return refs, nil
 }
 
+func (d *DerivationBuilder) findCUDA(ctx context.Context, out *packageFS) error {
```
mikeland73, line 325: Food for future thought: Is there a way to do this in a more plugin-like way? Maybe we allow built in plugins to use go code (or in the future compile plugins as needed).

gcurtis: I like the idea of aligning plugins and patching. This code runs as part of the flake builder (in the nix build sandbox), so plugins would just need something similar.

PR #2299, opened 2024-09-24
[services] Let users set the port for process-compose

Lagoja:
## Summary

Users can set the port that process-compose runs on when they use `devbox services`. If they do not specify a port, we default to randomly picking an open port with low risk of conflict

Users can specify their process-compose port by:

1. Passing the `--pcport, -p` flag when running `devbox services up`. 
2. Setting the `PC_PORT_NUM` environment variable. This variable can be set in the projects `devbox.json` if you want to always use the same port for process-compose. This is overridden by the `--pcport` flag.

## How was it tested?

In the MySQL example: 

1. When I run `devbox services up -b -p 8080`, or
2. When I run `PC_PORT_NUM devbox services up -b`,

I can verify that process-compose is running on my 8080 with `curl localhost:8080/processes` and seeing the list of expected processes.

### DIFF
--- a/internal/services/manager.go
+++ b/internal/services/manager.go
@@ -44,6 +44,7 @@ type ProcessComposeOpts struct {
 	BinPath    string
 	ExtraFlags []string
 	Background bool
+	PCPort     int
 }
 
 func newGlobalProcessComposeConfig() *globalProcessComposeConfig {
@@ -128,10 +129,13 @@ func StartProcessManager(
 	config := readGlobalProcessComposeJSON(configFile)
 	config.File = configFile
 
-	// Get the port to use for this project
-	port, err := getAvailablePort()
+	port, err := selectPort(processComposeConfig.PCPort)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to select port: %v", err)
+	}
+
+	if !isPortAvailable(port) {
+		return fmt.Errorf("port %d is already in use", port)
 	}
 
 	// Start building the process-compose command
--- a/internal/services/manager.go
+++ b/internal/services/manager.go
@@ -41,10 +41,10 @@ type globalProcessComposeConfig struct {
 }
 
 type ProcessComposeOpts struct {
-	BinPath    string
-	ExtraFlags []string
-	Background bool
-	PCPort     int
+	BinPath            string
+	ExtraFlags         []string
+	Background         bool
+	ProcessComposePort int
 }
 
 func newGlobalProcessComposeConfig() *globalProcessComposeConfig {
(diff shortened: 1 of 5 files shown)


internal/services/manager.go
```diff
@@ -44,6 +44,7 @@ type ProcessComposeOpts struct {
 	BinPath    string
 	ExtraFlags []string
 	Background bool
+	PCPort     int
```
mikeland73, line 47: `ProcesssComposePort`

internal/services/ports.go
```diff
@@ -60,6 +63,34 @@ func getAvailablePort() (int, error) {
 	return 0, errors.New("no available port")
 }
 
+func selectPort(configPort int) (int, error) {
+	if configPort != 0 {
+		return configPort, nil
+	}
+
+	if portStr, exists := os.LookupEnv("PC_PORT_NUM"); exists {
```
mikeland73, line 71: Thoughts on naming this `DEVBOX_PC_PORT_NUM`

Reasons:
* If user sets `PC_PORT_NUM` outside of devbox, it's going to make starting devbox services on multiple projects fail. Since this is the env name set by process compose itself, it's not unreasonable that people will set it.
* Keeping track of env variables is easier if we prefix them
* Avoids collisions

internal/services/ports.go
```diff
@@ -60,6 +63,34 @@ func getAvailablePort() (int, error) {
...
+	}
+
+	return getAvailablePort()
+}
+
 func isAllowed(port int) bool {
 	return port > 1024 && disallowedPorts[port] == ""
 }
+
+func isPortAvailable(port int) bool {
```
mikeland73, line 89: Make `getAvailablePort` call this.

internal/services/ports.go
```diff
@@ -60,6 +63,34 @@ func getAvailablePort() (int, error) {
...
+
+	if portStr, exists := os.LookupEnv("PC_PORT_NUM"); exists {
+		port, err := strconv.Atoi(portStr)
+		if err != nil {
+			return 0, fmt.Errorf("invalid PC_PORT_NUM environment variable: %v", err)
+		}
+		if port <= 0 {
+			return 0, fmt.Errorf("invalid PC_PORT_NUM environment variable: ports cannot be less than 0")
+		}
+		return port, nil
```
mikeland73, line 79: `return port, isPortAvailable(port)` so you don't have to call it outside.

internal/services/manager.go
```diff
@@ -128,10 +129,13 @@ func StartProcessManager(
...
-	// Get the port to use for this project
-	port, err := getAvailablePort()
+	port, err := selectPort(processComposeConfig.PCPort)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to select port: %v", err)
+	}
+
+	if !isPortAvailable(port) {
+		return fmt.Errorf("port %d is already in use", port)
```
mikeland73, line 138: remove, call inside selectPort

mikeland73 (approved):

PR #2303, opened 2024-09-25
[nix] Add DetSys installer behind feature flag

mikeland73:
## Summary

* Adds detsys installer.
* Puts it behind disabled flag. Add CICD matrix to test.
* Special casing: Handles missing systemd by just skipping it.

## How was it tested?

* Tested `devbox setup nix` in linux docker container.
* Will test in CICD
* Much more testing needed before removing feature flag

### DIFF
--- a/internal/nix/install.go
+++ b/internal/nix/install.go
@@ -29,7 +29,7 @@ const rootError = "warning: installing Nix as root is not supported by this scri
 
 // Install runs the install script for Nix. daemon has 3 states
 // nil is unset. false is --no-daemon. true is --daemon.
-func Install(writer io.Writer, daemon *bool) error {
+func Install(writer io.Writer, daemon func() *bool) error {
 	if isRoot() && build.OS() == build.OSWSL {
 		return usererr.New("Nix cannot be installed as root on WSL. Please run as a normal user with sudo access.")
 	}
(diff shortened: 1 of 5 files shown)


gcurtis (approved):

internal/nix/install.go
```diff
@@ -39,7 +40,14 @@ func Install(writer io.Writer, daemon *bool) error {
 	defer r.Close()
 
 	installScript := "curl -L https://releases.nixos.org/nix/nix-2.24.7/install | sh -s"
-	if daemon != nil {
+	if featureflag.UseDetSysInstaller.Enabled() {
+		// Should we pin version? Or just trust detsys
+		installScript = "curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install"
+		if isLinuxWithoutSystemd() {
+			installScript += " linux --init none"
```
savil, line 47: I presume this is to manage the nix daemon.
How does the nix daemon get started without this?

mikeland73: My understanding from the DetSys [docs](https://github.com/DeterminateSystems/nix-installer) is that this makes it root only but will still work. 

There are a few alternatives to this:

* Warn and give user options (e.g. install systemd or use new `--root-only`)
* Warn and install anyway

savil: Seems like they'll then need to run `sudo` to run devbox. I think emitting a warning (with a suggestion to install systemd and re-using the detsys installer manually) and continuing to install seems okay to minimize friction.  WDYT?

mikeland73: Not quite, user needs sudo only if not root, so at least some cases (e.g. ubuntu dockerfile) work pretty well as is.

That said, printing a message is very low friction. I'm not 100% sure re-running detsys installation would fix it, I need to experiment a bit more.

PR #2308, opened 2024-09-27
Build and Install Devbox via Flake

Lagoja:
## Summary

This adds a flake.nix to our repo for building and distributing Devbox. This will make it easier for developers who install Devbox with Nix to use pre-releases, and keep Devbox up-to-date on their system. 

This changes our workflow in a few small ways:

1. Whenever `go.mod` changes, we'll need to run `devbox run update-hash` to update the vendorHash for the flake. You can also run this script when you run `go mod tidy`
2. Releasing a new version will also require us to update the version number in the flake.nix. This should be done with a commit ("Bump version to x.y") that also gets tagged with the release number.
3. For debugging purposes, the flake will add the shortened commit hash to the version string. This also helps identify if a user built the binary from a flake.
4. I've added a test to `cli-tests` that will build the flake. I've confirmed that it will fail if the vendorHash is out of date or incorrect
5. We'll periodically need to run `nix flake update` to update the flake.lock file. 

## How was it tested?

1. Ran `nix build .` in the Devbox repo, confirmed that the resulting build worked
2. Tested the Github Action by pushing an invalid vendorHash

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -43,11 +43,11 @@
       // The hash is saved to the `vendor-hash` file, which is then
       // read into the Nix Flake.
       "update-hash": [
-        "go mod vendor",
-        "hash=$(nix hash path vendor)",
-        "echo \"$hash\" > vendor-hash",
-        "rm -rf vendor",
-        "echo \"Vendor Hash: $hash\""
+        // realpath to work-around nix hash not liking symlinks
+        "vendor=$(realpath $(mktemp -d))",
+        "trap \"rm -rf $vendor\" EXIT",
+        "go mod vendor -o $vendor",
+        "nix hash path $vendor >vendor-hash",
       ],
       "tidy": ["go mod tidy", "devbox run update-hash"],
       // docker-testscripts runs the testscripts with Docker to exercise
--- a/devbox.json
+++ b/devbox.json
@@ -4,7 +4,7 @@
   "packages": {
     "go": "latest",
     "runx:golangci/golangci-lint": "latest",
-    "runx:mvdan/gofumpt": "latest",
+    "runx:mvdan/gofumpt": "latest"
   },
   "env": {
     "GOENV": "off",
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -0,0 +1,69 @@
...
+          postInstall = ''
+            installShellCompletion --cmd devbox \
+              --bash <($out/bin/devbox completion bash) \
+              --fish <($out/bin/devbox completion fish) \
+              --zsh <($out/bin/devbox completion zsh)
+          '';
+
+          meta = with pkgs.lib; {
+            description = "Instant, easy, predictable shells and containers";
+            homepage = "https://www.jetpack.io/devbox";
```
gcurtis, line 62:
```suggestion
            homepage = "https://www.jetify.com/devbox";
```

flake.nix
```diff
@@ -0,0 +1,69 @@
...
+
+          postInstall = ''
+            installShellCompletion --cmd devbox \
+              --bash <($out/bin/devbox completion bash) \
+              --fish <($out/bin/devbox completion fish) \
+              --zsh <($out/bin/devbox completion zsh)
+          '';
+
+          meta = with pkgs.lib; {
+            description = "Instant, easy, predictable shells and containers";
```
gcurtis, line 61: Should we match the repo description?

```suggestion
            description = "Instant, easy, and predictable development environments";
```

devbox.json
```diff
@@ -1,45 +1,55 @@
...
+      // Updates the Flake's vendorHash: First run `go mod vendor` to vendor
+      // the dependencies, then hash the vendor directory with Nix.
+      // The hash is saved to the `vendor-hash` file, which is then
+      // read into the Nix Flake.
+      "update-hash": [
+        "go mod vendor",
+        "hash=$(nix hash path vendor)",
+        "echo \"$hash\" > vendor-hash",
+        "rm -rf vendor",
+        "echo \"Vendor Hash: $hash\""
```
gcurtis, line 50: Could we just vendor into a temp dir? I'm not sure what something like gopls will do if it suddenly sees a new vendor directory.

```suggestion
        // realpath to work-around nix hash not liking symlinks
        "vendor=$(realpath $(mktemp -d))",
        "trap \"rm -rf $vendor\" EXIT",
        "go mod vendor -o $vendor",
        "nix hash path $vendor >vendor-hash",
```

gcurtis (approved):
This is awesome, thanks!

PR #2310, opened 2024-09-28
Fix flake rev issue

Lagoja:
## Summary

Nix flake fails to build with:

```nix
       error: attribute 'dirtyShortRev' missing
       at /nix/store/m5mqwkz0ss2cav2c9663mx8vwsmvcas3-source/flake.nix:21:28:
           20|               then "${x}-${self.shortRev}"
           21|               else "${x}-${self.dirtyShortRev}")
             |                            ^
           22|         ];
```

This fixes the issue

## How was it tested?

```
nix build github:jetify-com/devbox/jl/fix-flake-rev
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,7 @@
           (x: "${lastTag}")
           (x: if (self ? shortRev)
               then "${x}-${self.shortRev}"
-              else "${x}-${if self ? dirtyShortRev then self.dirtyShortRev else "dirty"}")
+              else "${x}-${self.dirtyShortRev or "dirty"}")
         ];
 
         # Run `devbox run update-flake` to update the vendorHash
--- a/flake.nix
+++ b/flake.nix
@@ -13,15 +13,14 @@
 
         lastTag = "0.13.2";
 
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -16,9 +16,9 @@
         # Add the commit to the version string, in case someone builds from main
         getVersion = pkgs.lib.trivial.pipe self [
           (x: "${lastTag}")
-          (x: if (self ? revCount)
+          (x: if (self ? shortRev)
               then "${x}-${self.shortRev}"
-              else "${x}-${self.dirtyShortRev}")
+              else "${x}-${if self ? dirtyShortRev then self.dirtyShortRev else "dirty"}")
```
gcurtis, line 21: There's also an [or syntax] that provides an alternative value when an attribute doesn't exist.

```suggestion
              else "${x}-${self.dirtyShortRev or "dirty"}")
```

[or syntax]: https://nix.dev/manual/nix/2.24/language/operators#attribute-selection

gcurtis (approved):

PR #2325, opened 2024-10-06
[zeroconfig] Implement --autodetect flag

mikeland73:
## Summary

Closes DEV-151

(re) implements something similar to our old "planner" framework but only for python/poetry. It's only opt-in.

Using the `--autodetect` with `init` flag generates a devbox.json with any detection results.

Currently the framework only only uses the most relevant detector, but could be modified to use multiple detectors.

## How was it tested?

`devbox init --autodetect --dry-run`

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -1,14 +1,14 @@
 {
-  "name": "devbox",
+  "name":        "devbox",
   "description": "Instant, easy, and predictable development environments",
   "packages": {
-    "go": "latest",
+    "go":                          "latest",
     "runx:golangci/golangci-lint": "latest",
-    "runx:mvdan/gofumpt": "latest"
+    "runx:mvdan/gofumpt":          "latest",
   },
   "env": {
     "GOENV": "off",
-    "PATH": "$PATH:$PWD/dist"
+    "PATH":  "$PATH:$PWD/dist",
   },
   "shell": {
     "init_hook": [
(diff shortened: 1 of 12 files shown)


devbox.json
```diff
@@ -68,8 +68,8 @@
...
         "GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go test -c -o testscripts-linux-arm64",
         "image=$(docker build --quiet --tag devbox-testscripts-ubuntu:noble --platform linux/amd64 .)",
-        "docker run --rm --mount type=volume,src=devbox-testscripts-amd64,dst=/nix --platform linux/amd64 -e DEVBOX_RUN_FAILING_TESTS -e DEVBOX_RUN_PROJECT_TESTS -e DEVBOX_DEBUG $image \"$@\""
-      ]
-    }
-  }
+        "docker run --rm --mount type=volume,src=devbox-testscripts-amd64,dst=/nix --platform linux/amd64 -e DEVBOX_RUN_FAILING_TESTS -e DEVBOX_RUN_PROJECT_TESTS -e DEVBOX_DEBUG $image \"$@\"",
+      ],
+    },
+  },
```
savil, line 74: no concerns with the change. Seems a no-op, right? 
Is there a setting or something that we can all use to align our tools?

internal/autodetect/detector/detector.go
```diff
@@ -0,0 +1,8 @@
+package detector
+
+import "context"
+
+type Detector interface {
+	IsRelevant(path string) (float64, error)
```
savil, line 6: could we rename `IsRelevant` to `Score`, or if you'd like `RelevancyScore` ?

EDIT: `Relevance() (float64, error)` also seems good

internal/boxcli/init.go
```diff
@@ -19,16 +26,32 @@ func initCmd() *cobra.Command {
 			"You can then add packages using `devbox add`",
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInitCmd(args)
+			return runInitCmd(cmd, args, flags)
 		},
 	}
 
+	command.Flags().BoolVar(&flags.autoDetect, "autodetect", false, "Automatically detect packages to add")
```
savil, line 33: thoughts on just `auto`?

savil (approved):

devbox.json
mikeland73: devbox does this automatically if I ever run any devbox command that mutates the config. I guess we haven't done that in a while in this repo.

PR #2332, opened 2024-10-07
[docs] Update deps

mikeland73:
## Summary

I wanted to update dependencies to see if we can get rid of some Dependabot warnings. Also wanted to modernize our devbox.json a bit to make the onboarding process easier. Not totally working though.

## How was it tested?

I'm able to do `devbox run serve` but I only get 404s.

If I do build/start I get:

`Error: Cannot mix different versions of joi schemas` 

Not sure if I introduced this issue.

### DIFF
--- a/docs/app/devbox.json
+++ b/docs/app/devbox.json
@@ -1,3 +1,18 @@
 {
-  "packages": ["nodejs@latest"]
+  "packages": [
+    "nodejs@20.12.2",
+  ],
+  "env": {
+    "DEVBOX_COREPACK_ENABLED": "true",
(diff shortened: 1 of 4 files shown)


Lagoja (approved):

Lagoja:
I'm not seeing the conflicting schema error -- RE: the 404, that's probably because there is no page served by Docusaurus at the root, you have to navigate to `localhost/docs/[devbox|cloud|nixhub]`

Just made one small change to pin the pnpm version in the config, to make sure we avoid any issues there

Issue #2085, opened 2024-05-25
Fleek is deprecated - remove it from docs


kosciolek:
### What problem are you trying to solve?

[Fleek](https://github.com/ublue-os/fleek) is deprecated now. It probably should be removed from [the docs](https://www.jetify.com/devbox/docs/devbox_global/#using-fleek-with-devbox-global)

### What solution would you like?

-

PR #2338, opened 2024-10-08
Remove Fleek from Docs

Lagoja:
## Summary

Since Fleek is deprecated, removing it from Global docs

## How was it tested?

devbox run build


Lagoja:
Resolves #2085

mikeland73 (approved):

Issue #2335, opened 2024-10-08
Add a clean command to inverse `devbox init`


trya2l:
### What problem are you trying to solve?

Sometimes, I start using devbox in a folder, but I later on decide to migrate to another one, for different considerations.

In this case, I'd like to rapidly clean up the folder, but then I need to do that manually, even if this is not a complex task.

### What solution would you like?

I'd love something like `devbox clean` to remove devbox from the current folder (including direnv when the `devbox generate direnv` has been invoked)

### Alternatives you've considered

Currently, I remove everything manually

`rm -rf .devbox devbox.* .envrc`

Lagoja:
@DerArkeN @trya2l -- are you sure that you'd want to delete the `devbox.json` as well? I worry people might run this to clear their `.devbox` and `devbox.lock` state and lose their `devbox.json` unexpectedly.

DerArkeN:
I am with you on this one. Maybe add a "--hard" or similar if you want to delete the config as well?

trya2l:
@DerArkeN thanks for the PR, that was fast!

@Lagoja You're right, I didn't consider the potential side effects.

I'm unfamiliar with devbox under the hood, but as @DerArkeN mentionned, a flag `--hard`, `--all`, `--force`, ... could be a good way to go, I guess.

Now that I regard this more seriously, even if a lot of software doesn't provide the ability to clean/revert the changes applied, as a user, this still looks like a desirable feature, even more so if it doesn't conflict with future enhancements.

DerArkeN:
can we link the right pr to this?

PR #2341, opened 2024-10-09
boxcli: add clean command

DerArkeN:
## Summary
This if a first draft for #2335, I am not sure about the behaviour or features of the command yet so I only added basic functionality.
The issue also mentioned deleting files like .envrc, I didn't include this yet since these files might be generated by a user instead of the devbox, maybe this should be added by an optional parameter? or are generated files tracked?
I also wasn't sure if the clean command should exit the devbox shell or not. In case it should, is the pid of the devbox saved anywhere so it can be killed, or is there a better way to do so?

## How was it tested?
- devbox init
- devbox add go
- devbox shell
- devbox clean

=> expected result: .devbox, devbox.json, devbox.lock are deleted and the user remains in the shell.

### DIFF
--- a/internal/boxcli/clean.go
+++ b/internal/boxcli/clean.go
@@ -1,0 +1,34 @@
+// Copyright 2024 Jetify Inc. and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package boxcli
+
+import (
+	"github.com/spf13/cobra"
+
+	"go.jetpack.io/devbox/internal/devbox"
+)
+
+type cleanFlags struct{}
+
+func cleanCmd() *cobra.Command {
+	flags := &cleanFlags{}
+	command := &cobra.Command{
+		Use:   "clean",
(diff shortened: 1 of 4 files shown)


trya2l:
Hi, what about `--all`, or `--force` ?

DerArkeN:
All might be an option. Imo force is something else, it means perform an action even though it might break things, an doesn't fit the action here.

adamdicarlo0:
How about `devbox uninit` or `devbox eject`?

A command named `clean` deleting `devbox.json` and `devbox.lock` would be very surprising to me.

bjungs:
Hello! Any reason this hasn't gone forward?

truearken:
@trya2l is this not of interest anymore? i can either fix the conflicts or close the pr.

trya2l:
@truearken Hi, honestly, it’s been a long time since I last used devbox, so I’m not sure. I guess this functionality could still be useful, but it’s up to you.

PR #2346, opened 2024-10-10
[init] init config should return error if already exists

mikeland73:
## Summary

This fixes a NPE in autodetect. Previously when initializing a config with a directory that already contained a config we would return `(nil, nil)` but this is a mistake and a bit inconsistent with typical go. This changes it so that if config already exists we return an error instead.

IMO this also provides better UX displaying a message indicating config already exists.

This does have a minor backward compatibility issue. If users where used to calling `devbox init` and having it not fail they would now need to first test if devbox.json already exists.

### DIFF
--- a/internal/boxcli/init.go
+++ b/internal/boxcli/init.go
@@ -10,8 +10,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"go.jetpack.io/devbox/internal/boxcli/usererr"
 	"go.jetpack.io/devbox/internal/devbox"
+	"go.jetpack.io/devbox/internal/ux"
 	"go.jetpack.io/devbox/pkg/autodetect"
 )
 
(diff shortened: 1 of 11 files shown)


Lagoja:
If we want to be really cautious/backwards compatible here, we could have it emit a warning for 1 release, and then make it a hard error in a subsequent release

gcurtis (approved):

internal/boxcli/init.go
```diff
@@ -28,7 +30,15 @@ func initCmd() *cobra.Command {
...
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInitCmd(cmd, args, flags)
+			err := runInitCmd(cmd, args, flags)
+			if errors.Is(err, os.ErrExist) {
+				path := pathArg(args)
+				if path == "" || path == "." {
+					path, _ = os.Getwd()
+				}
+				return usererr.New("devbox.json already exists in %q.", path)
```
savil, line 39: Could we make switch this to a warning emitted instead and make it a non-error? 
I don't think its _that_ bad to run `devbox init` if a devbox.json already exists. Its useful for scripts.

savil (commented):
Please lets switch this to a no-op error if we can?

PR #2347, opened 2024-10-10
[clean] Add clean command

mikeland73:
## Summary

fixes https://github.com/jetify-com/devbox/pull/2341

### DIFF
--- a/internal/boxcli/clean.go
+++ b/internal/boxcli/clean.go
@@ -1,0 +1,38 @@
+// Copyright 2024 Jetify Inc. and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package boxcli
+
+import (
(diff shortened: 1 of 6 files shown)


DerArkeN:
I am a little confused about 2 PRs for the same thing being open at once, also I don't know what this "fixes".
I also noticed this will delete the config file again, which we discussed not to delete for the default behavior.
Shouldn't we keep the discussion at one single point instead of working on the same thing in parallel?

mikeland73:
@DerArkeN sorry about this, I hadn't see your PR and drafted this change when I saw a comment from @Lagoja. I'll close this and focus on your PR instead.

Also, my description was careless and wrong. What I meant to write was "fixes https://github.com/jetify-com/devbox/issues/2335" which causes github to associate the PR to the issue. I somehow copied your PR and linked to it instead.

PR #2355, opened 2024-10-15
[env] Fixed env_from bug when it is called from a subdirectory

mohsenari:
## Summary
Fixes this bug:
https://github.com/jetify-com/devbox/pull/2174/files#r1659165116

## How was it tested?
- use compiled devbox
- put `export FOO = 'BBBAR'` in .env file
- put the following in a devbox.json file
```json
  {
    "packages": [],
    "shell": {
      "init_hook": [
      ],
      "scripts": {
        "test": [
          "echo $FOO"
        ]
      }
    },
    "env_from": ".env"
  }
```
- create a subdirectory besides devbox.json called `test`
- cd test && devbox run test

### DIFF
--- a/internal/devconfig/configfile/env.go
+++ b/internal/devconfig/configfile/env.go
@@ -26,7 +26,12 @@ func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
 	if !c.IsdotEnvEnabled() {
 		return nil, fmt.Errorf("env file does not have a .env extension")
 	}
-	envFileAbsPath := path.Join(filepath.Dir(c.AbsRootPath), c.EnvFrom)
+	envFileAbsPath := filepath.Dir(c.AbsRootPath)
+	if filepath.IsAbs(c.EnvFrom) {
+		envFileAbsPath = path.Join(envFileAbsPath, path.Base(c.EnvFrom))
+	} else {
+		envFileAbsPath = path.Join(envFileAbsPath, c.EnvFrom)
+	}
 	file, err := os.Open(envFileAbsPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to open file: %s", envFileAbsPath)
--- a/internal/devconfig/configfile/env.go
+++ b/internal/devconfig/configfile/env.go
@@ -3,7 +3,6 @@ package configfile
 import (
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 
 	"github.com/hashicorp/go-envparse"
@@ -26,11 +25,10 @@ func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
 	if !c.IsdotEnvEnabled() {
 		return nil, fmt.Errorf("env file does not have a .env extension")
 	}
-	// We don't allow absolute path for env_from as it breaks reproducibility
-	if filepath.IsAbs(c.EnvFrom) {
-		return nil, fmt.Errorf("`env_from` field can't take an absolute path")
+	envFileAbsPath := c.EnvFrom
+	if !filepath.IsAbs(c.EnvFrom) {
+		envFileAbsPath = filepath.Join(filepath.Dir(c.AbsRootPath), c.EnvFrom)
 	}
-	envFileAbsPath := path.Join(filepath.Dir(c.AbsRootPath), c.EnvFrom)
 	file, err := os.Open(envFileAbsPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to open file: %s", envFileAbsPath)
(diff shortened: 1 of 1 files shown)


internal/devconfig/configfile/env.go
```diff
@@ -25,10 +26,10 @@ func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
 	if !c.IsdotEnvEnabled() {
 		return nil, fmt.Errorf("env file does not have a .env extension")
 	}
-
-	file, err := os.Open(c.EnvFrom)
+	envFileAbsPath := path.Join(filepath.Dir(c.AbsRootPath), c.EnvFrom)
```
mikeland73, line 29: This breaks with absolute paths. I think we need to treat absolute and relative paths differently.

mikeland73 (changes requested):

internal/devconfig/configfile/env.go
mohsenari: pushed a commit to handle this case.

internal/devconfig/configfile/env.go
```diff
@@ -25,10 +26,15 @@ func (c *ConfigFile) ParseEnvsFromDotEnv() (map[string]string, error) {
 	if !c.IsdotEnvEnabled() {
 		return nil, fmt.Errorf("env file does not have a .env extension")
 	}
-
-	file, err := os.Open(c.EnvFrom)
+	envFileAbsPath := filepath.Dir(c.AbsRootPath)
+	if filepath.IsAbs(c.EnvFrom) {
+		envFileAbsPath = path.Join(envFileAbsPath, path.Base(c.EnvFrom))
```
savil, line 31: should this scenario be `envFileAbsPath = c.EnvFrom`?

if the user has set
```
// devbox.json
{
   "env_from": "/my/absolute/path/foo.env`
}
```

Then, I'd think we want to just obey the absolute path directive. 

Perhaps we should error if the env_from path is absolute, because that would mess up any reproducibility of the devbox.json config. But I'll be open to just letting it fly. We can expect our users to use their judgement for when they want to make their config a bit less reproducible. For example, for the sake of some quick changes they may want to experiment with.

mohsenari: I agree with the absolute path erroring out case. Reproducibility should be considered here. I'll push a commit to change this case to an error.

savil (approved):

internal/devconfig/configfile/env.go
mikeland73: We allow absolute paths for flakes and plugins. I don't think env should be different. Absolute paths can be important for reproducibility as well.

mikeland73: Thinking about it even more, I don't think relative paths are any more reproducible than absolute paths! relative paths are more reproducible only if `.env` file is in the repo, but in that case you are better off just adding to the `env` field in devbox.json

mohsenari: @mikeland73 the idea is to prevent unintended mistakes that comes with putting absolute paths. For example, if user A puts `/Users/A/projects/Aproject/.env` and user B has the repo clone in `/Users/B/...` it will break the reproducibility. 
Do you think this is an acceptable risk and we should allow absolute path here?

mikeland73: I don't understand this logic at all. Why would the path of the config matter?

internal/devconfig/configfile/env.go
mikeland73: What you want is:

```golang
envfileAbsPath := c.EnvFrom
if !filepath.IsAbs(c.EnvFrom) {
  envfileAbsPath = filepath.Join(c.AbsRootPath, c.EnvFrom)
}
```

mikeland73 (changes requested):

mikeland73 (changes requested):
NACK

breaking absolute paths would be non-backward compatible

internal/devconfig/configfile/env.go
mohsenari: This was handling the case when env_from was absolute. Your suggestion below works too, it's just handling if abs() vs if !abs()

mikeland73: 2 reasons:

* erroring on abs path is not backward compatible. This reason alone is a blocker. path of least resistance is always to preserve backward compatibility, especially if it's easier! 
* relative poths also allow you to make it non reproducible! `../path/to/my/.env` would also live outside the repo. Or just `.env` when `.env` is in gititnore.

Why would folks use this feature with `.env` file that is checked in to their repo? (I guess some convenience maybe, but seems low value). The use case for this feature is for people to have _different_ .env files, so by definition reproducibility is not as important.

mikeland73: If `env_from` is absolute you don't need to change it at all.

mikeland73 (approved):

internal/devconfig/configfile/env.go
mohsenari: yeah I agree, I overcomplicated the logic there.

PR #2390, opened 2024-10-28
[auto] Add php detector

mikeland73:
## Summary

Add php detector

## How was it tested?

- [x] unit tests
- [ ] devbox init --auto

### DIFF
--- a/pkg/autodetect/detector/php_test.go
+++ b/pkg/autodetect/detector/php_test.go
@@ -1,0 +1,164 @@
+package detector
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


pkg/autodetect/detector/php_test.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+		{
+			name: "with composer.json",
+			setup: func(t *testing.T) string {
+				dir := t.TempDir()
+				err := os.WriteFile(filepath.Join(dir, "composer.json"), []byte(`{
+					"require": {
+						"php": "^8.1"
+					}
+				}`), 0644)
```
savil, line 35: suggest looking into using https://pkg.go.dev/testing/fstest

savil (approved):
thanks for adding tests!

PR #2391, opened 2024-10-28
Add --recompute flag for commands: shell and run

### LINKED ISSUE #2315: Option to skip package check on devbox run or devbox shell

Network issues when connecting to cache.nixos.org are a big blocker for me. I have to run `devbox run` multiple times for it to work even though the packages have not changed since the last time.

A flag to skip the check during devbox run would help mitigate such issues.

I was digging through the issues and came to https://github.com/jetify-com/devbox/issues/1882 but I can't seem to find anything about the recompute flag in the docs or how to disable it.

savil:
## Summary

This PR continues the work from #2013 to also add the `--recompute` flag option for `devbox run` and `devbox shell`.
For some users on bad networks, this can save them annoyance and time for when they _know_ their devbox environment is up-to-date.

Fixes #2315

## How was it tested?
This PR affects 3 commands: `run`, `shell` and `shellenv`.

1. For `run`:
Added `"hello": "latest",` to devbox.json of this project.

```
devbox run --recompute=false -- echo "hello world"
Warning: Your devbox environment may be out of date. Run with --recompute=true to update it.
hello world
```

then
```
devbox run -- echo "hello world"
hello world
```

2. For `shell`. Ran similar commands as above.

3. For `shellenv`. Followed test plan of #1963.

Changed the `.envrc` to be:
```
.envrc
@@ -1,7 +1,13 @@
 # Automatically sets up your devbox environment whenever you cd into this
 # directory via our direnv integration:

-eval "$(devbox generate direnv --print-envrc)"
+#eval "$(devbox generate direnv --print-envrc)"

+ # output of `devbox generate direnv --print-envrc` to modify it
+use_devbox() {
+    watch_file devbox.json devbox.lock
+    # eval "$(devbox shellenv --init-hook --install --no-refresh-alias)"
+    eval "$(devbox shellenv --init-hook --no-refresh-alias --recompute=false)"
+}
+use devbox
 # check out https://www.jetify.com/devbox/docs/ide_configuration/direnv/
 # for more details
```

Then modified devbox.json and saw the warning get printed.

### DIFF
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -21,10 +21,11 @@ import (
 
 type runCmdFlags struct {
 	envFlag
-	config      configFlags
-	omitNixEnv  bool
-	pure        bool
-	listScripts bool
+	config       configFlags
+	omitNixEnv   bool
+	pure         bool
+	listScripts  bool
+	recomputeEnv bool
 }
 
 // runFlagDefaults are the flag default values that differ
@@ -62,6 +63,7 @@ func runCmd(defaults runFlagDefaults) *cobra.Command {
 		"shell environment will omit the env-vars from print-dev-env",
 	)
 	_ = command.Flags().MarkHidden("omit-nix-env")
+	command.Flags().BoolVar(&flags.recomputeEnv, "recompute", true, "recompute environment if needed")
 
 	command.ValidArgs = listScripts(command, flags)
 
@@ -84,6 +86,7 @@ func listScripts(cmd *cobra.Command, flags runCmdFlags) []string {
 }
 
 func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
+	ctx := cmd.Context()
 	if len(args) == 0 || flags.listScripts {
 		scripts := listScripts(cmd, flags)
 		if len(scripts) == 0 {
@@ -111,9 +114,9 @@ func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
 	// Check the directory exists.
 	box, err := devbox.Open(&devopt.Opts{
 		Dir:         path,
+		Env:         env,
 		Environment: flags.config.environment,
 		Stderr:      cmd.ErrOrStderr(),
-		Env:         env,
 	})
 	if err != nil {
 		return redact.Errorf("error reading devbox.json: %w", err)
@@ -122,8 +125,12 @@ func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
 	envOpts := devopt.EnvOptions{
 		OmitNixEnv: flags.omitNixEnv,
 		Pure:       flags.pure,
+		RecomputeEnv: &devopt.RecomputeEnvOpts{
+			Disabled:              !flags.recomputeEnv,
+			StateOutOfDateMessage: fmt.Sprintf(devbox.StateOutOfDateMessage, "with --recompute=true"),
+		},
 	}
-	if err := box.RunScript(cmd.Context(), envOpts, script, scriptArgs); err != nil {
+	if err := box.RunScript(ctx, envOpts, script, scriptArgs); err != nil {
 		return redact.Errorf("error running script %q in Devbox: %w", script, err)
 	}
 	return nil
--- a/internal/boxcli/run.go
+++ b/internal/boxcli/run.go
@@ -17,6 +17,7 @@ import (
 	"go.jetpack.io/devbox/internal/devbox"
 	"go.jetpack.io/devbox/internal/devbox/devopt"
 	"go.jetpack.io/devbox/internal/redact"
+	"go.jetpack.io/devbox/internal/ux"
 )
 
 type runCmdFlags struct {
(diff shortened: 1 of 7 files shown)


internal/devbox/devbox.go
```diff
@@ -819,23 +804,34 @@ func (d *Devbox) ensureStateIsUpToDateAndComputeEnv(
...
 
-	// When ensureStateIsUpToDate is called with ensure=true, it always
-	// returns early if the lockfile is up to date. So we don't need to check here
-	if err := d.ensureStateIsUpToDate(ctx, ensure); isConnectionError(err) {
-		if !fileutil.Exists(d.nixPrintDevEnvCachePath()) {
-			ux.Ferrorf(
+	if envOpts.RecomputeEnv.Disabled {
+		upToDate, _ := d.lockfile.IsUpToDateAndInstalled(isFishShell())
+		if !upToDate {
+			ux.FHidableWarning(
```
mikeland73, line 810: This won't work. Hidable warnings need the format string to match. You are using a non format string as a format string (which is why govet is complaining). That means this string won't match the `HideMessage` call earlier on.

One way to fix:

Have `StateOutOfDateMessage` be only the part that differs. So something like:


```golang
ux.FHidableWarning(
	ctx,
	d.stderr,
	StateOutOfDateMessage, 
        envOpts.RecomputeEnv.StateOutOfDateMessage,
)
```

Not a blocker, but I this feels a bit fragile and connects the cobra command to the devbox code in a funky way. Maybe one way to improve this in the future would be to use lifecycle hooks:

```golang
envOpts.Lifecycle.OnEnvNotUpToDate = func() { ux.Warning } 
```

internal/boxcli/run.go
```diff
@@ -122,8 +125,12 @@ func runScriptCmd(cmd *cobra.Command, args []string, flags runCmdFlags) error {
 	envOpts := devopt.EnvOptions{
 		OmitNixEnv: flags.omitNixEnv,
 		Pure:       flags.pure,
+		RecomputeEnv: &devopt.RecomputeEnvOpts{
+			Disabled:              !flags.recomputeEnv,
+			StateOutOfDateMessage: fmt.Sprintf(devbox.StateOutOfDateMessage, "with --recompute=true"),
```
mikeland73, line 130: A few questions/comments:

* Is this message needed at all? It feels like if someone is voluntarily using --recompute=false, they know what they want. 
* This is copy pasted a bunch of times, maybe create a `var` for this?
* If this message is not needed, maybe inside devbox package it can simply skip the warning if `StateOutOfDateMessage` is empty. That way you only have to set this field in one place (in shellenv)

internal/devbox/devbox.go
```diff
@@ -819,23 +804,34 @@ func (d *Devbox) ensureStateIsUpToDateAndComputeEnv(
 ) (map[string]string, error) {
 	defer debug.FunctionTimer().End()
 
-	// When ensureStateIsUpToDate is called with ensure=true, it always
-	// returns early if the lockfile is up to date. So we don't need to check here
-	if err := d.ensureStateIsUpToDate(ctx, ensure); isConnectionError(err) {
-		if !fileutil.Exists(d.nixPrintDevEnvCachePath()) {
-			ux.Ferrorf(
+	if envOpts.RecomputeEnv.Disabled {
```
mikeland73, line 807: Changes in `ensureStateIsUpToDateAndComputeEnv` look OK, but please double triple check. This is very core functioning stuff.

internal/devbox/devopt/devboxopts.go
```diff
@@ -76,4 +75,10 @@ type EnvOptions struct {
 	OmitNixEnv        bool
 	PreservePathStack bool
 	Pure              bool
+	RecomputeEnv      *RecomputeEnvOpts
```
mikeland73, line 78: Why pointer? Default to structs unless you a) need mutating methods b) struct is massive and you want to avoid copying data.

mikeland73 (changes requested):

internal/devbox/devbox.go
savil: Yeah, this feels very icky. I like the lifecycles hooks approach. 👍🏾

internal/devbox/devopt/devboxopts.go
savil: In general, I go with pointers because its easier for the zero value when writing instantiations of the struct. This is especially relevant for test-cases. The downside is that the caller has to be more defensive for the nil-case.

mikeland73: This file uses all non-pointers, so in the interest of consistency we should keep that.

I don't follow why zero value is easier with non-pointer. Structs default to their zero values, no need to initialize anything.

The downside is quite large. A nil pointer use is a panic.

internal/devbox/devopt/devboxopts.go
```diff
@@ -62,18 +62,24 @@ type UpdateOpts struct {
...
 // - The struct is designed for the "common case" to be zero-initialized as `EnvOptions{}`.
 type EnvOptions struct {
+	Hooks             EnvLifecycleHooks
 	OmitNixEnv        bool
 	PreservePathStack bool
 	Pure              bool
+	SkipRecompute     bool
+}
+
+type EnvLifecycleHooks struct {
```
mikeland73, line 82: nit just `LifecycleHooks`

internal/devbox/devbox.go
```diff
@@ -819,23 +804,32 @@ func (d *Devbox) ensureStateIsUpToDateAndComputeEnv(
...
-	if err := d.ensureStateIsUpToDate(ctx, ensure); isConnectionError(err) {
-		if !fileutil.Exists(d.nixPrintDevEnvCachePath()) {
-			ux.Ferrorf(
+	if envOpts.SkipRecompute {
+		upToDate, _ := d.lockfile.IsUpToDateAndInstalled(isFishShell())
+		if !upToDate {
+			if envOpts.Hooks.OnStaleStateWithSkipRecompute != nil {
+				envOpts.Hooks.OnStaleStateWithSkipRecompute()
+			}
+		}
```
mikeland73, line 813: I was playing around with putting this outside the `envOpts.SkipRecompute` if statement.

It allows the hook to be more generic (i.e. `OnStaleState`). The downside is it can be a bit wasteful (calling `IsUpToDateAndInstalled` if we don't need to). `IsUpToDateAndInstalled` is fast (~1ms) but can be slower for large projects with remote plugins etc. So maybe we're all good as is. Just food for thought.

mikeland73 (approved):

internal/devbox/devbox.go
savil: Discussed in slack that this should be okay (for perf) to make generic into OnStaleState

PR #2395, opened 2024-10-30
[zeroconfig] Implement nodejs detector

mikeland73:
## Summary

TSIA

Possible improvement: should we default `DEVBOX_COREPACK_ENABLED` to true? (cc: @LucilleH )

## How was it tested?

- [x] unit tests
- [x] Manually created directory with package.json


LucilleH (approved):
Yes `DEVBOX_COREPACK_ENABLED` should be true by default

PR #2396, opened 2024-10-30
[zeroconfig] Default nodejs to corepack enabled

mikeland73:
## Summary

* Adds `DEVBOX_COREPACK_ENABLED: "1"` to nodejs detector.
* Adds ability to set env programmatically on config. (This probably doesn't work with interspaced comments)

## How was it tested?

* Unit tests
* Manually created project with package.json and then did `devbox init --auto`

### DIFF
--- a/internal/devconfig/configfile/ast.go
+++ b/internal/devconfig/configfile/ast.go
@@ -425,3 +425,25 @@ func (c *configAST) beforeComment(path ...any) []byte {
 		}),
 	)
 }
+
+func (c *configAST) setEnv(env map[string]string) {
+	members := make([]hujson.ObjectMember, 0, len(env))
+	for k, v := range env {
+		members = append(members, hujson.ObjectMember{
+			Name:  hujson.Value{Value: hujson.String(k)},
+			Value: hujson.Value{Value: hujson.String(v)},
+		})
+	}
+	i := c.memberIndex(c.root.Value.(*hujson.Object), "env")
+	if i == -1 {
+		c.root.Value.(*hujson.Object).Members = append(c.root.Value.(*hujson.Object).Members, hujson.ObjectMember{
+			Name:  hujson.Value{Value: hujson.String("env")},
+			Value: hujson.Value{Value: &hujson.Object{Members: members}},
+		})
+	} else {
+		c.root.Value.(*hujson.Object).Members[i].Value.Value = &hujson.Object{
+			Members: members,
+		}
+	}
+	c.root.Format()
+}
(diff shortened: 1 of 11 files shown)


internal/devconfig/configfile/ast.go
```diff
@@ -425,3 +425,25 @@ func (c *configAST) beforeComment(path ...any) []byte {
...
+		c.root.Value.(*hujson.Object).Members = append(c.root.Value.(*hujson.Object).Members, hujson.ObjectMember{
+			Name:  hujson.Value{Value: hujson.String("env")},
+			Value: hujson.Value{Value: &hujson.Object{Members: members}},
+		})
+	} else {
+		c.root.Value.(*hujson.Object).Members[i].Value.Value = &hujson.Object{
+			Members: members,
+		}
+	}
+	c.root.Format()
```
LucilleH, line 448: The above logic seems so convoluted - what's `hujson.ObjectMember`?

Is there a simpler representation? Something like
```
env := [key, val, key, val]
members := make(map[string]string)

for i := 0; i < len(env); i += 2 {
    members[env[i]] = env[i+1]
}
```

LucilleH (approved):

internal/devconfig/configfile/ast.go
mikeland73: I'm not sure if we can get rid of any of the code because hujson api is pretty generic so everything needs to be explicit. Maybe  I can refactor some portions out:


```golang
func (c *configAST) createMemberIfMissing(key string) *hujson.ObjectMember {
	i := c.memberIndex(c.root.Value.(*hujson.Object), key)
	if i == -1 {
		c.root.Value.(*hujson.Object).Members = append(c.root.Value.(*hujson.Object).Members, hujson.ObjectMember{
			Name: hujson.Value{Value: hujson.String(key)},
		})
		i = len(c.root.Value.(*hujson.Object).Members) - 1
	}
	return &c.root.Value.(*hujson.Object).Members[i]
}

func mapToObjectMembers(env map[string]string) []hujson.ObjectMember {
	members := make([]hujson.ObjectMember, 0, len(env))
	for k, v := range env {
		members = append(members, hujson.ObjectMember{
			Name:  hujson.Value{Value: hujson.String(k)},
			Value: hujson.Value{Value: hujson.String(v)},
		})
	}
	return members
}

func (c *configAST) setEnv(env map[string]string) {
	c.createMemberIfMissing("env").Value.Value = &hujson.Object{
		Members: mapToObjectMembers(env),
	}
	c.root.Format()
}
```

PR #2406, opened 2024-11-08
Fix flake version string issue

### LINKED ISSUE #2401: Python autopatching does not work when Devbox is installed via Flake
### What happened?

As of Devbox 0.13.6, we attempt to add Devbox to the Nix store so we can use it in the patch flake when auto-patching Python. When using Devbox installed from the Flake, the patching fails with the following error: 

```
❯ devbox add python@3.11
Info: Adding package "python@3.11" to devbox.json

Error: There was an error installing nix packages
source: nix: command error: nix --extra-experimental-features ca-derivations --option experimental-features 'nix-command flakes fetch-closure' print-dev-env --json path:/Users/johnlago/src/cling-notebook/.devbox/gen/flake: unable to download 'https://api.github.com/repos/jetify-com/devbox/commits/0.13.6-0bc66cb': HTTP error 422: exit code 1

(devbox)
```

Simplest fix is to change the version string in the Flake to match our standard installer, though we'll probably need a different approach for non-release builds?

### Steps to reproduce

1. Install devbox with `nix profile install github:jetify-com/devbox/latest`
2. Create a new Devbox project
3. Add Python with `devbox add python`

### Command

add

### devbox.json

_No response_

### Devbox version

0.13.6-0bc66cb

### Nix version

2.24.7

macOS (Apple Silicon)

### Debug logs

_No response_

Lagoja:
## Summary

Fixes #2401 

## How was it tested?
 
nix build . + CI/CD

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,17 +13,19 @@
 
         lastTag = "0.13.6";
 
-        revision = if (self ? shortRev)
-                   then "${self.shortRev}"
-                   else "${self.dirtyShortRev or "dirty"}";
+        revision =
+          if (self ? shortRev)
+          then "${self.shortRev}"
+          else "${self.dirtyShortRev or "dirty"}";
 
         # Add the commit to the version string for flake builds
-        version = "${lastTag}-${revision}";
+        version = "${lastTag}";
 
         # Run `devbox run update-flake` to update the vendor-hash
-        vendorHash = if builtins.pathExists ./vendor-hash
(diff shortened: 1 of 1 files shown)


gcurtis:
I think the original run was flakey PHP composer installs. There's another flakey test that should be fixed in #2417.

Lagoja:
Ok, i'll rebase once that's merged

gcurtis (approved):

PR #2418, opened 2024-11-21
[bug] plugin env PATH fix

### LINKED ISSUE #2138: Cannot set PATH env var in both a plugin and devbox.json
### What happened?

Setting `PATH` in a project's `devbox.json#env` will override that setting in made in any plugins. I haven't tested what happens if two plugins try to set `PATH` but expect it would be a "last wins" situation.

A workaround is to set `PATH` in `shell.init_hook` rather than in `env`. This is a bit more verbose but appears to always work as expected. A quick fix for the ruby plugin might be to do this, i.e.

```json
 "shell": {
      "init_hook": [
          "export PATH=\"{{ .Virtenv }}/bin:$PATH\""
      ]
  }
```

Ideally env vars in plugins would be expanded sequentially so that `PATH` could be modified by multiple plugins. But I'm not sure how difficult that would be to implement.

### Steps to reproduce

1. setup temp project: `cd $(mktemp -d) && devbox init && devbox generate direnv`
2. add ruby (the only built in plugin that sets `env.PATH`): `devbox add ruby`
3. OBSERVE ruby plugin has prepended a dir to the PATH: `echo $PATH | grep ".devbox/virtenv/ruby/bin"`
4. add `"env": { "PATH": "/path/to/my/bin:$PATH" }` to `devbox.json`
5. OBSERVE path has been prepended as expected `echo $PATH | grep "/path/to/my/bin"`
6. OBSERVE BUG ruby plugin's change to PATH has been overwritten: `echo $PATH | grep ".devbox/virtenv/ruby/bin"` (no result)

### Command

_No response_

### devbox.json

```jsonc
{
  "packages": ["ruby@latest"],
  "env": {
    "PATH": "/path/to/my/bin:$PATH"
  }
}
```

### Devbox version

0.11.0

### Nix version

2.18.1

macOS (Apple Silicon)

### Debug logs

_No response_

mohsenari:
## Summary
When a plugin modifies PATH and also config in devbox.json modifies PATH via `env:{}`, the env overwrites the plugin and so PATH modification from the plugin is lost. This is because we do `maps.Copy()` to merge the env from plugin to the one from config. 
The copy is fine for all other env variables but for PATH we need to handle merging the two rather than overwriting.
Fixes #2138 

## How was it tested?
bug recreate:
- devbox init
- devbox add ruby (ruby plugin modifies PATH)
- devbox shell (see PATH is prepended with ruby plugin)
- modify devbox.json and add `"env": {"PATH": "/my/config/bin:$PATH"}`
- re-enter devbox shell and see PATH prepend from ruby plugin is lost, only `"/my/config/bin"` is prepended.
fix recreate:
- follow same steps, but in last step both PATH prepends from devbox.json and from plugin are present.

### DIFF
--- a/internal/devconfig/config.go
+++ b/internal/devconfig/config.go
@@ -424,7 +424,7 @@ func createIncludableFromPluginConfig(pluginConfig *plugin.Config) *Config {
 //	/config/env/bin/:/plugin2/bin/:/plugin1/bin/:$PATH
 //
 // because config env should take priority over plugins
-func mergePATHsFromTwoEnvs(currentEnv map[string]string, newEnv map[string]string) map[string]string {
+func mergePATHsFromTwoEnvs(currentEnv, newEnv map[string]string) map[string]string {
 	if currentEnv["PATH"] != "" {
 		slog.Debug("A Plugin or Config wants to modify PATH. Processing the merge", "AddedPATH", newEnv["PATH"])
 		newEnv["PATH"] = strings.Replace(newEnv["PATH"], "$PATH", currentEnv["PATH"], 1)
--- a/internal/devconfig/config.go
+++ b/internal/devconfig/config.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 	"syscall"
 	"time"
 
@@ -347,18 +346,11 @@ func (c *Config) NixPkgsCommitHash() string {
 func (c *Config) Env() map[string]string {
 	env := map[string]string{}
 	for _, i := range c.included {
-		// Here, in new pluginPath we replace the keyword "$PATH"
-		// with what we have for PATH so far. If so far we have /plugin1/bin:$PATH
-		// and new plugin adds /plugin2/bin/:$PATH the result becomes
-		// /plugin2/bin/:/plugin1/bin/:$PATH
-		pluginEnv := mergePATHsFromTwoEnvs(env, i.Env())
-		maps.Copy(env, pluginEnv)
+		expandedEnvFromPlugin := OSExpandIfPossible(i.Env(), env)
+		maps.Copy(env, expandedEnvFromPlugin)
 	}
-	// Here we merge the processed PATH from all plugins to PATH in config env.
-	// So if config env has /config/env/:$PATH merging with the plugin PATH
-	// it becomes /config/env/:/plugin2/bin/:/plugin1/bin/:$PATH
-	rootEnv := mergePATHsFromTwoEnvs(env, c.Root.Env)
-	maps.Copy(env, rootEnv)
+	rootConfigEnv := OSExpandIfPossible(c.Root.Env, env)
+	maps.Copy(env, rootConfigEnv)
 	return env
 }
 
@@ -417,17 +409,18 @@ func createIncludableFromPluginConfig(pluginConfig *plugin.Config) *Config {
 	return includable
 }
 
-// Instead of overwriting PATH modifications we need to merge them
-// so merging of /plugin1/bin/:$PATH and /plugin2/bin/:$PATH
-// and /config/env/bin/:$PATH should result in
-//
-//	/config/env/bin/:/plugin2/bin/:/plugin1/bin/:$PATH
-//
-// because config env should take priority over plugins
-func mergePATHsFromTwoEnvs(currentEnv, newEnv map[string]string) map[string]string {
-	if currentEnv["PATH"] != "" && newEnv["PATH"] != "" {
-		slog.Debug("A Plugin or Config wants to modify PATH. Processing the merge", "AddedPATH", newEnv["PATH"])
-		newEnv["PATH"] = strings.Replace(newEnv["PATH"], "$PATH", currentEnv["PATH"], 1)
+func OSExpandIfPossible(env, existingEnv map[string]string) map[string]string {
+	mapping := func(value string) string {
+		// If the value is not set in existingEnv, return the value wrapped in ${...}
+		if existingEnv == nil || existingEnv[value] == "" {
+			return fmt.Sprintf("${%s}", value)
+		}
+		return existingEnv[value]
 	}
-	return newEnv
+
+	res := map[string]string{}
+	for k, v := range env {
+		res[k] = os.Expand(v, mapping)
+	}
+	return res
 }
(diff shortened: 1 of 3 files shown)


internal/devconfig/config.go
```diff
@@ -346,9 +347,18 @@ func (c *Config) NixPkgsCommitHash() string {
...
+		// /plugin2/bin/:/plugin1/bin/:$PATH
+		pluginEnv := mergePATHsFromTwoEnvs(env, i.Env())
+		maps.Copy(env, pluginEnv)
 	}
-	maps.Copy(env, c.Root.Env)
+	// Here we merge the processed PATH from all plugins to PATH in config env.
+	// So if config env has /config/env/:$PATH merging with the plugin PATH
+	// it becomes /config/env/:/plugin2/bin/:/plugin1/bin/:$PATH
+	rootEnv := mergePATHsFromTwoEnvs(env, c.Root.Env)
+	maps.Copy(env, rootEnv)
```
mikeland73, line 361: Can you use `OSExpandEnvMap` instead? 

Basically take the environment provided by the plugins and use it to expand the environment of the config.

internal/devconfig/config.go
```diff
@@ -346,9 +347,18 @@ func (c *Config) NixPkgsCommitHash() string {
 func (c *Config) Env() map[string]string {
 	env := map[string]string{}
 	for _, i := range c.included {
-		maps.Copy(env, i.Env())
+		// Here, in new pluginPath we replace the keyword "$PATH"
+		// with what we have for PATH so far. If so far we have /plugin1/bin:$PATH
+		// and new plugin adds /plugin2/bin/:$PATH the result becomes
+		// /plugin2/bin/:/plugin1/bin/:$PATH
+		pluginEnv := mergePATHsFromTwoEnvs(env, i.Env())
+		maps.Copy(env, pluginEnv)
```
mikeland73, line 355: Could you use `OSExpandEnvMap` instead of `mergePATHsFromTwoEnvs`?

mikeland73 (changes requested):

internal/devconfig/config.go
mohsenari: @mikeland73 after your suggestion, I tried it and it it doesn't work. The issue with `OSExpandEnvMap` is it merges the two PATHs correctly but it applies the (non-nix) system PATH to the `:$PATH` at the end. So with `OSExpandEnvMap` PATH becomes: `/my/config:/my/plugin:/opt/homebrew/sbin:/usr/local/bin:...`
with `mergePATHsFromTwoEnvs` PATH becomes: `/my/config:/my/plugin:/Users/mohsenansari/code/go.jetpack.io/devbox/examples/stacks/rails/.devbox/nix/profile/default/bin:...:/opt/homebrew/sbin:/usr/local/bin:...`

So OSExpandEnvMap can't work here because I don't want to expand $PATH env yet, I just want to merge them correctly without losing `:$PATH` suffix or `$PATH:` prefix to it.

mikeland73: Would something like 

```golang
func OSExpandIfPossible(env, existingEnv map[string]string) map[string]string {
	mapping := func(value string) string {
		// If the value is not set in existingEnv, return the value wrapped in ${...}
		if existingEnv == nil || existingEnv[value] == "" {
			return fmt.Sprintf("${%s}", value)
		}
		return existingEnv[value]
	}

	res := map[string]string{}
	for k, v := range env {
		res[k] = os.Expand(v, mapping)
	}
	return res
}
```

work?

mikeland73: And some unit tests if you decide to go this route:

```golang
package conf

import "testing"

func TestOSExpandIfPossible(t *testing.T) {
	tests := []struct {
		name        string
		env         map[string]string
		existingEnv map[string]string
		want        map[string]string
	}{
		{
			name: "basic expansion",
			env: map[string]string{
				"FOO": "$BAR",
				"BAZ": "${QUX}",
			},
			existingEnv: map[string]string{
				"BAR": "bar_value",
				"QUX": "qux_value",
			},
			want: map[string]string{
				"FOO": "bar_value",
				"BAZ": "qux_value",
			},
		},
		{
			name: "missing values remain as template",
			env: map[string]string{
				"FOO": "$BAR",
				"BAZ": "${QUX}",
			},
			existingEnv: map[string]string{
				"BAR": "bar_value",
				// QUX is missing
			},
			want: map[string]string{
				"FOO": "bar_value",
				"BAZ": "${QUX}",
			},
		},
		{
			name: "nil existing env",
			env: map[string]string{
				"FOO": "$BAR",
				"BAZ": "${QUX}",
			},
			existingEnv: nil,
			want: map[string]string{
				"FOO": "${BAR}",
				"BAZ": "${QUX}",
			},
		},
		{
			name: "empty existing env",
			env: map[string]string{
				"FOO": "$BAR",
			},
			existingEnv: map[string]string{},
			want: map[string]string{
				"FOO": "${BAR}",
			},
		},
		{
			name: "mixed literal and variable",
			env: map[string]string{
				"FOO": "prefix_${BAR}_suffix",
			},
			existingEnv: map[string]string{
				"BAR": "bar_value",
			},
			want: map[string]string{
				"FOO": "prefix_bar_value_suffix",
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got := OSExpandIfPossible(tt.env, tt.existingEnv)
			if len(got) != len(tt.want) {
				t.Errorf("OSExpandIfPossible() got %v entries, want %v entries", len(got), len(tt.want))
			}
			for k, v := range tt.want {
				if got[k] != v {
					t.Errorf("OSExpandIfPossible() for key %q = %q, want %q", k, got[k], v)
				}
			}
		})
	}
}
```

mikeland73 (commented):
I don't think we should treat $PATH different from other env vars. Also, a minor nit is that you are not really "merging" paths, you are expanding, but only if $PATH already exists in the new environment you are creating. 

I think what you want is something like `OSExpandIfPossible` (I added a possible implementation).

Also, for stuff like this, adding some unit tests and/or some integration tests in example dir is really useful, since reasoning about what the final environment will look like is hard. See https://github.com/jetify-com/devbox/blob/163ae999a46e6ff4b946437b22a782557e708eeb/examples/plugins/v2-local/devbox.json#L1C1-L36C1 for an example of an integration test for plugins.

mikeland73 (changes requested):
RC to send back to your queue

internal/devconfig/config.go
mohsenari: `OSExpandIfPossible` works. I included this suggestion along with the unit test and added a case for path to the unit test as well.

mikeland73 (approved):

PR #2435, opened 2024-12-04
feat(cli): improve search result view for large version lists

irdaislakhuafa:
## Summary
This is a cool package, i was looking for something like this and just found it. 
As we know when searching for a package the results will be like this
image

but the results will be a bit difficult to read when there are many versions are available like this.
image

I suggest it would be easier to read the available versions if it were like this
image

or if you have a more readable option I'm open to that. Thanks for making such a cool package.


gcurtis (approved):

gcurtis:
LGTM. Thanks!

PR #2437, opened 2024-12-05
devbox: export API for devbox install

gcurtis:
Add a top-level package that makes a subset of the `internal/devbox` functionality (currently just install) public.

There are some places where we want to run devbox commands (such as in the prefetcher) which requires us to run the CLI externally. Exporting a package means we can run the command implementation directly and don't need to install the CLI (similar to what we're doing with autodetect).

### DIFF
--- a/devbox.go
+++ b/devbox.go
@@ -1,0 +1,32 @@
+// Package devbox creates and configures Devbox development environments.
+package devbox
+
+import (
+	"context"
+	"io"
+
+	"go.jetpack.io/devbox/internal/devbox"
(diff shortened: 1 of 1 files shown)


savil (approved):

mikeland73 (approved):
For future thought: should we move devbox package out of `internal` and into `pkg`?

gcurtis:
I don't think I would move the whole package because there's a lot of cruft in there. If we do it piecemeal, we can keep the API clean and only expose the useful parts.

PR #2447, opened 2024-12-11
nix: make DetSys installer the default

gcurtis:
Remove the feature flag and make the DetSys installer the default.

- Download the installer directly in case the system doesn't have curl.
- Expose the auto-install functionality in the public nix package so we can reuse it elsewhere.
- Remove some checks that were only relevant to the old installer (WSL2 limitations and warnings about root).

The install logic is also now part of an `Installer` struct to provide some flexibility around what install program is run and when it's downloaded.

	type Installer struct{ ... }
		func (i *Installer) Download(ctx context.Context) error
		func (i *Installer) Run(ctx context.Context) error

### DIFF
--- a/internal/nix/install.go
+++ b/internal/nix/install.go
@@ -4,97 +4,22 @@
 package nix
 
 import (
-	"bytes"
+	"context"
 	_ "embed"
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
-	"strings"
+	"time"
 
+	"github.com/briandowns/spinner"
 	"github.com/fatih/color"
 	"github.com/mattn/go-isatty"
-	"github.com/pkg/errors"
 
-	"go.jetpack.io/devbox/internal/boxcli/featureflag"
 	"go.jetpack.io/devbox/internal/boxcli/usererr"
-	"go.jetpack.io/devbox/internal/build"
 	"go.jetpack.io/devbox/internal/cmdutil"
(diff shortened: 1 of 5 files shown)


internal/nix/install.go
```diff
@@ -154,31 +75,35 @@ func EnsureNixInstalled(writer io.Writer, withDaemonFunc func() *bool) (err erro
...
 	if isatty.IsTerminal(os.Stdout.Fd()) {
 		color.Yellow("Press enter to continue or ctrl-c to exit.\n")
 		fmt.Scanln() //nolint:errcheck
-	}
 
-	if err = Install(writer, withDaemonFunc); err != nil {
-		return err
-	}
+		spinny := spinner.New(spinner.CharSets[11], 100*time.Millisecond, spinner.WithWriter(writer))
+		spinny.Suffix = " Downloading the Nix installer..."
```
savil, line 84: nit: drop the prefix space?

savil (approved):
yay!

internal/nix/install.go
gcurtis: I think this puts a space between the spinner and the message. I copied what we have for [`Computing the Devbox environment`](https://github.com/jetify-com/devbox/blob/d2957086a837cd7fe93ce8fe8556e446835d5f0a/internal/devbox/devbox.go#L592).

PR #2449, opened 2024-12-12
devbox: clean up profile history after sync

gcurtis:
After syncing the flake packages to the profile, remove all old generations of the Nix profile. This allows the Nix garbage collector to eventually remove any old packages.

Opted for a Go implementation instead of calling `nix profile wipe-history` because deleting the history is pretty simple and this is faster.

### DIFF
--- a/internal/devbox/nixprofile.go
+++ b/internal/devbox/nixprofile.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/samber/lo"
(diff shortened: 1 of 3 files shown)


mohsenari (approved):
Let's write a couple of unit tests for this as well to make sure regression is caught in the future.

savil (approved):
I feel mixed about re-implementing a nix command, but perf is somewhat persuasive :)

savil:
>  to make sure regression is caught in the future.

@mohsenari I don't follow. What regression are you referring to?

mohsenari:
> >  to make sure regression is caught in the future.
> 
> @mohsenari I don't follow. What regression are you referring to?

@savil I'm talking about potential changes to this function in the future, causing regression

gcurtis:
@mohsenari I added an assertion to a testscripts test to ensure that the history is pruned after adding/removing packages.

PR #2464, opened 2024-12-20
flake: add narHash and lastModified attributes

gcurtis:
Add support for the `narHash` and `lastModified` attributes of flakerefs. The new `Ref.Locked` method uses these attributes to report whether a flake is locked.

### DIFF
--- a/nix/flake/flakeref.go
+++ b/nix/flake/flakeref.go
@@ -2,7 +2,7 @@
 package flake
 
 import (
-	"maps"
+	"cmp"
 	"net/url"
 	"path"
 	"slices"
@@ -167,8 +167,9 @@ func parseURLRef(ref string) (parsed Ref, fragment string, err error) {
 			parsed.Path = refURL.Path
 		}
 
-		parsed.NARHash = refURL.Query().Get("narHash")
-		parsed.LastModified, err = atoiOmitZero(refURL.Query().Get("lastModified"))
+		query := refURL.Query()
+		parsed.NARHash = query.Get("narHash")
+		parsed.LastModified, err = atoiOmitZero(query.Get("lastModified"))
 		if err != nil {
 			return Ref{}, "", redact.Errorf("parse flake reference URL query parameter: lastModified=%s: %v", redact.Safe(parsed.LastModified), redact.Safe(err))
 		}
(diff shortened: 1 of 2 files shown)


nix/flake/flakeref.go
```diff
@@ -156,24 +166,64 @@ func parseURLRef(ref string) (parsed Ref, fragment string, err error) {
...
+		parsed.NARHash = refURL.Query().Get("narHash")
+		parsed.LastModified, err = atoiOmitZero(refURL.Query().Get("lastModified"))
+		if err != nil {
+			return Ref{}, "", redact.Errorf("parse flake reference URL query parameter: lastModified=%s: %v", redact.Safe(parsed.LastModified), redact.Safe(err))
+		}
 
+		// lastModified and narHash get stripped from the query
+		// parameters, but dir stays.
+		query.Del("lastModified")
+		query.Del("narHash")
```
savil, line 208: seems like these lines are repeated in many case-bodies. Pull into a function and call that function instead?

nix/flake/flakeref.go
```diff
@@ -262,15 +345,26 @@ func parseGitHubRef(refURL *url.URL, parsed *Ref) error {
...
+
+		url, err := url.Parse("file+" + r.URL)
+		if err != nil {
+			// This should be rare and only happen if the caller
+			// messed with the parsed URL.
+			return ""
+		}
+		url.RawQuery = buildQueryString(url.Query(),
+			"lastModified", itoaOmitZero(r.LastModified),
+			"narHash", r.NARHash,
```
savil, line 365: what happens if lastModified or NarHash are zero?

nix/flake/flakeref.go
```diff
@@ -292,26 +386,35 @@ func (r Ref) String() string {
...
 		if r.Owner == "" || r.Repo == "" {
 			return ""
 		}
 		url := &url.URL{
-			Scheme:   "github",
-			Opaque:   buildEscapedPath(r.Owner, r.Repo, r.Rev, r.Ref),
-			RawQuery: buildQueryString("host", r.Host, "dir", r.Dir),
+			Scheme: "github",
+			Opaque: buildEscapedPath(r.Owner, r.Repo, r.Rev, r.Ref),
+			RawQuery: buildQueryString(nil,
```
savil, line 398: what is `nil` for?

savil (approved):

nix/flake/flakeref.go
gcurtis: `buildQueryString` will omit them.

nix/flake/flakeref.go
gcurtis: It's the initial `url.Values` query. In this case it's nil because we're building the query string from scratch.

nix/flake/flakeref.go
gcurtis: Are you referring to the `query.Del` calls? I dunno, it seems like overkill to make those two lines a separate function.

PR #2468, opened 2024-12-21
MySQL Plugin: create and link my.cnf to allow configuring MySQL

clessg:
## Summary

Adds a `$MYSQL_CONF` env var to the MySQL Plugin to allow a custom `my.cnf` (MySQL config file). MySQL does not provide an environment variable, command line flag, or anything else to change the location (it's always computed as `$MYSQL_HOME/my.cnf`), so we simply create a symlink there pointing to the user's config.

## How was it tested?

- Forked devbox, made changes to the mysql plugin, and ensured everything worked in a custom project.


Lagoja:
Thanks, this is an awesome addition! 

I added a similar feature for MariaDB, and also tweaked the detection logic for MySQL vs. MariaDB

clessg:
Hey John, thank you, looks perfect to me!

Lagoja (approved):

PR #2470, opened 2024-12-22
Feature: Update to JSON Plugin Schema to Allow Private Sources

Bee-Mar:
## Summary

The pull request introduces the ability to add plugins from private sources, which includes GitHub, GitLab, Bitbucket and arbitrary hosts. The JSON schema was reworked to be more declarative, which forces the user declare components of the addressed location. Changing the structure allows for future flexibility, and was needed in my case when working with repositories only accessible through an AWS tunnel.

In short, the plugin declaration syntax goes from something like this:

```json
  "include": [
    "github:<org>/<repo>?dir=<plugin-dir>"
  ]
```

to

```json
"include": [
  {
    "type": "github",
    "owner": "org",
    "repo": "repository",
    "dir": "subdir"
  }
]
```

An example of privately hosted plugins accessible through SSH:

```json
"include": [
  {
    "type": "ssh",
    "host": "localhost":
    "port": 1234,
    "owner": "org",
    "repo": "repository",
    "dir": "subdir"
  }
]
```

## How was it tested?

Correcting broken unit tests is still a work in progress, however this has been tested in by creating several repos to use as a test-bed, and ensuring they can be installed by declaring them in the `devbox.json`. Additionally, I have used this for my desired purpose, which is to access Devbox plugins only accessible in a privately hosted GitLab repo behind an AWS tunnel. 

Addtional unit tests can be added, however, this pull request is currently in draft so reviewers can begin looking through the changes and identify where they would like to see changes/improvements.


Bee-Mar:
Does anyone have interest in reviewing this pull request?

Bee-Mar:
Following up on this again. Is anyone interested in this feature?

Bee-Mar:
Is there anyone at all willing or interested in reviewing this PR? There are several people I've spoken with that would love to see this feature added to Devbox. I'm sure there are issues that need to be addressed within the PR, and I'm more than happy to make any suggested change. 

@savil  @mikeland73

Issue #1991, opened 2024-04-12
The order of appearance of scripts in generated readme should have same order as defined in devbox.json


ametad:
### What problem are you trying to solve?

The command `devbox generate readme` ignores the order of the scripts and re-orders it in the readme file to alphabetical order. I use the order explicitly for the humans who read it, so they have a clue what to do first. 

### Example of the problem

**Scripts**

The `devbox.json` with the scripts in a particular order: 

```json
{
  "shell": {
    "init_hook": [],
    "scripts": {
      "step-one": [
        "echo \"step-one\""
      ],
      "step-two": [
        "echo \"step-two\""
      ],
      "step-three": [
        "echo \"step-three\""
      ],
      "step-four": [
        "echo \"step-four\""
      ]
    }
  }
}
```

**Generated README.md**

This is the generated readme file created with `devbox generate readme`:

## Scripts
Scripts are custom commands that can be run using this project's environment. This project has the following scripts:

* [step-four](#devbox-run-step-four)
* [step-one](#devbox-run-step-one)
* [step-three](#devbox-run-step-three)
* [step-two](#devbox-run-step-two)

(etc.)

### What solution would you like?

The generated Markdown in the `README.md` file should have the same order as in the `devbox.json` file, like so:

## Scripts
Scripts are custom commands that can be run using this project's environment. This project has the following scripts:

* [step-one](#devbox-run-step-one)
* [step-two](#devbox-run-step-two)
* [step-three](#devbox-run-step-three)
* [step-four](#devbox-run-step-four)

(etc.)

### Alternatives you've considered

For now I prepend the script name with a number, but this is cumbersome for maintenance. Because when an extra step is requirement somewhere in between, the next script names also has to change. This is not great...

### Example of alternative

```json
{
  "shell": {
    "init_hook": [],
    "scripts": {
      "01-step-one": [
        "echo \"step-one\""
      ],
      "02-step-two": [
        "echo \"step-two\""
      ],
      "03-step-three": [
        "echo \"step-three\""
      ],
      "04-step-four": [
        "echo \"step-four\""
      ]
    }
  }
}
```

**Generated README.md**

## Scripts
Scripts are custom commands that can be run using this project's environment. This project has the following scripts:

* [01-step-one](#devbox-run-01-step-one)
* [02-step-two](#devbox-run-02-step-two)
* [03-step-three](#devbox-run-03-step-three)
* [04-step-four](#devbox-run-04-step-four)

(etc.)

[...]

PR #2477, opened 2025-01-07
fix: sort scripts list

### LINKED ISSUE #1631: [Bug]: devbox run returns scripts in random order
**Current Behavior (bug)**
Running `devbox run` returns the scripts list in a random sort order. It should return the list in the order of devbox.json. Its especially problematic if the list of scripts is very long. I am currently working around it by using `devbox run | grep '*' | sort` 

**Expected Behavior (fix)**
`devbox run` should return the same sort order as devbox.json

**Additional context**
devbox version: 0.8.1

guerinoni:
This allows to have a consistent output from `devbox run`.

Closes #1631 and #1991.

## How was it tested?

I tested running multiple times `devbox run` in different project where I use it.

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -345,6 +345,9 @@ func (d *Devbox) ListScripts() []string {
 		keys[i] = k
 		i++
 	}
+
+	slices.Sort(keys)
+
 	return keys
 }
 


mohsenari:
@mikeland73 @Lagoja PTAL. This is an improvement imo. Ideally, we list the scripts with the same order as defined in devbox.json but if that's not possible, next best thing is alphabetical sort.

mohsenari (approved):

PR #2478, opened 2025-01-10
fix(python plugin): set UV_PYTHON to python in .venv

### LINKED ISSUE #2411: Error installing python packages using `uv`
### What happened?

When settings up devbox shell, setting up uv as a package to use as an installer, and then trying to use it to install numpy (or any other package) it complains with

```
error: Failed to install: numpy-2.1.3-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (numpy==2.1.3)
  Caused by: failed to create directory `/nix/store/m85llqyd79hjm4cbgx8hds2rarbslc1s-python3-3.11.2/lib/python3.11/site-packages/numpy`
  Caused by: Permission denied (os error 13)
  ```

### Steps to reproduce

1. Setup a python devbox with uv
2. `devbox shell`
3. `uv pip install numpy`

### Command

_No response_

### devbox.json

```jsonc
{
    "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.13.6/.schema/devbox.schema.json",
      "packages": [
        "python@3.11.2",
        "uv@latest",
        "git@latest"
      ],

    "shell": {
    "init_hook": [
      "source $VENV_DIR/bin/activate",
    ],
      "scripts": {
        "test": [
        ]
      }
    }
  }
```

### Devbox version

0.13.6

### Nix version

nix (Nix) 2.24.7

Linux (x86-64)

### Debug logs

_No response_

yemaney:
Fixes #2411

## Summary
The UV_PYTHON env variables sets which python executable uv will use. 

The current plugin sets the UV_PYTHON value  to `.devbox/nix/profile/default/bin/python` which means `uv pip install ` commands will try to install and save libraries under the  `.devbox/nix/profile/default/lib/python3.x/site-packages` directory. 

This will cause an error as the directory doesn't have write permissions because it exists under a folder that is a symlinked to a folder under the `/nix/store` path.

This change makes sure UV_PYTHON points to the python executable under .venv. 
This will make `uv pip install` commands install libraries to the site-packages directory  under .venv just like normal `pip install ` would.

## How was it tested?

1. devbox create uvtest --template python-pip
2. cd uvtest
3. devbox add uv
4. uv pip install fastapi

### DIFF
--- a/plugins/python.json
+++ b/plugins/python.json
@@ -4,7 +4,7 @@
   "description": "Python in Devbox works best when used with a virtual environment (venv, virtualenv, etc.). Devbox will automatically create a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `. $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
   "env": {
     "VENV_DIR": "{{ .DevboxProjectDir }}/.venv",
-    "UV_PYTHON": "{{ .DevboxProjectDir }}/.devbox/nix/profile/default/bin/python"
+    "UV_PYTHON": "{{ .DevboxProjectDir }}/.venv/bin/python"
   },
   "create_files": {
     "{{ .Virtenv }}/bin/venvShellHook.sh": "python/venvShellHook.sh"


mdomke:
It would be awesome to get this merged! All my Python projects suffer from this problem.

pinage404:
Maybe related to https://github.com/jetify-com/devbox/issues/2428

Lagoja (approved):
Thanks! I verified that this works and fixes the issue as expected. Appreciate the contribution!

PR #2486, opened 2025-01-21
Optimize Docker Image Size by Cleaning Nix Store

artemklevtsov:
## Summary

This MR modifies the Dockerfile to reduce the final image size by optimizing the Nix store after installing packages. Added `nix-store --gc` and `nix-store --optimise` commands to clean up and optimize the Nix store after package installation.

## How was it tested?

Built the Docker image locally and verified the image size reduction. Confirmed that the application runs as expected after the changes.


mohsenari (approved):

mohsenari:
Thank you for making this PR. Would you mind adding how much you were able to reduce the image size in your testing?

PR #2495, opened 2025-01-30
fix: pass a ctx to client searcher `search`

guerinoni:
## Summary

This allows to cancel the search request.

## How was it tested?

Mostly manual testing


savil (approved):
Thanks! Will merge once CICD passes

PR #2497, opened 2025-01-31
feat: implement outdated command

guerinoni:
## Summary
This allow to present all new versions available for the pkg installed. It is a lot useful because otherwise you need to check this manually.

## How was it tested?

In a different repository, this is the output.

Screenshot 2025-01-31 at 17 54 25


guerinoni:
For sure it is something we can move under another menu, no problem!

Lagoja:
I think this might be best to move under `devbox list`? I think something like `devbox list --outdated` probably fits well with the current top level commands.

guerinoni:
@savil @Lagoja updated code to be under `list --outdated`

internal/boxcli/list.go
```diff
@@ -57,6 +62,28 @@ func listCmd() *cobra.Command {
...
 		},
 	}
+
+	cmd.Flags().BoolVar(&flags.outdated, "outdated", false, "List outdated packages")
 	flags.config.register(cmd)
 	return cmd
 }
+
+// outdatedPackage prints a list of outdated packages.
+func outdatedPackage(cmd *cobra.Command, box *devbox.Devbox) error {
```
savil, line 72: nit: can we rename this to `printOutdatedPackages`?

internal/boxcli/list.go
```diff
@@ -57,6 +62,28 @@ func listCmd() *cobra.Command {
...
+	results, err := box.Outdated(cmd.Context())
```
savil, line 73: nit: can we rename the api to `box.OutdatedPackages`?

this could be more clear as `nameToUpdateVersion` but don't feel strongly

internal/devbox/packages.go
```diff
@@ -43,6 +44,48 @@ const StateOutOfDateMessage = "Your devbox environment may be out of date. Run %
...
+func (d *Devbox) Outdated(ctx context.Context) (map[string]UpdateVersion, error) {
+	ctx, task := trace.NewTask(ctx, "devboxOutdated")
+	defer task.End()
+
+	outdatedPackages := map[string]UpdateVersion{}
+
+	for _, pkg := range d.AllPackages() {
+		if strings.HasSuffix(pkg.Versioned(), "latest") {
+			continue
+		}
```
savil, line 62: `@latest` packages have their versions pinned in the devbox.lock file. I think you could get their current version via:

```
lockfile := d.Lockfile()
lockfile.Packages[p.Raw].Version
```

I think its useful to print if they're outdated as well. A user may want to know that.

internal/devbox/packages.go
```diff
@@ -43,6 +44,48 @@ const StateOutOfDateMessage = "Your devbox environment may be out of date. Run %
...
+
+		result, err := searcher.Client().Search(ctx, pkg.CanonicalName())
+		if err != nil {
+			return nil, err
+		}
```
savil, line 67: @gcurtis does this reliably give us the latest version of the package?

For example, when I run it with `go` I get:
```
 devbox ls --outdated
The following packages can be updated:
 * go                             1.23.4 -> 1.24rc2
```

which doesn't seem to be what we want

internal/devbox/packages.go
```diff
@@ -43,6 +44,48 @@ const StateOutOfDateMessage = "Your devbox environment may be out of date. Run %
...
+					if v.Version > vv {
+						outdatedPackages[p.Name] = UpdateVersion{Current: vv, Latest: v.Version}
+						break
+					}
+				}
+			}
+		}
+	}
+
+	return outdatedPackages, nil
```
savil, line 86: @guerinoni I think we need this function to be rewritten as so. I created this patch via `git diff --patch`, so you could try applying it: https://gist.github.com/savil/7a55e959c5eb5c5e240f57dedd5fd7ad

copying here for clarity:
```
type UpdateVersion struct {
	Current string
	Latest  string
}

// Outdated returns a map of package names to their available latest version.
func (d *Devbox) Outdated(ctx context.Context) (map[string]UpdateVersion, error) {
	ctx, task := trace.NewTask(ctx, "devboxOutdated")
	defer task.End()

	lockfile := d.Lockfile()
	outdatedPackages := map[string]UpdateVersion{}

	for _, pkg := range d.AllPackages() {
		// For non-devbox packages, like flakes or runx, we can skip for now
		if !pkg.IsDevboxPackage {
			continue
		}

		lockPackage, err := lockfile.FetchResolvedPackage(pkg.Versioned())
		if err != nil {
			return nil, errors.Wrap(err, "failed to fetch resolved package")
		}
		existingLockPackage := lockfile.Packages[pkg.Raw]
		if lockPackage.Version == existingLockPackage.Version {
			continue
		}

		outdatedPackages[pkg.Versioned()] = UpdateVersion{Current: existingLockPackage.Version, Latest: lockPackage.Version}
	}

	return outdatedPackages, nil
}
```

wdyt @gcurtis @mikeland73 ?

The benefit of this is that it also works for `@latest`. For example, when I run this on the devbox repo itself, I see:
```
$ devbox run build
$  dist/devbox ls --outdated
The following packages can be updated:
 * go@latest                      1.23.0 -> 1.23.4
 * runx:golangci/golangci-lint@latest v1.60.2 -> v1.63.4
```
I don't think the `runx` one should be getting printed, but its cause we return true for `IsDevboxPackage` for runx packages!

internal/devbox/packages.go
guerinoni: I guess this is correct, is the latest version available in the store :)

internal/devbox/packages.go
guerinoni: I aligned with this patch :)

internal/devbox/packages.go
```diff
@@ -43,6 +43,40 @@ const StateOutOfDateMessage = "Your devbox environment may be out of date. Run %
...
+	Latest  string
+}
+
+// Outdated returns a map of package names to their available latest version.
+func (d *Devbox) Outdated(ctx context.Context) (map[string]UpdateVersion, error) {
+	ctx, task := trace.NewTask(ctx, "devboxOutdated")
+	defer task.End()
+
+	lockfile := d.Lockfile()
+	outdatedPackages := map[string]UpdateVersion{}
```
mikeland73, line 57: nit, would prefer to use a slice over map. We don't need random access or deduplication.

mikeland73: @gcurtis may not like this but we could just return `[]lo.Tuple2[*lockfile.Package, *lockfile.Package]`

that way we don't have to create a new struct at all.

internal/devbox/packages.go
mikeland73: Why not runx? The versions are correct. Ideally we could also do it for flakes as well but our current flake update mechanism is a bit broken.

mikeland73 (approved):
minor nits, but this LGTM

internal/devbox/packages.go
savil: oh good point, I forgot runx updating works properly. So all good

internal/devbox/packages.go
savil: hm I don't mind the struct. Its nice to have the names

internal/devbox/packages.go
savil: But its a release-candidate. In `devbox update` we wouldn't actually update you to that version if you had `go@latest`

internal/devbox/packages.go
guerinoni: How you prefer guys, here map doesn't create any performance problems, it is just a matter of some seconds to run the command, I chose readability over little optimization

PR #2499, opened 2025-01-31
[deps] Update x/crypto dependency

mikeland73:
## Summary

fixes https://github.com/jetify-com/devbox/security/dependabot/197


gcurtis (approved):
I think this just needs `devbox run update-hash`.

mikeland73:
> I think this just needs `devbox run update-hash`.

thanks. fixed it.

Feature request:

github action to autofix issues like this.

PR #2528, opened 2025-02-19
fix: outdated cmd wasn't check for real new versions

guerinoni:
## Summary

Previously this only checks on lockfile.

## How was it tested?

Manually on a project I'm working on, and I figured out it was not working correctly.

### DIFF
--- a/internal/devbox/packages.go
+++ b/internal/devbox/packages.go
@@ -13,11 +13,15 @@ import (
 	"path/filepath"
 	"runtime/trace"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/samber/lo"
+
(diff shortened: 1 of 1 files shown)


internal/devbox/packages.go
```diff
@@ -80,6 +90,32 @@ func (d *Devbox) Outdated(ctx context.Context) (map[string]UpdateVersion, error)
 	return outdatedPackages, nil
 }
 
+// isGreater returns true if v1 is greater than v2
+func isGreater(v1, v2 string) bool {
```
mikeland73, line 94: I wish it was this simple :(

Unfortunately, nix package versions don't follow any convention. Some (many) use semver, but many do not.

We have some pretty complicated logic to try to do this on the search server. I can look into open sourcing it so you can use it here.

guerinoni: Let me know, for now this can be enough for doing this dirty job :)

Issue #2560, opened 2025-03-17
Have nixpkgs use registry version


tlelson:
I set my nix registry to the version used to build the system so that anytime i build a flake it doesn't download a new tar ball.

However when I `devbox init` i can see it does not follow the system nixpkgs and fetches an unknown hash. In fact it seems to get a couple of version of nixpkgs.

Here is my registry:

```sh
› nix registry list | grep nixpkgs
system flake:nixpkgs path:/nix/store/vawkv67jxh8kl4flrqgpcsmn9inqgvjv-source?lastModified=1741862977&narHash=sha256-prZ0M8vE/ghRGGZcflvxCu40ObKaB%2Bikn74/xQoNrGQ%3D&rev=cdd2ef009676ac92b715ff26630164bb88fec4e0
global flake:nixpkgs github:NixOS/nixpkgs/nixpkgs-unstable
```

Here is the output of a recent devbox init:

```sh
› devbox add python terraform tflint tflint-plugins.tflint-ruleset-google google-cloud-sdk
Ensuring nixpkgs registry is downloaded.
Downloaded 'github:NixOS/nixpkgs/0d534853a55b5d02a4ababa1d71921ce8f0aee4c?narHash=sha256-IDtwLhWMKvydW9GrB6ETwhij%2Bvma0hYHln8pR/rTd%2BU%3D' to '/nix/store/8r2xs1kc
cx7kk1kmgi24kivfjxx67gl7-source' (hash 'sha256-IDtwLhWMKvydW9GrB6ETwhij+vma0hYHln8pR/rTd+U=').
Ensuring nixpkgs registry is downloaded: Success
Info: Adding package "python@latest" to devbox.json
Ensuring nixpkgs registry is downloaded.
Downloaded 'github:NixOS/nixpkgs/573c650e8a14b2faa0041645ab18aed7e60f0c9a?narHash=sha256-4thdbnP6dlbdq%2BqZWTsm4ffAwoS8Tiq1YResB%2BRP6WE%3D' to '/nix/store/0yw40qnr
ar3lvc5hax5n49abl57apjbn-source' (hash 'sha256-4thdbnP6dlbdq+qZWTsm4ffAwoS8Tiq1YResB+RP6WE=').
Ensuring nixpkgs registry is downloaded: Success
Info: Adding package "terraform@latest" to devbox.json
Info: Adding package "tflint@latest" to devbox.json
[...]

PR #2576, opened 2025-04-08
[perf] Cache nixpkgs resolution

mikeland73:
## Summary

This caches resolutions for unlocked nixpkgs (the default) for 90 days on the user's machine.  Otherwise, everytime devbox tries to resolve `github:NixOS/nixpkgs/nixpkgs-unstable` it will download a new version of nixpkgs. This is slow (40+ seconds) and usually nothing changes.

### If a user wants to update nixpkgs there are two ways of doing it:

* `devbox update` without any arguments will update nixpkgs. This is existing functionality.
* `devbox update nixpkgs [...pkgs]` will also update nixpkgs.

I'm not convinced updating nixpkgs when using `update` without arguments is best, but I didn't want to break past functionality. It adds 40+ seconds to update operations when there is no package to update.

An alternative command syntax considered was `devbox update --nixpkgs`. The benefit of doing this would be to avoid conflicts in the future. 

I did not use `stdenv` because it already exists and it is a different package.

### Bug fixes:

* Updating a specific package (e.g. `devbox update go`) no longer updates nixpkgs

### Alternative approaches:

* We could try to search the local nix database to try to find existing nixpkgs, but this seemed more trouble than it was worth.

## How was it tested?

* Created a new project and logged the cache being used.
* Replaced the cache file data with an old version, created a new project and observed the old version used.
* Tried several permutations of `devbox update` with and without packages

the

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -2,8 +2,8 @@
   "lockfile_version": "1",
   "packages": {
     "fd@latest": {
-      "last_modified": "2025-02-07T11:26:36Z",
-      "resolved": "github:NixOS/nixpkgs/d98abf5cf5914e5e4e9d57205e3af55ca90ffc1d#fd",
+      "last_modified": "2025-03-11T17:52:14Z",
+      "resolved": "github:NixOS/nixpkgs/0d534853a55b5d02a4ababa1d71921ce8f0aee4c#fd",
       "source": "devbox-search",
       "version": "10.2.0",
       "systems": {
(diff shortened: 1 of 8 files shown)


gcurtis (approved):
Could we shorten the cache time to something less? Maybe about a week?

There are times where an outdated stdenv breaks packages. This recently happened on macOS where the stdenv was overhauled, leading to weird runtime errors about git or frameworks not being found.

gcurtis:
> This issue is slightly related: #2560
> 
> Is it possible to pin nixpkgs to a specific version ?

@Allsimon you can try setting the top-level `{ "nixpkgs": { "commit": "<hash>" } }` in devbox.json or manually editing devbox.lock. I know this is a bit hacky though. Devbox generally uses a search index to map package versions (e.g., `go@1.19`) to nixpkgs commit hashes, so pinning a specific hash is a little tricky.

tlelson:
@gcurtis I'm interested in your suggestion but am not quite sure how best to use it in my context.

I have been moving my team to devbox which has been great for reproducibility but bad for CICD build times.

For instance.  I have a simple job that uses 3 packages (python3.12, pip, zip) to download some packages and zip them up.

When I run this using the `devbox run` command in gitlab it spends 3.5 minutes just downloading the packages from `cache.nixos.org`.

When I run it in a python3.12 container and `apt-get install zip` it take 30 seconds for the whole job.

What I would like to do, is pin the nixpkgs-unstable version to a specific hash so that I can prebuild a container and ensure that the nix store has the common packages in it.

Is this a viable strategy?  Is there a better way to improve slow jobs times?

P.S
There are no builds happening in this 3.5 minutes, just copies.

Issue #2575, opened 2025-04-08
UV Python project is broken: No interpreter found for path


mskonovalov:
### What happened?

Works just fine in Devbox 0.13.6
But broken in 0.14.0

```
Info: Ensuring packages are installed.
✓ Computed the Devbox environment.
error: No interpreter found for path `.../.venv/bin/python` in managed installations or system path
Error: error running script "it-tests" in Devbox: exit status 2
```

The issue is I have a venv in the subfolder, but it always tries to look in the root of the project :(

Will try to prepare the reproduction example.

Likely related to https://github.com/jetify-com/devbox/pull/2478
https://github.com/jetify-com/devbox/issues/2411

cc @yemaney 

### Steps to reproduce

1. Create a project with python venv inside subfolder
2. try to run any command via UV
3. see the error

### Command

run

### devbox.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.13.7/.schema/devbox.schema.json",
  "packages": [
    "python@3.10",
    "uv@0.4.30"
  ],
  "shell": {
    "init_hook": [
      "alias cdt='cd src/packages/toolkit'",
      ". $VENV_DIR/bin/activate"
    ],
    "scripts": {
      "unit-tests": ["cdt", "uv run pytest -m unit \"$@\""],
      "it-tests": ["cdt", "uv run pytest -m integration \"$@\""]
    }
  },
  "env": {
    "VENV_DIR": "./src/packages/toolkit/.venv",
  }
}
```

### Devbox version

0.14.0

### Nix version

2.27.1

### What system does this bug occur on?

macOS (Apple Silicon)

mskonovalov:
here is the example repo https://github.com/mskonovalov/devbox_bug
clone it and try to run `devbox run it-tests`
then you will see the following error
```
error: No interpreter found for path `/Users/kms/work/devbox_bug/.venv/bin/python` in managed installations or system path
Error: error running script "it-tests" in Devbox: exit status 2
```

Same result if you do 
```
devbox shell
uv sync
```

Riezebos:
Did you try adding `"UV_PYTHON": "$PWD/.venv/bin/python"` as fix mentioned in this comment on the linked issue? https://github.com/jetify-com/devbox/issues/2411#issuecomment-2544569867

mskonovalov:
I did, it does not help, still looking in the wrong place

yemaney:
[...]

PR #2608, opened 2025-05-05
fix(python plugin): remove UV_PYTHON  env variable

yemaney:
## Summary
Fixes: #2575

The current python plugin sets `UV_PYTHON` to path of python executable under default `VENV_DIR` directory. However if someone sets a custom `VENV_DIR` in their `devbox.json` file, this will cause an error. Since `uv` will attempt and fail to find python in the default VENV_DIR path where the venv would no longer exist.

This change removes the `UV_PYTHON` environment variable from the plugin altogether. This can be done, since `uv` will use the python executable of the active virtual environment its running in (https://docs.astral.sh/uv/reference/cli/#uv-python), and the venv is automatically created and activated by the python plugin.

This change allows flexibility for users to set their own `VENV_DIR`. It also still works with pip (`uv pip install`) as expected. 

## How was it tested?

1. devbox create uvtest --template python-pip
2. cd uvtest
3. mkdir -p src/uvtest
4. add `"VENV_DIR": "./src/uvtest/.venv"` to `env` section of `devbox.json` file
5. devbox shell
6. `uv run main.py`:  (this step would cause an error before these changes)
7. `uv pip install fastapi`

the

### DIFF
--- a/plugins/python.json
+++ b/plugins/python.json
@@ -3,8 +3,7 @@
   "version": "0.0.4",
   "description": "Python in Devbox works best when used with a virtual environment (venv, virtualenv, etc.). Devbox will automatically create a virtual environment using `venv` for python3 projects, so you can install packages with pip as normal.\nTo activate the environment, run `. $VENV_DIR/bin/activate` or add it to the init_hook of your devbox.json\nTo change where your virtual environment is created, modify the $VENV_DIR environment variable in your init_hook",
   "env": {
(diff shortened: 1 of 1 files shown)


mikeland73 (approved):
LGTM, but I'm less familiar with this plugin. @Lagoja PTAL

Lagoja (approved):

PR #2639, opened 2025-07-01
Evaluate environment variables in the local plugin path

### LINKED ISSUE #2638: Evaluate environment variables in local plugin path

My company is looking at adopting devbox as a package manager and orchestrator for running our services locally. I'd like to accomplish this by setting up a global devbox file that will be version controlled and then using plugins to define various services in the actual repos where each service lives.

Where I ran into a problem is that we don't guarantee the location of source code on each individual's development machine, so we can't set a relative or absolute path in the version controlled global file and guarantee it to work for everyone.

I wanted to solve this by defining an environment variable like $SRC_DIR that could be used in plugin paths in the global file like this:
```
"include": [
    "path:$SRC_DIR/app/devbox/plugin.json"
    "path:$SRC_DIR/otherapp/devbox/plugin.json"
]
```

The problem I ran into is that it looks like environment variables aren't currently evaluated as part of the local plugin path resolution so this results in devbox treating this as a relative path but including the env var verbatim as well. That results in an error like this:
```
Error: open /Users/user/.local/share/devbox/global/default/$SRC_DIR/app/devbox/plugin.json: no such file or directory
```

I think there's a quick fix for this so I'll also submit a PR to propose a solution. Thanks for considering!

Evaluate environment variables when assembling the path for local plugins.

_No response_

adamjreed:
## Summary

Fixes #2638

See issue for description of the problem. This PR is just a one-line change to evaluate current environment variables as devbox assembles the path for local plugins. This will allow things like a dynamic root directory for plugin source.

## How was it tested?

Built devbox locally with code change and copied it to /usr/local/bin/devbox. Refreshed global environment and plugins defined with an environment variable in the path now resolve properly.

the

### DIFF
--- a/internal/plugin/local.go
+++ b/internal/plugin/local.go
@@ -55,7 +55,7 @@ func (l *LocalPlugin) LockfileKey() string {
 }
 
 func (l *LocalPlugin) Path() string {
-	path := l.ref.Path
+	path := os.ExpandEnv(l.ref.Path)
 	if !strings.HasSuffix(path, pluginConfigName) {
 		path = filepath.Join(path, pluginConfigName)
 	}


Lagoja:
Looks pretty straightforward to me. @mikeland73 any pitfalls you see with this?

mikeland73:
This is good. Only caveat is that environment variables defined in `devbox.json` or passed via `-e` won't be available. That's probably OK?

Another approach could be to do the expansion before determining plugin type (i.e. local, remote, etc). That way we could do something like:

```
"includes": "${GOPLUGIN:-github:jetify-com/devbox-plugins?dir=go-monorepo}"
```

which would make it nicer to develop remote plugins (currently at Jetify we use this go plugin in a private repo and it's a bit annoying to test plugin updates)

Anyway let's not let perfect be enemy of good, will merge this. We can improve in follow ups.

Issue #2041, opened 2024-05-08
Prompt (starship) looks broken in zsh and in fish


niklasravnsborg:
### What happened?

<img width="1512" alt="Screenshot 2024-05-08 at 22 18 40" src="https://github.com/jetify-com/devbox/assets/6717303/4246831c-8f18-487e-8b5b-333347543093">

### Steps to reproduce

1. Install [Starship](https://starship.rs/) prompt (I installed it via [home-manager](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.starship.enable)).
2. Run `devbox init` in  a project
3. Run `devbox shell`

### Command

shell

### devbox.json

```jsonc
{
  "$schema":  "https://raw.githubusercontent.com/jetify-com/devbox/0.10.5/.schema/devbox.schema.json",
  "packages": ["bun@latest"],
  "shell": {
    "init_hook": [
      "echo 'Welcome to devbox!' > /dev/null"
    ],
    "scripts": {
      "test": [
        "echo \"Error: no test specified\" && exit 1"
      ]
    }
  }
}
```

### Devbox version

0.10.5

### Nix version

nix (Nix) 2.18.2

### What system does this bug occur on?

macOS (Apple Silicon)

bohrium2b:
This is likely caused by the devbox prompt. You can turn it off (allegedly; I haven't been able to make it work) by setting `DEVBOX_NO_PROMPT` to true. On fish, you can do it using `set -gx DEVBOX_NO_PROMPT true`.

MacroPower:
I was able to make it work with fish by using lowercase:

`set -gx devbox_no_prompt true`

maulik13:
I am in the same boat, not sure what I need to do in zsh to not have devbox as a prefix for my prompt which disrupts the starship prompt.

maulik13:
I managed to solve this by setting the environment variable `DEVBOX_NO_PROMPT=0` in my zshrc

And with the following starship custom element:

```toml
[custom.devbox]
description = "Shows an indicator when inside a Devbox shell"
format = "[  devbox]($style)"
style = "fg:gold bg:overlay"
when = "env | grep -E '^DEVBOX_SHELL_ENABLED='"
ignore_timeout = true
```

PR #2775, opened 2026-02-03
Fix for #2041 for fish

anka-213:
## Summary

Fix for the fish part of #2041.
Piping the prompt through echo breaks things like newlines. Just use printf instead.

## How was it tested?

Manually changing the prompt on my machine and seeing that it resolves the issue.

the


anka-213:
It seems like I had not tested it properly. With this version I get a newline after "(devbox) "

PR #2779, opened 2026-02-06
Bump minimum Nix version to 2.18

Lagoja:
## Summary

Nixpkgs now requires Nix >= 2.18 to evaluate, so drop support for older versions. This updates the CI test matrices and MinVersion constant.

## How was it tested?

Run in CI to confirm Nix Version checks are running correctly


.github/workflows/cli-tests.yaml
```diff
@@ -273,7 +273,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        nix-version: [2.15.1, 2.16.1, 2.17.0, 2.18.0, 2.19.2, 2.24.7]
+        nix-version: [2.18.0, 2.19.2, 2.24.7]
```
savil: Also add `"2.30.2"` here?

PR #2832, opened 2026-05-20
[services] Bump process-compose version and fix extra stderr output

mikeland73:
## Summary

- Bumps `processComposeVersion` from 1.87.0 to 1.110.0.
- Skips `Add` for utilities whose exact versioned name is already in the utility devbox config. Calling `Add` anyway printed noisy `Package "process-compose@..." already in devbox.json` output on every services interaction. A version mismatch (e.g. after bumping `processComposeVersion`) still falls through to `Add`, which replaces the existing package by canonical name.

nginx.conf and devbox.lock have minor automated changes.

## Test plan

- [x] `devbox services up` / `devbox services ls` no longer prints the "already in devbox.json" line.
- [x] Bumping `processComposeVersion` triggers a clean replace path (no stale 1.87.0 left behind).
- [x] Fresh install (no existing utility config) still adds process-compose correctly.

### DIFF
--- a/examples/servers/nginx/devbox.d/nginx/nginx.conf
+++ b/examples/servers/nginx/devbox.d/nginx/nginx.conf
@@ -1,8 +1,8 @@
 events {}
 http{
 server {
-         listen       8081;
-         listen       [::]:8081;
+         listen       8080;
+         listen       [::]:8080;
          server_name  localhost;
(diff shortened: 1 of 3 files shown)


Lagoja (approved):

mikeland73:
same test is failing in main. Will merge this and try to fix that.

PR #2837, opened 2026-05-31
Add use-devbox agent skill

mikeland73:
## What

Adds a Claude Code [agent skill](https://code.claude.com/docs) at `.claude/skills/use-devbox/SKILL.md` that teaches AI coding agents how to detect and use a devbox environment.

## Why

Agents working in a repo that uses devbox often don't realize that project tools (psql, python, gcloud, node, etc.) live *inside* the devbox environment rather than on the bare shell. As a result they conclude a tool is "unavailable" when devbox actually provides it. This skill gives an agent a clear, ordered playbook so it reaches for devbox at the right time.

## What the skill covers

- **When to reach for it** — a `devbox.json` exists, a tool seems missing, or the user mentions devbox.
- **Install check** — `command -v devbox`, with the official non-interactive installer.
- **Detection** — finding `devbox.json` in the cwd, an ancestor, or a subdir root.
- **Running commands** — preferring `devbox run [cmd]` for one-off, non-interactive use over `devbox shell`.
- **Project scripts** — running named scripts (`devbox run test`, `lint`, `build`, …) defined in `devbox.json`.
- **Binary source ordering** — devbox-provided > system > `devbox search`/`devbox add` (only with the user's confirmation before mutating `devbox.json`).
- **Services** and **global packages** for completeness.
- A **key facts** recap.

Install URL and example scripts in the skill match this repo's `README.md` / `devbox.md` / `devbox.json`.

https://claude.ai/code/session_01K1cHquYtVhkYErHrKgNUHW

---
_Generated by [Claude Code](https://claude.ai/code/session_01K1cHquYtVhkYErHrKgNUHW)_


Lagoja (approved):
This looks great, thanks for adding!

Issue #244, opened 2022-10-26
feature request: add support for setting the .devbox directory path via config


mohsenari:
Following up the thread in #226 
(created this for tracking)

Right now the `.devbox/` directory is located at the same level as devbox.json (project root).
The proposal is to make this configurable.
example:
```bash
devbox --config .devcontainer/devbox.json --cache /devbox/.devbox
```

PR #2838, opened 2026-05-31
Support setting the --config flag via the DEVBOX_CONFIG env var

### LINKED ISSUE #245: feature request: Set config and cache flags via env variables
following up the thread in #226 
(created for tracking)

be able to set `--config` and `--cache` flags using environment variable since it would allow users to set them in a Dockerfile:

```Dockerfile
ENV DEVBOX_CONFIG=.decontainer/.devbox.json
ENV DEVBOX_CACHE=/devbox/.devbox
```

mikeland73:
## Summary

Fixes #245.

The `--config` flag lets users point devbox at a `devbox.json` that lives in a non-default location. Until now there was no way to set it from the environment, which is awkward in contexts where passing the flag to every invocation is repetitive — the motivating example in the issue is a Dockerfile:

```dockerfile
ENV DEVBOX_CONFIG=.devcontainer
RUN devbox install   # picks up DEVBOX_CONFIG automatically
```

This PR defaults the `--config` flag to the `DEVBOX_CONFIG` environment variable when the flag isn't passed explicitly. An explicitly passed `--config`/`-c` flag still takes precedence over the env var.

## Changes

- Add a `DevboxConfig` (`DEVBOX_CONFIG`) constant in `internal/envir`.
- Use `os.Getenv(DEVBOX_CONFIG)` as the default value for the `--config` flag in both `register` and `registerPersistent`, and mention the env var in the flag usage text.
- Add `TestConfigFlagDefaultsToEnv` covering: unset env, env-as-default, and explicit-flag-overrides-env.

## Notes / scope

Issue #245 also mentions a `--cache` flag. devbox no longer exposes a `--cache` flag for the `.devbox` directory location (`DEVBOX_CACHE` today is used for a different purpose — the nixpkgs cache base URL), so this PR scopes the change to the `--config` flag, which is the part that's actionable against the current CLI. Happy to follow up if a configurable `.devbox` path is still desired (tracked separately in #244).

## Testing

```
go test ./internal/boxcli/ -run TestConfigFlagDefaultsToEnv
go vet ./internal/boxcli/ ./internal/envir/
```

https://claude.ai/code/session_01UQSFvFRfQ4Ku4F4LXbKjFn

---
_Generated by [Claude Code](https://claude.ai/code/session_01UQSFvFRfQ4Ku4F4LXbKjFn)_

### DIFF
--- a/internal/boxcli/config.go
+++ b/internal/boxcli/config.go
@@ -4,7 +4,11 @@
 package boxcli
 
 import (
+	"os"
+
 	"github.com/spf13/cobra"
+
+	"go.jetify.com/devbox/internal/envir"
 )
 
 // to be composed into xyzCmdFlags structs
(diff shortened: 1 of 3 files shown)


mohsenari (approved):

mohsenari:
Idk why cli tests fail for this PR, worth a retry

PR #2841, opened 2026-05-31
feat(config): deprecate nested `shell` object; add top-level `init_hook`/`scripts` and `devbox config fmt`

mikeland73:
## Summary

Deprecates the nested `shell` object in `devbox.json` in favor of top-level `init_hook` and `scripts` fields, and adds a `devbox config fmt` command that migrates existing configs in place.

### Config schema
- `init_hook` and `scripts` are now first-class top-level fields in `ConfigFile` and in `.schema/devbox.schema.json`.
- The nested `shell` object is still parsed for backward compatibility but is marked `deprecated` in the schema.
- `InitHook()` and `Scripts()` accessors prefer the top-level fields and fall back to `shell.init_hook` / `shell.scripts`. If both are present, scripts are merged by name with the top-level entry winning on conflicts.
- `devbox init` now generates the default config with top-level `init_hook` and `scripts` instead of a `shell` block.

### Migration
- `ConfigFile.MigrateShell()` moves `shell.init_hook` and `shell.scripts` to the top level and removes `shell`, updating both the parsed struct and the hujson AST so comments and formatting survive the rewrite. Existing top-level fields are never overwritten; legacy scripts are merged per name into an existing top-level `scripts` object, and a comment attached to the `shell` member is carried over to the first migrated field rather than dropped.
- New `devbox config fmt` command (`internal/boxcli/config_fmt.go`) opens the config directly (not via `devbox.Open`, so no env setup or warnings), runs `MigrateShell()`, and saves. Supports `-c/--config`.
- `devbox.Open` now emits a one-time warning when a config still uses `shell`, pointing users at `devbox config fmt`. The warning respects `IgnoreWarnings`.

## How was it tested?

- `internal/devconfig/configfile/shell_migration_test.go`: reading legacy nested fields, reading top-level fields, precedence/merge when both are present, migration preserving values and comments (including the `shell` member's own comment), round-trip save/load, and no-op when `shell` is absent.
- `testscripts/config/fmt.test.txt`: end-to-end `devbox config fmt` on a legacy config, verifies `shell` is removed, top-level fields are present, output matches the expected JSON, and a second run is idempotent.
- `go test ./internal/devconfig/... ./internal/boxcli/...` and `go test ./testscripts -run TestScripts/config` pass after merging `main` (which removed the `auth`/`cache` commands; conflict in `root.go` resolved by keeping only the new `config` command registration).

the

https://claude.ai/code/session_017QEtLcEdV7hPXvH7XYkyai


mikeland73:
@claude please review this PR.

mikeland73:
Addressed in 8e3db0a9:

- `ConfigFile.ScriptOrder()` now reads both `shell.scripts` and top-level `scripts` (legacy first, then top-level), matching the merge precedence in `Scripts()`. `InOrder` already uses the last occurrence of a duplicated name, so a script defined in both places takes the top-level position. Added `TestScriptOrderTopLevelScripts`, `TestScriptOrderMergesLegacyAndTopLevelScripts`, and `TestScriptOrderAfterMigrateShell`.
- The deprecated-`shell` warning in `devbox.Open` now uses a `deprecatedShellWarningHasBeenShown` guard, matching the legacy-packages warning.

PR #2843, opened 2026-06-01
nodejs plugin: don't break the shell when corepack is unavailable (nodejs-slim)

### LINKED ISSUE #2791: Node.js plugin fails to run corepack when using nodejs-slim
### What happened?

```
  "packages": [
  ],
  "include": [
    "plugin:nodejs"
  ]
```

### Steps to reproduce

2. `cd node-yarn`
3. `devbox rm nodejs`
4. `devbox add nodejs-slim`
5. `devbox shell --pure`

Output:
```
Starting a devbox shell...
bash: yarn: command not found
```

### Command

shell

### devbox.json

```jsonc
{
  "env": {
  },
  "shell": {
    "init_hook": [
      "yarn install"
    ],
    "scripts": {
      "run_test": "yarn start"
    }
  }
}
```

### Devbox version

0.17.0

### Nix version

nix (Nix) 2.26.2

Linux (x86-64)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2791.

`nodejs-slim` recently [stopped shipping `corepack`](https://github.com/NixOS/nixpkgs/pull/487796) as a default output. The built-in `nodejs` plugin's `init_hook` unconditionally runs `corepack enable` whenever `DEVBOX_COREPACK_ENABLED` is set — and the `node-yarn` / `node-pnpm` templates set that variable by default. As a result, switching a project to `nodejs-slim` broke the shell:

```
bash: yarn: command not found
```

This guards the corepack invocation with a `command -v corepack` check:

- If corepack **is** available → behaves exactly as before (enables corepack and prepends the corepack-bin dir to `PATH`).
- If corepack **is not** available → the shell no longer fails. Instead a single warning is printed explaining that `nodejs-slim` no longer ships corepack and suggesting the `nodejs` package or adding corepack explicitly.
- If `DEVBOX_COREPACK_ENABLED` is unset → nothing runs, as before.

The plugin `version` is bumped `0.0.2` → `0.0.3` to follow convention.

## How was it tested?

Validated the JSON and exercised the new shell snippet in all three states (corepack present + enabled, corepack absent + enabled, disabled) to confirm it enables corepack when present, warns instead of erroring when absent, and stays silent when not enabled.

cc @apgrucza (issue reporter)

---
_Generated by [Claude Code](https://claude.ai/code/session_011Km2Yg8uC6BjZkSPKQE1uT)_


mikeland73:
Closing as superseded: main's nodejs plugin now runs corepack setup via setup-corepack.mjs (#2857), whose run() helper already catches and ignores failures when corepack is unavailable — so the shell no longer breaks. This PR's init_hook approach is obsolete.

PR #2846, opened 2026-06-03
fix(shellenv): preserve literal newlines in exported env values (ambiguous redirect)

### LINKED ISSUE #2814: Bash: ambiguous redirect on `eval "$(devbox global shellenv)"`
### What happened?

`eval "$(devbox global shellenv)"` causes the error message "bash: 1__bp_interactive_mode: ambiguous redirect" to appear at every new prompt. (And by "new prompt" I don't mean when I open a new instance of Bash, but whenever a command finishes execution!)

This is a fresh installation of Devbox on a system that had Devbox installed previously. I deleted all Devbox and Nix configuration files I am aware of. I followed the uninstallation guide and re-installed both Devbox and Nix to rule out a configuration error as best as I could.

I found that the culprit is the following export generated by `devbox global shellenv`:

```
export PROMPT_COMMAND="__bp_precmd_invoke_cmd\
dbus-send --type=method_call --session --dest=io.elementary.terminal /io/elementary/terminal io.elementary.terminal.ProcessFinished string:\$PANTHEON_TERMINAL_ID string:\"\$(fc -nl -1 | cut -c 3-)\" int32:\$__bp_last_ret_value >/dev/null 2>&1\
__bp_interactive_mode";
```

If I move the `__bp_interactive_mode";` to the end of the line above it, it works fine!

Therefore, as a workaround one could write the output of `devbox global shellenv` to a file, edit it accordingly, and add `eval "$(cat workaround_file_name)"` to the Bash config.

### Steps to reproduce

1. Find a system that has the same problem or download [error_snip.txt](https://github.com/user-attachments/files/26686091/error_snip.txt). (Note: the file is shortened to the problematic lines, it's not the entire output of the command.)
2. Execute `eval "$(cat error_snip)"`
3. The output will be: "bash: 1__bp_interactive_mode: ambiguous redirect"
4. Open a new shell and try `eval "$(`[fixed_snip.txt](https://github.com/user-attachments/files/26686275/fixed_snip.txt)`)"` instead.
5. There will be no output.

### Command

global

### devbox.json

```jsonc
{
  "$schema":  "https://raw.githubusercontent.com/jetify-com/devbox/0.17.1/.schema/devbox.schema.json",
  "packages": ["git@latest"],
  "shell": {
    "init_hook": [
    ],
    "scripts": {
      "test": [
      ]
    }
  }
}
```

### Devbox version

0.17.1

### Nix version

2.34.5

Linux (x86-64)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2814.

`eval "$(devbox global shellenv)"` produced the following error on every new prompt for users whose shell sets a multi-line `PROMPT_COMMAND` (e.g. via `bash-preexec`, as the elementary OS terminal does):

```
bash: 1__bp_interactive_mode: ambiguous redirect
```

### Root cause

`exportify` (`internal/devbox/envvars.go`) builds `export KEY="value";` statements and escaped newlines inside the value by writing a backslash before the newline:

```go
case '$', '`', '"', '\\', '\n':
    strb.WriteRune('\\')
```

Inside a bash **double-quoted** string, a backslash-newline is a *line continuation*: the shell removes **both** the backslash and the newline and joins the adjacent lines. So a multi-line value like:

```
__bp_precmd_invoke_cmd
dbus-send ... >/dev/null 2>&1
__bp_interactive_mode
```

was emitted as `... 2>&1\<newline>__bp_interactive_mode` and collapsed by the shell into `... 2>&1__bp_interactive_mode`. The redirect target became `1__bp_interactive_mode`, hence the *ambiguous redirect* error.

Newlines are **not** special inside double quotes (see the [POSIX shell quoting rules](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_02_03) already cited in the code) — they are preserved literally. The fix is simply to stop escaping them.

Quick demonstration of the difference:

```console
$ eval 'X="a 2>&1\
__bp"'; printf '[%s]\n' "$X"
[a 2>&1__bp]          # old behavior: newline eaten, lines joined

$ eval 'X="a 2>&1
__bp"'; printf '[%s]\n' "$X"
[a 2>&1
__bp]                # new behavior: newline preserved
```

### Changes

- `internal/devbox/envvars.go`: drop `'\n'` from the set of characters `exportify` escapes, with a comment explaining why newlines must stay literal. The other shell-special characters (`$`, `` ` ``, `"`, `\`) are still escaped.
- `internal/devbox/envvars_test.go`: new `TestExportify` covering simple values, special-character escaping, and the multi-line regression from this issue.

This affects both `devbox shellenv` output and the env exports written into the generated shellrc; in both contexts the output is sourced as bash, so literal newlines inside double quotes are correct.

## How was it tested?

- `go test ./internal/devbox/ -run TestExportify -v` — passes.
- `go build ./internal/devbox/` and `go vet ./internal/devbox/` — clean.
- Verified the bash line-continuation behavior directly (see demonstration above).

---

cc @proedie (reporter) — thanks for the detailed write-up and the precise pinpointing of the offending line; that made this a quick fix.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EeaBQU7QAgsiT5ziTVpnGb)_


mikeland73:

PR #2865, opened 2026-06-12
fix(generate): use relative paths for env vars in generated README

### LINKED ISSUE #2178: [Minor] Generate readme should not expand env variables paths
### What happened?

If you enable the `postgresql` plugin, it sets some environment variables.

When you generate the readme with `devbox generate readme`, those variables are expanded revealing the full path of the person who run the command.

For example, I could end up with this in the readme:

```
PGDATA="/home/carloratm/ooo/oha/.devbox/virtenv/postgresql/data"
PGHOST="/home/carloratm/ooo/oha/.devbox/virtenv/postgresql"
```

If it is possible, I would like to see 

```
PGDATA: "{{ .Virtenv }}/data",
PGHOST: "{{ .Virtenv }}"
```

I am not entirely sure this is a nice solution, but the user absolute path doesn't look nice either.

Any thoughts?

### Steps to reproduce

1. Enable the `postgresql` plugin (or any other that sets env variables)
2. Generate the readme with `devbox generate readme

### Command

generate

### devbox.json

_No response_

### Devbox version

0.12.0

### Nix version

_No response_

Linux (x86-64)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2178.

`devbox generate readme` rendered plugin-provided environment variables with their **fully expanded absolute paths**. For example, enabling the `postgresql` plugin produced a README containing:

```sh
PGDATA="/home/carloratm/ooo/oha/.devbox/virtenv/postgresql/data"
PGHOST="/home/carloratm/ooo/oha/.devbox/virtenv/postgresql"
```

This leaks machine-specific paths — including the user's home directory / username — into a file that is typically committed to the repository.

The generated README already strips the absolute project directory out of **scripts** via `configfile.Scripts.WithRelativePaths(projectDir)`. Environment variables were the one place this wasn't applied, because `Config().Env()` expands included-plugin env values to absolute paths before they reach the template.

## Fix

Apply the same relative-path treatment to env var values before rendering: replace the absolute project directory with `.`, so the example above becomes:

```sh
PGDATA="./.devbox/virtenv/postgresql/data"
PGHOST="./.devbox/virtenv/postgresql"
```

This is intentionally consistent with the existing `Scripts.WithRelativePaths` behavior (same `strings.ReplaceAll(value, projectDir, ".")` transformation), and matches what was requested on the issue:

> It would indeed be nice, when the generated README.md file only has relative references.

Values that don't contain the project dir (e.g. `PGPORT="5432"`) are left untouched.

## How was it tested?

- Added `internal/devbox/docgen/docgen_test.go` covering: project-dir substitution, that the input map is not mutated, and that an empty project dir returns env unchanged.
- `go test ./internal/devbox/docgen/` passes; `go vet` and `gofmt` are clean.

cc the original reporter (the issue's author account has since been deleted) and @ametad, who reconfirmed the request on the issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J7UpXTjfVU9iAZiRmuUPSy)_

### DIFF
--- a/internal/devbox/docgen/docgen.go
+++ b/internal/devbox/docgen/docgen.go
@@ -2,6 +2,7 @@ package docgen
 
 import (
 	_ "embed"
+	"maps"
 	"os"
 	"strings"
 	"text/template"
@@ -71,7 +72,7 @@ func GenerateReadme(
 // scripts in the generated README.
 func envWithRelativePaths(env map[string]string, projectDir string) map[string]string {
 	if projectDir == "" {
-		return env
+		return maps.Clone(env)
 	}
 	result := make(map[string]string, len(env))
 	for key, value := range env {
--- a/internal/devbox/docgen/docgen.go
+++ b/internal/devbox/docgen/docgen.go
@@ -3,6 +3,7 @@ package docgen
 import (
 	_ "embed"
 	"os"
+	"strings"
 	"text/template"
 
 	"go.jetify.com/devbox/internal/devbox"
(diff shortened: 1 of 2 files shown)


ametad:
Could someone with enough rights, review this? It would be very nice if this could be implemented.

mikeland73:

PR #2867, opened 2026-06-14
fix(plugin): regenerate local plugin create_files when their content changes

### LINKED ISSUE #2755: Local plugins' files in `VirtEnv` aren't updated on change
### What happened?

I'm working on a local plugin. I'd expect that if I enter the devbox shell and a file in `create_files` is missing/stale, it would be created/overwritten, but that doesn't seem to be the case. The file, even if deleted, remains unaffected.

I'd expect this to copy the file again on each `devbox shell` evaluation, ensuring a local plugin can manage the files in the virtenv correctly.

### Steps to reproduce

I've made a dummy repo to reproduce this: https://github.com/tomtaylor/devbox-bug

1. Clone the repo
2. `devbox shell`
3. The plugin will have created a `test.txt` at `.devbox/virtenv/test/test.txt` containing `123`
4. Change the source file in the plugin (`plugin/test/test.txt`) to something else
5. Exit and reenter the shell
6. `.devbox/virtenv/test/test.txt` is still `123`

### Command

shell

### devbox.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.16.0/.schema/devbox.schema.json",
  "packages": [],
  "shell": {
    "init_hook": ["echo 'Welcome to devbox!' > /dev/null"],
    "scripts": {
      "test": ["echo \"Error: no test specified\" && exit 1"]
    }
  },
  "include": [
    "path:./plugins/test"
  ] 
}
```

### Devbox version

0.16.0

### Nix version

2.33.0

macOS (Apple Silicon)

### Debug logs

n/a

mikeland73:
## Summary

Fixes #2755.

When developing a **local plugin** (included via `path:`), editing a file referenced by the plugin's `create_files` does not update the copy in `.devbox/virtenv/...`. Even deleting the generated file and re-entering the shell leaves it missing/stale, which makes local plugin development confusing.

### Root cause

Devbox skips recomputing the environment (and rewriting plugin files) when its **state hash** is unchanged (`internal/lock/statehash.go`). That state hash is derived in part from `Devbox.ConfigHash()`, which hashes each included plugin config via `plugin.Config.Hash()` — and that resolved to the embedded `configfile.ConfigFile.Hash()`, which only hashes **`plugin.json`**.

The files referenced by `create_files` are read from local disk separately and are **not** part of any hash. So editing a local plugin's source file (e.g. `plugin/test/test.txt`) changes nothing the state hash tracks:

- `devbox.json` — unchanged
- the plugin's `plugin.json` — unchanged
- `devbox.lock`, nix manifest, print-dev-env cache — unchanged

`isStateUpToDate` therefore returns `true`, `CreateFilesForConfig` is never called, and the stale virtenv copy stays in place.

(Note: `shouldCreateFile` already returns `true` for files under `.devbox/`, so once recomputation is triggered the file *is* rewritten correctly — the gap was purely in cache invalidation.)

### Fix

Add a `Hash()` method on `plugin.Config` that shadows the embedded `ConfigFile.Hash()`. For **local plugins**, it folds the contents of the files referenced by `create_files` into the hash (sorted for determinism), so editing them invalidates Devbox's cached state and regenerates the virtenv files on the next `devbox shell`/`run`.

This is scoped to local plugins on purpose:
- **Built-in** plugin files are embedded in the Devbox binary, already covered by the `DevboxVersion` field in the state hash.
- **Remote** (github/git) plugins are addressed by an immutable ref, and reading their content can require network/cache access we don't want on every command.

The only caller of `plugin.Config.Hash()` is `Devbox.ConfigHash()`, so the override is tightly scoped to the cache-invalidation path.

## How was it tested?

- `go build ./internal/...`, `go vet ./internal/plugin/`, and `gofumpt` — all clean.
- New unit tests in `internal/plugin/local_test.go`:
  - `TestLocalPluginHashIncludesCreateFilesContent` — hash changes when a `create_files` source file is edited (plugin.json untouched). Fails before this change, passes after.
  - `TestLocalPluginHashStableWithoutContentChange` — hash is stable across calls when nothing changes.

> Note: the nix-backed end-to-end testscripts could not be executed in the authoring environment (no `nix` binary), but the unit tests exercise the exact hashing logic that gates regeneration.

---

cc @tomtaylor (issue reporter) — thanks for the clear repro repo, it made this easy to pin down.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EAJ8ZxjbxPrxrVpx4Atddd)_

### DIFF
--- a/internal/plugin/local_test.go
+++ b/internal/plugin/local_test.go
@@ -1,0 +1,102 @@
+package plugin
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"go.jetify.com/devbox/nix/flake"
+)
+
+// newTestLocalPlugin writes a plugin.json (with the given create_files content
+// file) into a temp dir and returns a *LocalPlugin pointing at it along with the
+// path of the referenced content file.
+func newTestLocalPlugin(t *testing.T, fileContents string) (*LocalPlugin, string) {
+	t.Helper()
+	pluginDir := t.TempDir()
+
+	pluginJSON := `{
+		"name": "testplugin",
+		"create_files": {
+			"{{ .Virtenv }}/test.txt": "test.txt"
+		}
+	}`
+	if err := os.WriteFile(filepath.Join(pluginDir, pluginConfigName), []byte(pluginJSON), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	contentPath := filepath.Join(pluginDir, "test.txt")
+	if err := os.WriteFile(contentPath, []byte(fileContents), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	local, err := newLocalPlugin(flake.Ref{Type: flake.TypePath, Path: pluginDir}, pluginDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return local, contentPath
+}
+
+// TestLocalPluginHashIncludesCreateFilesContent verifies that editing a file
+// referenced by a local plugin's create_files changes the plugin config hash,
+// even though devbox.json and the plugin.json are unchanged. This is what allows
+// Devbox to detect the change and regenerate the file in the virtenv.
+// See https://github.com/jetify-com/devbox/issues/2755.
+func TestLocalPluginHashIncludesCreateFilesContent(t *testing.T) {
+	local, contentPath := newTestLocalPlugin(t, "123")
+
+	pluginJSON, err := local.Fetch()
+	if err != nil {
+		t.Fatal(err)
+	}
+	cfg, err := buildConfig(local, filepath.Dir(contentPath), string(pluginJSON))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	before, err := cfg.Hash()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Edit the referenced content file without touching plugin.json.
+	if err := os.WriteFile(contentPath, []byte("456"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	after, err := cfg.Hash()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if before == after {
+		t.Errorf("hash did not change after editing create_files content: %q", before)
+	}
+}
+
+// TestLocalPluginHashStableWithoutContentChange verifies the hash is stable when
+// the referenced files don't change.
+func TestLocalPluginHashStableWithoutContentChange(t *testing.T) {
+	local, contentPath := newTestLocalPlugin(t, "123")
+
+	pluginJSON, err := local.Fetch()
+	if err != nil {
+		t.Fatal(err)
+	}
+	cfg, err := buildConfig(local, filepath.Dir(contentPath), string(pluginJSON))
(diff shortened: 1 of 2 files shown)


mikeland73:
The one red check (`test (not-main, ubuntu-latest, project-tests-off, 2.19.2)`) is a **pre-existing flaky test unrelated to this change**, not a regression from this PR.

- The only failing testscript is `add_platforms_flakeref.test`, which installs an external GitHub flake (`github:F1bonacc1/process-compose/v1.87.0`) over the network.
- In the **same commit (07682de)**, the identical suite **passed on nix 2.30.2** and only failed on the 2.19.2 job. A test that passes on one nix version and fails on another within the same commit is environmental (network fetch on a slower runner), not a code issue.
- This change only touches local-plugin config hashing (`plugin.Config.Hash()`); it does not affect `devbox add`, flake refs, or this test, which uses no local plugins. All plugin-related testscripts (`plugin.test`, `disable-plugin.test`, `plugin.cycle.test`, `nodejs_corepack_autodetect.test`) and all unit/lint/build jobs are green.

I re-triggered CI once already (the flake recurred only on the 2.19.2 job). I don't have permission to re-run individual jobs via the API, so a maintainer re-run of that single job should clear it.

---

mikeland73:
Heads up on the latest red check (`test (not-main, ubuntu-latest, project-tests-only)` on 33921e8): the only failing case is `TestExamples/development_elixir_elixir_hello_run_test.test`, and it's **unrelated to this PR's change**.

- That example's `init_hook` runs `mix deps.get` (a network fetch from hex.pm) with `elixir@latest`, and it failed fast (~7s) — an external/network dependency, not plugin logic.
- It uses no local plugins. Every plugin-related example and testscript passed, including `plugins_local`, `plugins_v2-local`, `plugins_builtin`, `plugins_github`, `plugins_v2-github`, and `plugins_github-with-revision`, plus all other language examples (ruby, node, python, php, go, rust, java, haskell, zig, etc.).
- This change only affects local-plugin `create_files` config hashing, which this example doesn't exercise.

Likely a transient hex.pm fetch or fallout from the recent `main` merges rather than a regression here. A re-run of just this job should tell — I don't have permission to re-run individual jobs via the API, and since you're actively merging into this branch I've avoided pushing re-trigger commits that could interfere.

---

mikeland73:
Update / correction to my previous note: the same `TestExamples/development_elixir_elixir_hello_run_test.test` failed again on the new commit `82bdfcb` (the 3rd `main` merge). So this is **deterministic, not a transient hex.pm fetch** — it reproduced identically across two different commits, each with a fresh `main` merge. A re-run won't clear it; the Elixir example looks broken on `main` itself (likely `elixir@latest` / `mix deps.get`), independent of this PR.

This change is still uninvolved: the Elixir example uses no local plugins, and every plugin example/testscript continues to pass. Nothing actionable on my end here — flagging so the Elixir example can be looked at separately from this PR. I'll stop re-flagging further Elixir repeats to avoid noise.

---

mikeland73:
Closing as a duplicate of #2871, which addresses the same issue with the same fix.

PR #2871, opened 2026-06-15
fix(plugin): re-create create_files when a local plugin's source changes

### LINKED ISSUE #2755: Local plugins' files in `VirtEnv` aren't updated on change

mikeland73:
## Summary

Fixes #2755.

When iterating on a **local plugin** (`include: ["path:./plugins/..."]`), editing one of the files referenced by the plugin's `create_files` had no effect: the file in `.devbox/virtenv/...` was never updated, and even deleting it didn't get it re-created on the next `devbox shell`.

### Root cause

`devbox shell` short-circuits in `ensureStateIsUpToDate` when the project's **state hash is unchanged** (`File.IsUpToDateAndInstalled` → `Devbox.ConfigHash` → per-plugin `Config.Hash`).

For plugins, `Config.Hash` resolved to the embedded `configfile.ConfigFile.Hash`, which only hashes the **plugin.json** itself. It never looked at the *contents of the files referenced by `create_files`*. So when a local plugin's source file changed (without any edit to `devbox.json` or `plugin.json`), the state hash stayed the same, the project was considered up to date, and `CreateFilesForConfig` was never called — leaving the stale (or missing) file in place.

`LocalPlugin.Hash` similarly only hashes the plugin's *path*, not its content, so it didn't help either.

### Fix

Add a `Config.Hash` method on `plugin.Config` that starts from `ConfigFile.Hash` and additionally folds in the content of each `create_files` source file (read via the plugin's `Source.FileContent`). The map is walked in sorted order so the hash is deterministic.

- **Local (`path:`) plugins** — source edits now change the hash, invalidate the state, and trigger a re-create of the virtenv files. ✅
- **Built-in / git / github plugins** — their file contents are stable for a given version (embedded FS / cached fetch), so the hash is unchanged and there is no behavior change or spurious recompute.

A missing/unreadable source file is tolerated (the file path is still part of the hash) so hashing never hard-fails shell startup.

## How was it tested?

- Added `internal/plugin/plugin_test.go::TestConfigHashIncludesCreateFilesContent`, which builds a local plugin `Config`, asserts the hash is stable across repeated calls with no change, and asserts it changes when a `create_files` source file's content changes.
- `go test ./internal/plugin/ -run TestConfigHashIncludesCreateFilesContent` passes.
- `go build ./...` and `gofmt`/`go vet` are clean for the changed package.

(The two unrelated `TestGitPluginFileContentCache*` failures in this sandbox are pre-existing — they `git commit` and the sandbox's commit-signing server rejects it; they don't touch this code path.)

cc @tomtaylor (issue reporter)

---
_Generated by [Claude Code](https://claude.ai/code/session_01BURbKokeurXFkBYFuCd7k8)_

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/tailscale/hujson"
+	"go.jetify.com/devbox/internal/cachehash"
 	"go.jetify.com/devbox/internal/devconfig/configfile"
 	"go.jetify.com/devbox/internal/devpkg"
 	"go.jetify.com/devbox/internal/lock"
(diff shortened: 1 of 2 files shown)


mikeland73:
Heads-up on the one red check — `test (not-main, ubuntu-latest, project-tests-only, 2.18.0)`: the only failing test is `TestExamples/development_elixir_elixir_hello_run_test.test`. All plugin tests passed, including `plugins_local`, `plugins_v2-local`, and `plugins_builtin`.

This failure is unrelated to this PR. The elixir example uses only the built-in `elixir@latest` package (no local `path:` plugin) and its `init_hook` runs `mix deps.get`, which fetches hex deps over the network — a known flaky point. This PR's change is a strict no-op for non-local plugins (`Config.Hash` returns the unchanged `ConfigFile.Hash` unless the source is a `*LocalPlugin`), so it cannot affect the elixir example.

I don't have permission to re-run the failed job from here (API returns 403). A maintainer re-running the failed job should turn it green.

---

mikeland73:

PR #2873, opened 2026-06-17
Fix nodejs plugin corepack setup failing in "type": "module" projects

### LINKED ISSUE #2856: setup-corepack.js fails when "module" is specified in root package.json
### What happened?

devbox with nodejs does not start anymore due to root `package.json` configuration spilling into `.devbox` internal working.

### Steps to reproduce

1. Create a folder
2. Add the file `package.json` with `{"type":"module"}`
3. Add the file `devbox.json` with `{"include":["plugin:nodejs"]}`
4. Run `devbox shell`

(OS: WSL Debian)
(nodejs@24 plugin_version: 0.0.3)

### Command

shell

### devbox.json

```jsonc
{
"include": ["plugin:nodejs"]
}
```

### Devbox version

0.17.3

### Nix version

_No response_

Other (please include in the description above)

### Debug logs

```
Starting a devbox shell...
file:///.../devboxtest/.devbox/virtenv/nodejs/bin/setup-corepack.js:12
const { execFileSync } = require("node:child_process");
                         ^

ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/.../devboxtest/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///.../devboxtest/.devbox/virtenv/nodejs/bin/setup-corepack.js:12:26
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

mikeland73:
## Summary

Fixes #2856.

The `nodejs` plugin runs a Corepack setup script as its `init_hook`. The script uses CommonJS `require()`, but it was named `setup-corepack.js`. When a project's root `package.json` declares `"type": "module"`, Node treats every `.js` file in that package — including this one — as an ES module, so the script crashes with:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
```

This breaks `devbox shell` entirely for any Node project using ES modules, which is increasingly common.

## Fix

Rename the script to `setup-corepack.cjs`. The `.cjs` extension forces Node to always interpret it as CommonJS, regardless of the surrounding `package.json` `type` field. The plugin's `init_hook` and `create_files` references are updated accordingly, and the plugin version is bumped `0.0.3` → `0.0.4`.

## Testing

- Added a third case to `testscripts/plugin/nodejs_corepack_autodetect.test.txt` that copies a `package.json` declaring `"type": "module"` and verifies shell init still succeeds.
- Verified at the Node level that a `.js` file with `require()` fails under `"type": "module"` while the equivalent `.cjs` file loads correctly.

## Notes

cc @AdaptivChris (issue reporter) — thanks for the clear reproduction.

https://claude.ai/code/session_015e5foES7T8gDQRHK3hJSdH

---
_Generated by [Claude Code](https://claude.ai/code/session_015e5foES7T8gDQRHK3hJSdH)_

### DIFF
--- a/plugins/nodejs.json
+++ b/plugins/nodejs.json
@@ -1,6 +1,6 @@
 {
     "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/main/.schema/devbox-plugin.schema.json",
-    "version": "0.0.3",
+    "version": "0.0.4",
     "name": "nodejs",
     "readme": "Devbox automatically configures Corepack for Nodejs when DEVBOX_COREPACK_ENABLED=1. You can install Yarn or Pnpm by adding them to your `package.json` file using `packageManager`\nCorepack binaries will be installed in your local `.devbox` directory\n\nWhen Corepack is enabled, Devbox also activates the package manager pinned in your `package.json` `packageManager` field automatically. Set DEVBOX_DISABLE_NODEJS_PACKAGE_MANAGER_AUTODETECT=1 to disable this behavior.",
     "env": {
@@ -9,11 +9,11 @@
     },
     "shell": {
         "init_hook": [
-            "node \"{{ .Virtenv }}/bin/setup-corepack.js\""
+            "node \"{{ .Virtenv }}/bin/setup-corepack.cjs\""
         ]
     },
     "create_files": {
       "{{ .Virtenv }}/corepack-bin": "",
-      "{{ .Virtenv }}/bin/setup-corepack.js": "nodejs/setup-corepack.js"
+      "{{ .Virtenv }}/bin/setup-corepack.cjs": "nodejs/setup-corepack.cjs"
     }
 }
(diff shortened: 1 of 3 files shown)


mikeland73:
Heads-up on CI: the only failing check (`test (not-main, ubuntu-latest, project-tests-only)`) is failing on `TestExamples/development_elixir_elixir_hello_run_test.test`, which is **unrelated to this change**. It fails while installing Hex from the network:

```
** (Mix) httpc request failed with: {:failed_connect, [{:to_address, {~c"builds.hex.pm", 443}}, {:inet, [:inet], {:tls_alert, {:unsupported_certificate, ...}}}]}
Could not install Hex because Mix could not download metadata at https://builds.hex.pm/installs/hex-1.x.csv.
```

This is an environmental/network flake reaching `builds.hex.pm`, not a code regression. All Node.js / Corepack example and plugin tests (the actual subject of this PR) pass. It reproduced on a re-run; I don't have permission to re-run jobs via the API, so a maintainer re-run should clear it.

---

PR #2888, opened 2026-06-22
fix(nodejs plugin): warn when Corepack is enabled but unavailable

### LINKED ISSUE #2791: Node.js plugin fails to run corepack when using nodejs-slim

mikeland73:
## Summary

Fixes #2791.

When `DEVBOX_COREPACK_ENABLED` is set, the nodejs plugin's `init_hook` runs `corepack enable`. Some Node packages no longer bundle Corepack — notably `nodejs-slim` ([nixpkgs#487796](https://github.com/NixOS/nixpkgs/pull/487796)) and Node.js 25+. In those cases the `corepack` binary is missing, the enable step failed silently, and the user later hit a cryptic error with no indication of the cause:

```
bash: yarn: command not found
```

## Fix

- **`plugins/nodejs/setup-corepack.mjs`**: detect the missing `corepack` binary (`ENOENT`) when running `corepack enable` and print an actionable warning pointing the user to add the `corepack` package, instead of failing silently. Package-manager activation is skipped when Corepack couldn't be enabled. Other failures (e.g. being offline) are still ignored so they never block shell initialization.

  ```
  [devbox] Warning: DEVBOX_COREPACK_ENABLED is set but the `corepack` command
  was not found. Some Node packages (such as nodejs-slim, and Node.js 25+) no
  longer bundle Corepack. Add the `corepack` package to your devbox.json (for
  example, run `devbox add corepack`) to use Yarn or pnpm.
  ```

- **`plugins/nodejs.json`**: bump the plugin version to `0.0.5` (so existing projects regenerate the script) and clarify the readme to call out `nodejs-slim` alongside Node.js 25+.

- **`testscripts/plugin/nodejs_corepack_autodetect.test.txt`**: add a case that switches the project to `nodejs-slim` and asserts the shell still initializes successfully and the warning is printed.

## How was it tested?

- `go test ./plugins/...` passes.
- `node --check` on the script passes, and I exercised it directly under Node 22:
  - no-op when `DEVBOX_COREPACK_ENABLED` is unset;
  - with `corepack` on `PATH`: installs the `yarn`/`pnpm`/... shims into the bin dir, exit 0;
  - with `corepack` removed from `PATH`: prints the warning, exits 0, does not crash.
- Added testscript coverage for the `nodejs-slim` (no bundled Corepack) path.

cc @apgrucza (issue reporter)

---
_Generated by [Claude Code](https://claude.ai/code/session_0112TqTA3ttT2HsVuz7gWUY3)_

### DIFF
--- a/plugins/nodejs.json
+++ b/plugins/nodejs.json
@@ -1,8 +1,8 @@
 {
     "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/main/.schema/devbox-plugin.schema.json",
-    "version": "0.0.4",
+    "version": "0.0.5",
     "name": "nodejs",
-    "readme": "Devbox automatically configures Corepack for Nodejs when DEVBOX_COREPACK_ENABLED=1. You can install Yarn or Pnpm by adding them to your `package.json` file using `packageManager`\nCorepack binaries will be installed in your local `.devbox` directory\n\nWhen Corepack is enabled, Devbox also activates the package manager pinned in your `package.json` `packageManager` field automatically. Set DEVBOX_DISABLE_NODEJS_PACKAGE_MANAGER_AUTODETECT=1 to disable this behavior.\n\nNote: newer versions of Nodejs (25+) no longer bundle Corepack, so you must add the `corepack` package to your devbox.json separately for Corepack to be available.",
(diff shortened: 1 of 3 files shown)


mikeland73:

PR #2892, opened 2026-06-24
fix(nodejs plugin): warn when Corepack is unavailable instead of failing silently

### LINKED ISSUE #2791: Node.js plugin fails to run corepack when using nodejs-slim

mikeland73:
## Summary

Fixes #2791.

The nodejs plugin's `setup-corepack.mjs` init_hook runs `corepack enable` and silently swallows any failure. On `nodejs-slim` and Node.js 25+ — which [no longer bundle Corepack](https://github.com/NixOS/nixpkgs/pull/487796) — this leaves users with a cryptic, unexplained failure at shell startup:

```
bash: yarn: command not found
```

…and no indication of *why* Corepack/yarn are missing or how to fix it.

## Fix

`run()` now detects the missing-binary case (`ENOENT`) and prints actionable guidance pointing the user at the documented fix, then cleanly skips package-manager activation:

```
[devbox] nodejs plugin: `corepack` was not found, so Corepack-managed package managers (such as yarn and pnpm) will be unavailable.
[devbox] Your Node.js package does not bundle Corepack (e.g. nodejs-slim, or Node.js 25+). Add it explicitly with `devbox add corepack` to enable it.
```

This matches the guidance already present in the plugin's `readme`. Shell initialization still succeeds (the warning is non-fatal) and the **happy path is unchanged**: when Corepack is present, `corepack enable` and `corepack prepare --activate <packageManager>` run exactly as before.

The plugin `version` is bumped `0.0.4` → `0.0.5` so cached copies in users' `virtenv` are refreshed.

## How was it tested?

- `node --check plugins/nodejs/setup-corepack.mjs` passes.
- Ran the script against a `PATH` with `node` but no `corepack`: it prints the guidance above and exits 0 (non-fatal).
- Ran it with a stub `corepack` on `PATH` and a pinned `packageManager` in `package.json`: it still calls `corepack enable --install-directory …` followed by `corepack prepare --activate yarn@4.0.0`, confirming the happy path is unaffected.

cc @apgrucza (issue reporter)

---

https://claude.ai/code/session_01MEAWvyABicqkLncj6irnQq

---
_Generated by [Claude Code](https://claude.ai/code/session_01MEAWvyABicqkLncj6irnQq)_

### DIFF
--- a/plugins/nodejs.json
+++ b/plugins/nodejs.json
@@ -1,6 +1,6 @@
 {
     "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/main/.schema/devbox-plugin.schema.json",
-    "version": "0.0.4",
+    "version": "0.0.5",
     "name": "nodejs",
     "readme": "Devbox automatically configures Corepack for Nodejs when DEVBOX_COREPACK_ENABLED=1. You can install Yarn or Pnpm by adding them to your `package.json` file using `packageManager`\nCorepack binaries will be installed in your local `.devbox` directory\n\nWhen Corepack is enabled, Devbox also activates the package manager pinned in your `package.json` `packageManager` field automatically. Set DEVBOX_DISABLE_NODEJS_PACKAGE_MANAGER_AUTODETECT=1 to disable this behavior.\n\nNote: newer versions of Nodejs (25+) no longer bundle Corepack, so you must add the `corepack` package to your devbox.json separately for Corepack to be available.",
     "env": {
--- a/plugins/nodejs/setup-corepack.mjs
+++ b/plugins/nodejs/setup-corepack.mjs
@@ -27,11 +27,22 @@ if (!corepackBinDir) {
   process.exit(0);
 }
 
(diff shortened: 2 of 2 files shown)


mikeland73:
Closing as a duplicate of #2888, which addresses the same issue with the same fix.

PR #2894, opened 2026-06-26
fix(shellenv): only export env vars Devbox adds or changes

### LINKED ISSUE #2826: Why `devbox shellenv` and `devbox global shellenv` export all existing envvars ?

Currently, both `devbox shellenv` and `devbox global shellenv` export not only the environment variables that actually need to be modified, but also unrelated variables such as `HOSTNAME` and `LANG`.  
Depending on the distribution, some environment variables are set as read‑only (for example, openSUSE marks variables like `PROFILEREAD` as read‑only).  
When running `eval "$(devbox global shellenv)"` in an environment where such read‑only variables exist, an unexpected error occurs, such as:

```
(eval):55: read-only variable: PROFILEREAD
```

Redefining environment variables that are unrelated to devbox with their original values should not be necessary.  
Therefore, I believe that `devbox shellenv` and `devbox global shellenv` should export **only the environment variables that are strictly required by devbox**, and avoid exporting unrelated ones.

_No response_

mikeland73:
## Summary

Fixes #2826.

`devbox shellenv` and `devbox global shellenv` export the **entire** environment, including variables that Devbox never touches such as `HOSTNAME` and `LANG`. On distributions that mark some variables read-only (e.g. openSUSE marks `PROFILEREAD` read-only), `eval "$(devbox global shellenv)"` then fails with errors like:

```
(eval):55: read-only variable: PROFILEREAD
```

Re-exporting a variable to the value it already has is a no-op at best, and at worst breaks the user's prompt/shell on every command.

## Fix

`EnvExports` now restricts the exported variables to those Devbox actually **adds or changes** relative to the current shell environment. Variables whose value already matches the ambient environment are dropped, since they are already present in the shell — so there is no reason to re-emit them (and no reason to trip over read-only ones).

- `internal/devbox/envvars.go`: add `onlyModifiedEnvVars(env, ambient)` helper that returns only new/changed variables.
- `internal/devbox/devbox.go`: in `EnvExports`, filter the computed env against `os.Environ()` when the new option is set. Skipped in **pure** mode, where the intent is to emit a complete, self-contained environment rather than a diff against the current shell.
- `internal/devbox/devopt/devboxopts.go`: add `OnlyModifiedEnv` to `EnvExportsOpts`.
- `internal/boxcli/shellenv.go`: enable `OnlyModifiedEnv` for the `shellenv` command (the reported path). The `devbox shell --print-env` / direnv path is intentionally left unchanged.

Variables Devbox relies on (`PATH`, the `DEVBOX_*` helpers, the shellenv hash, plugin/config env, etc.) are always new or changed, so they continue to be exported. Re-running `shellenv` inside an already-active environment now emits only what differs, which is idempotent.

## Testing

- Added `TestOnlyModifiedEnvVars` in `internal/devbox/envvars_test.go` covering unchanged (dropped), changed (kept), and new (kept) variables, including a read-only-style `PROFILEREAD` case.
- `go test ./internal/devbox/ -run 'TestOnlyModifiedEnvVars|TestExportify'` passes.
- `go build ./internal/...`, `go vet`, and `gofmt` are clean.

(The remaining failures in `./internal/devbox/` unit tests are pre-existing and environmental — they require the `nix` binary, which isn't available in CI sandbox; they are unrelated to this change.)

cc @gasuketsu (issue reporter)

---

https://claude.ai/code/session_01E6tBNHN6DFv5VabVTAosNq

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6tBNHN6DFv5VabVTAosNq)_

### DIFF
--- a/internal/boxcli/shellenv.go
+++ b/internal/boxcli/shellenv.go
@@ -149,9 +149,10 @@ func shellEnvFunc(
 			Pure:              flags.pure,
 			SkipRecompute:     !flags.recomputeEnv,
 		},
-		NoRefreshAlias: flags.noRefreshAlias,
-		RunHooks:       flags.runInitHook,
-		ShellFormat:    shellFormat,
+		NoRefreshAlias:  flags.noRefreshAlias,
+		OnlyModifiedEnv: true,
+		RunHooks:        flags.runInitHook,
+		ShellFormat:     shellFormat,
 	})
 	if err != nil {
 		return "", err
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -379,6 +379,16 @@ func (d *Devbox) EnvExports(ctx context.Context, opts devopt.EnvExportsOpts) (st
 		return "", err
 	}
 
+	// For `devbox shellenv`, only emit the variables that Devbox actually adds
+	// or changes relative to the current shell. Re-exporting unrelated variables
+	// (e.g. HOSTNAME, LANG) is redundant, and can fail when the user's shell
+	// marks some of them read-only (e.g. PROFILEREAD on openSUSE). See #2826.
+	// In pure mode we keep the full environment, since the goal there is a
+	// complete, self-contained environment rather than a diff.
+	if opts.OnlyModifiedEnv && !opts.EnvOptions.Pure {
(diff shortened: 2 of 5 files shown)


mikeland73:

PR #2903, opened 2026-07-05
fix(nix): detect nix at known locations, not just $PATH (#2787)

### LINKED ISSUE #2787: fish shell: nix unknown.
### Package name

devbox add error

### What changes are you requesting?

cannot using command devbox add, list and nix.
Error: We found a /nix directory but nix binary is not in your PATH and we were not able to find it in the usual locations. Your nix installation might be broken. If restarting your terminal or reinstalling nix doesn't work, please create an issue at https://github.com/jetify-com/devbox/issues

### Nixhub link

_No response_

mikeland73:
## Summary

Fixes #2787.

Users on setups where nix is installed but not on the interactive shell's `$PATH` — notably **NixOS**, and shells that don't source the Nix profile the way devbox expects (e.g. **fish**) — hit this even though nix is present and working:

```
Error: We found a /nix directory but nix binary is not in your PATH and we were not able to find it in the usual locations. Your nix installation might be broken.
```

The root cause is that devbox used **two different mechanisms** to find nix:

- To *run* nix commands, `nix.(*Nix).resolvePath()` searches `$PATH` and then falls back to well-known install locations, invoking nix by **absolute path**. This works even when nix isn't on `$PATH`.
- To *detect* whether nix is installed, `BinaryInstalled()` used a bare `exec.LookPath("nix")` — a `$PATH`-only check.

So detection could report nix as missing and abort in `EnsureNixInstalled`, even though devbox would have been perfectly able to run nix by absolute path.

On top of that, `resolvePath`'s fallback list had a latent bug: the NixOS entry was the bin **directory** `/run/current-system/sw/bin`, but the loop only accepts a path that is a non-directory executable — so that entry could never match, and the NixOS system-profile location was effectively never searched.

## Fix

- `internal/nix/install.go`: `BinaryInstalled()` now uses the same resolver that runs nix commands, via a new exported `nix.(*Nix).LookPath` / `nix.LookPath`. Detection and execution can no longer disagree.
- `nix/nix.go`: fix the fallback list so the NixOS entry points at the nix **binary** (`/run/current-system/sw/bin/nix`), and extract the list into `nixBinaryFallbackPaths()`.
- The happy path is unchanged: nix on `$PATH` is still found first; behavior only differs when nix is off-`$PATH` but present at a known location, where devbox now proceeds instead of erroring.

## How was it tested?

- `go build ./nix/... ./internal/nix/... ./internal/boxcli/...` passes.
- `go test ./nix/...` passes, including a new `TestNixBinaryFallbackPaths` regression test asserting every fallback entry is an absolute path to the nix binary (guarding against the directory-vs-binary bug) and that the NixOS system-profile path is covered.
- `go test ./internal/nix/...`: the added/affected tests pass. (`TestConfigIsUserTrusted` fails in this sandbox only because it shells out to a real `nix`, which isn't installed here — it fails identically on the unchanged base branch and is unrelated to this change.)

cc @simulatedcode (issue reporter)

the

https://claude.ai/code/session_01BXJuDH6keqvU1ydywuyizq

---
_Generated by [Claude Code](https://claude.ai/code/session_01BXJuDH6keqvU1ydywuyizq)_

### DIFF
--- a/internal/nix/install.go
+++ b/internal/nix/install.go
@@ -16,13 +16,18 @@ import (
 	"github.com/mattn/go-isatty"
 
 	"go.jetify.com/devbox/internal/boxcli/usererr"
-	"go.jetify.com/devbox/internal/cmdutil"
 	"go.jetify.com/devbox/internal/fileutil"
 	"go.jetify.com/devbox/nix"
 )
 
 func BinaryInstalled() bool {
-	return cmdutil.Exists("nix")
+	// Use the same resolver that Devbox uses to run nix commands (searching
+	// $PATH and well-known install locations), rather than a bare $PATH lookup.
+	// Otherwise Devbox can wrongly report nix as missing when it exists at a
+	// known location but isn't on $PATH — e.g. on NixOS, or when a non-POSIX
+	// login shell such as fish hasn't sourced the Nix profile (issue #2787).
+	_, err := nix.Default.LookPath()
+	return err == nil
 }
 
 func dirExistsAndIsNotEmpty(dir string) bool {
--- a/nix/nix.go
+++ b/nix/nix.go
@@ -34,6 +34,11 @@ func System() string {
 	return Default.System()
 }
 
+// LookPath calls [Nix.LookPath] on the default Nix installation.
+func LookPath() (string, error) {
+	return Default.LookPath()
+}
+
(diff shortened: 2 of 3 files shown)


mikeland73:

PR #2922, opened 2026-07-14
fix(nix): check per-output binary-cache status in InputAddressedPathForOutput (#2921)

### LINKED ISSUE #2921: 0.17.3 regression: InputAddressedPathForOutput fails to fall back to secondary substituters
## Description

Devbox 0.17.3 introduced a regression where `InputAddressedPathForOutput` (called during flake.nix template generation) fails with `"Package X cannot be fetched from binary cache store"` when a package's store path is:
1. Not present in the local nix store, AND
2. Not present in the **first** configured nix substituter

Devbox 0.17.2 handles this case correctly by falling through to secondary substituters (e.g. `cache.nixos.org`).

## Steps to Reproduce

1. Configure nix with multiple substituters where the first one is a private/partial cache:
   ```
   # /etc/nix/nix.conf
   substituters = https://my-private-cache.s3.amazonaws.com https://cache.nixos.org
   require-sigs = false
   ```

2. Create a `devbox.json` with standard packages:
   ```json
   {
     "packages": {
       "dnsutils": "latest",
       "tzdata": "latest",
       "curl": "latest"
     }
   }
   ```

3. Ensure the packages are NOT in the local nix store (fresh machine or after `nix-collect-garbage`)

4. Ensure the packages are NOT in the first substituter (`my-private-cache`) — they only exist in `cache.nixos.org`

5. Run `devbox shellenv` with a clean `.devbox` directory

## Expected Behavior (0.17.2)

Devbox successfully resolves packages. When the first substituter doesn't have a narinfo, nix falls through to `cache.nixos.org` and fetches the package.

## Actual Behavior (0.17.3)

```
Info: Installing the following packages to the nix store: dnsutils@latest, ...
Error: Package "dnsutils@latest" cannot be fetched from binary cache store
```

The error originates from `flake.nix.tmpl:52` at `$pkg.InputAddressedPathForOutput`. The stack trace:

```
go.jetify.com/devbox/internal/shellgen.writeFromTemplate
  go.jetify.com/devbox/internal/shellgen/generate.go:90
go.jetify.com/devbox/internal/shellgen.makeFlakeFile
  go.jetify.com/devbox/internal/shellgen/generate.go:171
go.jetify.com/devbox/internal/shellgen.GenerateForPrintEnv
  go.jetify.com/devbox/internal/shellgen/generate.go:60
go.jetify.com/devbox/internal/devbox.(*Devbox).recomputeState
  go.jetify.com/devbox/internal/devbox/packages.go:408
```

## Key Observation

The `nix path-info --offline` check that devbox runs **before** this step correctly reports the packages as missing. But instead of letting nix handle the fetch (which would use all configured substituters), `InputAddressedPathForOutput` appears to do its own binary cache lookup that only checks a subset of substituters.

## Environment

- Nix 2.18.8 (multi-user daemon install)
- OS: CentOS Stream 10 (x86_64)
- nix.conf has `binary-caches = https://private-cache.s3.amazonaws.com https://cache.nixos.org`
- `trusted-users` includes the running user

## Workaround

Pin to devbox 0.17.2 (`DEVBOX_USE_VERSION=0.17.2`) which does not have this regression.

mikeland73:
## Summary

Fixes #2921.

Devbox 0.17.3 regressed so that `devbox shellenv` (and any command that regenerates the environment) fails during flake generation with:

```
Error: Package "dnsutils@latest" cannot be fetched from binary cache store
```

for multi-output packages (e.g. `dnsutils`, `curl`) when the requested output is available in the binary cache but not *every* one of the package's default outputs is.

### Root cause

The error is raised by `InputAddressedPathForOutput` (`internal/devpkg/package.go`), called from the flake template at `internal/shellgen/tmpl/flake.nix.tmpl:52` — exactly where the reported stack trace points.

That template line only runs **inside** `{{ if $output.CacheURI }}`, and `CacheURI` is set **per output** by `GetOutputsWithCache` → `fetchNarInfoStatusOnce(name)`, which checks *that single output*.

But `InputAddressedPathForOutput` internally guarded on `IsInBinaryCache()`, which is the **all-default-outputs** check (`areExpectedOutputsInCacheOnce(useDefaultOutputs)` requires `len(found) == len(defaultOutputs)`).

So for a package whose requested output is cached while some other default output is not, the template committed to emitting a `builtins.fetchClosure` for the cached output, yet the stricter internal guard reported the whole package as not-cached and returned the error. The per-output check (template) and the all-outputs check (guard) disagreed. This is the only code path that produces that error message at that template line.

### Fix

`InputAddressedPathForOutput` returns the input-addressed path for a **single** output, so its precondition should concern only that output. Switch its guard from `IsInBinaryCache()` to `IsOutputInBinaryCache(output)`. This mirrors `InstallableForOutput`, which already gates the very same call on the per-output check:

```go
func (p *Package) InstallableForOutput(output string) (string, error) {
    inCache, err := p.IsOutputInBinaryCache(output)   // per-output
    ...
    if inCache {
        installable, err := p.InputAddressedPathForOutput(output)
```

The plural `InputAddressedPaths()` is unchanged: it still guards on `IsInBinaryCache()` (it legitimately wants all default outputs) before iterating, so relaxing the singular guard cannot weaken it — each per-output check there is a subset of the all-outputs check the caller already passed.

The error message now names the specific output for clearer diagnostics.

## How was it tested?

- `go build ./...` and `go vet ./internal/devpkg/` — clean.
- `go test ./internal/devpkg/` — passes.
- Verified by code analysis that this is the sole path producing `"... cannot be fetched from binary cache store"` at `flake.nix.tmpl:52`, and that the fix restores agreement between the template's per-output caching decision and the guard.

Note: the binary-cache resolution path gates on a Nix ≥ 2.17 install (`sysInfoIfExists` → `nix.AtLeast`), so an end-to-end reproduction requires a Nix environment with a partially-populated cache (as in the issue). I was unable to run that scenario in the sandbox used here; reviewers with a Nix toolchain can confirm against the issue's repro steps.

cc @sstarcher (issue reporter) — thanks for the detailed diagnosis and stack trace.

the

https://claude.ai/code/session_017rMbKrge1aJW6mNbGuPGDS

---
_Generated by [Claude Code](https://claude.ai/code/session_017rMbKrge1aJW6mNbGuPGDS)_

### DIFF
--- a/internal/devpkg/package.go
+++ b/internal/devpkg/package.go
@@ -602,11 +602,23 @@ func (p *Package) InputAddressedPaths() ([]string, error) {
 }
 
 func (p *Package) InputAddressedPathForOutput(output string) (string, error) {
-	if inCache, err := p.IsInBinaryCache(); err != nil {
+	// Check that this specific output is in the binary cache, rather than
+	// requiring every one of the package's default outputs to be cached.
+	//
+	// The flake template calls this per-output, and only for outputs it has
+	// already determined are cached (via GetOutputsWithCache, which checks each
+	// output individually). A multi-output package can have the requested
+	// output cached while some other default output is not. Guarding with the
+	// stricter IsInBinaryCache (which requires all default outputs) would
+	// spuriously fail in that case with "cannot be fetched from binary cache
+	// store", even though the output being fetched here is present. This also
+	// matches InstallableForOutput, which already gates on IsOutputInBinaryCache.
+	// See issue #2921.
+	if inCache, err := p.IsOutputInBinaryCache(output); err != nil {
 		return "", err
 	} else if !inCache {
 		return "",
-			errors.Errorf("Package %q cannot be fetched from binary cache store", p.Raw)
+			errors.Errorf("Package %q output %q cannot be fetched from binary cache store", p.Raw, output)
 	}
 
 	entry, err := p.lockfile.Resolve(p.LockfileKey())


sstarcher:
Anything to get movement here.  We are blocked on upgrading devbox without this fix.

internal/devpkg/package.go
```diff
@@ -602,11 +602,23 @@ func (p *Package) InputAddressedPaths() ([]string, error) {
...
+	//
+	// The flake template calls this per-output, and only for outputs it has
+	// already determined are cached (via GetOutputsWithCache, which checks each
+	// output individually). A multi-output package can have the requested
+	// output cached while some other default output is not. Guarding with the
+	// stricter IsInBinaryCache (which requires all default outputs) would
+	// spuriously fail in that case with "cannot be fetched from binary cache
+	// store", even though the output being fetched here is present. This also
+	// matches InstallableForOutput, which already gates on IsOutputInBinaryCache.
+	// See issue #2921.
```
savil, line 616: This makes sense, but I don't think this comment should be added here in the source code. Outside of this specific issue's bug fix, we don't need this comment's information for understanding the functionality of this function.

savil (approved):

sstarcher:
Anything else needed to get this merged in?  It blocks us from ugprading.

savil:
@sstarcher sorry for the delay - I've merged it now

internal/devpkg/package.go
savil: oof, I forgot to ensure this is removed prior to merge 🤦🏾

ipince: This is overly wordy Claude Opus being overly wordy.

PR #2926, opened 2026-07-18
fix(env): expose Devbox profile share dir via XDG_DATA_DIRS so package completions work

### LINKED ISSUE #2776: package completions not active in devbox shells (e.g. kubectl )
### What happened?

The kubectl nix package seems to have builtin completion enabled, see https://github.com/NixOS/nixpkgs/blob/d04d8548aed39902419f14a8537006426dc1e4fa/pkgs/by-name/ku/kubectl/package.nix#L24-L29

However, following a `devbox shell`, the kubectl completion is not active.

I'm running plain fresh ubuntu 24.0.4 with bash shell.

Same problem with all packages, e.g. kind also supporting completion https://github.com/NixOS/nixpkgs/blob/d04d8548aed39902419f14a8537006426dc1e4fa/pkgs/by-name/ki/kind/package.nix#L41-L46

### Steps to reproduce

1. devbox shell
2. kubectl <tab>
3.

### Command

_No response_

### devbox.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/jetpack-io/devbox/0.10.4/.schema/devbox.schema.json",
  "packages": [
    "kubectl@1.33.4"
  ],
  "shell": {
    "init_hook": [
    ],
    "scripts": {
      "test": [
      ]
    }
  }
}
```

### Devbox version

0.16.0

### Nix version

nix (Nix) 2.18.1

Linux (x86-64)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2776.

Package-provided shell completions were never active inside a Devbox shell. For
example, on a fresh Ubuntu + bash setup, `kubectl <TAB>` does nothing even though
the `kubectl` Nix package ships a completion script.

### Root cause

Nix packages install their bash completions under
`share/bash-completion/completions/<cmd>`, and `bash-completion`'s dynamic loader
discovers them by scanning the `bash-completion/completions` subdirectory of
**every** entry in `XDG_DATA_DIRS` at completion time.

When computing the environment (`internal/devbox/devbox.go`), Devbox only
preserved the host's original `XDG_DATA_DIRS`:

```go
if !envOpts.Pure {
    // preserve the original XDG_DATA_DIRS by prepending to it
    env["XDG_DATA_DIRS"] = envpath.JoinPathLists(env["XDG_DATA_DIRS"], os.Getenv("XDG_DATA_DIRS"))
}
```

It never added the Devbox profile's own `share` directory
(`.devbox/nix/profile/default/share`), which is where all of a project's
installed packages aggregate their data files. As a result the completions — and
other XDG data such as man pages and icons — shipped by those packages were
invisible to the shell.

### Fix

Prepend the profile's `share` directory to `XDG_DATA_DIRS` while computing the
environment:

```go
profileShareDir := filepath.Join(d.projectDir, nix.ProfilePath, "share")
env["XDG_DATA_DIRS"] = envpath.JoinPathLists(profileShareDir, env["XDG_DATA_DIRS"])
```

- Done for both regular and `--pure` environments, so completions keep working
  in pure shells too.
- `envpath.JoinPathLists` cleans, dedupes, and drops empty/relative segments, so
  this is a no-op if the entry is already present and it never introduces a
  duplicate or an unsafe relative path.
- This mirrors how `nix-env` / home-manager expose the same data via
  `XDG_DATA_DIRS`.

## How was it tested?

- `go build ./...` and `go vet ./internal/devbox/` — clean.
- Added `testscripts/run/xdg_data_dirs.test.txt`, which asserts that
  `devbox run echo '$XDG_DATA_DIRS'` includes
  `.devbox/nix/profile/default/share`.

Manual verification of the underlying mechanism: with the profile's `share`
directory on `XDG_DATA_DIRS`, `bash-completion`'s `__load_completion` finds
`share/bash-completion/completions/kubectl` in the Devbox profile and activates
`kubectl <TAB>` completion; the change is additive and does not alter any other
environment variable.

/cc @gberche-orange (issue reporter) — thanks for the clear report and the
upstream package links.

the

https://claude.ai/code/session_01EuAos6xjPuVdUWNEVvCYh1

---
_Generated by [Claude Code](https://claude.ai/code/session_01EuAos6xjPuVdUWNEVvCYh1)_

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -820,6 +820,18 @@ func (d *Devbox) computeEnv(
 
 	slog.Debug("computed environment PATH", "path", env["PATH"])
 
+	// Expose the Devbox profile's share directory through XDG_DATA_DIRS so that
+	// data files installed by packages are discoverable inside the Devbox
+	// environment. Most notably this includes shell completions (which Nix
+	// packages install under share/bash-completion/completions), but also man
+	// pages, icons, and other XDG data. Tools such as bash-completion look these
+	// up via XDG_DATA_DIRS, so without the profile's share directory the
+	// completions shipped by packages like kubectl are never loaded. This is
+	// done even in --pure mode so completions keep working there too.
+	// See https://github.com/jetify-com/devbox/issues/2776
+	profileShareDir := filepath.Join(d.projectDir, nix.ProfilePath, "share")
+	env["XDG_DATA_DIRS"] = envpath.JoinPathLists(profileShareDir, env["XDG_DATA_DIRS"])
+
 	if !envOpts.Pure {
 		// preserve the original XDG_DATA_DIRS by prepending to it
 		env["XDG_DATA_DIRS"] = envpath.JoinPathLists(env["XDG_DATA_DIRS"], os.Getenv("XDG_DATA_DIRS"))
--- a/testscripts/run/xdg_data_dirs.test.txt
+++ b/testscripts/run/xdg_data_dirs.test.txt
@@ -7,6 +7,11 @@
 exec devbox run echo '$XDG_DATA_DIRS'
 stdout '\.devbox/nix/profile/default/share'
 
+# The profile's share directory must also be exposed in --pure mode, where the
+# host's XDG_DATA_DIRS is not inherited, so completions keep working there too.
+exec devbox run --pure echo '$XDG_DATA_DIRS'
+stdout '\.devbox/nix/profile/default/share'
+
 -- devbox.json --
 {
   "packages": []
--- a/testscripts/run/xdg_data_dirs.test.txt
+++ b/testscripts/run/xdg_data_dirs.test.txt
@@ -1,0 +1,13 @@
+# The Devbox profile's share directory must be exposed through XDG_DATA_DIRS so
+# that data files installed by packages are discoverable inside the Devbox
+# environment. Most importantly this lets tools like bash-completion find the
+# shell completions that packages install under share/bash-completion/completions.
+# https://github.com/jetify-com/devbox/issues/2776
+
+exec devbox run echo '$XDG_DATA_DIRS'
+stdout '\.devbox/nix/profile/default/share'
+
+-- devbox.json --
+{
+  "packages": []
+}


mikeland73:

mikeland73:
@mikeland73 Reviewed for correctness and merge-readiness. Summary: **the change is correct and I found no blocking bugs. The only thing stopping a merge is that the PR is currently a draft.**

### Correctness ✅
The mechanism is right. `bash-completion`'s dynamic loader resolves `$XDG_DATA_DIRS` **at completion time** and scans `<dir>/bash-completion/completions/<cmd>` for each entry; Nix packages install completions under `share/bash-completion/completions/`, and the Devbox profile aggregates them at `.devbox/nix/profile/default/share`. That directory was never on `XDG_DATA_DIRS`, so the completions were invisible — the fix adds it.

Things I verified specifically:
- **`d.projectDir` is absolute** (`filepath.Dir(cfg.Root.AbsRootPath)`), which matters because `envpath.JoinPathLists` silently drops relative paths — so the profile-share entry is actually retained, not discarded.
- **Applies uniformly** via `computeEnv`, so `devbox shell`, `devbox run`, `devbox shellenv`, and `global` all get it consistently.
- **`--pure` is handled** — the entry is added before the `!Pure` host-append, so pure shells keep completions (covered by the added `--pure` test assertion).
- **Idempotent/safe** — `JoinPathLists` cleans and dedupes, so no duplicates and no-op if already present.
- **Builds & merges** — merges cleanly into current `main` (no conflicts; `#2941`'s `devbox.go` edits don't overlap), and `go build ./...` + `go vet` are clean on the merged tree.

### Merge blockers
1. **Draft status (the actual blocker).** The PR was converted to draft by @savil, so it can't merge as-is. I have **not** marked it ready, since another maintainer set it and I don't want to override that unprompted — say the word and I'll mark it ready for review.
2. `mergeable_state: blocked` otherwise just reflects required review approval; **CI is fully green** (all 25 checks incl. the `test-result` gate, every `test`/`test-nix-versions` shard, lint, builds, flake build) and there are **no merge conflicts**.

### Two non-blocking notes for your call
- **Edge case — empty `XDG_DATA_DIRS`.** Per the XDG spec, once `XDG_DATA_DIRS` is set to a non-empty value the implicit defaults (`/usr/local/share:/usr/share`) no longer apply. On a host where it was previously *unset/empty* (e.g. a minimal container), we now set it to the profile share, so consumers like `bash-completion` would search only that and stop falling back to `/usr/share`. On normal desktop distros this is a non-issue (the session sets `XDG_DATA_DIRS`), but if you want to be defensive I can append the XDG default dirs when the resulting value would otherwise be empty. Happy to add that if you'd prefer.
- **Slight broadening.** Exposing the profile `share/` also surfaces other XDG data from installed packages (desktop entries, GSettings schemas, icons, mime). This is standard nix-profile behavior and generally what you want, just noting it's a touch beyond completions.

Net: correct and ready to go once it's out of draft. Let me know if you'd like me to (a) mark it ready and/or (b) add the empty-`XDG_DATA_DIRS` fallback.

---

PR #2906, opened 2026-07-06
fix(mariadb,mysql plugin): ignore /etc/mysql config files, for better isolation (#2904)

jefft:
## Summary

Fixes #2904.

By default, mariadbd reads /etc/mysql/mariadb.cnf, and if the system has another version of MariaDB installed, those config files can cause the Devbox mariadbd to fail (e.g. requesting loading plugins not available).

Devbox services ought to be isolated from their environment. This commit causes devbox mariadbd (and mysqld) to only read from $MYSQL_CONF (typically devbox.d/mariadb/my.cnf).

Additionally, wrap the client/admin binaries (mariadb, mariadb-admin, mariadb-dump, mysqldump, mysql, mysqladmin) with --defaults-file and --socket so they also ignore /etc/mysql and connect to the correct socket.

### Mariadb

Optionally, create a deliberately broken config option in the default `mariadb.cnf` as a canary:
```sh
echo -e "[server]\nunknown_param_to_break_mariadbd_devbox_testing=true" | sudo tee -a /etc/mysql/mariadb.cnf
```
Then run:
```sh
mkdir /tmp/mariadbtest
cd /tmp/mariadbtest
devbox init
devbox add mariadb
devbox run mariadbd --verbose --help | grep ^port   # prints 3306
echo "port = 13306" >> devbox.d/mariadb/my.cnf  # Set a nonstandard port
devbox run mariadbd --verbose --help | grep ^port   # prints 13306, showing my.cnf is used
devbox services up -b
devbox run mariadb -e "show variables where variable_name in ('datadir', 'port', 'log_error', 'socket');"
devbox run mariadb -e "status;"
```
For me the latter prints:
```
jturner@jturner-desktop:/tmp/mariadbtest$ devbox run mariadb -e "show variables where variable_name in ('datadir', 'port', 'log_error', 'socket');"
Info: Running script "mariadb" on /tmp/mariadbtest
+---------------+---------------------------------------------------------+
| Variable_name | Value                                                   |
+---------------+---------------------------------------------------------+
| datadir       | /tmp/mariadbtest/.devbox/virtenv/mariadb/data/          |
| log_error     | /tmp/mariadbtest/.devbox/virtenv/mariadb/run/mysql.log  |
| port          | 13306                                                   |
| socket        | /tmp/mariadbtest/.devbox/virtenv/mariadb/run/mysql.sock |
+---------------+---------------------------------------------------------+

[...]

Issue #2904, opened 2026-07-06
mariadb reads and get confused by /etc/mysql/ config files


jefft:
### What happened?

Mariadb 11.4 (`mariadb@latest`) always seems to read `/etc/mysql/mariadb.cnf`, unless `--no-defaults` is specified.  This causes mariadbd to fail to start if one has another version of MariaDB installed via system packages, with more extensive plugins. E.g. with debian `mariadb-server-10.5` installed:

```sh
$ devbox run mariadbd --verbose 
Info: Running script "mariadbd" on /tmp/mariadbtest
2026-07-06 18:52:55 0 [Note] Starting MariaDB 11.0.4-MariaDB-log source revision 48af85db217a88409e6aac51bf86e3bf525d8f07 as process 938257
2026-07-06 18:52:55 0 [ERROR] mariadbd: Can't open shared library '/nix/store/23nb6k2y7cp4g46gbhpkq6pcc8l61y35-nix-shell-env/lib/mysql/plugin/provider_bzip2.so' (errno: 20, cannot open shared object file: Not a directory)
2026-07-06 18:52:55 0 [ERROR] Couldn't load plugins from 'provider_bzip2.so'.
2026-07-06 18:52:55 0 [ERROR] mariadbd: Can't open shared library '/nix/store/23nb6k2y7cp4g46gbhpkq6pcc8l61y35-nix-shell-env/lib/mysql/plugin/provider_lz4.so' (errno: 20, cannot open shared object file: Not a directory)
2026-07-06 18:52:55 0 [ERROR] Couldn't load plugins from 'provider_lz4.so'.
2026-07-06 18:52:55 0 [ERROR] mariadbd: Can't open shared library '/nix/store/23nb6k2y7cp4g46gbhpkq6pcc8l61y35-nix-shell-env/lib/mysql/plugin/provider_lzma.so' (errno: 20, cannot open shared object file: Not a directory)
2026-07-06 18:52:55 0 [ERROR] Couldn't load plugins from 'provider_lzma.so'.
2026-07-06 18:52:55 0 [ERROR] mariadbd: Can't open shared library '/nix/store/23nb6k2y7cp4g46gbhpkq6pcc8l61y35-nix-shell-env/lib/mysql/plugin/provider_lzo.so' (errno: 20, cannot open shared object file: Not a directory)
2026-07-06 18:52:55 0 [ERROR] Couldn't load plugins from 'provider_lzo.so'.
[...]

PR #2930, opened 2026-07-23
fix(mariadb plugin): wrap mysql/mariadb clients with project socket (#2522)

### LINKED ISSUE #2522: mariadb (but not mysql) does not respect the MYSQL_UNIX_PORT env variable
### What happened?

To replicate:

```sh
cd /tmp
devbox create mariadbtest --template mariadb
cd mariadbtest
devbox run bash -c 'echo $MYSQL_UNIX_PORT'  # Prints /tmp/mariadbtest/.devbox/virtenv/mariadb/run/mysql.sock
# Note: do not start mariadb. This means we expect the connect attempt to fail
devbox run mariadb
```
The output ends with:
```
ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
```
This shows that the socket path, `MYSQL_UNIX_PORT`, is not being read.

To show that this is buggy, run the exact same steps with mysql instead of mariadb:

```sh
cd /tmp
devbox create mysqltest --template mysql
cd mysqltest
devbox run bash -c 'echo $MYSQL_UNIX_PORT'  # Prints /tmp/mysqltest/.devbox/virtenv/mysql80/run/mysql.sock
# Note: do not start mysql. This means we expect the connect attempt to fail
devbox run mysql
```
This prints
```
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/tmp/mysqltest/.devbox/virtenv/mysql80/run/mysql.sock' (2)
```
which is correct - the MYSQL_UNIX_PORT socket is being used.

--- 

Perhaps this is an upstream mariadb bug. A fix/workaround would be for Devbox's mariadb flake.nix to generate `mysql` and `mariadb` wrappers that set `--socket=$MYSQL_UNIX_PORT`.

### Steps to reproduce

_No response_

### Command

_No response_

### devbox.json

```jsonc

```

### Devbox version

0.14.0

### Nix version

nix (Nix) 2.18.5

Linux (x86-64)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2522.

The MariaDB client does not honor the `MYSQL_UNIX_PORT` environment variable, so running the client inside a Devbox environment (`devbox run mariadb` / `devbox run mysql`) falls back to its compile-time default socket (e.g. `/run/mysqld/mysqld.sock`) instead of the project's socket at `$MYSQL_UNIX_PORT`:

```
ERROR 2002 (HY000): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
```

This differs from the `mysql` plugin, whose (Oracle MySQL) client *does* honor `MYSQL_UNIX_PORT`, which is why the same steps work there.

This change wraps the `mysql` and `mariadb` client binaries in the mariadb plugin's `flake.nix` so they default to `--socket=$MYSQL_UNIX_PORT`, mirroring how the daemon binaries (`mysqld`, `mariadbd`, `mysqld_safe`, ...) are already wrapped. It's the fix suggested by the reporter in the issue.

Details:
- Each wrap is guarded by an existence check (`[ -e ... ]`) because the available client names vary across MariaDB versions (`mysql` is typically a symlink to `mariadb`).
- `--add-flags` prepends the socket flag, so users can still override it explicitly (e.g. connecting to a remote host over TCP is unaffected).
- The generated flake lives under `.devbox/`, which Devbox always regenerates, so existing projects pick up the fix on the next shell entry (the changed derivation rebuilds the wrapped package).

cc @jefft (issue reporter)

## How was it tested?

- `go build ./internal/plugin/... ./plugins/...` passes; the plugin's `flake.nix` is embedded via `//go:embed` in `plugins/builtins.go`.
- Reviewed the generated Nix `postBuild` output for shell correctness (proper `[ -e ... ]` guards, `''$MYSQL_UNIX_PORT` escaping consistent with the surrounding wrappers).

Note: I don't have a Nix-enabled environment here to run the full `devbox create --template mariadb` reproduction end to end, so a maintainer verification of `devbox run mariadb` connecting to `$MYSQL_UNIX_PORT` would be appreciated.

the

---
_Generated by [Claude Code](https://claude.ai/code/session_01D8QQU1dVaNycyxQWcRyBhs)_

### DIFF
--- a/plugins/mariadb/flake.nix
+++ b/plugins/mariadb/flake.nix
@@ -28,6 +28,21 @@
               --add-flags '--basedir=$out --datadir=''$MYSQL_DATADIR --pid-file=''$MYSQL_PID_FILE --socket=''$MYSQL_UNIX_PORT';
           fi
 
+          # Point the client binaries at the project's socket. Unlike the MySQL
+          # client, the MariaDB client does not honor the MYSQL_UNIX_PORT
+          # environment variable, so without this it falls back to its
+          # compile-time default (e.g. /run/mysqld/mysqld.sock).
+          # https://github.com/jetify-com/devbox/issues/2522
+          if [ -e "$out/bin/mysql" ]; then
+            wrapProgram "$out/bin/mysql" \
+              --add-flags '--socket=''$MYSQL_UNIX_PORT';
+          fi
+
+          if [ -e "$out/bin/mariadb" ]; then
+            wrapProgram "$out/bin/mariadb" \
+              --add-flags '--socket=''$MYSQL_UNIX_PORT';
+          fi
+
           wrapProgram "$out/bin/mysql_install_db" \
             --add-flags '--basedir=$out --datadir=''$MYSQL_DATADIR --pid-file=''$MYSQL_PID_FILE --basedir=''$MYSQL_BASEDIR';
 


jefft:
I fixed #2522 properly in https://github.com/jetify-com/devbox/pull/2906. This Claude PR is untested and broken:

 - it only fixes `mariadb`, leaving `mariadb-admin` and `mariadb-dump` to use the wrong socket. 
- `mysql` should not be wrapped (contrary the suggestion in my bug report) because in the native package, `mysql` is just an symlink for `mariadb` and as such behaves correctly when `mariadb` is fixed. In fact in my experience, wrapping `mysql` et al actively breaks it (as noted in comment https://github.com/jetify-com/devbox/pull/2906#issuecomment-4916942931 - I wish I'd recorded how/why it breaks things!).

There's also little use fixing #2522 without also fixing #2904, because both result in commands like `mariadb` interacting with the wrong instance.

mikeland73:
Thanks for the detailed review, @jefft — you're right on all counts, and I'm closing this in favor of #2906.

I missed that #2906 already fixes #2522 (its title only references #2904, so my search for open PRs against #2522 didn't surface it). Your fix is clearly the better one:

- It wraps `mariadb`, `mariadb-admin`, and `mariadb-dump` — mine only covered the interactive client, so `mariadb-admin`/`mariadb-dump` would still have used the wrong socket.
- It correctly leaves `mysql` as a symlink rather than wrapping it. Good to know that wrapping the `mysql` symlink actively breaks things — my PR wrapped it, so that was a real bug.
- It also addresses the `/etc/mysql` isolation from #2904 via `--defaults-file`, which as you note is what makes the socket fix actually useful.

Closing #2930. #2906 supersedes it.

---

PR #2935, opened 2026-07-28
fix(env): don't leak nixpkgs SOURCE_DATE_EPOCH into the shell (#2597)

### LINKED ISSUE #2597: SOURCE_DATE_EPOCH variable causes Docker images to shows created to be "45 years ago"
### What happened?

When using the devbox the variable `SOURCE_DATE_EPOCH` get's set to '315532800' (I assume for reproducability reasons) but causes my Docker images to appear created 45 years ago.

This causes our automatic cleanup to delete just recently created images in the registry.

Perhaps this is a Docker issue instead but maybe you can give me a reason for the variable being set in the devbox shell.

### Steps to reproduce

1. Be in a devbox shell
2. Build a Docker image
3. Run `docker images`  and check the "CREATED" column

### Command

_No response_

### devbox.json

```jsonc

```

### Devbox version

0.14.2

### Nix version

_No response_

macOS (Intel)

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2597.

The nixpkgs stdenv sets `SOURCE_DATE_EPOCH` to a fixed timestamp
(`315532800` = 1980-01-01) so that *builds* are reproducible. Devbox sources the
variables from `nix print-dev-env` into the interactive shell, so this value
leaked into every Devbox environment.

As a result, any timestamp-respecting tool run inside the shell stamps its output
with 1980 instead of the current time. The most visible symptom is Docker images
showing up as created **"45 years ago"** — which, as the reporter notes, can trip
age-based registry cleanup and delete freshly built images. It also affects
`tar`, `gzip`, and other build tools.

**Root cause:** `computeNixEnv` (`internal/devbox/devbox.go`) copies every
`exported` variable from `nix print-dev-env` except those in `ignoreDevEnvVar`.
That denylist mirrors Nix's own `nix develop` ignore list, which does **not**
include `SOURCE_DATE_EPOCH`, so the stdenv value passed straight through.

**Fix:** add `SOURCE_DATE_EPOCH` to `ignoreDevEnvVar`, mirroring how `HOME` and
`TMPDIR` (other stdenv-provided values that are wrong for an interactive shell)
are already ignored. Devbox now drops the Nix-provided value and falls back to
the current environment — normally unset, so tools use the real time. Users who
genuinely want reproducible builds can still set `SOURCE_DATE_EPOCH` explicitly
via `devbox.json`'s `env` block, which is layered on top of these variables.

Worth noting: the `python/pip` and `temporal` examples already carry a
`SOURCE_DATE_EPOCH=$(date +%s)` workaround in their generated `venvShellHook.sh`,
which corroborates that the 1980 leak has been a real-world nuisance.

```diff
 	"SHELLOPTS":          true,
+	"SOURCE_DATE_EPOCH":  true,
 	"TEMP":               true,
```

## How was it tested?

- `go build ./internal/devbox/`, `go vet ./internal/devbox/`, and `gofmt -l` — all clean.
- Added `testscripts/run/source_date_epoch.test.txt`, which asserts that
  `devbox run echo '$SOURCE_DATE_EPOCH'` does not emit the `315532800` (1980)
  value. This fails on the old code and passes with the fix.

cc @truearken (issue reporter)

the

---
_Generated by [Claude Code](https://claude.ai/code/session_017YBWVPRVeEoftDjUW6WUR4)_

### DIFF
--- a/internal/devbox/devbox.go
+++ b/internal/devbox/devbox.go
@@ -1079,8 +1079,9 @@ var ignoreCurrentEnvVar = map[string]bool{
 // ignoreDevEnvVar contains environment variables that Devbox should remove from
 // the slice of [Devbox.PrintDevEnv] variables before sourcing them.
 //
-// This list comes directly from the "nix develop" source:
+// Most of this list comes directly from the "nix develop" source:
 // https://github.com/NixOS/nix/blob/f08ad5bdbac02167f7d9f5e7f9bab57cf1c5f8c4/src/nix/develop.cc#L257-L275
+// Entries not in that list are called out below.
 var ignoreDevEnvVar = map[string]bool{
 	"BASHOPTS":           true,
 	"HOME":               true,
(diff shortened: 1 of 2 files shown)


mikeland73:

PR #2936, opened 2026-07-29
fix(services): re-invoke devbox by executable path, not hardcoded name (#1321)

### LINKED ISSUE #1321: [Bug][minor]: services commands rely on the binary being called "devbox". Any other name will break.
**Current Behavior (bug)**

**Expected Behavior (fix)**

**Additional context**

mikeland73:
## Summary

Fixes #1321.

`devbox services start|stop|ls|restart` (and the non-current-shell path of `devbox services up`) work by **re-invoking devbox inside the computed environment**. `runDevboxServicesScript` builds a command and hands it to `RunScript`, which writes a small script that `eval`s the command. The command began with the hard-coded string `"devbox"`:

```go
// internal/devbox/services.go
func (d *Devbox) runDevboxServicesScript(ctx context.Context, cmdArgs []string) error {
	cmdArgs = append([]string{"services"}, cmdArgs...)
	return d.RunScript(ctx, devopt.EnvOptions{}, "devbox", cmdArgs)
}
```

So the generated script effectively runs `eval "devbox services ..."`, which requires the binary to be named exactly `devbox` **and** be resolvable on `PATH`. If the binary is installed or renamed to anything else (a versioned install, a wrapper, `devbox-cli`, etc.), services commands fail with `devbox: command not found`.

This was the only real self-invocation of the binary — the process-compose manager already shells out via `processComposeConfig.BinPath`, and every other `devbox ...` occurrence in the services code is user-facing help text.

### Fix

Reference the currently running binary via `os.Executable()` instead of the literal `"devbox"`. The path is shell-quoted (via `strconv.Quote`, matching how the surrounding code already quotes arguments before they are `eval`'d) so it survives the eval even if it contains spaces. If the executable path can't be determined, it falls back to `"devbox"` — the previous PATH-lookup behavior.

```go
func devboxBinaryForSelfInvocation() string {
	exe, err := os.Executable()
	if err != nil {
		return "devbox"
	}
	return strconv.Quote(exe)
}
```

This mirrors the existing use of `os.Executable()` in `internal/devbox/pure_shell.go`.

## How was it tested?

- `go build ./...` — clean.
- `go vet ./internal/devbox/` — clean.
- `gofmt` — no diffs.
- Added `TestDevboxBinaryForSelfInvocation` in `internal/devbox/services_test.go`, which asserts the helper returns the shell-quoted path to the running binary rather than the literal `"devbox"`. It passes:
  ```
  --- PASS: TestDevboxBinaryForSelfInvocation (0.00s)
  ```

cc @mikeland73 (issue reporter)

the

https://claude.ai/code/session_01BQ4KvbTQb7oJ2M3tc79DuG

---
_Generated by [Claude Code](https://claude.ai/code/session_01BQ4KvbTQb7oJ2M3tc79DuG)_

### DIFF
--- a/internal/devbox/services.go
+++ b/internal/devbox/services.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"strconv"
 	"text/tabwriter"
 
@@ -271,7 +272,24 @@ func (d *Devbox) StartProcessManager(
 // defaults for the `devbox services` scenario.
 func (d *Devbox) runDevboxServicesScript(ctx context.Context, cmdArgs []string) error {
 	cmdArgs = append([]string{"services"}, cmdArgs...)
-	return d.RunScript(ctx, devopt.EnvOptions{}, "devbox", cmdArgs)
+	return d.RunScript(ctx, devopt.EnvOptions{}, devboxBinaryForSelfInvocation(), cmdArgs)
+}
+
+// devboxBinaryForSelfInvocation returns a shell-quoted reference to the
+// currently running devbox binary. `devbox services ...` re-invokes devbox
+// inside the computed environment, and using the actual executable path (rather
+// than a hardcoded "devbox") ensures this works even when the binary has been
+// installed or renamed to something other than "devbox". If the executable path
+// can't be determined, it falls back to "devbox", relying on a PATH lookup as
+// before. See https://github.com/jetify-com/devbox/issues/1321.
+func devboxBinaryForSelfInvocation() string {
+	exe, err := os.Executable()
+	if err != nil {
+		return "devbox"
+	}
+	// Quote the path so it survives being eval'd by the generated run script
+	// (e.g. when the path contains spaces).
+	return strconv.Quote(exe)
 }
 
 func (d *Devbox) ShowProcessComposePort(ctx context.Context, writer io.Writer) error {
--- a/internal/devbox/services_test.go
+++ b/internal/devbox/services_test.go
@@ -1,0 +1,32 @@
+// Copyright 2024 Jetify Inc. and contributors. All rights reserved.
+// Use of this source code is governed by the license in the LICENSE file.
+
+package devbox
+
+import (
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDevboxBinaryForSelfInvocation verifies that `devbox services ...`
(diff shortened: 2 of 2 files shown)


mikeland73:

mikeland73:
Reviewed for correctness. Summary: I believe this is correct and I don't see anything that blocks merging. Details below, including one behavioral note worth a maintainer's eye and one hardening change I just pushed.

## Correctness

- **Coverage is complete.** Every non-`--run-in-current-shell` service subcommand — `start`, `stop`, `ls`, `restart`, and `up` (via `StartProcessManager`) — routes through `runDevboxServicesScript`, which is the single place devbox re-invokes itself. That was the only real self-invocation: the process-compose manager already execs via `processComposeConfig.BinPath`, and every other `devbox …` string in the services code is user-facing help text.
- **The generated Dockerfile/devcontainer templates were intentionally left alone.** Those `["devbox", …]` entries run inside a container where `devbox` on `PATH` is the right reference, not the host's `os.Executable()`.
- **`os.Executable()` is an established pattern here** (already used in `pure_shell.go`, `telemetry`, `nixcache/setup`, `setup`).

## Behavioral note (not a blocker)

Re-invoking `os.Executable()` runs the currently-executing binary directly, bypassing a `PATH`-based `devbox` launcher wrapper if one is installed. For a self-invocation this is the desired behavior — it guarantees the *same* version that's already running (no redundant launcher re-resolution), and `RunScript` passes the inherited environment (launcher env vars included) through to the nested call. Since the launcher has already selected the correct (possibly project-pinned) binary before we reach this code, there's no version regression. Flagging it only so it's a conscious choice.

## Hardening pushed (`eeed220`)

My first commit quoted the path with `strconv.Quote`. Since the path is `eval`'d by a POSIX `sh` run script, `strconv.Quote`'s Go-style `\uXXXX` escapes wouldn't be interpreted by the shell inside double quotes — so a devbox binary under a non-ASCII path would have broken. Switched to `shellescape.Quote` (already used in `shell.go`), which emits POSIX single-quoted output that round-trips correctly for spaces, unicode, and shell metacharacters. Test updated accordingly.

## Verification

`go build ./...`, `go vet ./internal/devbox/`, and `gofmt` are clean; `TestDevboxBinaryForSelfInvocation` passes.

One pre-existing item I did **not** touch (out of scope): the `cmdArgs` in `RunScript` are still `strconv.Quote`d and share the same theoretical unicode limitation, but those are service names/flags rather than filesystem paths.

Note: the PR is currently in **draft** (converted by @savil). I haven't flipped it back to ready — leaving that to you.

---

PR #2939, opened 2026-07-31
fix(nix): don't report a broken install for a freshly mounted /nix volume (#2601)

### LINKED ISSUE #2601: Devbox is failing when an empty /nix folder exists
### What happened?

I want to use devbox in docker and mount the /nix volume.
Initially the volume is empty and I would expect that devbox can handle that.
But it fails with:
`Error: We found a /nix directory but nix binary is not in your PATH and we were not able to find it in the usual locations. Your nix installation might be broken. If restarting your terminal or reinstalling nix doesn't work, please create an issue at https://github.com/jetify-com/devbox/issues`

### Steps to reproduce

1. create an empty /nix folder
2. start devbox

### Command

_No response_

### devbox.json

```jsonc

```

### Devbox version

0.14.2

### Nix version

_No response_

Linux

### Debug logs

_No response_

mikeland73:
## Summary

Fixes #2601.

When running Devbox in a container (e.g. Docker or Kubernetes) with `/nix` mounted as a persistent volume, the volume starts out effectively empty. Devbox is supposed to install Nix into it, but instead it fails with:

> Error: We found a /nix directory but nix binary is not in your PATH and we were not able to find it in the usual locations. Your nix installation might be broken. [...]

### Root cause

`EnsureNixInstalled` decided whether `/nix` already contained an installation by checking whether the directory was non-empty:

```go
if dirExistsAndIsNotEmpty("/nix") { ... return "broken installation" error ... }
```

A truly empty `/nix` already falls through to the installer, but a freshly **mounted** volume is usually not truly empty: `ext4` (and other filesystems) create a `lost+found` directory at the root of the mount. So `/nix` looks non-empty, Devbox assumes a pre-existing (broken) install, and refuses to install Nix.

### Fix

Detect an existing installation by ignoring known filesystem cruft (`lost+found`, `.DS_Store`) rather than treating any non-empty `/nix` as installed. A `/nix` that contains only those entries is now treated as a fresh install target, so Devbox proceeds to install Nix as expected. A `/nix` that contains real content (e.g. a `store` directory) is still treated as an existing installation, preserving the original "broken install" diagnostics.

The generic `dirExistsAndIsNotEmpty` helper was only used here, so it's replaced by a `/nix`-specific `nixDirIsInstalled` helper.

## How was it tested?

- `go test ./internal/nix/ -run TestNixDirIsInstalled -v` — new table-driven test covering: empty dir, dir with files, dir with a nix `store`, dir with hidden files, **dir with only `lost+found`** (the regression case → treated as not installed), dir with `lost+found` + `store` (→ installed), and non-existent dir. All pass.
- `go build ./...` / `go vet ./internal/nix/` / `gofmt` clean.

(Pre-existing `TestConfigIsUserTrusted` failures in the package are unrelated — they require a `nix` binary that isn't present in the CI sandbox and are not touched by this change.)

cc @ascknx (issue reporter)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Upz2KchxkA6erYatxPzVjg)_

### DIFF
--- a/internal/nix/install.go
+++ b/internal/nix/install.go
@@ -17,7 +17,6 @@ import (
 
 	"go.jetify.com/devbox/internal/boxcli/usererr"
 	"go.jetify.com/devbox/internal/cmdutil"
-	"go.jetify.com/devbox/internal/fileutil"
 	"go.jetify.com/devbox/nix"
 )
 
@@ -25,9 +24,33 @@ func BinaryInstalled() bool {
 	return cmdutil.Exists("nix")
 }
 
-func dirExistsAndIsNotEmpty(dir string) bool {
-	empty, err := fileutil.IsDirEmpty(dir)
-	return err == nil && !empty
+// nonNixDirEntries are directory entries that can appear inside /nix without
+// indicating an existing Nix installation. Container and VM users commonly
+// mount an otherwise-empty volume at /nix, and many filesystems (for example
+// ext4) create a lost+found directory at the root of such a mount. A /nix that
+// contains only these entries should still be treated as a fresh install
+// target rather than a broken installation.
+var nonNixDirEntries = map[string]bool{
+	"lost+found": true,
+	".DS_Store":  true,
+}
+
+// nixDirIsInstalled reports whether dir looks like it already contains a Nix
+// installation. It returns false when dir is missing, empty, or contains only
+// filesystem cruft (see nonNixDirEntries) so that Devbox installs Nix into a
+// freshly mounted /nix volume instead of reporting a broken installation.
+// See https://github.com/jetify-com/devbox/issues/2601.
+func nixDirIsInstalled(dir string) bool {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false
+	}
+	for _, entry := range entries {
+		if !nonNixDirEntries[entry.Name()] {
+			return true
+		}
+	}
+	return false
 }
 
 var ensured = false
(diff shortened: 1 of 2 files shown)


mikeland73:

mikeland73:
Reviewed for correctness and merge-readiness.

## Correctness — looks good ✅

The change swaps the generic `dirExistsAndIsNotEmpty("/nix")` check for a `/nix`-specific `nixDirIsInstalled` that ignores known filesystem cruft (`lost+found`, `.DS_Store`). I checked behavior parity across every path in `EnsureNixInstalled`:

| `/nix` contents | old behavior | new behavior |
|---|---|---|
| missing / empty | install | install (unchanged) |
| real content (`store`, `var`, …) | "broken install" error | "broken install" error (unchanged) |
| unreadable (permission error) | install¹ | install¹ (unchanged) |
| **only `lost+found`** (mounted volume) | **"broken install" error** | **install (fixed)** |

¹ `os.ReadDir` erroring returns `false` here, mirroring the old `fileutil.IsDirEmpty` error path — so no regression for root-owned `/nix` read by a non-root user.

The only intended behavior change is the last row, which is exactly #2601. `go build ./...`, `go vet`, `gofmt`, and the new table-driven `TestNixDirIsInstalled` all pass locally. The now-unused `dirExistsAndIsNotEmpty` helper and its `fileutil` import are cleanly removed (no other references).

## Blocking merge 🚧

1. **The PR is in draft** (converted by @savil) — GitHub won't merge a draft, and the conversion dropped any auto-merge/merge-queue membership. It needs to be marked *Ready for review* by a maintainer. I'm intentionally not flipping it back myself.
2. **CI + approval** — status is currently `blocked` / checks pending. Needs CI green and a maintainer approval under branch protection.

## Non-blocking notes

- This fixes Devbox's *own* premature guard. Whether the underlying Nix installer then installs cleanly into a mounted `/nix` that contains a `lost+found` is a separate concern outside this function — worth an end-to-end check in a real Docker volume before closing the issue, but it applied equally to the pre-existing empty-`/nix` install path, so it's not a regression from this PR.
- `.DS_Store` is included defensively (harmless on macOS); happy to drop it if you'd prefer to scope the ignore-list strictly to `lost+found`.

---

PR #2940, opened 2026-08-04
feat!: remove the Jetify Nix cache

mikeland73:
First of five stacked PRs making Devbox account-free. Each PR removes one login-gated feature; the last one drops the dependencies.

## The stack

Each PR is based on the one above it — review and merge bottom-up, starting here.

- `main`
  - **1.** [#2940 — remove the Jetify Nix cache](https://github.com/jetify-com/devbox/pull/2940) &nbsp;⬅&nbsp; **you are here**
    - **2.** [#2941 — remove Jetify Cloud secrets](https://github.com/jetify-com/devbox/pull/2941)
      - **3.** [#2942 — remove global push/pull to Jetify Cloud](https://github.com/jetify-com/devbox/pull/2942)
        - **4.** [#2943 — remove devbox auth and the identity provider](https://github.com/jetify-com/devbox/pull/2943)
          - **5.** [#2944 — drop dependencies left over from the account features](https://github.com/jetify-com/devbox/pull/2944)

## Why

The Jetify Nix cache required a Jetify Cloud account: cache URIs and the S3 credentials backing them were fetched from the Jetify API using the logged-in session. Without an account it did nothing; with one it changed `nix build` behavior silently on every `add`/`install`/`shell`.

## What's removed

- The `devbox cache` command tree: `upload`/`copy`, `configure`, `credentials`, `enable`, `info`.
- `internal/devbox/providers/nixcache`, which fetched AWS credentials and cache URIs from the Jetify API.
- The substituter path in `installNixPackagesToStore`, so `nix build` no longer receives `--extra-substituters` or AWS credentials — along with the build-from-source retry that existed only to recover from a failed cache build.
- The S3 narinfo probe in `internal/devpkg`. Cache lookups now query only the public `https://cache.nixos.org`.
- `internal/setup`, the sudo-task framework whose only task was the nix cache host setup, plus `nix.IncludeDevboxConfig`/`restartDaemon`, which added the user to nix `trusted-users` and wrote `~root/.aws/config` with a `credential_process` pointing back at `devbox cache credentials`.
- The nightly `cache-upload` workflow.

## Kept

`nix.CurrentConfig`/`Config.IsUserTrusted` stay — they are generic nix.conf helpers with their own tests, not cache-specific.

## Verification

`go build`, `go vet`, `go test ./...` (including the full testscript suite), and `golangci-lint run` all pass on this branch on its own.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2941, opened 2026-08-04
feat!: remove Jetify Cloud secrets

mikeland73:
Second of five stacked PRs making Devbox account-free. **Based on #2940** — review that one first; this PR's diff is only the second commit.

## The stack

Each PR is based on the one above it — review and merge bottom-up, starting at #2940.

- `main`
  - **1.** [#2940 — remove the Jetify Nix cache](https://github.com/jetify-com/devbox/pull/2940)
    - **2.** [#2941 — remove Jetify Cloud secrets](https://github.com/jetify-com/devbox/pull/2941) &nbsp;⬅&nbsp; **you are here**
      - **3.** [#2942 — remove global push/pull to Jetify Cloud](https://github.com/jetify-com/devbox/pull/2942)
        - **4.** [#2943 — remove devbox auth and the identity provider](https://github.com/jetify-com/devbox/pull/2943)
          - **5.** [#2944 — drop dependencies left over from the account features](https://github.com/jetify-com/devbox/pull/2944)

## Why

`devbox secrets` and `env_from: jetify-cloud` stored secrets in Jetify Cloud, which required an org-scoped login plus a `.jetify/project.json` linking the project to a Jetify org.

## What's removed

- The `devbox secrets` command tree (alias `envsec`): `init`, `set`, `remove`/`rm`, `list`/`ls`, `download`, `upload`.
- `Devbox.Secrets`/`UninitializedSecrets`, the envsec bindings in `internal/devbox/secrets.go`.

## Behavior change worth a look

Projects that still set `env_from` to `jetify-cloud` (or the legacy `envsec` / `jetpack-cloud` spellings) **keep working**. `configEnvs` now warns and skips those secrets rather than failing, so an existing `devbox.json` does not have to be edited before the shell will start:

```
Warning: Ignoring env_from = "jetify-cloud". Jetify Cloud secrets are no longer supported by Devbox.
```

An `env_from` value that is neither a `.env` file nor a legacy cloud value is still an error. Both paths are covered by a new testscript, `testscripts/run/envfrom_jetify_cloud.test.txt`.

`ConfigFile.IsEnvsecEnabled` is renamed to `IsJetifyCloudEnvFrom`, since detecting the value is now all it does.

## Why `devbox auth whoami` goes here

`whoami` is removed in this PR rather than in the auth PR (#2943): its implementation *was* envsec's `WhoAmI`, so there is nothing left to call once the envsec bindings are gone. Deferring it would have meant writing a throwaway reimplementation.

## Verification

`go build`, `go vet`, `go test ./...` (including the full testscript suite), and `golangci-lint run` all pass on this branch on its own.


loreto (approved):
LGTM

PR #2942, opened 2026-08-04
feat!: remove global push/pull to Jetify Cloud

mikeland73:
Third of five stacked PRs making Devbox account-free. **Based on #2941** — this PR's diff is only the third commit.

## The stack

Each PR is based on the one above it — review and merge bottom-up, starting at #2940.

- `main`
  - **1.** [#2940 — remove the Jetify Nix cache](https://github.com/jetify-com/devbox/pull/2940)
    - **2.** [#2941 — remove Jetify Cloud secrets](https://github.com/jetify-com/devbox/pull/2941)
      - **3.** [#2942 — remove global push/pull to Jetify Cloud](https://github.com/jetify-com/devbox/pull/2942) &nbsp;⬅&nbsp; **you are here**
        - **4.** [#2943 — remove devbox auth and the identity provider](https://github.com/jetify-com/devbox/pull/2943)
          - **5.** [#2944 — drop dependencies left over from the account features](https://github.com/jetify-com/devbox/pull/2944)

## Why

`devbox global push` and `devbox global pull` with no argument synced the global profile to a Jetify-hosted S3 bucket, authenticating by exchanging the logged-in ID token for AWS credentials via `sts:AssumeRoleWithWebIdentity`.

## What changes

Both commands now require an argument and keep working for every source that needs no account: git repos, http(s) archives, plain `devbox.json` URLs, and local file paths.

- `push` takes a git repo, `pull` takes a git repo / URL / file.
- Cobra rejects the no-argument form (`accepts 1 arg(s), received 0`) instead of falling through to the cloud, so the failure is immediate and legible.
- `internal/pullbox/s3` is deleted, along with `devopt.Credentials`, which existed only to carry the ID token into it.

`push`'s help text previously read "Leave empty to use jetify cloud. Can be a git repo for self storage." — now just "Push a [global] config to a git repo".

## Verification

`go build`, `go vet`, `go test ./...` (including the full testscript suite), and `golangci-lint run` all pass on this branch on its own.

Manually checked against a build of `main` that the file-path and URL pull paths behave identically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2943, opened 2026-08-04
feat!: remove devbox auth and the identity provider

mikeland73:
Fourth of five stacked PRs making Devbox account-free. **Based on #2942** — this PR's diff is only the fourth commit.

## The stack

Each PR is based on the one above it — review and merge bottom-up, starting at #2940.

- `main`
  - **1.** [#2940 — remove the Jetify Nix cache](https://github.com/jetify-com/devbox/pull/2940)
    - **2.** [#2941 — remove Jetify Cloud secrets](https://github.com/jetify-com/devbox/pull/2941)
      - **3.** [#2942 — remove global push/pull to Jetify Cloud](https://github.com/jetify-com/devbox/pull/2942)
        - **4.** [#2943 — remove devbox auth and the identity provider](https://github.com/jetify-com/devbox/pull/2943) &nbsp;⬅&nbsp; **you are here**
          - **5.** [#2944 — drop dependencies left over from the account features](https://github.com/jetify-com/devbox/pull/2944)

## Why

With the nix cache (#2940), cloud secrets (#2941), and cloud global push/pull (#2942) gone, nothing is left that needs a Jetify account — so the login itself goes. This is the load-bearing PR: after the first three, `internal/devbox/providers/identity` has no callers left except telemetry.

## What's removed

- `devbox auth login` / `logout` / `tokens new`, and the `AUTH` feature flag that gated them. (`whoami` went in #2941, where its envsec implementation lived.)
- `internal/devbox/providers/identity`, which ran the browser OAuth/PKCE flow, stored sessions under `$XDG_CACHE_HOME/jetify/auth`, and exchanged `DEVBOX_API_TOKEN` for an access token.
- The identity lookups in telemetry. `user_id` now comes only from the `GITHUB_USER_NAME` hash and the `org_id` property is dropped, so telemetry no longer reads any token from disk. Telemetry is otherwise untouched — `deviceID`, Sentry, and `DO_NOT_TRACK` all behave as before.
- The Jetify endpoints in `internal/build`: `Issuer`, `ClientID`, `JetpackAPIHost`, `SuccessRedirect`, `Audience`, `DashboardHostname`.

## Result

Devbox no longer contacts `api.jetpack.io` or `accounts.jetify.com`, and no longer reads or writes an auth token anywhere. Verified by running the suite with a real, valid token still sitting in `~/Library/Caches/jetify/auth/` — nothing touches it. `devbox all` (which lists hidden commands too) shows no `auth`, `cache`, or `secrets`.

## Verification

`go build`, `go vet`, `go test ./...` (including the full testscript suite), and `golangci-lint run` all pass on this branch on its own.


loreto (approved):
LGTM

PR #2950, opened 2026-08-16
feat(config): recognize devbox.jsonc as an alternate config filename

### LINKED ISSUE #2602: Support `devbox.jsonc` for better editor & diff highlighting

`devbox.json` supports comments, which is quite helpful. By default though, IDEs and GitHub diffs don't expect `.json` files to have comments, and highlight them as errors.

If devbox supported searching for `devbox.jsonc` files as well as `.json`, I could just name them accurately and most syntax highlighters would just work.

In VSCode, I've added this to my `settings.json`, which at least helps in the IDE:
```json
    "files.associations": {
      "devbox.json": "jsonc"
    }
```
However, even with `devbox.json linguist-language=JSON-with-Comments` in `.gitattributes`, I can't get GitHub diffs to not highlight comments in red in diffs of the `devbox.json`, which makes PR review annoying.

mikeland73:
## Summary

Fixes #2602 (cc @gjoseph92).

`devbox.json` is parsed as JSONC and already supports comments, but editors and GitHub diffs treat comments in a `.json` file as syntax errors. This PR lets a project name its config **`devbox.jsonc`** so standard tooling highlights it correctly, with no per-repo `files.associations` or `.gitattributes` workarounds.

Changes (all in `internal/devconfig`):

- Add `configfile.AltName` (`"devbox.jsonc"`) and `configfile.ValidNames`, and have `searchDir` try both names during discovery. `devbox.json` is listed first so it still wins when a directory contains both files; existing projects are unaffected.
- Add `ConfigFile.FileName()` and make `SaveTo` use it, so `devbox add`/`rm`/etc. write back to `devbox.jsonc` instead of creating a stray `devbox.json`. Configs with no on-disk path (e.g. loaded from a URL by `devbox global pull`) still fall back to `devbox.json`.

### Scope / known follow-ups

This PR covers config discovery and save round-tripping. A few code paths still assume `devbox.json` and are left for a follow-up so this change stays small:

- `devbox run --all-projects` / `devbox update --all-projects` (`internal/boxcli/multi`) only match `devbox.json` when walking subdirectories.
- Generated `Dockerfile` / `.envrc` templates (`devbox generate`) hardcode `devbox.json`.
- `devbox global pull` "is default config" checks (`internal/pullbox`) only consider `devbox.json`.
- Docs don't yet mention `devbox.jsonc`.

## How was it tested?

`go build ./...`, `go vet ./internal/devconfig/...`, and `go test ./internal/devconfig/...` pass. New unit tests:

- `TestJSONCConfig` (`internal/devconfig/config_test.go`): `Open` and `Find` discover `devbox.jsonc`; `devbox.json` wins when both exist; `SaveTo` writes back to `devbox.jsonc` without creating a `devbox.json`.
- `TestFileName` / `TestSaveToPreservesFileName` (`internal/devconfig/configfile/file_test.go`): `FileName()` maps each path to the right basename and falls back to `devbox.json`; `SaveTo` writes the expected filename.

the

### DIFF
--- a/internal/devconfig/config.go
+++ b/internal/devconfig/config.go
@@ -142,8 +142,7 @@ func Find(path string) (*Config, error) {
 // searchDir looks for a config file in dir. It does not search parent
 // directories.
 func searchDir(dir string) (*Config, error) {
-	try := []string{configfile.DefaultName}
-	for _, name := range try {
+	for _, name := range configfile.ValidNames {
 		path := filepath.Join(dir, name)
 		slog.Debug("trying config file", "path", path)
 
@@ -156,7 +155,7 @@ func searchDir(dir string) (*Config, error) {
 		if errors.Is(err, os.ErrNotExist) {
 			continue
 		}
-		// Ignore directories named devbox.json.
+		// Ignore directories that happen to share a config filename.
 		if errors.Is(err, errIsDirectory) {
 			continue
 		}
--- a/internal/devconfig/config_test.go
+++ b/internal/devconfig/config_test.go
@@ -281,6 +281,86 @@ func TestFindError(t *testing.T) {
 	})
 }
 
+func TestJSONCConfig(t *testing.T) {
+	const jsonc = "{\n  // devbox lets you comment your config\n  \"packages\": []\n}\n"
+
+	t.Run("OpenDiscoversJSONC", func(t *testing.T) {
+		dir := t.TempDir()
+		path := filepath.Join(dir, configfile.AltName)
+		if err := os.WriteFile(path, []byte(jsonc), 0o644); err != nil {
+			t.Fatal(err)
+		}
+
+		cfg, err := Open(dir)
+		if err != nil {
+			t.Fatalf("Open(%q) error: %v", dir, err)
+		}
+		if cfg.Root.AbsRootPath != path {
+			t.Errorf("cfg.Root.AbsRootPath = %q, want %q", cfg.Root.AbsRootPath, path)
+		}
+	})
+
+	t.Run("FindDiscoversJSONCInParent", func(t *testing.T) {
+		root, child, _ := mkNestedDirs(t)
+		path := filepath.Join(root, configfile.AltName)
+		if err := os.WriteFile(path, []byte(jsonc), 0o644); err != nil {
+			t.Fatal(err)
+		}
+
+		cfg, err := Find(child)
+		if err != nil {
+			t.Fatalf("Find(%q) error: %v", child, err)
+		}
+		if cfg.Root.AbsRootPath != path {
+			t.Errorf("cfg.Root.AbsRootPath = %q, want %q", cfg.Root.AbsRootPath, path)
+		}
+	})
+
+	t.Run("DefaultNameWinsWhenBothExist", func(t *testing.T) {
+		dir := t.TempDir()
+		jsonPath := filepath.Join(dir, configfile.DefaultName)
+		if err := os.WriteFile(jsonPath, []byte(`{"packages": []}`), 0o644); err != nil {
+			t.Fatal(err)
+		}
+		if err := os.WriteFile(filepath.Join(dir, configfile.AltName), []byte(jsonc), 0o644); err != nil {
+			t.Fatal(err)
+		}
+
+		cfg, err := Open(dir)
+		if err != nil {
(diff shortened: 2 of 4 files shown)


mikeland73:
@claude please review this PR.

mikeland73:
## Review

Reviewed at `2201bf8` (my original change + the follow-up hardening commit). **LGTM** — the approach is sound and the follow-up closes the loose ends I'd originally left out of scope.

**Verified locally** (merged head, on top of current `main`):
- `go build ./...`, `go vet`, and `gofmt -l` on all touched files are clean.
- `go test ./internal/devconfig/... ./internal/devbox/generate/... ./internal/boxcli/multi/... ./internal/pullbox/...` passes. The only failures are `TestFindError/Permissions` and `/ExactFilePermissions`, which fail **identically on `main`** when the suite runs as root (uid 0 bypasses `0o000`); GitHub's non-root runners are unaffected. Not introduced here.

**Correctness of the core change:**
- Discovery is centralized in `searchDir`, now iterating `configfile.ValidNames` with `devbox.json` first, so a directory containing both files still resolves to `devbox.json` — existing projects are unaffected (`DefaultNameWinsWhenBothExist` covers this).
- `SaveTo` now writes back via `FileName()` (basename of `AbsRootPath`, falling back to `DefaultName` when unset), so `devbox add` round-trips to `devbox.jsonc` instead of dropping a stray `devbox.json`. URL-loaded configs (`AbsRootPath == ""`, e.g. `pullbox`) still default correctly.

**Follow-up commit — completeness check.** I grepped the tree for other spots that hardcoded the config name; the commit caught all the functional ones:
- `devconfig.Init` now refuses when *any* recognized name exists, preventing a new `devbox.json` from silently shadowing an existing `devbox.jsonc` (`TestInitRefusesWhenJSONCExists`). The returned `&fs.PathError{Err: fs.ErrExist}` still satisfies the caller's `errors.Is(err, os.ErrExist)` check in `boxcli/init.go`. ✅
- `boxcli/multi` and `pullbox/files.go` use `slices.Contains(configfile.ValidNames, …)` — faithful generalizations of the prior `== DefaultName` checks. ✅
- Generated **dev** Dockerfile copies `{{.ConfigFileName}}` (defaulting to `devbox.json` via `cmp.Or`); the **prod** Dockerfile uses `COPY . .`, so it already includes `devbox.jsonc` — correctly left untouched. ✅
- Generated `.envrc` now `watch_file`s both names. ✅

**Nits (non-blocking):**
- `generate/devcontainer_util.go:88` still hardcodes `"devbox.json"`, but it's only inside the "prod Dockerfile needs a 'start' script" **error message** — cosmetic, not a path.
- Optional: a `testscripts/` end-to-end case (rename to `devbox.jsonc`, then `devbox add` + `devbox shell`) would lock in the round-trip at the CLI layer, but the unit coverage across `Open`/`Find`/`SaveTo`/`Init`/Dockerfile is already solid.

Nothing here blocks merge.

---

mikeland73:
Good catch on the `multi.Open` finding — it's a real regression from the `ValidNames` change and I've fixed it in `4865407`.

`filepath.WalkDir` visited both files, so a directory containing **both** `devbox.json` and `devbox.jsonc` was opened twice, and `devbox run/update --all-projects` would run against it twice — inconsistent with the "devbox.json wins" precedence used everywhere else.

Fix:
- Deduplicate on the containing directory (a `seenDirs` set), so each project is opened once.
- Open by **directory** (`optsCopy.Dir = filepath.Dir(path)`) instead of by the specific file, so `devconfig.searchDir` applies the documented filename precedence rather than `multi` picking whichever file `WalkDir` happened to hit first.
- Added `internal/boxcli/multi/multi_test.go::TestOpenDeduplicatesConfigsInSameDir`, which sets up one project with both config names and one with only `devbox.json` and asserts exactly two projects are opened (fails at 3 without the fix).

Verified locally on the updated head: `go build ./...`, `go vet`, and `gofmt -l` are clean, and `go test ./internal/boxcli/... ./internal/devconfig/... ./internal/devbox/generate/...` passes (the only failures are the pre-existing `TestFindError` permission subtests that also fail on `main` under root).

On the nits: the `devcontainer_util.go:88` string and the other CLI help strings are cosmetic wording, left as the PR's documented follow-up.

---

PR #2958, opened 2026-08-17
fix(poetry plugin): activate the in-project virtualenv so scripts land on PATH

### LINKED ISSUE #2676: Poetry shell missing scripts
### What happened?

Scripts from the python package are not available on the path of the devbox shell since updating from 1.7.1.

### Details

The example in the devbox repo runs on poetry version [1.7.1](https://github.com/jetify-com/devbox/blob/86e9b6a0e0ac62ba344e1df1a7c10dc1887074a6/examples/development/python/poetry/poetry-demo/devbox.lock#L13)

Poetry is a major version ahead of this now and `poetry shell` is implemented by a plugin. [poetry docs](https://python-poetry.org/docs/cli/#shell)

The command itself shows:

> Looks like you're trying to use a Poetry command that is not available.
> 
> Since Poetry (2.0.0), the shell command is not installed by default. You can use,
> 
>   - the new env activate command (recommended); or
>   - the shell plugin to install the shell command

Maybe its just the examples that are out of date because I see that the [plugin code](https://github.com/jetify-com/devbox/blob/main/plugins/poetry/initHook.sh) uses: `poetry env use ...` instead.

This [fork](https://github.com/jetify-com/devbox/compare/main...tlelson:devbox:main) shows 3 steps:
1. Update poetry
2. Add a python script to the pyproject.toml
3. Hack the `init_hook` to get the script working in the devbox shell.

---

What the the best way to fix this?
Is there another way to do thing without using the poetry installed python?

I am happy to work on the fix.

### Steps to reproduce

1. Follow the 3 small commits in my fork

### Command

_No response_

### devbox.json

... see the fork

### Devbox version

0.14.2

### Nix version

nix (Nix) 2.28.4

macOS (Intel)

mikeland73:
## Summary

Fixes #2676.

Since **Poetry 2.0.0** removed the bundled `poetry shell` command, the poetry plugin relies on `poetry env use` (in `initHook.sh`) to point Poetry at the Devbox-provided Python. That *selects* the virtualenv but never *activates* it, so console scripts and entry points installed by `poetry install` live in `<project>/.venv/bin` and are **not on the Devbox shell's PATH**. Users are forced to prefix everything with `poetry run`, which is the regression reported in the issue:

> Scripts from the python package are not available on the path of the devbox shell since updating from 1.7.1.

### Fix

The plugin already sets `POETRY_VIRTUALENVS_IN_PROJECT=true`, so the virtualenv is always created at `<pyproject-dir>/.venv`. I activate it from the plugin's `init_hook` — which runs in the *sourced* shellrc context, unlike the *executed* `initHook.sh`, so it can actually modify the shell's `PATH`:

```jsonc
"init_hook": [
    "\"{{ .Virtenv }}/bin/initHook.sh\"",
    "if [ -d \"${DEVBOX_PYPROJECT_DIR:-$DEVBOX_DEFAULT_PYPROJECT_DIR}/.venv\" ]; then export VIRTUAL_ENV=\"${DEVBOX_PYPROJECT_DIR:-$DEVBOX_DEFAULT_PYPROJECT_DIR}/.venv\"; export PATH=\"$VIRTUAL_ENV/bin:$PATH\"; fi"
]
```

Ordering makes this robust:
- Plugin init hooks run **before** the user's own hooks (`Config.InitHook()` appends included-plugin hooks ahead of the root config's hooks).
- `poetry env use` inside `initHook.sh` creates `.venv`, so the `[ -d … ]` guard already passes when the activation line runs.
- By the time the user's `poetry install` executes, `.venv/bin` is already on `PATH`, so the freshly installed scripts resolve by name.

The path is derived from env vars and every expansion is double-quoted, so project directories containing spaces are handled correctly. The plugin `version` is bumped `0.0.5` → `0.0.6` so existing environments regenerate with the new hook, matching the convention used by prior plugin behavior changes.

## How was it tested?

- `go test ./plugins/ ./internal/plugin/` passes, including `plugins/init_hook_quoting_test.go` (the new line uses no `{{ }}` templates, and all path expansions are quoted).
- Validated the JSON parses and `bash -n` syntax-checked the rendered activation line.
- Simulated a spaced project dir end-to-end: exported `DEVBOX_DEFAULT_PYPROJECT_DIR="/tmp/pdir with space"`, placed an executable at `.venv/bin/myscript`, ran the activation line, and confirmed `VIRTUAL_ENV` was set correctly and `myscript` resolved and ran by name from `PATH`.

cc @tlelson — thanks for the clear root-cause analysis and reproduction in the issue.

the

---
_Generated by [Claude Code](https://claude.ai/code/session_012aiRPVpidRdyKDJiUQJhxN)_

### DIFF
--- a/plugins/poetry.json
+++ b/plugins/poetry.json
@@ -1,6 +1,6 @@
 {
     "name": "poetry",
-    "version": "0.0.5",
+    "version": "0.0.6",
     "description": "This plugin automatically configures poetry to use the version of python installed in your Devbox shell, instead of the Python version that it is bundled with. The pyproject.toml location can be configured by setting DEVBOX_PYPROJECT_DIR (defaults to the devbox.json's directory).",
     "env": {
         "DEVBOX_DEFAULT_PYPROJECT_DIR": "{{ .DevboxProjectDir }}",
@@ -13,7 +13,8 @@
     },
     "shell": {
         "init_hook": [
-            "\"{{ .Virtenv }}/bin/initHook.sh\""
+            "\"{{ .Virtenv }}/bin/initHook.sh\"",
+            "if [ -d \"${DEVBOX_PYPROJECT_DIR:-$DEVBOX_DEFAULT_PYPROJECT_DIR}/.venv\" ]; then export VIRTUAL_ENV=\"${DEVBOX_PYPROJECT_DIR:-$DEVBOX_DEFAULT_PYPROJECT_DIR}/.venv\"; export PATH=\"$VIRTUAL_ENV/bin:$PATH\"; fi"
         ]
     }
 }


mikeland73:

mikeland73:
## Review: correctness & merge-blockers

I reviewed the change adversarially. **No correctness blockers** — a few non-blocking notes below for your call.

### Verified correct
- **Hook ordering holds.** `Config.InitHook()` (`internal/devconfig/config.go:363-369`) appends included-plugin hooks *before* the root config's hooks, so this plugin's `initHook.sh` (which runs `poetry env use`, creating `<project>/.venv` in in-project mode) and the new activation line both run before a user's `poetry install`. By the time `poetry install` populates `.venv/bin`, that dir is already on `PATH`.
- **Env vars are available.** `DEVBOX_DEFAULT_PYPROJECT_DIR` is exported in the plugin's `env`, and `shellrc.tmpl` emits `ExportEnv` before sourcing the hooks file, so the expansion resolves.
- **Spaces are handled.** Every expansion is double-quoted; simulated end-to-end with a project dir containing a space and a script in `.venv/bin` resolved and ran by name.
- **Graceful degradation.** The `[ -d … ]` guard means that if `poetry env use` fails (no `pyproject.toml`, etc.) the line is a no-op rather than prepending a bogus path.
- **Checks:** `go build ./...`, `go test ./plugins/ ./internal/plugin/` (incl. `init_hook_quoting_test.go` — the new line uses no `{{ }}` templates and is fully quoted), `go vet`, and `gofmt` are clean. No snapshot/golden/testscript records the poetry `init_hook`, so nothing to regenerate.

### Non-blocking considerations
1. **`.venv` is hardcoded.** It's correct because the plugin forces `POETRY_VIRTUALENVS_IN_PROJECT=true`, but if a user overrides that to `false`, Poetry creates the venv under `POETRY_VIRTUALENVS_PATH` and the guard silently skips activation. That's not a regression (no worse than today), but the fix won't help that case. A more robust alternative is deriving the path from `poetry env info --path` at the cost of one extra Poetry invocation per shell start — your call on the trade-off.
2. **PATH duplication on refresh.** The hook re-runs on shell refresh / nested `devbox run`, unconditionally prepending `.venv/bin`, so `PATH` can accumulate duplicates within a session. Functionally harmless, and it mirrors how the existing `python` plugin's `venvShellHook` auto-activates, but a "skip if already present" guard would keep `PATH` tidy.
3. **Behavior change (intentional).** Every poetry-plugin user now gets the in-project venv auto-activated (`VIRTUAL_ENV` + `PATH`). This is the fix and it matches the `python` plugin's convention; the `0.0.5 → 0.0.6` bump reflects it.

### CI caveat
No CI checks have reported on this PR (status is `pending` with 0 checks). The shell logic was validated by simulation, but I could not exercise it against a live Poetry 2.x install (no Nix in my environment). If the repo's full integration suite runs `devbox shell` with the poetry example, that's the remaining real-world signal worth confirming before merge.

---

mikeland73:
Good catch on the fish-shell breakage — that's a real blocker, and I've verified and fixed it in `b6d4906`.

Confirmed the mechanism: fish sources the combined `.hooks` file natively (`internal/devbox/shellrc_fish.tmpl:62`), and fish has no `then`/`fi` (blocks close with `end`), so the inlined `if … then … fi` was an unterminated-`if` parse error that would abort the *entire* hooks file for any fish user with the poetry plugin — not just poetry's own line.

Rather than rewrite it as `[ … ] && …` short-circuits (fish also lacks `export` and `${VAR:-default}`, so that would still not be portable), I moved the PATH change to the plugin's **`env`** field, which is the existing shell-portable convention:

```jsonc
"env": {
    ...
    "PATH": "{{ .DevboxProjectDir }}/.venv/bin:$PATH"
}
```

This matches what the `nodejs` (`"{{ .Virtenv }}/corepack-bin:$PATH"`) and `ruby` plugins already do. Devbox expands the value with `os.Expand` (`internal/conf/env.go:27`) and `exportify` emits it as a single literal export, so it applies uniformly across bash, zsh, and fish with no inline shell syntax. The `init_hook` is back to just the executed `initHook.sh` line. `go test ./plugins/ ./internal/plugin/`, `go vet`, and `gofmt` are clean, and `grep` confirms no plugin `init_hook` contains `if`/`then`/`fi`.

One deliberate trade-off worth flagging: because `os.Expand` doesn't support `${VAR:-default}`, the `env` value uses the static `{{ .DevboxProjectDir }}` rather than the runtime `DEVBOX_PYPROJECT_DIR` override. So a user who relocates `pyproject.toml` via `DEVBOX_PYPROJECT_DIR` won't get that subdir's `.venv/bin` auto-added to PATH (they can still use `poetry run`). This covers the default case — which is what #2676 reports — and is a no-op, not a regression, for the override case. Happy to fold in `VIRTUAL_ENV` too, or handle the override, if you'd prefer.

---

mohsenari (approved):

PR #2959, opened 2026-08-17
Add devbox global info command

### LINKED ISSUE #2275: devbox info only works in devbox projects

When using devbox solely as a global package manager, `devbox info` doesn't work.  It returns the error:
```
```

I don't see any reason why `devbox info` can't work outside of devbox projects (like `devbox search` does).

Allow `devbox info` to work outside of devbox projects.

_No response_

### LINKED ISSUE #2500: Be able to run `devbox info xxx` without a project `devbox.json` file

I would like to use the command line to get info about a nix package without having a `devbox.json` file in my current directory (as I use `devbox` only for installing system wide tools (`devbox global xxx`))

I would like to be able to always run `devbox info xxx` and see the info of `xxx` package.

_No response_

humtta:
## Summary

This PR adds `devbox global info`, reusing the existing `info` command with the global configuration. This keeps it consistent with other project-scoped commands such as `add` and `rm`, which provide global variants.

Closes #2275 (cc @andbleo).
Closes #2500 (cc @yonil7).

## How was it tested?

Ran it from a directory with no `devbox.json` in the current directory or any parent directory.

Verified that `devbox info cowsay` failed because no project configuration was found, while `devbox global info cowsay` succeeded and displayed the package information.

All community contributions in this pull request are licensed to the project maintainers under the terms of the [Apache 2 License].

By creating this pull request, I represent that I have the right to license the contributions to the project maintainers under the Apache 2 License as stated in the [Community Contribution License].

[apache 2 license]: https://www.apache.org/licenses/LICENSE-2.0
[community contribution license]: https://github.com/jetify-com/opensource/blob/main/CONTRIBUTING.md#community-contribution-license

### DIFF
--- a/internal/boxcli/global.go
+++ b/internal/boxcli/global.go
@@ -29,6 +29,7 @@ func globalCmd() *cobra.Command {
 	}
 
 	addCommandAndHideConfigFlag(globalCmd, addCmd())
+	addCommandAndHideConfigFlag(globalCmd, infoCmd())
 	addCommandAndHideConfigFlag(globalCmd, installCmd())
 	addCommandAndHideConfigFlag(globalCmd, pathCmd())
 	addCommandAndHideConfigFlag(globalCmd, pullCmd())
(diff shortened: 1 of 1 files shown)


mikeland73 (approved):
regular info should probably work in non-devbox dir context as well, but this is reasonable too.

PR #2971, opened 2026-09-15
fix(release): exit cleanly on flake bump stop, don't block on in-progress cli-tests

mikeland73:
## Summary

When `publish-release` opens the flake bump PR and has to stop, it now exits 0 (it's a planned pause, not a failure, so `devbox run` no longer appends `Error: ... exit status 1`) and prints the exact command to continue with, e.g. `devbox run publish-release --version 0.18.2`. The release scripts in `devbox.json` now forward `"$@"` so that `--version` actually reaches `release.ts`. The "cli-tests on main" check only stops on a completed red run — a queued/in-progress run (the normal state right after the bump PR merges) warns and continues, since `cli-release` runs the same suite itself. Docs and `--help` text updated to match.

## How was it tested?

`node --check scripts/release.ts`; a simulated halt prints the resume command and exits 0; `devbox run release-status 0.18.0` confirms args are forwarded through `devbox run`.

## Community Contribution License

All community contributions in this pull request are licensed to the project
maintainers under the terms of the
[Apache 2 License](https://www.apache.org/licenses/LICENSE-2.0).

By creating this pull request, I represent that I have the right to license the
contributions to the project maintainers under the Apache 2 License as stated in
the
[Community Contribution License](https://github.com/jetify-com/opensource/blob/main/CONTRIBUTING.md#community-contribution-license).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2973, opened 2026-09-15
fix(devpkg): treat binary cache narinfo network errors as a cache miss

mikeland73:
## Summary

Before running a script or shell, devbox HEADs `https://cache.nixos.org/<hash>.narinfo` for each package with a 5s timeout to decide whether it can `fetchClosure` instead of building. A timeout, DNS failure, or connection reset from that check propagated up and failed the whole command (seen as a flaky `script.test` failure in CI on #2971). Since the check is only an optimization, transport errors now return "not in cache" (with a debug log) so the caller falls back to letting Nix build/substitute the package — a slow cache makes the command slower, not fail. Non-2xx responses were already treated this way; the result is still memoized per hash so all callers in one command agree.

## How was it tested?

Added `narinfo_cache_test.go` covering 200 → hit, 404 → miss, a server that hangs past the timeout → miss with no error, and a refused connection → miss with no error. `go build ./...`, `go vet`, and `go test ./internal/devpkg/ ./internal/shellgen/` pass.

the


mikeland73:
@claude ptal

Issue #2703, opened 2025-09-23
`devbox integrate` does NOT run init_hook


tm-michael:
### What happened?

When the VSCode extension is run and on selects "Reopen devbox", it only sets the env variables but it doesn't run the init hook. Init hooks can include important steps and set other environment variables that aren't set using the `env` key in the `devbox.json`. When using things like Cursor or extensions that open their own shells and don't get the `devbox shell` injection, they thus lack important variables

### Steps to reproduce

1. create a devbox.json with an init_hook that sets an environment variable e.g `TEST=true`
2. open the project in VSCode with the extension installed and the "devbox shell" injection deactivated <img width="725" height="406" alt="Image" src="https://github.com/user-attachments/assets/63ffe2df-2e1b-47bf-97ba-e083e41a1c83" />
3. use the "reopen in devbox" feature
4. open a terminal and run `echo $TEST`

It will not be set.

### devbox.json

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.16.0/.schema/devbox.schema.json",
  "packages": [],
  "env": {
    "SOMETHING": "hello"
  },
  "shell": {
    "init_hook": [
      "export TEST=true"
    ],
    "scripts": {
      "test": [
        "echo \"Error: no test specified\" && exit 1"
      ]
    }
  }
}
```

### Devbox version

0.16.0

### Nix version

2.31.0

### What system does this bug occur on?

Linux (x86-64)

tm-michael:
This is also the case for `devbox shellenv`, but `devbox run` does.

PR #2975, opened 2026-09-15
feat(vscode): add opt-in runInitHookOnReopen setting

mikeland73:
Stacked on #2874 (adds the `--run-init-hook` flag to `devbox integrate vscode`). Merge that first.

## Summary

- New extension setting **`devbox.runInitHookOnReopen`** (boolean, default `false`). When enabled, "Devbox: Reopen in Devbox shell environment" passes `--run-init-hook` to `devbox integrate vscode`, so the project's `init_hook` runs and any environment variables it exports are present in the reopened editor. See #2703.
- The flag is only appended when the setting is on, so users on an older CLI that doesn't know the flag keep working with default settings.
- Bumps the extension to **0.1.9** so `vsce publish --skip-duplicate` in the release workflow actually publishes the change.
- README and CHANGELOG updated.

## How was it tested?

- `yarn compile` and `yarn lint` pass in `vscode-extension/`.


apgrucza:
Looks like the vscode release workflow hasn't run yet: https://github.com/jetify-com/devbox/actions/workflows/vscode-ext-release.yaml

mikeland73:
> Looks like the vscode release workflow hasn't run yet: https://github.com/jetify-com/devbox/actions/workflows/vscode-ext-release.yaml

on it!

@mohsenari let me know if anything else I should do

mohsenari:
I think we're good.

